PR #260, opened 2020-10-09
feat(publishing): preserve abbreviations

LukeCarrier:
The Markdown source:

    HTML

    *[HTML]: HyperText Markup Language

Will now be preserved, allowing renderers (e.g. Jekyll) to emit:

    <abbr title="HyperText Markup Language">HTML</abbr>

### DIFF
--- a/packages/engine-server/src/topics/markdown/remark-abbr.d.ts
+++ b/packages/engine-server/src/topics/markdown/remark-abbr.d.ts
@@ -1,0 +1,14 @@
+declare module "remark-abbr" {
+  import { Plugin } from "unified";
+
+  namespace remarkAbbr {
+    type Abbr = Plugin<[Options?]>;
+
+    interface Options {
+      expandFirst?: boolean;
+    }
+  }
+
+  const plugin: remarkAbbr.Abbr;
+  export = plugin;
+}
--- a/packages/dendron-cli/src/commands/__tests__/__snapshots__/build-site.spec.ts.snap
+++ b/packages/dendron-cli/src/commands/__tests__/__snapshots__/build-site.spec.ts.snap
@@ -2,6 +2,7 @@
 
 exports[`buildSite basic: bond 1`] = `
 "- [lbl](refactor.one)
+
 "
 `;
 
@@ -9,16 +10,19 @@ exports[`buildSite ids are converted: converted 1`] = `
 "Build site home
 
 - [build-site.one](notes/id.build-site.one)
+
 "
 `;
 
 exports[`buildSite image prefix and no copy assets: sample.image-link.md 1`] = `
 "![link-alt](fake-s3.com/link-path.jpg)
+
 "
 `;
 
 exports[`buildSite image prefix: sample.image-link.md 1`] = `
 "![link-alt](fake-s3.com/link-path.jpg)
+
 "
 `;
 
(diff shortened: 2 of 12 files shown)


packages/engine-server/src/topics/markdown/remark-abbr.d.ts
```diff
@@ -0,0 +1,14 @@
+declare module "remark-abbr" {
```
LukeCarrier, line 1: I'm happy to get this into either Definitely Typed or directly upstream, if the maintainer is happy to accommodate the types.

kevinslin: i think definitely typed is open to anyone contributing so you might just be able to submit a pull request directly

kevinslin (commented):
this looks great! can you add a unit test in [markdown.spec.ts](https://github.com/dendronhq/dendron/blob/master/packages/engine-server/src/topics/markdown/markdown.spec.ts#L47:L47) so future generations can preserve this functionality? :)

packages/engine-server/src/topics/markdown/remark-abbr.d.ts
LukeCarrier: Yup, I think so too.

I'll _try_ raising a PR against [zmarkdown](https://github.com/zestedesavoir/zmarkdown/tree/master/packages/remark-abbr) first, but since there are quite a few plugins there I suspect they'll be hesitant.

LukeCarrier:
It looks like `remark-abbr` introduces some additional whitespace into rendered output. I couldn't see anything obvious, but I'll try to get to the bottom of it tomorrow.

kevinslin:
> It looks like `remark-abbr` introduces some additional whitespace into the processed output, which has updated some other Jest snapshots used by `markdown.spec.ts`. I couldn't see an obvious cause but I'll try to get to the bottom of it tomorrow.

if its just newline , then it's not a big deal since markdown preview will squash consecutive newlines. if its not a substantial change, you are free to run the jest test with the `-u` flag and add it to the pull request :)

kevinslin:
Hey @LukeCarrier , this looks great. one last thing, could you also sign the CLA? 

tldr: the CLA helps us be flexible in the future for different business models. it allows us to re-license dendron for specific scenarios. 
- for example, when deploying to enterprise (some of them, eg. google, don't allow for software with gpl licenses), dendron can be deployed under a custom license instead.

LukeCarrier:
@kevinslin signed!

If you're able to hold off merging I'm in the process of getting the types into DefinitelyTyped. I think I need to raise a PR against the tooling to get Unified and Remark whitelisted so I can use their types.

kevinslin:
Roger, no rush. Will hold off

kevinslin:
checking if there are any updates on this?

LukeCarrier:
Hey @kevinslin, sorry for the hold up. Would you be happy to merge with the definition inline for now and I'll get a PR in to zmarkdown later this week? I've got to give some thought to how to test the TS definitions I'm submitting there and I think I'll have to cover a couple of other modules.

kevinslin:
inline definition is fine and we can remove it later when it gets merged upstream. one more addition, could you also add the abbreviation to [utilsv2](https://github.com/dendronhq/dendron/blob/master/packages/engine-server/src/topics/markdown/utilsv2.ts#L45) which is used by build/publish for server mode?

LukeCarrier:
I've added the plugin to `utilsv2`, rebased and updated the snapshots. I noticed that the whitespace changes leak into the `buildPod` output and I'd like to dig into this when I write up the TS definitions for zmarkdown, but I hope this is okay in the meantime?

Thanks for your patience!

kevinslin:
All good. Thanks for the contribution. Will make sure to get it out with the next release!

LukeCarrier:
I've raised a PR to upstream the `remark-abbr` types: zestedesavoir/zmarkdown#419

PR #285, opened 2020-10-21
Don't throw an exception in genNoteNameOnly when opts are passed

### LINKED ISSUE #122: Create daily journal note without requiring active text editor
## Context

Creating a daily journal will fail if you are currently not in an open editor (eg. the welcome page of VSCode).

This is a carryover from how scratch notes and journal notes work because they get added relative to a given hierarchy. The daily journal note does not require this.

## Proposal

Don't throw error when creating a daily journal note and you are not currently in a note

## Changes Required

- [CreateNote](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/commands/CreateNote.ts#L81)
    - [ ] `genFname` needs to take an additional property in opt, `requireActiveEditor?: boolean`
    - [ ] `genFname` should not throw an error if no active editor is present and the above flag is set

bs:
Fixes #122

- Don't throw an exception in genNoteNameOnly when called from outside of a note if opts.overrides.domain is passed.
- Call DendronClientUtilsV2 rather than VSCodeUtils when creating a journal or scratch from LookupController

### DIFF
--- a/packages/plugin-core/src/components/lookup/LookupController.ts
+++ b/packages/plugin-core/src/components/lookup/LookupController.ts
@@ -7,6 +7,7 @@ import { LookupCommandOpts } from "../../commands/LookupCommand";
 import { CONFIG } from "../../constants";
 import { Logger } from "../../logger";
 import { VSCodeUtils } from "../../utils";
+import { DendronClientUtilsV2 } from "../../utils";
 import { DendronWorkspace } from "../../workspace";
 import {
   ButtonType,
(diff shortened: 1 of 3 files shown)


kevinslin:
This looks great. Can you also add an test [here](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/LookupCommand.test.ts#L876:L876) to assert that no errors are thrown when the `asOwnDomain` flag is set?

bs:
Yes! Slammed tomorrow but will do this weekend :)

kevinslin:
well, now i have to find another **good first issue** for folks. thanks for the pr, will go out in the next release :)

PR #293, opened 2020-10-23
Avoid doc-ception and doc-ception encouraging test

bs:
Add an emoji (and indicate that the Doctor command has initiated)


kevinslin:
looks good. appreciate the fix and the infusion of cultural references :)

PR #316, opened 2020-10-29
fix(publishing): use remark-abbr types from DT

LukeCarrier:
Follow up to #260 🎉


kevinslin:
Awesome, glad to see that merged :)

PR #323, opened 2020-11-01
Fixing minor typos

seedublancaster:
Fixing two minor transpositions.


seedublancaster:
The only reason for the merge pull was that I accidentally forked two separate patches for the same correction to the same single file.

kevinslin:
thanks for the pr!

PR #335, opened 2020-11-10
Fix typo in engine-server/src/workspace.ts

bunchesofdonald:


kevinslin:
Thanks for catching this!

PR #360, opened 2020-11-17
Fix formatting

konrad-jamrozik:


kevinslin:
Haha, of course you would notice since you've been submitting the issues. Thanks!

PR #372, opened 2020-11-23
Reference dendron gem for publishing

cldwalker:
Hi. While [running the publishing jekyll server locally](https://www.dendron.so/notes/63b184f7-1901-4580-af40-97ef87069a04.html), I noticed two jekyll files were out of date. I think these changes are correct but perhaps I'm mistaken. Cheers


kevinslin:
You are correct. Thanks for the fix!

PR #373, opened 2020-11-25
Minor changes to clean up build.sh + TS errors

kpathakota:
dendron-cli was looking for dendron-pods/seeds-core so I had to interchange those values in the build script. 

I think the getConfigData was just a copy error from a few commits ago and was failing a type script check for version number.


kpathakota:
Build errors and I seem to have used the wrong user when committing this.

PR #384, opened 2020-11-27
Removed duplicate version

kpathakota:
Strong TSX linter through a build error when I ran this. Not sure which version you wanted here @kevinslin


kevinslin:
ah, its okay. i have an updated version that will fix this. currently doing quite a big refactor which is why i haven't pushed the code yet

kpathakota:
Cool, I'll close this one out then.

PR #386, opened 2020-11-27
Toggle plugin views using 'when' contexts

cscorley:
Hi team!

I started off a bit frustrated and almost requested a feature, but I figured it was simple enough I might be able to figure it out on my own and learn a bit of extension writing.  

Essentially, I did not like "Dendron Tree View" and "Backlinks" in my view pane while working in non-Dendron workspaces.  I know these can be toggled off, but I thought it better if Dendron managed this itself.  This PR enables Dendron view panes selectively (i.e., when Dendron is actually active) [using 'when' contexts](https://code.visualstudio.com/api/extension-guides/command#using-a-custom-when-clause-context).  I am not entirely sure if this covers every place Dendron becomes active, but the manual testing I completed shows the idea works.

I am not sure how to write a test for this.  Ideally, I would also store these keys into `constants.ts`, but I am also unsure if there should be a new enum defined or if this cleanly fits elsewhere.  Additionally, I think this should be configurable somehow, but I'm not very familiar with how to accomplish that either.  Thanks!

### DIFF
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -33,12 +33,12 @@
                 {
                     "id": "dendronTreeView",
                     "name": "Dendron Tree View",
-                    "visibility": "visible"
+                    "when": "dendron:showTreeView"
                 },
                 {
                     "id": "dendron.backlinksPanel",
                     "name": "Backlinks",
-                    "visibility": "visible"
+                    "when": "dendron:showBacklinksPanel"
                 }
             ]
         },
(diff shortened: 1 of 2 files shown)


kevinslin:
First, thanks for the PR. appreciate you taking the time to send a PR instead of just filing an issue (though issues are welcomed to). 

As far as testing, Dendron basically mocks the vscode extension context to test activation because tests are interrupted whenever vscode reloads a workspace. We're currently refactoring the tests to have a uniform interface (we've evolved our testing harness over the past few months) so I'm going to merge this as is. 

As for adding using a constants keys - yes, that would be desired. We'll add that in when we finish refactoring the test.

Issue #344, opened 2020-11-14
Lookup Note Creation does not Respect Title Case


HeuFui:
**Describe the bug**
You're creating a new note through lookup. You carefully type its title using Title Case (sometimes it's the name of a person or author). When said note is created, the title case isn't respected in this new note's title.

This is not a big problem _per se_. Typing Something Like This and getting 'Typing something like this' is not so bad. It's worse when you have as title the name of a person or author. For instance, when you create a new with something like 'Georg Wilhelm Friedrich Hegel's Science of Logic circularity' and get as its title 'Georg wilhelm friedrich hegel's science of logic circularity', it becomes, well, weird.

Example:
![1](https://user-images.githubusercontent.com/7608736/99144970-74dc6100-2649-11eb-8074-7bda04d3576d.png)

![2](https://user-images.githubusercontent.com/7608736/99144975-802f8c80-2649-11eb-99d6-5c4e80dd1e6f.png)

**To Reproduce**
1. Go to lookup;
2. Create a new note with Title Case;
3. Check new note's title.

**Expected behavior**
What's expected is that the casing used during the note creation is respected in the title of the new note.

kevinslin:
Thanks for the issue. Will be adding it to this months backlog!

kevinslin:
shipped in 0.20

PR #404, opened 2020-12-13
fixed note creation not respecting title case issue #344

brainjos:
removed the post-formatting for the title to remain true to the original input casing.

### DIFF
--- a/packages/common-all/src/nodev2.ts
+++ b/packages/common-all/src/nodev2.ts
@@ -532,7 +532,7 @@ export class NoteUtilsV2 {
   }
 
   static genTitle(fname: string) {
-    return _.capitalize(DNodeUtilsV2.basename(fname, true));
+    return DNodeUtilsV2.basename(fname, true);
   }
 
   static getNotesByFname({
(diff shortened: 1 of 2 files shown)


kevinslin (changes requested):
Thanks for moving quickly on this pull request :)

A few comments:
- we still want capitalization on the title if the user doesn't have custom capitalization. instead of removing capitalization, can we do a check to only capitalize if the input is all lowercase?
- can we add a test for this?
    - the note creation tests are [here](https://github.com/dendronhq/dendron/blob/master/packages/engine-server/src/__tests__/enginev2.spec.ts#L149:L149)
    - would add a test case [here](https://github.com/dendronhq/dendron/blob/master/packages/common-test-utils/src/presets/engine-server/write.ts#L106:L106) about mixed capitalization
- this reminded me to update our [pull request checklist](https://www.dendron.so/notes/81da87be-2d4e-47b5-a1d6-c0d647e1ab00.html#pull-request-checklist) :)

brainjos:
Ah, that makes sense. Just to make sure I understand:
1. if a title is 'Title example' it remains 'Title example'
2. If a title is 'title example' it changes to 'Title example'
3. If a title is 'Title Example' it remains 'Title Example'
4. If a tile is 'title Example' it remains as 'title Example'
Most unsure about case 4, seems like the convention before was to make sure that the first letter was capitalized.

PR #426, opened 2020-12-27
fix broken FAQ link

Newman5:
replace 
https://github.com/dendronhq/dendron-template/blob/master/vault/dendron.faq.md
with
https://dendron.so/notes/683740e3-70ce-4a47-a1f4-1f140e80b558.html


kevinslin:
thanks for caching this!

PR #449, opened 2021-01-12
GitHub actions

LukeCarrier:
Sorry for the delay in getting this over the line!

I'm going to continue working on the Windows issues and aim to get both a Windows and a Cygwin on Windows build working soon.

Could you take a look at the tests skipped in `chore: skip leaky tests` and see if you're able to make sense of what's happening? I can then rebase out that commit before you merge this.


kevinslin:
This looks good. I'm going to go ahead and merge then diagnose the skipped tests post merge

This isn't a blocker but just a question: did the ci/cd test fail when jest was running in parallel? noticed that we changed it to run serially

LukeCarrier:
@kevins8 sweet, thanks for reviewing this. I switched to running serially to make the Jest output easier to read when troubleshooting failures. Happy to switch to running them in parallel if you prefer? I could try to split out the test output per-module and make that available as an artifact.

kevinslin:
> @kevins8 sweet, thanks for reviewing this. I switched to running serially to make the Jest output easier to read when troubleshooting failures. Happy to switch to running them in parallel if you prefer? I could try to split out the test output per-module and make that available as an artifact.

Got it. no worries since ci-cd minutes are (for now) free on github for open source. I might add a separate local parallel option since I usually run it manually before committing. 

One last thing, could you also update the developer [getting started docs](https://dendron.so/notes/64f0e2d5-2c83-43df-9144-40f2c68935aa.html) with the new scripts?

Issue #417, opened 2020-12-20
Use new markdown parser when publishing Markdown


kevinslin:
## Context

We made an update to how Dendron parses markdown but they are a bunch of places where we still use the old md parsing toolchain 

## Proposal
- replace the parser used for markdown publishing [here](https://github.com/dendronhq/dendron/blob/master/packages/pods-core/src/builtin/MarkdownPod.ts#L240:L240)
- [new parser](https://github.com/dendronhq/dendron/blob/master/packages/engine-server/src/markdown/utils.ts#L89:L89 )
- example usage [here](https://github.com/dendronhq/mume/blob/master/src/markdown-engine.ts#L2949:L2949)
- add test cases [here](https://github.com/dendronhq/dendron/blob/master/packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts#L10:L10)

hikchoi:
Looks like you already swapped out the parser here while doing the pairing session :smile: : https://github.com/dendronhq/dendron/commit/c73f6647adb8c5e3b21aab049989d44745448a39#diff-061146d73ca415bb58ed56495fe5d04ee9fb665cd8f86930e6f893139d6de3db

Should I go ahead and add additional tests with the rest of the `ENGINE_HOOKS` or do you think the basic one there should be sufficient?

kevinslin:
heh, you're right. I would say add a few more tests and we can call it
good 😅
Meanwhile, you can check the github issues for other good first tasks when
you feel up for the next one:
https://github.com/dendronhq/dendron/labels/size.small

PR #458, opened 2021-01-19
chore: add more tests for MarkdownPod

hikchoi:
- Added additional tests for `MarkPublishPod`
Closes https://github.com/dendronhq/dendron/issues/417

### DIFF
--- a/packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts
+++ b/packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts
@@ -79,41 +79,37 @@ describe("MarkdownPublishPod", () => {
     );
   });
 
-  test("schema with template", async () => {
+  test("empty note", async () => {
     await runEngineTestV5(
       async ({ engine, wsRoot, vaults }) => {
+        const fname = "empty";
         const vault = vaults[0];
         const vaultName = VaultUtils.getName(vault);
-        const fnames = ["bar.template.ch1", "bar.template.ch2"];
-        const resp = await Promise.all(
-          fnames.map(async (fname) => {
-            return new MarkdownPublishPod().execute({
-              engine,
-              vaults,
-              wsRoot,
-              config: {
-                fname,
-                vault: vaultName,
-                dest: "stdout",
-              },
-            });
-          })
-        );
-        expect(resp).toEqual(["ch1 template", "ch2 template"]);
+        const resp = await new MarkdownPublishPod().execute({
+          engine,
+          vaults,
+          wsRoot,
+          config: {
+            fname,
+            vault: vaultName,
+            dest: "stdout",
+          },
+        });
+        expect(resp).toEqual("");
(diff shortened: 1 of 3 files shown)


packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts
```diff
@@ -78,4 +78,129 @@ describe("MarkdownPublishPod", () => {
...
+        expect(resp).toEqual([fooOneBody, fooTwoBody]);
+      },
+      {
+        expect,
+        preSetupHook: ENGINE_HOOKS.setupNoteRefRecursive,
+      }
+    );
+  });
+
+  test("journal", async () => {
```
kevinslin, line 171: since this case isn't really testing anything `setupJournals` was a preset to test that sibling navigation worked, we don't need to include this as a test case

packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts
```diff
@@ -78,4 +78,129 @@ describe("MarkdownPublishPod", () => {
       }
     );
   });
+
+  test("schema with template", async () => {
```
kevinslin, line 82: `schemaPreset` was to test that template schemas worked. I would use these tests more to test special cases for markdown (eg. what happens when we publish an empty note? a note with a wiki link? a note with a relative link? etc)

packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts
```diff
@@ -78,4 +78,129 @@ describe("MarkdownPublishPod", () => {
...
+        expect(resp).toEqual(["ch1 template", "ch2 template"]);
+      },
+      {
+        expect,
+        preSetupHook: ENGINE_HOOKS.setupSchemaPreseet,
+      }
+    );
+  });
+
+  test("schema with namespace template", async () => {
```
kevinslin, line 111: same comment as above. this tests for schema functionality but doesn't test something specific about Publish

kevinslin (changes requested):
see comments

packages/pods-core/src/builtin/__tests__/MarkdownPod.spec.ts
hikchoi: Will think about some edge cases then 👍

hikchoi:
I've removed the unnecessary tests and added tests for dendron specific markdowns (wikilinks, various refs). 
I tried to cover everything I could find that are dendron specific markdown, but I might have missed some things, so please let me know if that's the case.

Also bundled related cases into one test block, but I'm not sure if that's how you would usually lay out your tests.
If it's better to have them separately, I'll do that.

kevinslin (approved):

kevinslin:
awesome, thanks for the changes. just merged :)

PR #472, opened 2021-02-08
Add description option to dendron config

### LINKED ISSUE #464: Initialize dendron config with better defaults
## Context

When you publish for the first time, certain seo metadata will show up as `undefined` if you don't fill in a value. We should have a default for `description` field 

- defaults for config: https://github.com/dendronhq/dendron/blob/master/packages/engine-server/src/config.ts#L16:L16
- seo options: https://dendron.so/notes/f2ed8639-a604-4a9d-b76c-41e205fb8713.html#seo-options
- tests: https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/Extension.test.ts#L103:L103

## Proposal

Add a default value for `description` field.

nikitavoloboev:
Fixes #464 

Not sure if the description default is good though.

### DIFF
--- a/packages/engine-server/src/config.ts
+++ b/packages/engine-server/src/config.ts
@@ -24,6 +24,7 @@ export class DConfig {
         siteRootDir: "docs",
         usePrettyRefs: true,
         title: "Dendron",
+        description: "Personal knowledge space",
       },
     };
   }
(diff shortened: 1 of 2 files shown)


nikitavoloboev:
will add tests

PR #478, opened 2021-02-11
Render 403 pages as popover

nikitavoloboev:
depends on https://github.com/dendronhq/dendron-11ty/pull/10


kevinslin (changes requested):
looks like this broke some tests. we have jest snapshots that do an exact match of `dendronPub`. you can fix it by updating the snapshots. they are automatically updated if you run the test for that specific file: https://dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#run-a-specific-test-from-non-dendron-plugin-package

kevinslin (approved):

PR #479, opened 2021-02-11
feat: add option to configure human readable ts formatting

### LINKED ISSUE #468: Add Day of the Week to the Epoch Timestamps
**Is your feature request related to a problem? Please describe.**
For Journal notes and others for which the date is an important factor, it would be useful to have the additional context of the day of the week included with the "created" and "updated" frontmatter.

**Describe the solution you'd like.**
At the moment, the epoch timestamps are presented as such:

1612190405514 (Feb 1, 2021, 9:40 AM)

This could be changed to include the day of the week, such as: 

1612190405514 (Mon, Feb 1, 2021, 9:40 AM)

**Describe alternatives you've considered**
Alternatively, a user could look up the associated day of the week manually using an epoch converter website, or, as suggested on the Dendron site, install the timestamp converter extension. Both of these options seem a waste however, as Dendron already converts the majority of its epoch timestamps to a human readable format. 

Thank you.

hikchoi:
- add dendron workspace option to choose how human readable timestamps are formatted.


packages/plugin-core/src/windowWatcher.ts
```diff
@@ -52,11 +54,15 @@ export class WindowWatcher {
       const ts = _.toInteger(_.trim(tsMatch[0].split(":")[1], `'" `));
 
       const dt = DateTime.fromMillis(ts);
+      const tsConfig = getConfigValue(
+        CodeConfigKeys.DEFAULT_TIMESTAMP_DECORATION_FORMAT
+      ) as string;
+      const formatOption = (DateTime as any)[tsConfig];
```
hikchoi, line 60: I'm not entirely sure if I should be doing this here. Would it defeat the purpose of using typescript?

hikchoi:
Also, I have no idea how to go about testing this 😅  
`vscode.window` doesn't expose anything about the currently set decoration.

packages/plugin-core/package.json
```diff
@@ -125,6 +125,34 @@
...
+                        "DATETIME_HUGE",
+                        "DATETIME_HUGE_WITH_SECONDS",
+                        "DATETIME_MED",
+                        "DATETIME_MED_WITH_SECONDS",
+                        "DATETIME_SHORT",
+                        "DATETIME_SHORT_WITH_SECONDS",
+                        "DATE_FULL",
+                        "DATE_HUGE",
+                        "DATE_MED",
+                        "DATE_MED_WITH_WEEKDAY",
```
hikchoi, line 144: This is not exactly what was requested in #468 but close. 

weird, the type declaration from luxon doesn't include `DateTime.DATETIME_MED_WITH_WEEKDAY` (which is exactly what was requested).

Maybe this PR shouldn't be closing the issue.

packages/plugin-core/src/windowWatcher.ts
kevinslin: the way to get around this is to make a `DateTimeFormat` type that consists of the union of all the static members.

```ts
type DateTimeFormat = "DATETIME_FULL" | "..." | "...";
```

kevinslin (changes requested):
looks good, just on ecomment about the types

packages/plugin-core/src/windowWatcher.ts
hikchoi: Lovely.

hikchoi:
Just pushed a fix that removes type assertion and defines a `DateTimeFormat` type. One day I'll get used to typing 😇 

Any idea why luxon has `DATETIME_MED_WITH_WEEKDAY` implemented but doesn't have a declaration for it?

kevinslin (approved):
looks good :)

PR #506, opened 2021-02-24
Show backlinks at the end of pages

nikitavoloboev:


packages/engine-server/src/markdown/remark/backlinks.ts
```diff
@@ -0,0 +1,49 @@
...
+import u from "unist-builder";
+
+// Plugin that adds backlinks at the end of each page if they exist
+const plugin: Plugin = function (this: Unified.Processor) {
+  const proc = this;
+  function transformer(tree: Node) {
+    let root = tree as Root;
+    const { fname, vault } = MDUtilsV4.getDendronData(proc);
+    const { engine } = MDUtilsV4.getEngineFromProc(proc);
+    const note = NoteUtilsV2.getNotesByFname({
```
kevinslin, line 16: should use `getNoteByFnameV5` for getting a single note. `getNotesByFname` is for when you only have the file name

kevinslin (changes requested):
can we add some tests as well as make sure the type checks pass?

nikitavoloboev:
I added the test and used `getNoteByFnameV5`.

For the test, I am not sure how I can create multiple files. I thought doing:

```
        const resp = await MDUtilsV4.procRehype({
          proc: proc(engine, { ...dendronData, fname: "alpha", vault }),
        }).process("[[beta]]");
        const resp2 = await MDUtilsV4.procRehype({
          proc: proc(engine, { ...dendronData, fname: "beta", vault }),
        }).process("");
```

Would work but it doesn't and I didn't find other examples. In my mind above code first creates a file with name `alpha` and content `[[beta]]`. And then a file with name `beta` with empty content. I then try check that file `beta` has the content with `backlinks` inside.

Also this type error, I am not sure how to fix it.

image

kevinslin (approved):

Issue #329, opened 2020-11-06
Link previews // Pretty URLs


tmladek:
**Is your feature request related to a problem? Please describe.**
A significant portion of my notes is composed of briefly annotated URLs. My workflow right now is that I just paste the URL in a note, then write the comment below, but the bare URL is a bit ugly, and it can be quite difficult to tell at a glance what it contains. It's not as bad when the URL follows a reasonable format (https://example.com/posts/how-i-solved-my-note-taking-woes), but oftentimes that's not the case (https://twitter.com/tweeter/status/15246681041231653377).

**Describe the solution you'd like**
One solution that might be nice is to add an (optional) link preview feature to the Markdown preview sidebar, and pop up a box below the URL with some basic [Open Graph](https://ogp.me/) metadata. This would take care of the vast majority of social media posts, and a great deal of other documents too. The OG format is easy enough to parse (provided there's a HTML parser available), but one issue that I see is that it incurs a couple additional requests on each preview and I'm not sure how that would work out once the underlying document would start updating.

Alternatively, there would be no need for a whole "infobox", but just replacing the URL with the title of the page (either by `<title>` or Open Graph) would go a long way.

Another solution would be to add a command that inserts the URL in the format I described above, but as an "already baked-in" markdown snippet. This may not reside in core Dendron and could exist in a separate extension - as of now though, I wasn't able to find any VS Code extension with this functionality.

**Describe alternatives you've considered**
I checked the VS Code extension Marketplace, and I could not find an extension that matches either of the above proposed solutions.

I could of course just copy and paste the title manually below the URL, but I feel that goes a bit against the kind of frictionless UX that I love Dendron for.

Thanks for your consideration!

kevinslin:
thanks for the feature request! i think having this makes sense and i appreciate that you took the time to list out specific solutions. 

based on the various implementations, i'm leaning towards being able to insert the url in a **pretty** format. 

one of the dendron [principles](https://dendron.so/notes/7fcebd7d-6411-4c9d-8baf-65629dc018a1.html#interoperability) is interoperability so having the entirety of the information in the text is preffered. 

this could be done as a separate command (eg. `Pretty URL`) and as an automatic formatting option when you're editing a document. 

going to mark this as a `good first task` and schedule it as a December milestone item if no one gets to this beforehand :)

kevinslin:
Changes to make:
- create `Dendron: Paste Link` command
- should read link from clipboard and attach metadata from open graph
[...]

PR #512, opened 2021-02-28
[feature][plugin-core][commands] Add PasteLink Command

hydrosquall:
# Motivation

- Reduce the effort needed to create a new note about a web resource. See #329 for more details.

# Changes

- Added the `open-graph-scraper` package + types
- Added and registered a `PasteLink` command without keyboard shortcuts (P and L are both taken, so I didn't want to overthink it). It searches the opengraph meta tags for "titles" to use in markdown
- See https://a.cl.ly/9Zu8yXgr for a demo. 

# Testing

- Not sure if testing this command is required since we just depends on a network request - should that be mocked? If not, do we need a test for just the "string building" capability?

## Development notes

<details>

<summary>(click to expand)</summary>

As this is my 1st Dendron PR, I saved notes about things I'm likely to look up someday / may be useful to other first time contributors.

- Commands to add new packages

```bash
yarn run lerna  add --dev @types/open-graph-scraper --scope @dendronhq/plugin-core
yarn run lerna  add open-graph-scraper --scope @dendronhq/plugin-core

# Run locally - make sure to do this before running the vscode "build" command.
./bootstrap/scripts/watch.sh
```
- Where to go to run a development version of VScode. (recommend doing so with other extensions disabled for better performance): https://a.cl.ly/wbu9L7qm. Note that you can "refresh" running instances rather than closing and re-opening previous windows (see lower rectangle for where you can do this).

- I didn't have `lerna` in my global namespace, so I edited the bootstrap file to be prefixed with `yarn run`, e.g. `yarn run lerna ...` rather than bare `lerna`. 

</details

### DIFF
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -494,6 +494,13 @@
                 "desc": "Generate diagnostics report",
                 "docLink": "dendron.topic.commands.md",
                 "docPreview": ""
+            },
+            {
+                "command": "dendron.pasteLink",
+                "title": "Dendron: Paste Link",
+                "desc": "Fetch metadata for a web link and format as markdown",
+                "docLink": "dendron.topic.commands.md",
+                "docPreview": ""
             }
         ],
         "keybindings": [
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -609,6 +609,7 @@
         "@types/mocha": "^7.0.2",
         "@types/node": "^13.11.0",
         "@types/open": "^6.2.1",
+        "@types/open-graph-scraper": "^4.7.0",
         "@types/semver": "^7.3.1",
         "@types/vscode": "^1.45.0",
         "@typescript-eslint/eslint-plugin": "^2.30.0",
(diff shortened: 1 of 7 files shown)


kevinslin (changes requested):
changes look great. can you also add a test for this?
you can use [this](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/CopyNoteUrl.test.ts#L39:L39) as a template.  testing instructions [here](https://dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#run-a-specific-test-for-dendron-plugin)

hydrosquall:
Hi @kevins8 ,

I took at pass at writing an integration test. It feels conceptually sound to me, but I'm missing something as it doesn't run as expected (https://a.cl.ly/mXu1v6zK )

I looked at a few different options for mocking `ogs` (https://github.com/sinonjs/sinon/issues/562#issuecomment-399090111 etc didn't seem to work), making an internal layer of indirection seemed preferable to installing a 3rd party dependency just to be able to mock a default export.

kevinslin (approved):
great job. excited to get this merged :)

PR #514, opened 2021-03-01
Changelog support

nikitavoloboev:


packages/engine-server/src/changelog/changelog.ts
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    const { stdout } = await execa(
+      "git",
+      [`log`, `--pretty=format:'%h'`, `-n`, `1`],
+      { cwd: path }
+    );
+    // use slice as there are quotes around the commit hash
+    commitHash = stdout.slice(1, -1);
+  } catch (error) {
+    console.log(error);
```
kevinslin, line 51: instead of logging error, doesn't this cause everything else to fail? should be a `throw` in that case

packages/engine-server/src/changelog/changelog.ts
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  // get files changed/added
+  try {
+    const { stdout } = await execa("git", ["show", "--name-status"], {
+      cwd: path,
+    });
+    let status = stdout.split("\n");
```
kevinslin, line 59: this might fail on windows which has a different newline delimeter

packages/engine-server/src/changelog/changelog.ts
```diff
@@ -0,0 +1,135 @@
...
+          filesChanged.push(filePath);
+          changes.push({
+            action: "Added",
+            fname: filePath,
+          });
+        }
+      }
+    });
+  } catch (error) {
+    console.log(error);
```
kevinslin, line 96: same comment as earlier, if there's an error, we don't want to move forward but do a hard fail

packages/engine-server/src/changelog/changelog.ts
```diff
@@ -0,0 +1,135 @@
...
+          encoding: "utf-8",
+        });
+      }
+    }
+    return changes;
+  });
+}
+
+// get files changed/added for a repo for the last commit
+async function getChanges(path: string, vaults: DVault[]) {
```
kevinslin, line 35: i wouldn't use `path` as a parameter since it conflicts with node's builtin `path` module

packages/engine-server/src/changelog/changelog.ts
```diff
@@ -0,0 +1,135 @@
...
+      try {
+        const { stdout } = await execa(
+          "git",
+          ["show", commitHash, "--", change.fname],
+          { cwd: path }
+        );
+        change.diff = Diff2Html.html(stdout);
+        return Diff2Html.html(stdout);
+      } catch (error) {
+        console.log(error);
```
kevinslin, line 110: hard fail on error

packages/engine-server/src/changelog/changelog.ts
```diff
@@ -0,0 +1,135 @@
...
+    const { stdout } = await execa("git", ["log", "-1", "--format=%cd"], {
+      cwd: path,
+    });
+    let date = stdout.split(/\s+/).slice(1, 5);
+    let day = date[0];
+    let month = date[1];
+    let year = date[3];
+    commitDate = `${day} ${month} ${year}`;
+  } catch (error) {
+    console.log(error);
```
kevinslin, line 127: hard fail

kevinslin (changes requested):

kevinslin (approved):

PR #515, opened 2021-03-02
Fix siteUrl typo in cli error throw output

ScriptAutomate:
Ran into this error today. Fixing the typo in the output.


kevinslin (approved):
thanks for the quick turnaround :)

PR #517, opened 2021-03-02
Clean up config after vault remove command

### LINKED ISSUE #505: Remove vault doesn't remove entry from payload in dendron.yml
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] UI
- [ ] Workbench
- [ ] Publishing
- [ ] Pods
- [ ] Markdown
- [ ] Hierarchy
- [ ] Graph
- [x] Misc (Choose this if your not sure)

## Describe the bug

I played with adding and removing vault and noticed that new vaults get added to 
```
duplicateNoteBehavior:
        action: useVault
        payload:
``` 
but nor removed when I delete the vault

## To Reproduce

1. Add a new local vault with `Vault Add`
2. Check that the new vault is added ot the list under payload in `dendron.yml`
3. Remove the vault
4. The entry is still present

## Expected behavior

The vault should be removed from the paylod list once it is removed

## Desktop (please complete the following information)

- OS: Win 10
- VSCode 1.53.2
- Dendron 0.30

hikchoi:
Closes #505 

- clean up `duplicateNoteBehavior` configuration after vault is removed
- test said fix
- move `stubVaultInput` to test utils. (Sorry if this is introducing noise to the PR, but I needed that stub to write the tests for my fixes 😅 )

### DIFF
--- a/packages/engine-server/src/workspace.ts
+++ b/packages/engine-server/src/workspace.ts
@@ -170,8 +170,10 @@ export class WorkspaceService {
     const config = this.config;
     config.vaults = _.reject(config.vaults, { fsPath: vault.fsPath });
 
-    if (!config.site.duplicateNoteBehavior) {
-    } else if (_.isArray(config.site.duplicateNoteBehavior.payload)) {
+    if (
+      config.site.duplicateNoteBehavior &&
+      _.isArray(config.site.duplicateNoteBehavior.payload)
+    ) {
       if (config.vaults.length == 1) {
         // if there is only one vault left, remove duplicateNoteBehavior setting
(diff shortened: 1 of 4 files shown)


kevinslin (changes requested):
This looks great. just one nitpick - we're aiming to consolidate on testing syntax and since most of the tests use `jest`, we're going to use the `expect().toEqual()` syntax everywhere instead of `deepStrictEqual`. you can see an example of it here: https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/PasteLink.test.ts#L8:L8

kevinslin (approved):

Issue #520, opened 2021-03-03
Consolidate testing convention


hikchoi:
## Context

Discussed [here](https://github.com/dendronhq/dendron/pull/517#pullrequestreview-602384709) that we should consolidate all test syntax to use jest nomenclature.

## Proposal

Find all use of bare `assert` and change it to use jest's [`expect`](https://jestjs.io/docs/en/expect)

kevinslin:
Agreed. Thanks for creating the issue :)

hikchoi:
@kevinslin 

<details>
<summary>These are all the bare asserts on the latest master branch</summary>
<p>

```
127 results - 28 files

root • packages/lsp-client/client/src/test/completion.test.ts:
  36  
  37:   assert.ok(actualCompletionList.items.length >= 2);
  38    expectedCompletionList.items.forEach((expectedItem, i) => {
  39      const actualItem = actualCompletionList.items[i];
  40:     assert.equal(actualItem.label, expectedItem.label);
  41:     assert.equal(actualItem.kind, expectedItem.kind);
  42    });

root • packages/lsp-client/client/src/test/diagnostics.test.ts:
  50  
  51:   assert.equal(actualDiagnostics.length, expectedDiagnostics.length);
  52  

  54      const actualDiagnostic = actualDiagnostics[i];
  55:     assert.equal(actualDiagnostic.message, expectedDiagnostic.message);
  56:     assert.deepEqual(actualDiagnostic.range, expectedDiagnostic.range);
  57:     assert.equal(actualDiagnostic.severity, expectedDiagnostic.severity);
  58    });

plugin-core • src/test/testUtilsv2.ts:
  431      toEqual: (value2: any) => {
  432:       assert.deepStrictEqual(value, value2);
  433      },
  434      toBeTruthy: () => {
  435:       assert.ok(value);
  436      },
  437      toBeFalsy: () => {
  438:       assert.ok(_.isUndefined(value) || !value);
  439      },

plugin-core • src/test/suite-integ/BacklinksTreeDataProvider.test.ts:
   61          const out = toPlainObject(await getChildren()) as any;
   62:         assert.strictEqual(
   63            out[0].command.arguments[0].path as string,

   65          );
   66:         assert.strictEqual(out.length, 1);
   67          done();

   92          const out = toPlainObject(await getChildren()) as any;
   93:         assert.strictEqual(
   94            out[0].command.arguments[0].path as string,

   96          );
[...]

PR #527, opened 2021-03-07
Consolidate testing convention

hikchoi:
closes: #520 

- Fix all bare asserts to jest-like expect
- Omit changes to
  - test harnesses
  - commented test codes that uses bare asserts
  - test code in lsp-client


hikchoi:
I disregarded the already failing/pending tests, but individually ran all other test files that already passed before the change.

kevinslin (approved):
this is great. thanks for taking of this :)

PR #607, opened 2021-04-05
Add CapitalizeSelection command

codetheweb:
As requested by @kpathakota and @kevinslin. 😃

### DIFF
--- a/packages/plugin-core/src/commands/CapitalizeSelectionCommand.ts
+++ b/packages/plugin-core/src/commands/CapitalizeSelectionCommand.ts
@@ -20,6 +20,10 @@ export class CapitalizeSelectionCommand extends BasicCommand<
     const maybeTextEditor = VSCodeUtils.getActiveTextEditor();
 
     if (maybeTextEditor) {
+      if (maybeTextEditor.selection.isEmpty) {
+        return;
+      }
+
       const text = maybeTextEditor.document.getText(maybeTextEditor.selection);
 
       const transformedText = text
--- a/packages/plugin-core/src/commands/CapitalizeSelectionCommand.ts
+++ b/packages/plugin-core/src/commands/CapitalizeSelectionCommand.ts
@@ -24,7 +24,7 @@ export class CapitalizeSelectionCommand extends BasicCommand<
 
       const transformedText = text
         .split(" ")
-        .map((token) => `${token[0].toUpperCase()}${token.slice(1)}`)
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/commands/CapitalizeSelectionCommand.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+  async execute() {
+    const maybeTextEditor = VSCodeUtils.getActiveTextEditor();
+
+    if (maybeTextEditor) {
+      const text = maybeTextEditor.document.getText(maybeTextEditor.selection);
+
+      const transformedText = text
+        .split(" ")
+        .map((token) => `${token[0].toUpperCase()}${token.slice(1)}`)
```
kevinslin, line 27: What would happen for unicode input? eg. 🌱

packages/plugin-core/src/test/suite-integ/CapitalizeSelectionCommand.test.ts
```diff
@@ -0,0 +1,49 @@
...
+          body: "hello world",
+        });
+      },
+      onInit: async ({ vaults, wsRoot }) => {
+        const notePath = path.join(wsRoot, vaults[0].fsPath, "gamma.md");
+        const editor = await VSCodeUtils.openFileInEditor(
+          vscode.Uri.file(notePath)
+        );
+
+        if (!editor) {
```
kevinslin, line 32: since this is a test, we can just assert editor using an `as` operator

packages/plugin-core/src/commands/CapitalizeSelectionCommand.ts
codetheweb: It's generally unicode safe, i.e.:

```
> '👍'.toUpperCase();
'👍'
```

It does look like there are [one or two very minor edge cases](https://stackoverflow.com/a/30763402/2129808) when dealing with other languages, so I changed it to toLocaleUpperCase() instead to be on the safe side: 2f3929be

packages/plugin-core/src/test/suite-integ/CapitalizeSelectionCommand.test.ts
codetheweb: 1949b152

codetheweb:
It's unclear why that one test is failing; I don't think it's related to anything I changed.
Says the snapshot is different but only for Node 14 on macOS which seems very odd.

kevinslin:
That test is known to be flaky. I need to change the snapshot so no worries.

PR #623, opened 2021-04-14
fix: go-down at root matches itself only

### LINKED ISSUE #619: "Go down" command doesn't work at the root level
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [x] UI
- [ ] Workbench
- [ ] Publishing
- [ ] Pods
- [ ] Markdown
- [x] Hierarchy
- [ ] Graph
- [ ] Misc (Choose this if your not sure)

## Describe the bug

When triggered at the root level, the "[Go down](https://wiki.dendron.so/notes/eea2b078-1acc-4071-a14e-18299fc28f47.html#go-down)" command opens lookup for `root` which matches only itself (unless there are other notes with "root" in the names).

## To Reproduce

Steps to reproduce the behaviour:

1. Create some notes
2. Open `root`
3. Press `Ctrl + Shift + 🔽`
4. Observe the lookup trying to match `root`

## Expected behaviour

"Go down" command should match direct children of the root.

## Screenshots

image
**Hierarchies**

image
**"Go Down" Lookup**
## Desktop (please complete the following information)

- OS: Win 10x64
- Dendron: 0.37.0

## Additional context

This is probably related to the `root` hierarchy being implicit for all the notes in a vault.

aleksey-rowan:
When the `go-down` command is executed at `root`, the value for the lookup is set to an empty string (will match everything) instead of `root.` (will match only itself and other notes with `root.` in the name):

https://github.com/dendronhq/dendron/blob/79580cdca8d34d2e8e0e57adc69811ddd8dcc2cc/packages/plugin-core/src/commands/GoDownCommand.ts#L22-L24

In the `LookupControllerV2`, there is a falsy check against the `quickPickValue` and when it matches the empty string, the filename of the active editor is used instead (which is `root.`):

https://github.com/dendronhq/dendron/blob/79580cdca8d34d2e8e0e57adc69811ddd8dcc2cc/packages/plugin-core/src/components/lookup/LookupControllerV2.ts#L272-L275

Seeing how `quickPickValue` is typed to be only `string` or `undefined`, I think the condition should explicitly check for `undefined` to preserve the empty string and let the `go-down` command match everything at the root level.

Closes #619

### DIFF
--- a/packages/plugin-core/src/components/lookup/LookupControllerV2.ts
+++ b/packages/plugin-core/src/components/lookup/LookupControllerV2.ts
@@ -269,7 +269,7 @@ export class LookupControllerV2 {
         break;
       }
       default:
-        if (quickPickValue) {
+        if (quickPickValue !== undefined) {
           onUpdateValue = quickPickValue;
         } else {
           let editorPath = vscode.window.activeTextEditor?.document.uri.fsPath;


kevinslin:
This is great. Thanks for the prompt fix!

PR #625, opened 2021-04-14
feat: add svg and rollup image

8brandon:


packages/dendron-design-system/package.json
```diff
@@ -18,8 +18,8 @@
     "prepare": "tsdx build",
     "size": "size-limit",
     "analyze": "size-limit --why",
-    "storybook": "start-storybook -p 6006 --no-manager-cache",
-    "build-storybook": "build-storybook",
+    "storybook": "start-storybook -s ./public -p 6006",
```
8brandon, line 21: so that storybook serves files from public folder

8brandon, line 21: so that storybook serves files from public folder

packages/dendron-design-system/tsdx.config.js
```diff
@@ -0,0 +1,8 @@
+import image from '@rollup/plugin-image';
+
+module.exports = {
+  rollup(config) {
+    config.plugins.push(image());
```
8brandon, line 5: https://www.npmjs.com/package/@rollup/plugin-image

8brandon: so that rollup can bundle assets such as svg

PR #634, opened 2021-04-16
Updated labler + templates to match new scenario-based areas

kpathakota:


.github/config/issue-labler/labler-v1.yml
```diff
@@ -1,21 +1,15 @@
...
-  - '- \[ ?(?:x|X) ?\] Workbench'
-area.publishing:
-  - '- \[ ?(?:x|X) ?\] Publishing'
-area.pods:
-  - '- \[ ?(?:x|X) ?\] Pods'
-area.markdown:
-  - '- \[ ?(?:x|X) ?\] Markdown'
-area.seeds:
-  - '- \[ ?(?:x|X) ?\] Seed'
+area.onboard:
```
flammehawk, line 1: You should not edit this file. 
Copy the content into `labler-v2.yml`

flammehawk (changes requested):
Since we don't want to invalidate the already existing issues with the new labler, 
you need to create the wanted Config inside a `labler-v2.yml`

And you need to change the corresponding Variable in the Templates. 
This block `issue_labeler_regex_version=1`will need to have the 1 changed to a two in each template.

.github/config/issue-labler/labler-v1.yml
kpathakota: ah, good to know, thanks.

kpathakota:
@flammehawk - hope the updated changes work now. Haven't figured out a way to test other than merging it in ... so would be good to know if you've got any tips there.

flammehawk (approved):
Looks all good to me.

flammehawk:
> @flammehawk - hope the updated changes work now. Haven't figured out a way to test other than merging it in ... so would be good to know if you've got any tips there.

The best bet to test would probably be to fork merge the changes and test it by creating an issue.

Issue #388, opened 2020-11-30
Create contributors page


kevinslin:
## Context

Dendron has lots of wonderful contributors. It's time to acknowledge all their hard work.

## Proposal

Make a contributions page to feature in the README similar to [this](https://github.com/foambubble/foam#contributors-)

kpathakota:
I think we should tackle this + an overall readme update next week.

kevinslin:
Agreed!

PR #651, opened 2021-04-20
bots for stale issues and issues with no response

kpathakota:


AleksueiR:
Seems like the stale bot is a bit too aggressive--even issues from the [Community Issue Board](https://github.com/orgs/dendronhq/projects/3) and ones like this https://github.com/dendronhq/dendron/issues/388 are getting marked as stale.

aleksey-rowan:
Seems like the stale bot is a bit too aggressive--even issues from the [Community Issue Board](https://github.com/orgs/dendronhq/projects/3) and ones like this https://github.com/dendronhq/dendron/issues/388 are getting marked as stale.

kevins8:
Agreed. Just updated the stalebot script to only act on issues with a
pre-defined tag

kpathakota:
Hmm... I think it's actually doing the right thing. We want to know/clean up issues that are getting too old. I think we can make it a little more lenient in terms of time but stalebot is meant to clean up old issues and the no-reply bot is intended to be used with a tag so we close issues that haven't gotten enough information. Will document this better in the wiki and we can revisit.

flammehawk:
Well, I think how it worked was good, but the label for Stale issues should be `status.stale` instead of `status.wont-fix`. 
`status.wont-fix` should only represent issues that aren't getting fixed. Stale does not mean that it won't get fixed, only that there was no work on it for a long time.

kevins8:
The issues being stale is no fault of the user but rather that we currently
lack the bandwidth to address the issues.
I think we should start with issues that need more information but haven't
been updated by the user. A smaller rollout also aligns to our iteration
principle :)
https://handbook.dendron.so/notes/b89ba854-72fb-4ebc-a8a0-55960b89e9dc.html#iteration

kpathakota:
Yeah, I'd like to build up to having a stale bot (which implies our backlog is groomed) but agreed for now to suspend it.

aleksey-rowan:
Renaming `wont-fix` to just `state` is a good idea as it will confuse people otherwise.

Also, I understand that the stale bot was to auto-close stale issues after a week if no further activity is recorded:

https://github.com/dendronhq/dendron/blob/6c45e6ed58db803a925c4bae4fa4532a1a6d492f/.github/stale.yml#L1-L4

I don't think the bot should indiscriminately close old issues just because no one commented on them for a while. Maybe the bot can only close issues that don't have any comments at all (or comments from the main contributors), or they don't have labels assigned, or upvotes, or some other indication that the issue is not only stale but moribund.

kpathakota:
@aleksey-rowan - I've suspended the bot for now so it doesn't auto-close. Community members can simply comment on the issue/remove the stale tag and the bot won't do anything - I still feel like this isn't too heavy a lift if the bot marks something as "stale". Once we get the backlog under control, we can consider re-enabling this.

PR #654, opened 2021-04-21
Updated label and removed restriction.

kpathakota:


kpathakota:
In case we do decide to enable the stale issue bot again, it will use the new label (status.stale).

Issue #485, opened 2021-02-13
Show progress bar when importing a remote vault


kevinslin:
**Is your feature request related to a problem? Please describe.**

Currently, when we add a remote vault, cloning happens in the background with no indication of progress.

**Describe the solution you'd like**

There should be a progress bar to show that progress is happening

**Additional context**

Code to add remote vault [here](https://github.com/dendronhq/dendron/blob/dev-kevin/packages/plugin-core/src/commands/VaultAddCommand.ts#L141:L141).
Example of progress bar [here](https://github.com/dendronhq/dendron/blob/dev-kevin/packages/plugin-core/src/commands/ImportPod.ts#L72:L72)

PR #668, opened 2021-04-25
Add progress to remote vault add

hikchoi:
This PR:
- Adds a progress bar when adding a remote vault.

closes: #485

### DIFF
--- a/packages/plugin-core/src/commands/VaultAddCommand.ts
+++ b/packages/plugin-core/src/commands/VaultAddCommand.ts
@@ -11,7 +11,7 @@ import { WorkspaceService } from "@dendronhq/engine-server";
 import fs from "fs-extra";
 import _ from "lodash";
 import path from "path";
-import { commands, QuickPickItem, window } from "vscode";
+import { commands, ProgressLocation, QuickPickItem, window } from "vscode";
 import { PickerUtilsV2 } from "../components/lookup/utils";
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -140,23 +140,39 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
-    await _.reduce(
-      vaults,
-      async (resp: any, vault: DVault) => {
-        await resp;
-        return this.addVaultToWorkspace(vault);
+    const vaults = await window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: "Adding remote vault",
+        cancellable: true,
```
kevinslin, line 147: if we want this to be cancellable, we need to pass in a token and add logic to actually cancel this operation. otherwise, i would just set it to false at this point

kevinslin (changes requested):

packages/plugin-core/src/commands/VaultAddCommand.ts
hikchoi: Ah okay. Completely blanked on that. Will change it to not be cancellable.

hikchoi: fixed :+1:

kevinslin (approved):

PR #669, opened 2021-04-26
feat: capitalize text command

hfellerhoff:
Changes made:
- Add command to capitalize text in all active selections
- Add tests

### DIFF
--- a/packages/plugin-core/src/commands/CapitalizeText.ts
+++ b/packages/plugin-core/src/commands/CapitalizeText.ts
@@ -1,0 +1,40 @@
+import { VSCodeUtils } from "../utils";
+import { window } from "vscode";
+import _ from "lodash";
+import { DENDRON_COMMANDS } from "../constants";
+import { BasicCommand } from "./base";
+
+type CommandOpts = {};
+
+type CommandInput = {};
+
+type CommandOutput = void;
+
+export class CapitalizeTextCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput
+> {
+  static key = DENDRON_COMMANDS.CAPITALIZE_TEXT.key;
+  async gatherInputs(): Promise<CommandInput | undefined> {
+    return {};
+  }
+  async execute() {
+    // Get active text editor
+    const textEditor = VSCodeUtils.getActiveTextEditor();
+    if (_.isUndefined(textEditor)) {
+      window.showErrorMessage("No active document found");
+      return;
+    }
+
+    await textEditor.edit((editBuilder) => {
+      textEditor.selections.forEach((selection) => {
+        // Get the text of the selection
+        const text = textEditor.document.getText(selection);
+        const capitalizedText = text.toUpperCase();
+
+        // Swap the old text with the capitalized text
+        editBuilder.replace(selection, capitalizedText);
+      });
+    });
+  }
+}
--- a/packages/plugin-core/src/commands/CapitalizeText.ts
+++ b/packages/plugin-core/src/commands/CapitalizeText.ts
@@ -1,5 +1,5 @@
 import { VSCodeUtils } from "../utils";
-import { window } from "vscode";
+import { window, env } from "vscode";
 import _ from "lodash";
 import { DENDRON_COMMANDS } from "../constants";
 import { BasicCommand } from "./base";
@@ -30,7 +30,7 @@ export class CapitalizeTextCommand extends BasicCommand<
       textEditor.selections.forEach((selection) => {
         // Get the text of the selection
         const text = textEditor.document.getText(selection);
-        const capitalizedText = text.toUpperCase();
+        const capitalizedText = text.toLocaleUpperCase(env.language);
 
         // Swap the old text with the capitalized text
         editBuilder.replace(selection, capitalizedText);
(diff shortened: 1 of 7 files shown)


flammehawk:
Question @hfellerhoff, could you please explain why this Command would need to be added by dendron? 
There are already many vscode plugins out there that do this and offer different Textmanipulations.

kevinslin:
@flammehawk this is part of Dendron's intern applications. it's not something that will be merged into Dendron 😅

flammehawk:
Okay, nevermind me asking then, but to understand it correctly @kevinslin 

> this is part of Dendron's intern applications

Do you mean Applications that get used internally by `Dendronhq`, or do you mean that it`s part of an application to become an Intern at `Dendronhg` xD

kpathakota:
> Do you mean Applications that get used internally by `Dendronhq`, or do you mean that it's part of an application to become an Intern at `Dendronhg` xD

Application to be an intern at DendronHQ :)

packages/plugin-core/src/commands/CapitalizeText.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (_.isUndefined(textEditor)) {
+      window.showErrorMessage("No active document found");
+      return;
+    }
+
+    await textEditor.edit((editBuilder) => {
+      textEditor.selections.forEach((selection) => {
+        // Get the text of the selection
+        const text = textEditor.document.getText(selection);
+        const capitalizedText = text.toUpperCase();
```
kevinslin, line 33: what would happen for different locale based text when running this?

packages/plugin-core/src/test/suite-integ/CapitalizeText.test.ts
```diff
@@ -0,0 +1,123 @@
+import { ENGINE_HOOKS } from "@dendronhq/common-test-utils";
+import _ from "lodash";
+import { ExtensionContext, Position, Selection } from "vscode";
+import { CapitalizeTextCommand } from "../../commands/CapitalizeText";
+import { VSCodeUtils } from "../../utils";
+import { expect } from "../testUtilsv2";
+import { runLegacyMultiWorkspaceTest, setupBeforeAfter } from "../testUtilsV3";
+
+suite("notes", function () {
```
kevinslin, line 9: can the name here be changed to the method under test?

packages/plugin-core/src/test/suite-integ/CapitalizeText.test.ts
```diff
@@ -0,0 +1,123 @@
...
+        const end = new Position(8, lowercaseText.length);
+        const selection = new Selection(start, end);
+
+        await editor.edit((editBuilder) => {
+          editBuilder.insert(start, lowercaseText);
+        });
+
+        await new CapitalizeTextCommand().run();
+
+        const changedText = editor.document.getText(selection);
```
kevinslin, line 36: this method name is confusing since the text isn't actually changed

packages/plugin-core/src/test/suite-integ/CapitalizeText.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+        expect(changedText).toEqual(uppercaseText);
+        done();
+      },
+    });
+  });
+
+  test("with multiple selections", (done) => {
```
kevinslin, line 77: nice! not many people would think to test for multiple selections :)

kevinslin (changes requested):

packages/plugin-core/src/commands/CapitalizeText.ts
hfellerhoff: The behavior may not be desired, and may not correctly transform the text depending on the language. One possible fix is to import `env` from `vscode` and refactor line 33 as:

```const capitalizedText = text.toLocaleUpperCase(env.language);```

 so that all languages are supported.

packages/plugin-core/src/test/suite-integ/CapitalizeText.test.ts
hfellerhoff: Yes - I'll update it shortly.

packages/plugin-core/src/test/suite-integ/CapitalizeText.test.ts
hfellerhoff: Would the name `CapitalizeSelectedTextCommand` be clearer?

packages/plugin-core/src/test/suite-integ/CapitalizeText.test.ts
hfellerhoff: Thanks! I use multiple selections in VSCode frequently, so it seemed like a natural use case to consider.

packages/plugin-core/src/commands/CapitalizeText.ts
hfellerhoff: One possible issue with this approach is if bilingual users attempt to use the command, as my understanding is that VSCode will prioritize their primary locale. However, this still supports a lot more users than the previous iteration.

kpathakota:
Congrats @hfellerhoff, welcome aboard :) Closing this one out.

Issue #614, opened 2021-04-08
Next/Previous sibling go out of order if more than 9


af4jm:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [x] UI
- [ ] Workbench
- [ ] Publishing
- [ ] Pods
- [ ] Markdown
- [x] Hierarchy
- [ ] Graph
- [ ] Misc (Choose this if your not sure)

## Describe the bug

with a set of files named x.1.md through x.n.md (where n >= 10), go through the siblings with Ctrl+Shift+[ or ] 

they go out of order, despite the order being correct in Workspace Explorer

Actual Order: 28, 29, 3, 30, 31

(I have 1 set that goes to 150... and if I start from 1, it goes 1, 10, 100, 101, 102 :(

## Expected behavior

28, 29, 30, 31

## Desktop (please complete the following information)

- OS: Win10
- Browser VSCode 1.55.1
- Version 0.36.2

kpathakota:
Yup, super clear no what needs to happen here and it is pretty annoying when navigating notes.

ognjenio:
I usually name files that I need in order with leading zeroes:
- 0001-file
- 0002-file

And for dates yyyy-mm-dd format works well.

kevinslin:
This is because we convert everything to string before sorting. We should have better detection for number only titles and sort by numbers instead

hikchoi:
@af4jm @ognjenio Just letting you know that the fix has been shipped in 0.40.
Now siblings with numeric names will be sorted correctly (i.e next sibling of 29 will be 30 not 3) 😃

af4jm:
just tried it & the command errors every time... and Dendron tree view still has the string sort order

hikchoi:
> just tried it & the command errors every time... 

I apologize. There seems to be a regression on the `Go Previous/Next Sibling` commands introduced in the latest update. I'll let you know once again when this is fixed.

> and Dendron tree view still has the string sort order

This issue (and the fix) does not address the dendron tree view sort order. Not exactly the same, but looks like there is an open issue on document display order in the tree view already : https://github.com/dendronhq/dendron/issues/440 
If the direction that issue is going doesn't work with you, could you open a separate issue for the dendron tree view sort order?

kevinslin:
@af4jm this has just been fixed in 0.42

af4jm:
cool, thx... also just left some thoughts in the tree view issue

PR #684, opened 2021-04-30
Sort numeric siblings in `GoToSiblingCommand`

hikchoi:
This PR:
- sorts sibling note file names so that they don't go out of order during `GoToSiblingCommand`.
- adds tests for above change.

closes: #614

### DIFF
--- a/packages/plugin-core/src/utils.ts
+++ b/packages/plugin-core/src/utils.ts
@@ -74,6 +74,15 @@ export class FileUtils {
   static escape(fpath: string) {
     return fpath.replace(/(\s+)/g, "\\$1");
   }
+
+  /**
+   * Given a full fname, return the leaf
+   * @param fname fname
+   * @returns name of leaf node
+   */
+  static getLeafNameFromFname(fname: string) {
+    return _.split(fname, ".").pop();
+  }
 }
 
 // NOTE: used for tests
--- a/packages/plugin-core/src/utils.ts
+++ b/packages/plugin-core/src/utils.ts
@@ -74,15 +74,6 @@ export class FileUtils {
   static escape(fpath: string) {
     return fpath.replace(/(\s+)/g, "\\$1");
   }
-
-  /**
-   * Given a full fname, return the leaf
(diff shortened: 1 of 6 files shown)


packages/plugin-core/src/utils.ts
```diff
@@ -74,6 +74,15 @@ export class FileUtils {
   static escape(fpath: string) {
     return fpath.replace(/(\s+)/g, "\\$1");
   }
+
+  /**
+   * Given a full fname, return the leaf
+   * @param fname fname
+   * @returns name of leaf node
+   */
+  static getLeafNameFromFname(fname: string) {
```
kevinslin, line 83: could you add this to [dnode](https://github.com/dendronhq/dendron/blob/master/packages/common-all/src/dnode.ts#L102:L102)? It'll be a good place to consolidate all node path methods

packages/plugin-core/src/utils/strings.ts
```diff
@@ -13,3 +13,8 @@ export const matchAll = (
 
   return out;
 };
+
+// https://stackoverflow.com/questions/18082/validate-decimal-numbers-in-javascript-isnumeric/1830844#1830844
+export const isNumeric = (n: any) => {
```
kevinslin, line 18: can we add this in [common-all](https://github.com/dendronhq/dendron/blob/master/packages/common-all/src/utils.ts#L6:L6)? the goal is to embed most of the logic into the more foundational libraries so that non-vscode clients (eg. web) will eventually be able to call into them as well

kevinslin (changes requested):

kevinslin:
> did I break the tests? 🤔

no, this was some snapshot tests breaking because of a change i made for the early seed. if you push with changes, the tests should reflect that

hikchoi:
I'll make the changes and let you know tomorrow morning my time :+1:

hikchoi:
utils moved to suggested location, rebased to master :+1:

kevinslin (approved):

Issue #355, opened 2020-11-16
Find / create missing linked notes


ognjenio:
**Is your feature request related to a problem? Please describe.**
Sometimes adding a link to a note before it exists is convenient. This is particular true for tags. While writing, I often add links to tags. Only after do I realize that they don't exist yet. This has resulted in many notes linking to tags that don't exist. 

To solve this I'd have to go through them one by one and click one each tag, and create missing files. 

**Describe the solution you'd like**
A way to more efficiently create linked notes -- there might be a few ways to do it:
1. When adding a link, if a note is not found, add a suggestion to create it
2. When running Doctor, create all notes that are linked but don't exist (or at least show a list of notes that are linked but don't exist)
3. Add a new command that can find all broken links

**Describe alternatives you've considered**
`grep` all `[[]]` and then looping through them in bash to create files.

kevinslin:
Great suggestion. To start off, I think adding this to Doctor (2) makes sense. I also like 1. 

We can do either opt in or opt out via config (eg. create a file if link doesn't exist and then notify user that a file has been created - this would happen on file save to avoid creating spurious files). 

Would the opposite of this also be useful? (eg. delete empty notes that have been around for longer than one day)

regoje:
I don't think immediately deleting would be good because it's destructive,
but having a search that lists all empty files would be very helpful.

Another extension accomplished functionality similar to search results by
opening a new file and dumping the search results there. I can't recall
which one at the moment though but it worked well.

hikchoi:
I'm working my way towards adding a doctor command action `CREATE_MISSING_LINKED_NOTES` to start with. Once I get that to work, I'll look into adding a config to enable this function on save.

chmac:
I think that this has been implemented, maybe this issue can be closed?

aleksey-rowan:
You are correct, @chmac, 
- Doctor now has the [createMissingLinkedNotes](https://wiki.dendron.so/notes/ZeC74FYVECsf9bpyngVMU/#createmissinglinkednotes) command that will create notes that are linked to but don't exist,
- and [findBrokenLinks](https://wiki.dendron.so/notes/ZeC74FYVECsf9bpyngVMU/#findbrokenlinks) will show a list of broken links

In workspaces with two or more vaults the above commands only work on links the vault prefix though, if you have a multi-vault setup and links without the `dendron://` prefix, it won't work.

@ognjenio, if you would still like an option to show a suggestion to create a note when adding a link, please open a new issue for that.

PR #713, opened 2021-05-11
Add Doctor command to create missing linked notes

hikchoi:
This PR:
- adds a new Doctor command that goes through wikilinks in all notes and create notes that are linked but does not exist yet.

related to: #355

### DIFF
--- a/packages/dendron-cli/src/commands/doctor.ts
+++ b/packages/dendron-cli/src/commands/doctor.ts
@@ -4,6 +4,7 @@ import {
   VaultUtils,
   DLink,
   NoteUtils,
+  DVault,
 } from "@dendronhq/common-all";
 import {
   DendronASTDest,
@@ -73,7 +74,7 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
   }
 
   getWildLinkDestinations(notes: NoteProps[], wsRoot: string) {
-    let wildWikiLinks = [] as DLink[];
+    let wildWikiLinks: DLink[] = [];
     _.forEach(notes, (note) => {
       const links = note.links;
       if (_.isEmpty(links)) {
@@ -90,22 +91,22 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
             vault: destVault,
             notes: notes,
             wsRoot: wsRoot,
-          }) as NoteProps;
+          });
           return !noteExists;
         })
       );
       return true;
     });
-    const uniqueCandidates = _.map(
+    const uniqueCandidates: NoteProps[] = _.map(
       _.uniqBy(wildWikiLinks, "to.fname"),
       (link) => {
         const destVault = link.to!.vault ? link.to!.vault : link.from.vault;
-        return {
-          fname: link.to!.fname,
-          vault: destVault,
-        };
+        return NoteUtils.create({
+          fname: link.to!.fname!,
+          vault: destVault!,
+        });
       }
-    ) as NoteProps[];
+    );
     return uniqueCandidates;
   }
 
--- a/packages/dendron-cli/src/commands/doctor.ts
+++ b/packages/dendron-cli/src/commands/doctor.ts
@@ -72,6 +72,43 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
     });
   }
 
+  getWildLinkDestinations(notes: NoteProps[], wsRoot: string) {
+    let wildWikiLinks = [] as DLink[];
+    _.forEach(notes, (note) => {
+      const links = note.links;
+      if (_.isEmpty(links)) {
+        return;
+      }
+      wildWikiLinks = wildWikiLinks.concat(
+        _.filter(links, (link) => {
+          if (link.type !== "wiki") {
+            return false;
+          }
+          const destVault = link.to!.vault ? link.to!.vault : note.vault;
+          const noteExists = NoteUtils.getNoteByFnameV5({
+            fname: link.to!.fname as string,
+            vault: destVault,
+            notes: notes,
+            wsRoot: wsRoot,
+          }) as NoteProps;
+          return !noteExists;
+        })
+      );
+      return true;
+    });
+    const uniqueCandidates = _.map(
+      _.uniqBy(wildWikiLinks, "to.fname"),
+      (link) => {
+        const destVault = link.to!.vault ? link.to!.vault : link.from.vault;
+        return {
+          fname: link.to!.fname,
+          vault: destVault,
+        };
+      }
+    ) as NoteProps[];
+    return uniqueCandidates;
+  }
+
   async enrichArgs(args: CommandCLIOpts): Promise<CommandOpts> {
     const engineArgs = await setupEngine(args);
     return { ...args, ...engineArgs };
@@ -96,6 +133,11 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
     let engineDelete = dryRun
       ? () => {}
       : throttle(_.bind(engine.deleteNote, engine), 300, {
+          leading: true,
+        });
+    let engineGetNoteByPath = dryRun
+      ? () => {}
+      : throttle(_.bind(engine.getNoteByPath, engine), 300, {
           leading: true,
         });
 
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -67,6 +96,71 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+        let wildWikiLinks = [] as DLink[];
+        _.forEach(notes, (note) => {
+          const links = note.links;
+          if (_.isEmpty(links)) {
+            return false;
+          }
+          const wsRoot = DendronWorkspace.wsRoot();
+          wildWikiLinks = wildWikiLinks.concat(
+            _.filter(links, (link) => {
+              if (link.type != "wiki") {
```
kevinslin, line 111: since its javascript, we want to use strict equality check `!==` (see [here](https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons))

packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -67,6 +96,71 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+              fname: link.to!.fname,
+              vault: link.from.vault,
+            };
+          }
+        ) as NoteProps[];
+        if (uniqueCandidates.length > 0) {
+          // show preview before creating
+          await this.showPreview(uniqueCandidates);
+          const options = ["proceed", "cancel"];
+          const shouldProceed = await VSCodeUtils.showQuickPick(options, {
```
kevinslin, line 140: check for `undefined` here, it means the user pressed escape in which case we should cancel the command

packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
```diff
@@ -266,3 +270,37 @@ suite("notes", function () {
...
+          wsRoot,
+        });
+        const cmd = new DoctorCommand();
+        sinon.stub(cmd, "gatherInputs").returns(
+          Promise.resolve({
+            action: DoctorActions.CREATE_MISSING_LINKED_NOTES,
+          })
+        );
+        await cmd.run();
+        done();
```
kevinslin, line 302: don't forget to check if the note is created

packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
```diff
@@ -266,3 +270,37 @@ suite("notes", function () {
...
+suite("CREATE_MISSING_LINKED_NOTES", function () {
+  let ctx: vscode.ExtensionContext;
+
+  ctx = setupBeforeAfter(this, {
+    afterHook: () => {
+      sinon.restore();
+    },
+  });
+
+  test.only("basic", (done) => {
```
kevinslin, line 283: don't forget to remove `.only` in the final pull request

packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
```diff
@@ -11,7 +15,7 @@ import { DoctorCommand } from "../../commands/Doctor";
 import { ReloadIndexCommand } from "../../commands/ReloadIndex";
 import { onWSInit, setupDendronWorkspace } from "../testUtils";
 import { expect } from "../testUtilsv2";
-import { setupBeforeAfter } from "../testUtilsV3";
+import { runLegacySingleWorkspaceTest, setupBeforeAfter } from "../testUtilsV3";
 
 suite("notes", function () {
```
kevinslin, line 20: this was my mistake but if you could change the name here to `DoctorComamndTest` :)

packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -67,6 +96,71 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
         });
         break;
       }
+      case DoctorActions.CREATE_MISSING_LINKED_NOTES: {
```
kevinslin, line 99: can we have the bulk of this logic in `DoctorCLICommand`? that way, it will be accessible in the CLI as well. for the CLI, we might add a `dry-run` flag in place of the preview

packages/plugin-core/src/commands/Doctor.ts
hikchoi: wouldn't `shouldProceed !== "proceed"` check for `undefined` as well? If I run press escape it will cancel the command and show the information message correctly.

packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
hikchoi: ah yeah. this test file isn't complete 😄  I pushed it with the code mid way

packages/plugin-core/src/commands/Doctor.ts
hikchoi: Do you think it's better to make this a separate command (like `BackfillV2Command` for `DoctorActions.FIX_FRONTMATTER` does)?

or should this be a cli-first command?

flammehawk: Just my two cents, but wouldn't it be better if cli, plugin, or API endpoint on the engine are just different front ends for the Command? 
Since it should be executable from the cli,  plugin, or an API call to the engine? 

The plugin could call the API to get it done, cli could execute it without needing the engine or call the engine and format the result differently for a console.

packages/plugin-core/src/commands/Doctor.ts
kevinslin: your right, i missed that!

packages/plugin-core/src/commands/Doctor.ts
kevinslin: the CLI, plugin and API endpoint are just different frontends for the command. 
the dendron engine handles the core operations (CRUD and a few extras) and it is accessed through the API.

just published an updated startup diagram [here](https://wiki.dendron.so/notes/c160ddce-edec-4f6e-841b-418d6030fa37.html)

hikchoi:
- moved the core logic to cli.
- handled case where multiple vaults had wild links.
- handled cross-vault wild link cases.
- added cli and plugin side tests for:
  - basic functionality
  - on cancel / dry run
  - multiple vault case
  - cross-vault case
  - aliased wild links
 
I was planning to add tsdoc comments for the new feature but I am deferring this to a separate PR to favor a timely delivery.

packages/dendron-cli/src/commands/doctor.ts
```diff
@@ -66,6 +72,43 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+          if (link.type !== "wiki") {
+            return false;
+          }
+          const destVault = link.to!.vault ? link.to!.vault : note.vault;
+          const noteExists = NoteUtils.getNoteByFnameV5({
+            fname: link.to!.fname as string,
+            vault: destVault,
+            notes: notes,
+            wsRoot: wsRoot,
+          }) as NoteProps;
```
kevinslin, line 93: don't cast it here because it could be undefined. the !noteExists should still be valid without a cast

packages/dendron-cli/src/commands/doctor.ts
```diff
@@ -66,6 +72,43 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
     });
   }
 
+  getWildLinkDestinations(notes: NoteProps[], wsRoot: string) {
+    let wildWikiLinks = [] as DLink[];
```
kevinslin, line 76: instead of casting, specify in assignment

```
let wildWikiLinks: DLink[] = [];
```

packages/dendron-cli/src/commands/doctor.ts
```diff
@@ -66,6 +72,43 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+    const uniqueCandidates = _.map(
+      _.uniqBy(wildWikiLinks, "to.fname"),
+      (link) => {
+        const destVault = link.to!.vault ? link.to!.vault : link.from.vault;
+        return {
+          fname: link.to!.fname,
+          vault: destVault,
+        };
+      }
+    ) as NoteProps[];
```
kevinslin, line 108: move type to var decalration

```
uniqueCandidates: NoteProps[] = _.map
```

kevinslin (changes requested):
nice work. just some typescript type fixes but otherwise, this looks great!

hikchoi:
pushed the suggested changes. Still a bit of resistance coming from mainly using a non-typed language for a few years 😅

kevinslin (approved):

PR #721, opened 2021-05-14
feat: add graphviz pod to export notes as a graph in dot language

hfellerhoff:
This PR adds a Graphviz pod, which takes all notes and exports them as a graph in [DOT language](https://graphviz.org/doc/info/lang.html).

Pod features include:
- Note export as a graph
- Ability to filter exported notes. Options include:
    - Hierarchical connections only: Export only connections based on `parent -> child` relationships
    - Link connections only: Export only connections based on `[[linked-note]] -> linked-note` relationships
    - Hybrid connections: Export all connections, prioritizing hierarchical connections


packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  beforeAll(() => {
+    exportDest = tmpDir().name;
+  });
+
+  test("basic", async () => {
+    await runEngineTestV5(
+      async ({ engine, vaults, wsRoot }) => {
+        const pod = new GraphvizExportPod();
+        engine.config.useFMTitle = true;
```
kevinslin, line 43: this should be true by default. is it not?

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+        await pod.execute({
+          engine,
+          vaults,
+          wsRoot,
+          config: {
+            dest: exportDest,
+            includeBody: true,
+            includeStubs: true,
+            includeHierarchy: true,
```
kevinslin, line 98: since we use `include*` to indicate additional metadata for a notes, can we use a different verb for toggling between graph views? for example `showGraphByHierarchy` and `showGraphByEdges`

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+          "graphviz.dot",
+        ]);
+        expect(expectedFiles).toEqual(actualFiles);
+
+        // check contents of graphviz file
+        const dotFile = fs.readFileSync(path.join(exportDest, "graphviz.dot"), {
+          encoding: "utf8",
+        });
+
+        // TODO: IDs of files switch between test runs, which causes the snapshot
```
kevinslin, line 69: since we don't have a snapshot, it'll be good to have a comment at the top with how the basic graphs look like to make this easier to debug in the future

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+            dest: exportDest,
+            includeBody: true,
+            includeStubs: true,
+            includeHierarchy: false,
+            includeLinks: true,
+          } as GraphvizExportConfig,
+        });
+
+        // check contents of graphviz file
+        const foo = fs.readFileSync(path.join(exportDest, "graphviz.dot"), {
```
kevinslin, line 135: more descriptive name here

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+          encoding: "utf8",
+        });
+
+        // check for link-specific elements
+        await checkString(foo, "--", "dotted");
```
kevinslin, line 140: this looks same for hierarchy test. can we do two checks here? one for something that is only available in edges and one that is only available for hierarchies (the latter should evaluate to false)

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+          } as GraphvizExportConfig,
+        });
+
+        // check contents of graphviz file
+        const foo = fs.readFileSync(path.join(exportDest, "graphviz.dot"), {
+          encoding: "utf8",
+        });
+
+        // check for both hierarchical and link elements
+        await checkString(foo, "--", "dotted");
```
kevinslin, line 170: include test that would pass for both hierarchy and links

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+            dest: exportDest,
+            includeBody: true,
+            includeStubs: true,
+            includeHierarchy: true,
+            includeLinks: true,
+          } as GraphvizExportConfig,
+        });
+
+        // check contents of graphviz file
+        const foo = fs.readFileSync(path.join(exportDest, "graphviz.dot"), {
```
kevinslin, line 165: update name

packages/pods-core/src/builtin/GraphvizPod.ts
```diff
@@ -0,0 +1,161 @@
+import fs from "fs-extra";
+import _ from "lodash";
+import { DLink, NoteProps, NoteUtils } from "@dendronhq/common-all";
+import path from "path";
+import { ExportPod, ExportPodPlantOpts } from "../basev3";
+import { ExportPodConfig } from "../basev3";
+
+const ID = "dendron.graphviz";
+
+interface ParentDictionary {
```
kevinslin, line 10: prefer using `type` to `interface`

- types and interfaces are mostly interchangable, `types` are a bit more flexible and so we've standardized on using it
- see official typescript docs [here](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces)

kevinslin (changes requested):
good job. left some comments, mostly style and convention related. for this change, it would also be great to have a pull request submitted for the documentation. not a blocker to merge the code but should be ready when we ship this

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
hfellerhoff: To be perfectly honest, I kept this as it was the way a similar test was implemented. I can go ahead and remove it

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
hfellerhoff: Sounds good, I'll do that. Is there an easy way to maintain IDs between test runs so that we could use a snapshot here?

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
kevinslin: it wasn't default once upon a time ago. we should probably update the old tests :)

hfellerhoff: Would you like me to take a look at the other engine tests and update them accordingly?

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
hfellerhoff: My initial reason for doing it this way is that at the moment, the config allows for exporting hierarchical and link connections at the same time. I like the clarity of those new names, I just wonder if there would be confusion in the event that both were marked true

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
hfellerhoff: The current issue with this is the fact that linked notes are currently notated by a "dotted" line, whereas hierarchical connections are noted as a normal solid line. The "dotted" style is added on top of the base connection style, but nothing is added for hierarchical connections at the moment. I can look into a better of way of denoting the two so the difference is clearer

kevinslin: in that case can we check for the non-existence of the "dotted" line?

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
kevinslin: got it. in that case, we can go either way on the `include` syntax but we should clarify what we mean by hierarchy and links. if we do stick with the `include`, I would change it to `includeHierarchyLinks` and `includeEdgeLinks`

packages/engine-test-utils/src/__tests__/pods-core/GraphvizPod.spec.ts
kevinslin: so not yet - i'm thinking of making that an option to `runEngine`. currently, the non-root nodes have stable ids in the test but the root nodes do not

PR #738, opened 2021-05-17
feat: Workspace Sync command

SeriousBug:
This pull request completes the existing Sync command stub, and adds tests for it. It also makes a few small changes to `pullVaults` and `pushVaults` functions to make testing easier.


kevinslin (changes requested):
this looks great. could you also make an addition to `GitTestUtils` to add a `createRepoForRemoteWorkspace` method that creates a repo with the remote set to a local tmp dir?

we can use that to test the `push` case

SeriousBug:
@kevinslin Added the push test.

kevinslin (approved):

PR #741, opened 2021-05-18
triaged bot update

kpathakota:
Had an irritating issue happen every time I triaged something but then updated the issue, the untriaged label would get added to it again. Figured this would be a good way to address it.


kpathakota:
If you've got a better idea for this @flammehawk, I'm all ears but I think this should work okay for our purposes.

PR #742, opened 2021-05-19
feat: Basic block reference link functionality

SeriousBug:
This pull request adds basic functionality for linking to block anchors. This allows links to manually inserted block anchors to function. For example, with the following files:

In `note-alpha.md`:
```
...
Lorem ipsum dolor amet

^my-block-anchor
```

In `note-beta.md`:
```
...
Fuga voluptas veniam ab

[[note-alpha#^my-block-anchor]]
```

Here, using the link in `note-beta` should take the user to `note-alpha` at the line of the block anchor.

The integration into VSCode is not there yet, so the block anchor must be inserted manually, and the anchor portion of the link must be written manually. It should work to place the anchors anywhere in the document, with the only limitations are: they must only contain letters, numbers, dashes, or underscores; they must appear at the end of a line or on a separate line by themselves.

### DIFF
--- a/packages/engine-server/src/markdown/remark/blockAnchors.ts
+++ b/packages/engine-server/src/markdown/remark/blockAnchors.ts
@@ -50,7 +50,7 @@ function attachParser(proc: Unified.Processor) {
         type: "blockAnchor",
         value,
         id: match[1],
-      } as BlockAnchor);
+      });
     }
     return;
   }
--- a/packages/engine-server/src/markdown/remark/blockAnchors.ts
+++ b/packages/engine-server/src/markdown/remark/blockAnchors.ts
@@ -66,10 +66,10 @@ function attachParser(proc: Unified.Processor) {
 function attachCompiler(proc: Unified.Processor, _opts?: CompilerOpts) {
   const Compiler = proc.Compiler;
   const visitors = Compiler.prototype.visitors;
-  let { dest } = MDUtilsV4.getDendronData(proc);
 
   if (visitors) {
     visitors.blockAnchor = function (node: BlockAnchor) {
+      const { dest } = MDUtilsV4.getDendronData(proc);
       switch (dest) {
         case DendronASTDest.MD_DENDRON:
         case DendronASTDest.MD_REGULAR:
--- a/packages/engine-server/src/markdown/remark/blockAnchors.ts
+++ b/packages/engine-server/src/markdown/remark/blockAnchors.ts
@@ -1,0 +1,90 @@
+import _ from "lodash";
+import { Eat } from "remark-parse";
+import Unified, { Plugin } from "unified";
+import { BlockAnchor, DendronASTDest } from "../types";
+import { MDUtilsV4 } from "../utils";
+
+// Letters, digits, dashes, and underscores.
+// The underscores are an extension over Obsidian.
+export const LINK_REGEX = /^\^([\w-]+)$/;
+export const LINK_REGEX_LOOSE = /\^([\w-]+)/;
+
+/**
+ *
+ * @param text The text to check if it matches an block anchor.
+ * @param matchLoose If true, a block anchor anywhere in the string will match. Otherwise the string must contain only the anchor.
+ * @returns The identifier for the match block anchor, or undefined if it did not match.
+ */
+export const matchBlockAnchor = (
+  text: string,
+  matchLoose: boolean = true
+): string | undefined => {
+  const match = (matchLoose ? LINK_REGEX_LOOSE : LINK_REGEX).exec(text);
+  if (match && match.length == 1) return match[1];
+  return undefined;
+};
+
+type PluginOpts = {};
+
+type CompilerOpts = {};
+
+const plugin: Plugin<[CompilerOpts?]> = function (
+  this: Unified.Processor,
+  opts?: PluginOpts
+) {
+  attachParser(this);
+  if (this.Compiler != null) {
+    attachCompiler(this, opts);
+  }
+};
+
+function attachParser(proc: Unified.Processor) {
+  function locator(value: string, fromIndex: number) {
+    return value.indexOf("^", fromIndex);
+  }
+
+  function inlineTokenizer(eat: Eat, value: string) {
+    const match = LINK_REGEX.exec(value);
+    if (match) {
+      return eat(match[0])({
+        type: "blockAnchor",
+        value,
+        id: match[1],
+      } as BlockAnchor);
+    }
+    return;
+  }
+  inlineTokenizer.locator = locator;
+
+  const Parser = proc.Parser;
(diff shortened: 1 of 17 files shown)


SeriousBug:
This also adds same file links, both for headers and anchors. These links look like `[[#header]]` or `[[#^block]]`, where it's implied that it's referring to a header or anchor in the current file. This is something Obsidian uses (at least for block references), and I think it makes sense to add support to Dendron as well.

SeriousBug:
@kevinslin Could you have a look at the issue with imports? I'm not seeing it when building on my end, `yarn bootstrap:build` shows no errors.

packages/engine-server/src/markdown/remark/blockAnchors.ts
```diff
@@ -0,0 +1,90 @@
+import _ from "lodash";
+import { Eat } from "remark-parse";
+import Unified, { Plugin } from "unified";
+import { BlockAnchor, DendronASTDest } from "../types";
+import { MDUtilsV4 } from "../utils";
+
+// Letters, digits, dashes, and underscores.
+// The underscores are an extension over Obsidian.
+export const LINK_REGEX = /^\^([\w-]+)$/;
```
kevinslin, line 9: since we use LINK_REGEX in multiple places, can we rename to BLOCK_LINK_REGEX?

packages/engine-server/src/markdown/remark/blockAnchors.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  if (visitors) {
+    visitors.blockAnchor = function (node: BlockAnchor) {
+      const { dest } = MDUtilsV4.getDendronData(proc);
+      switch (dest) {
+        case DendronASTDest.MD_DENDRON:
+        case DendronASTDest.MD_REGULAR:
+        case DendronASTDest.MD_ENHANCED_PREVIEW:
+        case DendronASTDest.HTML: {
+          // Anything more to do here? Can we embed HTML into the preview?
```
kevinslin, line 78: yep, HTML can be embedded in the preview. that would be how I would implement blocks

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -138,7 +141,26 @@ export class LinkUtils {
     };
   }
 
-  static parseLinkV2(linkString: string) {
+  /**Either value or anchorHeader will always be present if the function did not
```
kevinslin, line 144: nitpick: space btw * and E

packages/engine-test-utils/src/__tests__/engine-server/markdown/blockAnchors.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  return { ...opts } as any;
+}
+
+/** Gets the descendent (child, or child of child...) node of a given node.
+ *
+ * @param node The root node to start descending from.
+ * @param indices Left-to-right indexes for children, e.g. first index is for the root, second is for the child of the root...
+ * @returns Requested child. Note that this function has no way of checking types, so the child you get might not be of the right type.
+ */
+function getDescendentNode<Child extends DendronASTNode>(
```
kevinslin, line 34: maybe rename to `getLeafNode`? that's a more widely used term for this

packages/engine-test-utils/src/__tests__/engine-server/markdown/blockAnchors.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+describe("blockAnchors", () => {
+  describe("parse", () => {
+    let engine: any;
+    let dendronData = {
+      fname: "placeholder.md",
+      dest: DendronASTDest.MD_REGULAR,
+    };
+
+    test("parses anchor by itself", () => {
+      const resp = proc(engine, genDendronData(dendronData)).parse(
```
kevinslin, line 61: I would add the following test utility which can return a *proper* type. this way, you can pass in an `DendronASTNode` to `getDescendentNode` and avoid the `@ts-ignore` statements
```ts
function invokeDendronProc(proc: Processor, text: string): DendronASTNode {
  return proc.parse(text) as DendronASTNode
}
```

packages/engine-test-utils/src/__tests__/engine-server/markdown/blockAnchors.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      );
+      expect(getBlockAnchor(resp).type).toEqual("inlineCode");
+    });
+
+    test("doesn't parse code block not at the end of the line", () => {
+      const resp = proc(engine, genDendronData(dendronData)).parse(
+        "^block-id Lorem ipsum"
+      );
+      expect(getDescendentNode(resp, 0, 0).type).toEqual("text");
+      // @ts-ignore
```
kevinslin, line 80: see previous comment

packages/engine-test-utils/src/__tests__/engine-server/markdown/blockAnchors.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      expect(resp.children[0].children.length).toEqual(1); // text only, nothing else
+    });
+
+    test("parses anchors at the end of the line", () => {
+      const resp = proc(engine, genDendronData(dendronData)).parse(
+        "Lorem ipsum ^block-id"
+      );
+      // @ts-ignore
```
kevinslin, line 88: see previous comment

kevinslin (changes requested):
awesome work. just a few comments

packages/engine-test-utils/src/__tests__/engine-server/markdown/blockAnchors.spec.ts
SeriousBug: I'll argue that "descendant" is the right term here. The node that gets selected won't necessarily be a leaf node, it may have children of it's own. It's a relatively widely used term as well, here are a few examples of the term being used:

https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator

https://api.jquery.com/descendant-selector/

I can change it of course if you don't agree that it's appropriate.

SeriousBug:
@kevinslin I made the changes you suggested, minus one I explained above. Let me know if you'd like me to change that as well.

packages/engine-test-utils/src/__tests__/engine-server/markdown/blockAnchors.spec.ts
kevinslin: your right, misread the function. this is good, thanks for the examples :)

kpathakota:
@all-contributors please add @SeriousBug for code

kpathakota:
Take 2: @all-contributors please add @SeriousBug for code

PR #762, opened 2021-05-25
enhance: Track anchors in note properties for efficient anchor lookups

SeriousBug:
This tracks both headers and block anchors.


packages/common-all/src/error.ts
```diff
@@ -115,3 +115,33 @@ export const error2PlainObject = (err: IDendronError): DendronErrorPlainObj => {
...
+ *       else if (name === "baz") { ... }
+ *       else assertUnreachable(name);
+ *     }
+ *
+ * Let's say someone changes the type Names to `type Names = "bar" | "baz" | "ham";`. Thanks to this
+ * assertion, the compiler will warn them that this branch is now reachable, and something is wrong.
+ *
+ * @param x
+ */
+export function assertUnreachable(_never: never): never {
```
kevinslin, line 142: nice addition. consider adding this to [[dendron.dev.style]] in `dendron-site`

packages/common-all/src/types/foundation.ts
```diff
@@ -45,6 +46,10 @@ export type DNodeProps<T = any, TCustom = any> = {
    * Node links (eg. backlinks, wikilinks, etc)
    */
   links: DLink[];
+  /**
+   * Anchors within the node (headings, block anchors)
+   */
+  anchors: { [index: string]: DNoteAnchorPositioned | undefined };
```
kevinslin, line 52: when would it be useful for an anchor to be undefined?

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -216,6 +219,47 @@ export class LinkUtils {
...
+            value,
+            line: line - 1,
+            column: column - 1,
+          },
+        ]);
+      } else if (anchor.type === DendronASTTypes.BLOCK_ANCHOR) {
+        anchors.push([
+          `^${anchor.id}`,
+          {
+            type: "block",
```
kevinslin, line 249: we should be able to use the value of `DendronASTTypes.BLOCK_ANCHOR` here - makes it easier to find and refactor

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -216,6 +219,47 @@ export class LinkUtils {
...
+    wsRoot: string;
+  }): Promise<{ [index: string]: DNoteAnchorPositioned }> {
+    if (opts.note.stub) return {};
+    const noteContents = await NoteUtils.readFullNote(opts);
+    const noteAnchors = RemarkUtils.findAnchors(noteContents);
+    const anchors: [string, DNoteAnchorPositioned][] = [];
+    noteAnchors.forEach((anchor) => {
+      if (_.isUndefined(anchor.position)) return;
+      const { line, column } = anchor.position.start;
+      if (anchor.type === "heading") {
```
kevinslin, line 234: to be consistent, we might just add this as a `DendronASTTypes` of heading

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -216,6 +219,47 @@ export class LinkUtils {
...
+        const value = getSlugger().slug(anchor.children[0].value as string);
```
kevinslin, line 235: we should instantiate `getSlugger` at the top. github slugger will add the correct suffix to the header if it encounters duplicates

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -260,6 +304,15 @@ export class RemarkUtils {
     return selectAll("blockAnchor", parsed) as BlockAnchor[];
   }
 
+  static findAnchors(content: string): Anchor[] {
+    const parser = MDUtilsV4.remark().use(blockAnchors);
+    const parsed = parser.parse(content);
+    return [
+      ...(selectAll("heading", parsed) as Heading[]),
+      ...(selectAll("blockAnchor", parsed) as BlockAnchor[]),
```
kevinslin, line 312: use the `DendronASTTypes` enum instead of the raw value here

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -31,32 +27,20 @@ export { CommandOpts as GotoNoteCommandOpts };
...
-      return getSlugger().slug(h.children[0].value as string) === anchorSlug;
-    });
-  }
+export const findAnchorPos = (opts: {
+  anchor: DNoteAnchor;
+  note: NoteProps;
+}): Position => {
+  const { anchor: findAnchor, note } = opts;
+  let key: string;
+  if (findAnchor.type === "header") key = getSlugger().slug(findAnchor.value);
```
kevinslin, line 36: we don't currently handle it today but we should be keeping track of anchor values so that we can add the proper suffix for duplicate headers. can we split that out in a separate task and issue?

kevinslin (changes requested):

packages/common-all/src/types/foundation.ts
SeriousBug: Lookups from the anchors have whatever type you give them, even though the key could be missing. i.e.:

```
const testMap: {[index: string]: string} = {};
const x: string = testMap["doesnt-exist"];
```

x has the type `string` even though the actual value can be (and is) `undefined`. So I added `undefined` to the type to reflect this. This seems to be a limitation of TypeScript.

packages/engine-server/src/markdown/remark/utils.ts
SeriousBug: The values for the AST types and anchor types are different though, `blockAnchor` and `heading` vs. `block` and `header`. Should I change them to match the AST type names? Or I can create an enum for anchor types if we just want to avoid magic numbers.

packages/engine-server/src/markdown/remark/utils.ts
SeriousBug: Added `HEADING` to `DendronASTTypes` and revised existing code to use it.

packages/plugin-core/src/commands/GotoNote.ts
SeriousBug: Created #768.

packages/common-all/src/types/foundation.ts
kevinslin: got it. makes sense

packages/engine-server/src/markdown/remark/utils.ts
kevinslin: unless there's a good reason not to, I would just have the names be the same. its already the case for `wikiLink` and the other dendron constructs

PR #766, opened 2021-05-26
feat: Add block anchor HTML generation

SeriousBug:


packages/engine-server/src/markdown/remark/blockAnchors.ts
```diff
@@ -65,28 +68,37 @@ function attachParser(proc: Unified.Processor) {
...
 
-function attachCompiler(proc: Unified.Processor, _opts?: CompilerOpts) {
+function attachCompiler(proc: Unified.Processor, _opts?: PluginOpts) {
   const Compiler = proc.Compiler;
   const visitors = Compiler.prototype.visitors;
 
   if (visitors) {
-    visitors.blockAnchor = function (node: BlockAnchor) {
+    visitors.blockAnchor = function (node: BlockAnchor): string | Element {
+      debugger;
```
kevinslin, line 77: remove `debugger`

packages/engine-server/src/markdown/remark/blockAnchors.ts
```diff
@@ -65,28 +68,37 @@ function attachParser(proc: Unified.Processor) {
...
       switch (dest) {
         case DendronASTDest.MD_DENDRON:
         case DendronASTDest.MD_REGULAR:
         case DendronASTDest.MD_ENHANCED_PREVIEW:
-        case DendronASTDest.HTML: {
-          // Anything more to do here? Can we embed HTML into the preview?
-          // Prints ^{\^block-id} so that it gets rendered small.
-          return `^\{\\^${node.id}\}`;
-        }
+        case DendronASTDest.HTML:
```
kevinslin, line 83: this is legacy from days past when we used to use `remark` to render HTML`. this should be a dead end at this point since rendering HTML means using rehype

packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -176,7 +177,7 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
         }
       }
       if (
-        node.type === "refLinkV2" &&
+        node.type === DendronASTTypes.REF_LINK_V2 &&
```
kevinslin, line 180: nice catch

packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -208,6 +209,12 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
           }
         }
       }
+      if (node.type === DendronASTTypes.BLOCK_ANCHOR) {
+        const procOpts = proc.data("procOpts") as any;
```
kevinslin, line 213: we should use `getProcOpts` [method](https://github.com/dendronhq/dendron/blob/integ-publish/packages/engine-server/src/markdown/utils.ts#L198:L198) for this

kevinslin (changes requested):

PR #781, opened 2021-06-01
Add redesigned note and schema graph views

hfellerhoff:
This pull request is currently **Work in Progress 🚧**

Changes/Additions:
- new Next.js routes for note and schema graphs (`graph` and `graph-schema`)
- new Next.js component for rendering note/schema graphs using Cytoscape.js
- new commands to show these graphs within VSCode (`Show Note Graph` and `Show Schema Graph`)


kevinslin (changes requested):
it looks like there are other changes in the pull request that came from an old version of master. can you make sure this is rebased from master and then submit a pull request against master?

PR #782, opened 2021-06-01
feat: Block reference support in HTML

SeriousBug:
This pull request adds support for block references in HTML. It includes support for:

* Single block anchor references `![[note#^anchor]]`
* Range block anchor references `![[note#^anchor1:#^anchor2]]`
* Same file block references `![[#^anchor]]` and `![[#header]]`

Range block anchor references are allowed to be mixed with headers too, so it's possible to do `![[#header:#^anchor]]` or `![[#^anchor:#header]]` to reference only a part of a section.

It's also possible to reference single items in lists, or parts of lists. Consider the following list

```
* Aut id architecto quia.
* Sapiente sed accusamus eum.
  * Ullam optio est quia. ^one
* Reprehenderit doloribus.
  * Sint minus fuga omnis non. ^two
  * Iure neque alias dolorem. ^three
```

The reference `![[#^one]]` would only reference the list element it's next to. `![[#^two:#^three]]` would reference both elements, and anything in between.

An interesting case is something like `![[#^one:#^two]]` where the referenced items are nested in different levels of the list. In this case, the reference refers to anything after the start anchor and before the end anchor, and also the parent of the start anchor up to a common ancestor with the end. In other words, the output looks like:

```
* Sapiente sed accusamus eum.
  * Ullam optio est quia. ^one
* Reprehenderit doloribus.
  * Sint minus fuga omnis non. ^two
```

This allows you to refer to regions of lists across different depths while preserving the required context for the items you selected within the region, and avoids ending up with a broken list.

### DIFF
--- a/packages/engine-server/src/markdown/remark/noteRefsV2.ts
+++ b/packages/engine-server/src/markdown/remark/noteRefsV2.ts
@@ -14,6 +14,7 @@ import {
 } from "@dendronhq/common-all";
 import { file2Note } from "@dendronhq/common-server";
 import _ from "lodash";
+import { List } from "mdast";
 import { html, paragraph, root } from "mdast-builder";
 import { Eat } from "remark-parse";
 import Unified, { Plugin } from "unified";
@@ -403,10 +404,12 @@ function convertNoteRefHelperAST(
 
   // TODO: can i just strip frontmatter when reading?
   let anchorStartIndex = bodyAST.children[0].type === "yaml" ? 1 : 0;
+  let secondaryStartIndex: number | undefined;
   let anchorEndIndex = bodyAST.children.length;
+  let secondaryEndIndex: number | undefined;
 
   if (anchorStart) {
-    anchorStartIndex = findAnchor({
+    [anchorStartIndex, secondaryStartIndex] = findAnchor({
       nodes: bodyAST.children,
       match: anchorStart,
     });
@@ -417,7 +420,7 @@ function convertNoteRefHelperAST(
   }
 
   if (anchorEnd) {
-    anchorEndIndex = findAnchor({
+    [anchorEndIndex, secondaryEndIndex] = findAnchor({
       nodes: bodyAST.children.slice(anchorStartIndex + 1),
       match: anchorEnd,
     });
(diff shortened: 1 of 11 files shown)


packages/engine-test-utils/src/__tests__/engine-server/markdown/noteRefv2.spec.ts
```diff
@@ -684,11 +742,836 @@ describe("noteRefV2", () => {
...
+            "Minima expedita vero.",
+            "Iure neque alias dolorem."
+          );
+        },
+      },
+    });
+
+    runAllTests({
+      name: "compile",
+      testCases: [
```
kevinslin, line 1556: can you also add a test case for what happens when we have a nested block reference link?

kevinslin: also a case when a block reference link is referencing to an invalid anchor

packages/engine-test-utils/src/__tests__/engine-server/markdown/remarkUtils.spec.ts
```diff
@@ -5,27 +5,61 @@ import {
...
+        }
+      );
+    });
+    test("one block anchor", async () => {
+      await runEngineTestV5(
+        async ({ engine }) => {
+          const body = engine.notes["foo"].body;
+          debugger;
+          const out = RemarkUtils.findAnchors(body);
+          expect(out).toMatchSnapshot("bond");
```
kevinslin, line 49: "bond" is a debug value to help quickly find a given snapshot. you can take a snapshot without parameters. if you see other snapshots using this value, please remove them

kevinslin (changes requested):

packages/engine-test-utils/src/__tests__/engine-server/markdown/noteRefv2.spec.ts
SeriousBug: > can you also add a test case for what happens when we have a nested block reference link?

Do you mean have a note with a reference to another note, that also has a reference? E.g. parsed note has `![[note1#^start:#^end]]`, then note1 has `![[note2#^start:#^end]]`?

SeriousBug:
@kevinslin Added the tests you suggested, and caught a bug in the process.

packages/engine-server/src/markdown/remark/noteRefsV2.ts
```diff
@@ -375,11 +377,150 @@ export function convertNoteRefASTV2(
...
+  };
+  let end: FindAnchorResult = null;
+
+  if (anchorStart) {
+    start = findAnchor({
+      nodes: bodyAST.children,
+      match: anchorStart,
+    });
+    if (_.isNull(start)) {
+      debugger;
```
kevinslin, line 465: remove debugger

kevinslin (changes requested):
nice! yeah, i remember running into this when doing the original note ref impl. just remove the debug statement and we should be good to go

SeriousBug:
@kevinslin Done, thanks!

kevinslin (approved):

PR #783, opened 2021-06-02
enhance: Improve startup speed by parallelizing schema loading

SeriousBug:
I tested this change on our org repo. With the schema files in disk cache, this is giving me around 5% improvement. The improvement should be much higher in real world conditions when loading from cold (files not in disk cache), with slow disks, and for users with lots of schemas.


kevinslin (approved):
nice optimization :)

PR #786, opened 2021-06-03
Check in pre-commit against `debugger;` statements

SeriousBug:
This is just a suggestion. Personally, I keep forgetting to take `debugger;` statements out when committing, so I would find this useful. I tested it with a commit and can confirm that it works.

### DIFF
--- a/hooks/common.js
+++ b/hooks/common.js
@@ -1,0 +1,36 @@
+const fs = require('fs');
+const path = require('path');
+const chalk = require('chalk');
+
+function checkToken({ filesToCheck, forbiddenTokens }) {
+  let status = 0;
+  for (let [term, value] of Object.entries(forbiddenTokens)) {
+    const { rgx, fileRgx, message } = value;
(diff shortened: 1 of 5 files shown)


flammehawk:
This would be better as a GitHub Action on "Push/ PR" to the Release Branches. 
On every other branch, I personally would not care about them. 

Reason being that, If I'm currently developing something and testing it, I would not want to need to remove the statements just because I'm committing to the current state to be able to return to it.  

But preventing them from getting onto the release Branches is in deed needed.

kevinslin:
I was actually just thinking about adding this. I think @flammehawk has a point about local changes. Can we turn this into a `pre-push` hook instead of a `pre-commit` hook? the `.only` check might also belong there as well

SeriousBug:
Thanks for the feedback! I added a `pre-push` hook that checks only the files that are getting pushed. I then moved the `.only` and `debugger;` checks into this hook.

kevinslin (approved):

PR #789, opened 2021-06-03
enhance: Workspace sync improvements

SeriousBug:
Workspace vault sync is now configurable, so users can choose whether they want workspace vaults to get synchronized, skipped, committed and pulled but not pushed, or pulled and pushed but not committed.

Non-repository vault handling has been improved. It fixes exceptions that would be thrown when:
* If some vaults were in repositories but not others.
* If a vault is not in a repository, but the Dendron workspace is.

### DIFF
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -156,6 +156,19 @@ export type DendronConfig = {
    * Development related options
    */
   dev?: DendronDevConfig;
+
+  /**
+   * How workspace vaults should be handled when using workspace "add and commit" and "sync" commands.
+   *
+   * Options are:
+   * * skip: Skip them entirely. You must manage the repository manually.
+   * * noPush: Commit any changes and pull updates, but don't push. You can watch the repository and make local changes without sharing them back.
+   * * noCommit: Pull and push updates if the workspace is clean, but don't commit. You manually commit your local changes, but automatically share them once you committed.
+   * * sync: Commit changes, and pull and push updates. Treats workspace vaults like regular vaults.
+   *
+   * Defaults to `noCommit`.
+   */
+  workspaceVaultSync?: "skip" | "noPush" | "noCommit" | "sync";
 };
 
 export type DendronDevConfig = {
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -14,6 +14,13 @@ export enum DVaultVisibility {
   PRIVATE = "private",
 }
 
+export enum DVaultSync {
+  SKIP = "skip",
+  NO_PUSH = "noPush",
+  NO_COMMIT = "noCommit",
+  SYNC = "sync",
+}
+
 export type DVault = {
   /** Name of vault */
   name?: string;
@@ -31,6 +38,20 @@ export type DVault = {
    * If this is enabled, don't apply workspace push commands
    */
   noAutoPush?: boolean;
+  /**
+   * How the vault should be handled when using "add and commit" and "sync" commands.
+   *
+   * Options are:
+   * * skip: Skip them entirely. You must manage the repository manually.
+   * * noPush: Commit any changes and pull updates, but don't push. You can watch the repository and make local changes without sharing them back.
+   * * noCommit: Pull and push updates if the workspace is clean, but don't commit. You manually commit your local changes, but automatically share them once you committed.
+   * * sync: Commit changes, and pull and push updates. Treats workspace vaults like regular vaults.
+   *
+   * This setting overrides the `workspaceVaultSync` setting for the vault, even if the vault is a workspace vault.
+   *
+   * Defaults to `sync`.
+   */
+  sync?: DVaultSync;
 };
 
 export type DWorkspace = {
@@ -168,7 +189,7 @@ export type DendronConfig = {
    *
    * Defaults to `noCommit`.
    */
-  workspaceVaultSync?: "skip" | "noPush" | "noCommit" | "sync";
+  workspaceVaultSync?: DVaultSync;
 };
 
 export type DendronDevConfig = {
--- a/packages/engine-server/src/workspace.ts
+++ b/packages/engine-server/src/workspace.ts
@@ -220,9 +220,17 @@ export class WorkspaceService {
         const git = new Git({ localUrl: root });
         if (!this.shouldWorkspaceVaultSync("commit", root)) return undefined;
         if (await git.hasChanges()) {
-          await git.addAll();
-          await git.commit({ msg: "update" });
-          return root;
+          try {
+            await git.addAll();
+            await git.commit({ msg: "update" });
+            return root;
+          } catch (err) {
+            const stderr = err.stderr ? `: ${err.stderr}` : "";
+            throw new DendronError({
+              message: `error adding and committing vault${stderr}`,
+              payload: { err, repoPath: root },
+            });
+          }
         }
         return undefined;
       })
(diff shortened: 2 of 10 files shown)


packages/common-all/src/types/workspace.ts
```diff
@@ -156,6 +156,19 @@ export type DendronConfig = {
...
    */
   dev?: DendronDevConfig;
+
+  /**
+   * How workspace vaults should be handled when using workspace "add and commit" and "sync" commands.
+   *
+   * Options are:
+   * * skip: Skip them entirely. You must manage the repository manually.
+   * * noPush: Commit any changes and pull updates, but don't push. You can watch the repository and make local changes without sharing them back.
+   * * noCommit: Pull and push updates if the workspace is clean, but don't commit. You manually commit your local changes, but automatically share them once you committed.
```
kpathakota, line 166: Oh man, love the noCommit option. Really clever.

packages/engine-server/src/workspace.ts
```diff
@@ -199,15 +199,38 @@ export class WorkspaceService {
     return;
   }
 
+  shouldWorkspaceVaultSync(
+    command: "commit" | "push" | "pull",
+    vaultRoot: string
+  ): boolean {
+    if (!WorkspaceService.isWorkspaceVault(vaultRoot)) return true;
+    let config = this.config.workspaceVaultSync;
+    if (_.isUndefined(config)) config = "noPush"; // default
```
kpathakota, line 208: I thought default should be noCommit? According to https://github.com/dendronhq/dendron/pull/789/files#diff-c34b22187b5e218639151cc49620e31d30fea1a5ababad8d17b8f9bf7efc3249R169

SeriousBug: Oh right, thanks for the reminder. I was actually going to ask what the default should be. I really like `noCommit`, but `noPush` preserves the current behavior. Any thoughts?

kpathakota: I thought current default is "sync". You have to work to add things to private and not have it sync by default. I *think* the simplest would actually be to default to sync and have the other options be available. Especially since it's called the workspace sync command and that's the behavior we'd like to encourage.

kevinslin: default should be commit and push unless otherwise specified by config

packages/engine-server/src/workspace.ts
```diff
@@ -425,39 +451,49 @@ export class WorkspaceService {
     const out = await Promise.all(
       allRepos.map(async (root) => {
         const git = new Git({ localUrl: root });
-        if (await git.hasRemote()) {
+        // It's impossible to pull if there is no remote, or if there are tracked files that have changes
+        if (
+          !(await git.hasRemote()) ||
+          (await git.hasChanges({ untrackedFiles: "no" }))
+        )
+          return undefined;
```
kevinslin, line 459: do we return some sort of error or warning on this?

packages/engine-server/src/workspace.ts
```diff
@@ -425,39 +451,49 @@ export class WorkspaceService {
...
   async pushVaults(): Promise<string[]> {
-    const allRepos = await this.getAllRepos();
     const vaults = this.config.vaults;
-    const wsRoot = this.wsRoot;
+    const vaultsPushed = new Set<string>();
     const out = await Promise.all(
-      allRepos.map(async (root) => {
+      vaults.map(async (vault) => {
+        const root = await this.getVaultRepo(vault);
+        if (root === "") return undefined; // The vault is not in a repository
```
kevinslin, line 483: since our convention has been to return undefined for invalid cases, we should stick to that. could you also add this to the [docs](https://wiki.dendron.so/notes/773e0b5a-510f-4c21-acf4-2d1ab3ed741e.html#summary) ?

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
```diff
@@ -7,24 +7,45 @@ import { GitTestUtils } from "@dendronhq/engine-test-utils";
...
   let ctx: vscode.ExtensionContext;
   ctx = setupBeforeAfter(this, {});
 
+  describe("no repo", () => {
+    test("do nothing", (done) => {
+      runLegacyMultiWorkspaceTest({
+        onInit: async ({}) => {
+          const out = await new SyncCommand().execute();
+          expect(out).toBeTruthy();
+          const { committed, pulled, pushed } = out as any;
```
kevinslin, line 24: instead of an `any` cast, can you use the `!` operator so that the output has stricter types?

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
```diff
@@ -67,4 +88,128 @@ suite("workspace sync command", function () {
...
+async function changeConfig(
+  wsRoot: string,
+  overrideConfig: Partial<DendronConfig>
+) {
+  // Get old config, and override it with the new config
+  const serv = getWS().workspaceService!;
+  const config = serv.config;
+  await serv.setConfig(_.merge(config, overrideConfig));
+
+  // Commit this change, otherwise it will be a tracked file with changes which breaks git pull
```
kevinslin, line 211: can you leave this comment at the function signature as well? this way it shows up in the IDE when people hover over the function signature

packages/common-all/src/types/workspace.ts
```diff
@@ -156,6 +156,19 @@ export type DendronConfig = {
...
+   * * sync: Commit changes, and pull and push updates. Treats workspace vaults like regular vaults.
+   *
+   * Defaults to `noCommit`.
+   */
+  workspaceVaultSync?: "skip" | "noPush" | "noCommit" | "sync";
```
kevinslin, line 171: can we make this config granular to a per vault level? (eg. for our dendron workspace, we want everything to sync except `handbook` and `dendron-site`.

I might add this as a property of `DVault` instead of a separate object (eg. `syncBehavior`, with the default being `sync`)

packages/engine-server/src/workspace.ts
SeriousBug: This configuration is only for [workspace vaults](https://wiki.dendron.so/notes/c6fd6bc4-7f75-4cbb-8f34-f7b99bfe2d50.html#workspace-vault). The previous behavior for workspace vaults was the same as `noPush`, this pull request makes that configurable. Regular vaults will continue to get the `sync` behavior and are not configurable with this pull request.

packages/common-all/src/types/workspace.ts
kevinslin: You should be able to use [getVaultByDirPath](https://github.com/dendronhq/dendron/blob/dev-kevin/packages/common-all/src/vault.ts#L72:L72) to match a directory to a vault

kevinslin (changes requested):

packages/engine-server/src/workspace.ts
SeriousBug: I'll add a per-vault configuration as @kevinslin suggested, which will apply to all vaults. This config option will continue to only affect workspace vaults, unless someone suggests otherwise.

SeriousBug: @kevinslin Just to avoid any miscommunication: previous behavior for workspace vaults was to commit and pull but [not push](https://github.com/dendronhq/dendron/blob/6f401d75f21122c84efd03f4307531fde719e37d/packages/engine-server/src/workspace.ts#L446). Should I be changing that behavior? To which option?

packages/engine-server/src/workspace.ts
SeriousBug: Revised the reporting so the Sync command is more informative. Please see `Sync.ts`.

PR #793, opened 2021-06-04
Create redesigned note and schema graphs

hfellerhoff:
This is a **Draft PR 🚧** and is work in progress.

Changes/additions made:
- Add note graph
  - Core display
  - Next.js graph route
  - VSCode integration
  - On node click, open note
- Add schema graph
  - Core display
  - Next.js graph route
  - VSCode integration


packages/dendron-next-server/components/graph.tsx
```diff
@@ -0,0 +1,156 @@
...
+import { useRouter } from "next/router";
+import { useThemeSwitcher } from "react-css-theme-switcher";
+import { Space, Typography } from "antd";
+import Head from "next/head";
+
+const getCytoscapeStyle = (themes: any, theme: string | undefined) => `
+      node {
+        width: 5;
+        height: 5;
+        background-color: ${theme === themes.dark ? "#807B7B" : "#999393"};
```
kevinslin, line 16: instead of hardcoding, can we add this to a constants file  (eg. `styles/theme-antd.tsx`).  in the future, we can use this file to generate the light/dark themes

kevinslin: specifically, themes would look like the below

```tsx
type Theme = {
 "background-color": string
...
}

const darkTheme = {...}

const themes = {
  darkTheme,
  ...
}

```

packages/dendron-next-server/components/graph.tsx
```diff
@@ -0,0 +1,156 @@
...
+          springCoeff: () => 0.0008,
+          mass: () => 4,
+          gravity: -1.2,
+          pull: 0.0001,
+          theta: 0.666,
+          dragCoeff: 0.02,
+          movementThreshold: 1,
+          timeStep: 20,
+          refresh: 10,
+          animate: false, //!isLargeGraph,
```
kevinslin, line 99: should this be uncommented?

packages/dendron-next-server/components/graph.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const logger = createLogger("Graph");
+  const graphRef = useRef<HTMLDivElement>(null);
+  const { switcher, themes, currentTheme, status } = useThemeSwitcher();
+
+  const [cy, setCy] = useState<Core>();
+
+  useEffect(() => {
+    if (graphRef.current && elements) {
+      // Naive check to prevent full graph re-renders when selecting a node
+      if (cy && cy.elements("*").length > 5) return;
```
kevinslin, line 60: can you elaborate on the `length > 5`?

packages/dendron-next-server/pages/vscode/graph.tsx
```diff
@@ -0,0 +1,105 @@
+import {
```
kevinslin, line 1: would rename to `graph-notes.tsx`

packages/dendron-next-server/pages/vscode/graph.tsx
```diff
@@ -0,0 +1,105 @@
...
+        //     });
+        //   }
+        // });
+
+        return [...childConnections, ...linkConnections];
+      })
+      .flat();
+
+    setElements({ nodes, edges });
+  }, [engine]);
```
kevinslin, line 80: do you want to update this on any change on engines or just when notes have changed?

packages/plugin-core/src/commands/ShowNoteGraph.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // If panel does not exist
+    const panel = window.createWebviewPanel(
+      "dendronIframe", // Identifies the type of the webview. Used internally
+      title, // Title of the panel displayed to the user
+      ViewColumn.Two, // Editor column to show the new webview panel in.
+      {
+        enableScripts: true,
+        retainContextWhenHidden: true,
+      } // Webview options. More on these later.
```
kevinslin, line 51: remove comment

packages/plugin-core/src/commands/ShowSchemaGraph.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    // If panel does not exist
+    const panel = window.createWebviewPanel(
+      "dendronIframe", // Identifies the type of the webview. Used internally
+      title, // Title of the panel displayed to the user
+      ViewColumn.Two, // Editor column to show the new webview panel in.
+      {
+        enableScripts: true,
+        retainContextWhenHidden: true,
+      } // Webview options. More on these later.
```
kevinslin, line 51: remove comment

packages/plugin-core/src/constants.ts
```diff
@@ -640,6 +640,18 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     docLink: "dendron.topic.commands.md",
     docPreview: `![](https://foundation-prod-assetspublic53c57cce-8cpvgjldwysl.s3-us-west-2.amazonaws.com/assets/images/workbench.help.gif)`,
   },
+  SHOW_NOTE_GRAPH: {
+    key: "dendron.showNoteGraph",
+    title: `${CMD_PREFIX} Show Note Graph`,
```
kevinslin, line 645: would currently add a `V2` suffix to this command so that people can use this in conjunction with `markdown links` while we're iterating

kevinslin (changes requested):
nice job. i think we can ship the new note graph cmd next week if we can get the changes in today. for schema graph, I would put an additional `when context` (    when: DendronContext.DEV_MODE) to hide it from the production build at the moment

packages/dendron-next-server/components/graph.tsx
hfellerhoff: This line is designed to prevent the graph from re-rendering when selecting a node. I added this check due to the way the tendency of the schema graph to render twice - once with the single "hard-coded" root node, and once again once the schema data is available. I can see if there's a more elegant solution where the schema graph only gets rendered after the schema data is available

packages/dendron-next-server/pages/vscode/graph.tsx
hfellerhoff: Good catch - updating it to only be on changes to note data

packages/dendron-next-server/components/graph.tsx
hfellerhoff: Just made these changes - let me know if the structure/design makes sense to you

packages/dendron-next-server/components/graph.tsx
kevinslin: makes sense. was confused about 5 vs 1 but see that you updated it to 1

packages/plugin-core/src/commands/ShowSchemaGraph.ts
```diff
@@ -0,0 +1,106 @@
...
+
+type CommandOpts = {};
+
+type CommandOutput = void;
+
+export class ShowSchemaGraphCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput
+> {
+  static key = DENDRON_COMMANDS.SHOW_SCHEMA_GRAPH.key;
```
kevinslin, line 22: this needs to be _V2

packages/plugin-core/src/commands/ShowNoteGraph.ts
```diff
@@ -0,0 +1,106 @@
...
+
+type CommandOpts = {};
+
+type CommandOutput = void;
+
+export class ShowNoteGraphCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput
+> {
+  static key = DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key;
```
kevinslin, line 22: this needs to be _V2

kevinslin (changes requested):
Some changes with keys. Also could you update the dark theme text color?

packages/dendron-next-server/components/graph.tsx
hfellerhoff: Yeah, that was just to have a buffer as I was working things out and I had forgotten to set it to a more understandable/accurate value

kevinslin (approved):

PR #794, opened 2021-06-04
enhance: Telemetry respects VSCode preferences

### LINKED ISSUE #711: Telemetry opt-out setting
## Please Select if your Request is Either something new or an Enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

Because only my `settings.json` is shared across VSCode installations, I need to opt out of telemetry on every system I use.
Additionally I already provide VSCode with a telemetry preference via the `telemetry.enableTelemetry` setting.

## Describe the solution you'd like ##

Dendron should respect the existing system for telemetry opt-out by using the [`isTelemetryEnabled API`](https://code.visualstudio.com/docs/getstarted/telemetry#_for-extension-authors).

It could provide an override, asking for explicit consent, to enable telemetry just for Dendron,
which would ideally be stored somewhere more persistent like `settings.json`.

## Describe alternatives you've considered ##

- Make telemetry opt-in, which goes against the documented rationale of making it opt-out
- Only add the opt-out to the `settings.json`, ignoring `isTelemetryEnabled`, which to me seems quite rude

## Additional context ##

My primary reason for disabling telemetry is, that I heavily dislike the collection of a UUID, as it reduces the anonymity of the data.

SeriousBug:
If telemetry is disabled in VSCode preferences, it will be respected by Dendron. Additionally, users can force enable Dendron telemetry even when it's disabled in VSCode preferences by running the command `Dendron: Enable Telemetry`.

closes #711 

Also completed the `DisposableStore` stub because I happened to come across it. Let me know if I should avoid it for now and I can pick that commit out.

### DIFF
--- a/packages/engine-test-utils/src/__tests__/engine-server/segmentClient.spec.ts
+++ b/packages/engine-test-utils/src/__tests__/engine-server/segmentClient.spec.ts
@@ -1,0 +1,50 @@
+import { SegmentClient, tmpDir } from "@dendronhq/common-server";
+import sinon from "sinon";
+import os from "os";
+
+function mockHome() {
+  const tmpHome = tmpDir().name;
+  sinon.stub(os, "homedir").returns(tmpHome);
+}
+
+describe("SegmentClient", () => {
+  test("enabled", (done) => {
+    mockHome();
+    SegmentClient.enable();
+    const instance = SegmentClient.instance({ forceNew: true });
+    expect(instance._hasOptedOut).toEqual(false);
+
+    sinon.restore();
+    done();
+  });
+
+  test("disabled by user", (done) => {
+    mockHome();
+    SegmentClient.disable();
+    const instance = SegmentClient.instance({ forceNew: true });
+    expect(instance._hasOptedOut).toEqual(true);
+
+    sinon.restore();
+    done();
+  });
+
+  test("disabled by config", (done) => {
+    mockHome();
+    SegmentClient.enable();
+    const instance = SegmentClient.instance({ forceNew: true, optOut: true });
+    expect(instance._hasOptedOut).toEqual(true);
+
+    sinon.restore();
+    done();
+  });
+
+  test("force enabled by user", (done) => {
+    mockHome();
+    SegmentClient.forceEnable();
+    const instance = SegmentClient.instance({ forceNew: true, optOut: true });
+    expect(instance._hasOptedOut).toEqual(false);
+
+    sinon.restore();
+    done();
+  });
+});
--- a/packages/engine-test-utils/src/__tests__/engine-server/segmentClient.spec.ts
+++ b/packages/engine-test-utils/src/__tests__/engine-server/segmentClient.spec.ts
@@ -11,7 +11,7 @@ describe("SegmentClient", () => {
   test("enabled", (done) => {
     mockHome();
     SegmentClient.enable();
-    const instance = SegmentClient.instance({ forceNew: true });
+    const instance = SegmentClient.instance({ forceNew: true, optOut: false });
     expect(instance._hasOptedOut).toEqual(false);
 
     sinon.restore();
@@ -21,7 +21,7 @@ describe("SegmentClient", () => {
   test("disabled by user", (done) => {
     mockHome();
     SegmentClient.disable();
-    const instance = SegmentClient.instance({ forceNew: true });
+    const instance = SegmentClient.instance({ forceNew: true, optOut: false });
     expect(instance._hasOptedOut).toEqual(true);
 
     sinon.restore();
@@ -30,7 +30,6 @@ describe("SegmentClient", () => {
 
   test("disabled by config", (done) => {
     mockHome();
-    SegmentClient.enable();
     const instance = SegmentClient.instance({ forceNew: true, optOut: true });
     expect(instance._hasOptedOut).toEqual(true);
 
@@ -38,9 +37,9 @@ describe("SegmentClient", () => {
     done();
   });
 
-  test("force enabled by user", (done) => {
+  test("enabled by user", (done) => {
     mockHome();
-    SegmentClient.forceEnable();
+    SegmentClient.enable();
     const instance = SegmentClient.instance({ forceNew: true, optOut: true });
     expect(instance._hasOptedOut).toEqual(false);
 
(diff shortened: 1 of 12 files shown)


packages/plugin-core/src/telemetry.ts
```diff
@@ -0,0 +1,44 @@
...
+    Logger.error({ msg: "Error when trying to access telemetry preference" });
+    return undefined;
+  }
+}
+
+/** Creates a SegmentClient for telemetry, if enabled, and listens for vscode telemetry settings to disable it when requested. */
+export function setupSegmentClient(ws: DendronWorkspace) {
+  function instantiateSegmentClient() {
+    SegmentClient.instance({
+      optOut: ws.config.noTelemetry || !isVSCodeTelemetryEnabled(),
```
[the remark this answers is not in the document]
SeriousBug: If I'm not misunderstanding you, I don't think that's a good idea. Then, when the user enabled telemetry in VSCode settings, we won't know if they manually disabled Dendron telemetry or if it was because they had the VSCode setting disabled before.

My main concern is that if a user has never touched "Dendron: Enable Telemetry" and "Dendron: Disable Telemetry" commands, then Dendron should always follow the VSCode settings. If they did, then that takes precedence.

SeriousBug:
> Instead of adding a force enable, can we re-use the existing disablement mechanism and add a `.dendron.no-telemetry`

Not sure what this comment means. This is still re-using the existing disablement mechanism, so if a user has already run the "Dendron: Disable Telemetry" command, this pull request will continue to respect that. The changes here are:
* for users who did not run either the disable or the enable command, we follow VSCode settings
* for users who ran the "Dendron: Enable Telemetry" command, we override VSCode settings and always do telemetry since we had explicit consent

packages/plugin-core/src/telemetry.ts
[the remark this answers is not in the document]
kevinslin: that's fair. one thing we should be able to track though is number of workspaces with telemetry disabled. this tells us the delta in our metrics. so whether its calling `DisableTelemetry` or calling `    AnalyticsUtils.track(VSCodeEvents.DisableTelemetry);` we should know how many workspaces have telemetry disabled

kevinslin (changes requested):

SeriousBug:
Built a more comprehensive enable/disable system that tracks why it was enabled or disabled too. This allows us to detect changes in configuration that enable or disable telemetry, so we can report the change.

It also recognizes the legacy tracking (no file means enabled, file means disabled), and should be able to recognize if a user who previously hadn't disabled Dendron's telemetry actually had VSCode telemetry disabled, and report this change.

packages/plugin-core/src/test/suite-integ/Telemetry.test.ts
```diff
@@ -0,0 +1,185 @@
...
+        },
+      });
+    });
+
+    test("enabled by vscode configuration, but disabled by workspace configuration", (done) => {
+      runLegacyMultiWorkspaceTest({
+        ctx,
+        preSetupHook: async ({ wsRoot }) => {
+          const config = DConfig.genDefaultConfig();
+          config.noTelemetry = true;
```
kevinslin, line 63: an easier way to write this is `TestConfigUtils.withConfig` and pass in a callback

packages/plugin-core/src/test/suite-integ/Telemetry.test.ts
```diff
@@ -0,0 +1,185 @@
...
```
kevinslin, line 84: see above. also applies for subsequent

packages/plugin-core/src/commands/DisableTelemetry.ts
```diff
@@ -21,7 +21,7 @@ export class DisableTelemetryCommand extends BasicCommand<
   }
   async execute() {
     AnalyticsUtils.track(VSCodeEvents.DisableTelemetry);
```
kevinslin, line 23: can we pass a `reason` property as a second property to this?

```ts
.track(DisableTelemetry, {reason: TelemetryStatus.DISABLED_BY_COMMAND})
```

packages/plugin-core/src/telemetry.ts
```diff
@@ -0,0 +1,82 @@
...
+        AnalyticsUtils.track(VSCodeEvents.EnableTelemetry);
+        SegmentClient.enable(TelemetryStatus.ENABLED_BY_CONFIG);
+      } else if (SegmentClient.isEnabled(status)) {
+        // was enabled, now disabled
+        if (ws.config.noTelemetry) {
+          Logger.info({
+            msg: "The user change workspace settings, so we are now disabling telemetry",
+          });
+          SegmentClient.disable(TelemetryStatus.DISABLED_BY_WS_CONFIG);
+          AnalyticsUtils.track(VSCodeEvents.DisableTelemetry);
```
kevinslin, line 44: can we pass a `reason` property to this

packages/plugin-core/src/telemetry.ts
```diff
@@ -0,0 +1,82 @@
...
+        } else if (!isVSCodeTelemetryEnabled()) {
+          Logger.info({
+            msg: "The user change VSCode settings, so we are now disabling telemetry",
+          });
+          SegmentClient.disable(TelemetryStatus.DISABLED_BY_VSCODE_CONFIG);
+          AnalyticsUtils.track(VSCodeEvents.DisableTelemetry);
```
kevinslin, line 50: can we pass a `reason` property to this

packages/engine-test-utils/src/__tests__/engine-server/segmentClient.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    const { stub, tmpHome: _tmpHome } = FileTestUtils.mockHome();
+
+    const instance = SegmentClient.instance({ forceNew: true });
+    expect(instance.hasOptedOut).toEqual(false);
+
+    stub.restore();
+    done();
+  });
+
+  test("enabled by command", (done) => {
```
kevinslin, line 17: might be good to have a test that enables it after explicitly disabling it

kevinslin (changes requested):
this looks great!  just a few more minor comments

kevinslin (approved):

PR #797, opened 2021-06-05
Add scoping capabilities to doctor commands

hikchoi:
This PR:
- Adds a scope toggle button to the Doctor plugin command quickpick.
- Can toggle between `workspace` and `file`.
- Defaults to `workspace` scope.
- `createMissingLinkedNotes` action is disabled for `workspace` scope.
- Above changes are also mirrored in DoctorCLICommand by adding a new command option `candidates` where you can explicitly pass in a list of notes.
- Fixed tests that broke due to above changes.
- Skipped tests for some `createMissingLinkedNotes` that should be re-enabled once we enable workspace scope in the future.
- Added tests to for the scoped behavior.

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
@@ -44,6 +44,7 @@ suite("DoctorCommandTest", function () {
       sinon.stub(cmd, "gatherInputs").returns(
         Promise.resolve({
           action: DoctorActions.FIX_FRONTMATTER,
+          scope: "workspace",
         })
       );
       await cmd.run();
@@ -301,6 +302,7 @@ suite("CREATE_MISSING_LINKED_NOTES", function () {
         sinon.stub(cmd, "gatherInputs").returns(
           Promise.resolve({
             action: DoctorActions.CREATE_MISSING_LINKED_NOTES,
+            scope: "workspace",
           })
         );
         const quickPickStub = sinon.stub(VSCodeUtils, "showQuickPick");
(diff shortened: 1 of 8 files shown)


packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
```diff
@@ -39,11 +39,14 @@ suite("DoctorCommandTest", function () {
     onWSInit(async () => {
       const testFile = path.join(root.name, "vault", "bond2.md");
       fs.writeFileSync(testFile, "bond", { encoding: "utf8" });
+      const testFile2 = path.join(root.name, "vault", "bond3.md");
+      fs.writeFileSync(testFile2, "bond3", { encoding: "utf8" });
```
kevinslin, line 43: nitpick: when testing code, we have a convention of adding "bond" to output values we want to debug - its easy to find in logs but please don't commit as it makes it harder to grep

kevinslin (changes requested):
The icons are a little confusing with our current UX for lookup (we show you an icon if the thing is not selected). my first instinct was to click the folder icon to enable doctor command to run for the workspace. can we switch the icons and give a hint of the current doctor scope in the quickpick title?

Screen Shot 2021-06-05 at 6 40 40 AM

packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts
hikchoi: ah. didn't know that. thanks for letting me know.
do you want me to change the one that was already there?

hikchoi:
- swapped codicons for on/off state
- renamed test file name
- added current scope to quickpick title

kevinslin (approved):

PR #808, opened 2021-06-09
Dev joshi

Harshita-mindfire:


packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -0,0 +1,159 @@
...
+    try {
+      const result = await base(tableName).create(records);
+      if (result) {
+        const timestamp = Time.now().toMillis();
+        fs.writeFileSync(checkpoint, timestamp.toString(), {
+          encoding: "utf8",
+        });
+      }
+    } catch (error) {
+      throw Error(error);
```
kevinslin, line 99: should log this error before throwing it

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -0,0 +1,159 @@
...
+    return notes.filter((note) => note.fname.includes(srcHierarchy));
+  }
+
+  async plant(opts: ExportPodPlantOpts) {
+    const { notes, config, dest } = opts;
+    const { apiKey, baseId, tableName, srcFieldMapping, srcHierarchy } =
+      config as AirtableExportConfig;
+    const checkpoint: string = this.verifyDir(dest);
+
+    if (_.isUndefined(srcHierarchy) || _.isEmpty(srcHierarchy)) {
```
kevinslin, line 126: you can mark properties as required when defining the config

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -0,0 +1,187 @@
...
+          );
+          total = total + result.data.records.length;
+          if (total === filteredNotes.length) {
+            const timestamp = filteredNotes[filteredNotes.length - 1].created;
+            fs.writeFileSync(checkpoint, timestamp.toString(), {
+              encoding: "utf8",
+            });
+          }
+        } catch (error) {
+          console.log("failed to export all the notes. Error: ", error);
```
kevinslin, line 125: use the logger that comes with the pod for logging. see https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.setup.md#L66:L66

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -0,0 +1,187 @@
...
+          throw Error("error");
```
kevinslin, line 126: can we wrap this in an dendron exception? https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.setup.md#L82:L82

kevinslin (changes requested):
this doesn't block the PR but can you also write up the documentation for this?
- instructions of updating docs here -> [[For Feature Owners|dendron.sop.early-seed#for-feature-owners]] (be sure to checkout the dev branch of dendron-site)
- the actual calender docs should go here -> dendron.topic.pod.builtin.airtable (doesn't exist yet)

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -0,0 +1,187 @@
...
+        filteredNotes,
+        apiKey,
+        baseId,
+        tableName,
+        srcFieldMapping,
+        srcHierarchy,
+        checkpoint,
+      });
+    } else {
+      throw Error(
```
kevinslin: instead of throwing an error, can we change the output of `plant` to return `{notes, status}` were notes is the notes that were updated and status is the following enum?

```ts
enum POD_STATUS {
	NO_NEW_ENTRIES = "no_new_entries",
	OK = "ok",
	ERROR = "ERROR"
}

```

kevinslin (changes requested):

packages/pods-core/src/builtin/AirtablePod.ts
Harshita-mindfire: if we do not throw error, the normal execution does not break and  ../plugin-core/src/commands/ExportPod.ts has 
```js
window.showInformationMessage(`done exporting. destination: ${dest}`);
```
 which gives msg as done exporting, even if it is not. 
Can we make changes in this file as well to display message as per status?

kevinslin: got it. yeah, lets save that for a next task then since we have some bigger structural changes as well to integrate

kevinslin (approved):

PR #812, opened 2021-06-10
feat: Copy block reference

SeriousBug:
"Copy Note Link", "Copy Note Reference", and "Copy Note URL" commands now support block anchors.

I preserved the existing behavior of these commands, but extended them so that:
1. If an existing block anchor is selected, it will be copied in the link
2. For "Copy Note Reference", if multiple lines are selected, the command will insert block anchors (or use existing block anchors or headers), which will be copied in the reference as a range.

Also fixed a bug where if there was a header and block anchor within the same line, the block anchor would be parsed as being part of the header, which means the link wouldn't work when used.


SeriousBug:
Also replaced the UUID package with nanoid as requested.

Additionally, I made 2 changes in the VSCode settings:
1. Set the tab width for the workspace as 2. Prettier already formats files to that width, and VSCode recognizes the width in the file and uses it automatically. Without this setting however new files will have 4-space width until prettier runs, which is a minor inconvenience.
2. Set debugging launch options to skip node internals and mocha library code. This makes debugging easier because VSCode will now skip them when stepping through functions, and will hide or de-emphasize them in the stack trace.

packages/plugin-core/.vscode/launch.json
```diff
@@ -15,6 +15,10 @@
         "--extensionDevelopmentPath=${workspaceFolder}"
       ],
       "outFiles": ["${workspaceFolder:plugin-core}/out/**/*.js"],
+      "skipFiles": [
```
kevinslin, line 18: nice addition!

packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -41,14 +41,29 @@ export class CopyNoteLinkCommand extends BasicCommand<
       wsRoot: DendronWorkspace.wsRoot(),
     }) as NoteProps;
     if (!note) {
-      throw Error(`${fname} not found in engine`);
+      throw Error(
+        `${fname} not found in engine! Try saving this file and running "Dendron: Reload Index"`
```
kevinslin, line 45: nice!

kevinslin (changes requested):
1. looks like copy note link is generating an anchor when the next line is selected? https://www.loom.com/share/8460eee51c354069878cc5d350739907
2. shouldn't this generate a block reference anchor?: https://www.loom.com/share/a04e3c2e60954b1bb5f8ebfada833184
3. this probably shouldn't generate a range. default behavior when you dbl click a line is to also select first part of next line : https://www.loom.com/share/8d15b55eb54b4516bf6d2bc4804c6f78

SeriousBug:
Thanks for the feedback!

> * this probably shouldn't generate a range. default behavior when you dbl click a line is to also select first part of next line : https://www.loom.com/share/8d15b55eb54b4516bf6d2bc4804c6f78

Good catch, I rarely use the mouse so I wasn't aware of this behavior. Fixed.

> * looks like copy note link is generating an anchor when the next line is selected? https://www.loom.com/share/8460eee51c354069878cc5d350739907

Same as above, should be fixed now.

> * shouldn't this generate a block reference anchor?: https://www.loom.com/share/a04e3c2e60954b1bb5f8ebfada833184

I was only generating them if multiple lines were selected. Revised to generate if there is some amount of text selected.

kevinslin (changes requested):
changes look good, just a few comments:
- for subsequent commits, can we avoid force pushing? it might be necessary if you've done a rebase from master but in that case, I would just merge from master instead. this is because we lose the ability to see incremental diffs from the last code review
- it looks like there's a bunch of debugger statements in the code. wonder why the pre-push script didn't catch it
- perhaps separate issue by i notice that `GoToNote` command on a relative link with an anchor no longer goes to the part of the document with the anchor. if its a completely separate issue, no need to do the fix in this PR

SeriousBug:
>     * for subsequent commits, can we avoid force pushing? it might be necessary if you've done a rebase from master but in that case, I would just merge from master instead. this is because we lose the ability to see incremental diffs from the last code review

Thanks, I didn't know how the review UI worked. I'll use merge commits instead.

>     * it looks like there's a bunch of debugger statements in the code. wonder why the pre-push script didn't catch it

This might have to do with rebasing actually. The hook diffs against what would get pushed to upstream, which might not be reliable when rebasing. Not sure how to get around that, but it shouldn't be a problem once I stop rebasing all the time :smile: 

>     * perhaps separate issue by i notice that `GoToNote` command on a relative link with an anchor no longer goes to the part of the document with the anchor. if its a completely separate issue, no need to do the fix in this PR

I think I have been seeing some intermittent issues with that outside of this branch too, usually "Reload Index" fixes it. I think it has something to do with notes not being updated soon enough, since we now rely on `note.anchors` to be updated (as opposed to the previous behavior of parsing the note at GoToNote`).

kevinslin (approved):

Issue #743, opened 2021-05-20
Random note


aleksey-rowan:
## Please Select if your Request is Either something new or an Enhancement ##

- [ ] Enhancement of an existing Feature.
- [x] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [x] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

I would like to have a way to open a random note from my vault, similar to how the Obsidian's [Random note](https://help.obsidian.md/Plugins/Random+note) command works.

## Describe the solution you'd like ##

A Dendron command to open a random note. 

## Describe alternatives you've considered ##

[vscode-memo](https://github.com/svsool/vscode-memo) extension has a "Open random note" command which seems to work fine on top of a Dendron vault, but using yet another extension for just this one feature is excessive.

## Additional context ##

Ideally, this command should be configurable to whitelist/blacklist certain hierarchies from which you do or do not want random notes to be picked up.

vicrdguez:
+1 I think this is something great to have to re-discover knowledge in your notes

kevinslin:
Think this is a great idea, especially with the addition of whitelisting/blacklisting

PR #813, opened 2021-06-10
feat:Random Note Command

jonathanyeung:
Fixes https://github.com/dendronhq/dendron/issues/743 

New Command to navigate to a random note.

### DIFF
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -1,752 +1,757 @@
 {
-  "name": "@dendronhq/plugin-core",
-  "private": true,
-  "displayName": "dendron",
-  "description": "Dendron is a hierarchal note taking tool that grows as you do. ",
-  "publisher": "dendron",
-  "version": "0.45.1",
-  "engines": {
-    "vscode": "^1.54.0"
-  },
-  "categories": [
-    "Other"
-  ],
-  "installConfig": {
-    "hoistingLimits": "none"
-  },
-  "keywords": [
-    "second brain",
-    "markdown",
-    "roam",
-    "notes",
-    "schema"
-  ],
-  "activationEvents": [
-    "*"
-  ],
-  "main": "./out/src/extension.js",
-  "extensionPack": [
-    "dendron.dendron-markdown-preview-enhanced",
-    "dendron.dendron-snippet-maker"
-  ],
-  "contributes": {
-    "languages": [
-      {
-        "id": "markdown",
-        "extensions": [
-          ".md"
-        ],
-        "aliases": [
-          "markdown"
-        ],
-        "configuration": "./language-configuration.json"
-      }
+    "name": "@dendronhq/plugin-core",
+    "private": true,
+    "displayName": "dendron",
+    "description": "Dendron is a hierarchal note taking tool that grows as you do. ",
+    "publisher": "dendron",
+    "version": "0.45.1",
+    "engines": {
+      "vscode": "^1.54.0"
+    },
+    "categories": [
+      "Other"
     ],
-    "views": {
-      "explorer": [
+    "installConfig": {
+      "hoistingLimits": "none"
+    },
+    "keywords": [
+      "second brain",
+      "markdown",
+      "roam",
+      "notes",
+      "schema"
+    ],
+    "activationEvents": [
+      "*"
+    ],
+    "main": "./out/src/extension.js",
+    "extensionPack": [
+      "dendron.dendron-markdown-preview-enhanced",
+      "dendron.dendron-snippet-maker"
+    ],
+    "contributes": {
+      "languages": [
         {
-          "id": "dendron.sample",
-          "name": "Sample View",
-          "when": "dendron:devMode",
-          "type": "webview"
-        },
-        {
-          "id": "dendron.treeView",
-          "name": "Tree View",
-          "when": "dendron:pluginActive && !dendron:webUIEnabled"
-        },
-        {
-          "id": "dendron.tree-view",
-          "name": "Tree View V2",
-          "when": "dendron:webUIEnabled",
-          "type": "webview"
-        },
-        {
-          "id": "dendron.backlinks",
-          "name": "Backlinks",
-          "when": "dendron:pluginActive"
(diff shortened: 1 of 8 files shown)


packages/plugin-core/package.json
```diff
@@ -1,752 +1,757 @@
 {
-  "name": "@dendronhq/plugin-core",
```
jonathanyeung, line 2: Oops - did we do some 'prettier' or whitespace formatting changes recently to this file? The only material difference should be the addition of 

```json
            {
                "command": "dendron.randomNote",
                "title": "Dendron: Random Note",
                "desc": "Open a Random Note"
            },
```

kevinslin: yeah, we did some formatting. no worries about this

packages/plugin-core/src/commands/RandomNote.ts
```diff
@@ -0,0 +1,81 @@
...
+import { VSCodeUtils } from "../utils";
+import { getWS } from "../workspace";
+import { BasicCommand } from "./base";
+import _ from "lodash";
+
+type CommandOpts = {};
+
+type CommandInput = {};
+
+type CommandOutput = void;
```
kevinslin, line 13: a convention we've been following is to return the note as an output. makes it easy to test

packages/plugin-core/src/commands/RandomNote.ts
```diff
@@ -0,0 +1,81 @@
...
+    // If no pattern is specified for include, then include all notes for the search set.
+    let includeSet: string[] = config?.include ?? [""];
+
+    let searchPredicate = function (note: NoteProps) {
+      if (note.stub == true) {
+        return false;
+      }
+
+      let isMatch = false;
+
```
kevinslin, line 39: nitpick: prefer `forEach` syntax for conciseness. eg. `includeSet.forEach( pattern => { ... } )`

packages/plugin-core/src/test/suite-integ/RandomNoteCommand.test.ts
```diff
@@ -0,0 +1,184 @@
...
+import { VSCodeUtils } from "../../utils";
+import { expect } from "../testUtilsv2";
+import {
+  runLegacyMultiWorkspaceTest,
+  setupBeforeAfter,
+  withConfig,
+} from "../testUtilsV3";
+
+// common template function for RandomNoteCommand testing
+function basicTest(
```
kevinslin, line 15: prefer object notation for function argument for readability. see https://wiki.dendron.so/notes/773e0b5a-510f-4c21-acf4-2d1ab3ed741e.html#prefer-using-object-notation-for-methods-ith-multiple-arguments

packages/plugin-core/src/test/suite-integ/RandomNoteCommand.test.ts
```diff
@@ -0,0 +1,184 @@
...
+  validateFn: () => void,
+  done: () => void,
+  includePattern?: string[],
+  excludePattern?: string[]
+) {
+  runLegacyMultiWorkspaceTest({
+    ctx,
+    preSetupHook: async ({ wsRoot, vaults }) => {
+      for (let name of noteNames) {
+        await CreateNoteFactory({ fname: name, body: "" }).create({
```
kevinslin, line 27: instead of exporting note factoring, you can use [NoteTestUtilsV4](https://github.com/dendronhq/dendron/blob/dev-kevin/packages/common-test-utils/src/noteUtils.ts#L55:L55)

kevinslin (changes requested):

packages/plugin-core/src/test/suite-integ/RandomNoteCommand.test.ts
jonathanyeung: Sorry, new to Typescript - when I tried to switch to object notation, the parameters can no longer be referenced within `runLegacyMultiWorkspaceTest()`.  Seems that the properties of opts aren't known in that context (Error is `Object literal may only specify known properties, and 'opts' does not exist in type 'SetupCodeConfigurationV2...`

packages/plugin-core/src/commands/RandomNote.ts
jonathanyeung: Hmm - doesn't seem like you can call `break` from within an interable.forEach() loop though: https://stackoverflow.com/questions/2641347/short-circuit-array-foreach-like-calling-break

packages/plugin-core/src/commands/RandomNote.ts
jonathanyeung: cool, will adjust

packages/plugin-core/src/test/suite-integ/RandomNoteCommand.test.ts
jonathanyeung: cool, will adjust and remove the export.

packages/plugin-core/package.json
jonathanyeung: is it ok to merge as is? or would you prefer me to re-align to the whitespace changes currently in master?

kevinslin: can you update the package generation code? specifically [this](https://github.com/dendronhq/dendron/blob/dev-kevin/packages/plugin-core/scripts/genConfig.ts#L85:L85) should be changed to 2 spaces

packages/plugin-core/src/test/suite-integ/RandomNoteCommand.test.ts
kevinslin: hmm... how did you do it? it should look like the following:
```ts
basicTest({ctx, preSetupHook, ...}: {
  ctx: ...
  preSetupHook: ...
})
```

jonathanyeung: I see what you mean now - adjusted

packages/plugin-core/package.json
jonathanyeung: updated

kevinslin (approved):

PR #814, opened 2021-06-10
Add initial note filtering, schema graph

hfellerhoff:
Changes made:
- Add Schema Graph V2 from dev mode
- When selecting schema node, open schema in VSCode
- Allow for note graph filtering based on hierarchy and linked note connections

### DIFF
--- a/packages/dendron-next-server/components/graph-filter-view.tsx
+++ b/packages/dendron-next-server/components/graph-filter-view.tsx
@@ -1,0 +1,114 @@
+import {
+  Typography,
+  Collapse,
+  Switch,
+  Space,
+  InputNumber,
+  Divider,
+  Input,
+} from "antd";
+import _, { values } from "lodash";
+import { GraphConfig } from "./graph";
+const { Panel } = Collapse;
+
+type FilterProps = {
+  type: "note" | "schema";
+  config: GraphConfig;
+  setField: (key: string, value: any) => void;
+};
+
+const GraphFilterView = ({ type, config, setField }: FilterProps) => {
+  const sections = new Set(Object.keys(config).map((key) => key.split(".")[0]));
+  const configEntries = Object.entries(config);
+
+  return (
+    <div
+      style={{
+        position: "absolute",
+        top: "1rem",
+        left: "1rem",
+        background: "#F5F6F8",
+        borderRadius: 4,
+        zIndex: 10,
+        minWidth: "12rem",
+      }}
+    >
+      <Collapse>
+        {Array.from(sections).map((section, i) => (
+          <Panel header={_.capitalize(section)} key={i}>
+            <Space direction="vertical" style={{ width: "100%" }}>
+              {configEntries
+                .filter(([key]) => key.includes(section))
+                .map(([key, entry]) => {
+                  const keyArray = key.split(".");
+                  const label = _.capitalize(keyArray[keyArray.length - 1]);
+
+                  return (
+                    <Space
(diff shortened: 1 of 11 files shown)


packages/dendron-next-server/components/graph-filter-view.tsx
```diff
@@ -0,0 +1,114 @@
...
+    });
+  };
+
+  return (
+    <div
+      style={{
+        position: "absolute",
+        top: "1rem",
+        left: "1rem",
+        background: currentTheme === "light" ? "#F5F6F8" : "#303030",
```
kevinslin, line 39: can we get these values from the theme file?

packages/dendron-next-server/components/graph.tsx
```diff
@@ -83,39 +174,45 @@ export default function Graph({
...
+      network.layout(getEulerConfig(isLargeGraph)).run();
 
       network.on("select", (e) => onSelect(e));
 
       setCy(network);
     }
-  }, [graphRef, elements]);
+  };
+
+  // const parseElements = (
```
kevinslin, line 185: are we still using this? if not, lets delete

packages/dendron-next-server/hooks/useGraphElements.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  const nodes: any[] = [];
+  const edges: GraphEdges = {
+    hierarchy: [],
+  };
+
+  if (_.isUndefined(vaults)) return { nodes, edges };
+
+  vaults.map((vault) => {
+    const VAULT_ID = `${vault.name}`;
```
kevinslin, line 90: there's a chance that a vault does not have a name in which case we default to fsPath. should use [VaultUtils.getName](https://github.com/dendronhq/dendron/blob/feature-graph/packages/common-all/src/vault.ts#L8:L8)

packages/dendron-next-server/hooks/useGraphElements.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    nodes.push({
+      data: {
+        id: VAULT_ID,
+        label: vault.name,
```
kevinslin, line 95: see above comment

packages/dendron-next-server/hooks/useGraphElements.tsx
```diff
@@ -0,0 +1,186 @@
...
+        group: "nodes",
+        selectable: false,
+      },
+    });
+
+    schemaArray
+      .filter((schema) => schema.vault.name === vault.name)
```
kevinslin, line 102: see above comment (also applies to subsequent `vault.name` occurrences)

kevinslin (changes requested):
1. looks like there's an extra node created in root? you can see it if you open up `test-workspace` with schema graph.  https://www.loom.com/share/9fb2b0bb6b0f492ea2ec552f2fefcd42
2. currently, every workspace will have a `root` schema. for the graph, can we show each root distinctly? (add the vault name next to its label to differentiate)

kevinslin (changes requested):
1. when I run this against test-workspace, I get three nodes when there should only be two (root of vault1 and root of vault2). also, clicking on vault doesn't open the schema.

see https://www.loom.com/i/c208042f90e1496ba24af5708030b5d9

2. name for root notes should indicate that their root (eg. root (vault1))

kevinslin (approved):

PR #819, opened 2021-06-12
feat: better rename

SeriousBug:
Finished making it work for note refs and added more tests.


kevinslin (approved):
this is a significant change with lots of moving parts. great job getting this over the finish line!

Issue #817, opened 2021-06-11
Block Link doesn't work properly.


agichim:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [x] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

[Block Links](https://wiki.dendron.so/notes/21bc1caa-d815-42db-bc16-27d6c16a3115.html#block-links) don't work properly.

Create a block link following the Dendron docs from the link above. Click on the block link in the preview, you should be taken to the referenced paragraph in the same `.md` file, but instead nothing happens. 

When holding CTRL and clicking on the block link in the editor you are taken to the beginning of the `.md` file and the typing cursor `|` is placed right a the beginning before the first `---` of the frontmatter.

When holding CTRL and hovering the cursor over the block link the small preview windows displays the UUID or `id` of the frontmatter and the whole document content. If hovering over the block link without holding CTRL the whole document content is displayed from the beginning of the document. 

## To Reproduce

1. In the same `.md` file place a few paragraphs
2. At the end of one paragraph place a block reference `^my-block`
3. Anywhere in the document place a block link referencing the block reference `[[#^my-block]]
4. Click on the block link in the preview (nothing happens)
   - Hold CTRL and hover over the block link (doesn't show the block reference)
   - Hold CTRL and click on the block link (it takes you to the beginning of the document)

## Expected behavior

1. Clicking on the block link in the preview should take you to the block reference.
2. Holding CTRL and hovering over the block link should show only the block reference in the small preview window.
3. Holding CTRL and clicking on the block link should take you to the block reference.

## Screenshots

![image](https://user-images.githubusercontent.com/51931067/121758144-1e200280-cb18-11eb-8560-39e076d3a593.png)
![image](https://user-images.githubusercontent.com/51931067/121758154-28da9780-cb18-11eb-9f64-4fb43f0d2775.png)

## Desktop (please complete the following information)

- OS: Windows 10. Ver. 2004 (Build 19041.985)
- Browser: Brave (not tested the block link and block reference in the published GH pages)
- Version → version of what software?

## Dendron Log file

[...]

PR #823, opened 2021-06-15
fix: anchors now update when note is modified

SeriousBug:
Should work for #817, as well as intermittent issues with GoToNote on anchors not working.

### DIFF
--- a/packages/plugin-core/src/fileWatcher.ts
+++ b/packages/plugin-core/src/fileWatcher.ts
@@ -9,6 +9,7 @@ import {
 import { file2Note, string2Note } from "@dendronhq/common-server";
 import { HistoryService, LinkUtils } from "@dendronhq/engine-server";
 import _ from "lodash";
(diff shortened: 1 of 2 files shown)


SeriousBug:
I can't reproduce the failing tests, although it looks like `master` is also failing on CI.

packages/plugin-core/src/fileWatcher.ts
```diff
@@ -91,6 +92,11 @@ export class VaultWatcher {
     note = NoteUtils.hydrate({ noteRaw: note, noteHydrated });
     const links = LinkUtils.findLinks({ note, engine: eclient });
     note.links = links;
+    const anchors = AnchorUtils.findAnchors(
```
kevinslin, line 95: nitpick on style: Promises are usually awaited on at the call site or chained with a `then` function.

kevinslin (changes requested):

kevinslin (approved):

PR #824, opened 2021-06-15
feat: filter note and schema graphs by vault

hfellerhoff:
Adds:
- show/hide different vaults in graph
- move graph config application into separate hook to separate concerns

### DIFF
--- a/packages/dendron-next-server/components/graph.tsx
+++ b/packages/dendron-next-server/components/graph.tsx
@@ -56,7 +56,7 @@ const getCytoscapeStyle = (themes: any, theme: string | undefined) => {
 `;
 };
 
-const getEulerConfig = (isLargeGraph: boolean) => ({
+const getEulerConfig = (shouldAnimate: boolean) => ({
   name: "euler",
   // @ts-ignore
   springLength: () => 80,
@@ -69,7 +69,7 @@ const getEulerConfig = (isLargeGraph: boolean) => ({
   movementThreshold: 1,
   timeStep: 20,
   refresh: 10,
-  animate: !isLargeGraph,
+  animate: shouldAnimate,
   animationDuration: undefined,
   animationEasing: undefined,
   maxIterations: 1000,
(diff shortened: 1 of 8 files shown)


packages/dendron-next-server/hooks/useApplyGraphConfig.tsx
```diff
@@ -0,0 +1,88 @@
+import cytoscape from "cytoscape";
+import { useEffect, useState } from "react";
+import { createLogger } from "../../common-frontend/lib";
```
kevinslin, line 3: watch out for relative imports into workspace packages. see [here](https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.troubleshooting.md#L12:L12)

packages/dendron-next-server/components/graph.tsx
```diff
@@ -17,10 +17,13 @@ import AntThemes from "../styles/theme-antd";
 import GraphFilterView from "./graph-filter-view";
 import {
   GraphConfig,
+  GraphConfigItem,
   GraphEdges,
   GraphElements,
   GraphNodes,
 } from "../lib/graph";
+import { VaultUtils } from "../../common-all/lib";
```
kevinslin, line 25: watch out for relative imports into workspace packages. see [here](https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.troubleshooting.md#L12:L12)

packages/dendron-next-server/components/graph.tsx
```diff
@@ -190,8 +171,35 @@ export default function Graph({
...
+          label: name,
+        };
+
+        return {
+          ...dict,
+          [key]: item,
+        };
+      }, {});
+
+      console.log(vaultConfigObject);
```
kevinslin, line 194: use `logger` instead of `console.log`

kevinslin (changes requested):
Fullscreen_6_15_21__9_46_AM

don't remember if this was the behavior before but it looks like there is no edge link between a note in vault2 that has a link to a note in vault1

kevinslin (approved):

PR #825, opened 2021-06-15
enhance: Workspace Sync checks for changes before pushing & better handling when there is no upstream

SeriousBug:
We now check if there are any changes to be pushed before pushing. This way, if the user adds a vault they can't push to, sync will continue to function without any configuration as long as they don't start making changes to the repository.

We now also check if there is an upstream branch configured for the current branch, in addition to checking if there are remotes. This can easily happen if the user checks out a new branch. If it does happen, we now continue syncing other vaults, and notify the user that we can't pull or push that repository because it has no upstream branch.


packages/engine-server/src/topics/git.ts
```diff
@@ -135,4 +135,41 @@ export class Git {
...
+      );
+      return _.trim(stdout);
+    } catch {
+      return undefined;
+    }
+  }
+
+  /**
+   *
+   * @param old: The old identifier (e.g. commit, tag, branch) that we are diffing against.
```
kevinslin, line 153: can the param name match the variable name? `eg. oldCommit`

packages/engine-server/src/topics/git.ts
```diff
@@ -135,4 +135,41 @@ export class Git {
...
+   * @param new: The new identifier (e.g. commit, tag, branch) that we are diffing from.
```
kevinslin, line 154: see above

kevinslin (commented):
minor doc string updates. do you know why the plugin test is failing?

SeriousBug:
>  do you know why the plugin test is failing?

Not sure, but seems to be passing now.

kevinslin (approved):

PR #826, opened 2021-06-16
enhance: Hover displays targeted area for links with `#headers` and `#^anchors`

SeriousBug:
Hover now supports headers and anchors. All of the following work with hover as one would expect:

```
[[note#header]]

[[note#^anchor]]

[[alias|note#header]]

[[dendron://vault/note#header]]

![[note#start:#*]]

![[note#start:#^end]]
```

This is done by converting the link to a reference, and using our processor stack to compile it to regular markdown. This allows all anchor processing done for references to be reused here.

Also includes slightly better handling for links (e.g. `http://example.com` or `mailto:user@example.com`) and missing files.

Finally, cleaned up several unnecessary `escapeForRegExp` expressions, most of which were unused. Replaced the used one with lodash's built-in function.


packages/plugin-core/src/features/ReferenceHoverProvider.ts
```diff
@@ -1,122 +1,115 @@
...
-          return out;
-        };
-        const content = await getContent();
+    const engine = getEngine();
+    const vault = PickerUtilsV2.getOrPromptVaultForOpenEditor();
 
-        return new vscode.Hover(content, hoverRange);
-      }
+    // Check if what's being referenced is a note.
+    const maybeNotes = NoteUtils.getNotesByFname({
```
kevinslin, line 83: we could have a xvault link in which case the vault also needs to be passed in. that would be `ref.vaultName` property

kevinslin (changes requested):
1. 

can we add tests for this? see `src/test/suite-integ/ReferenceProvider.test.ts` for current tests

2. 
looks like there's a few issues with xvault links and anchor references. see examples here: https://www.loom.com/share/c17084c40ad64d5c8960b4f825510c56

the note in question was made from `test-workspace`, instructions on connecting to it here -> [[Test Workspace|dendron.dev.qa#test-workspace]] (pull from dendron-site)

it looks like this isn't being recognized if there's no new line between the reference and the referent. in `dendron.ref.links.md`, the link `![[dendron.ref.links.target#^123]]` generates the following html. 

note how `Bar one` is on the same line as the closing `</div>`. you should disable pretty refs when rendering for markdown.

```
# Bar

<div class="portal-container">
<div class="portal-head">
<div class="portal-backlink" >
<div class="portal-title">From <span class="portal-text-title">bar</span></div>
<a href="bar" class="portal-arrow">Go to text <span class="right-arrow">→</span></a>
</div>
</div>
<div id="portal-parent-anchor" class="portal-parent" markdown="1">
<div class="portal-parent-fader-top"></div>
<div class="portal-parent-fader-bottom"></div>Bar one


</div></div>
```

SeriousBug:
> can we add tests for this? see `src/test/suite-integ/ReferenceProvider.test.ts` for current tests

Done.

> note how `Bar one` is on the same line as the closing `</div>`. you should disable pretty refs when rendering for markdown.

Looks like that was the issue, disabling it solved the problems demonstrated in the loom.

By the way, it looks like `vault2` is missing in the test workspace. It looks like it was left out of the commit because it's in the `.gitignore` file.

kevinslin:
> > can we add tests for this? see `src/test/suite-integ/ReferenceProvider.test.ts` for current tests
> 
> Done.
> 
> > note how `Bar one` is on the same line as the closing `</div>`. you should disable pretty refs when rendering for markdown.
> 
> Looks like that was the issue, disabling it solved the problems demonstrated in the loom.
> 
> By the way, it looks like `vault2` is missing in the test workspace. It looks like it was left out of the commit because it's in the `.gitignore` file.

doh. good catch, will fix!

kevinslin (approved):

Issue #541, opened 2021-03-14
Markdown Export Pod


kevinslin:
## Context

Dendron recently launched the ability to [export to plain markdown](https://wiki.dendron.so/notes/13c4a608-0a32-4c62-9c7e-2b7f9f2d18bf.html#export). This is still an experimental feature and we are collecting feedback on how to best implement this. 

## Proposal

This issue is to collect feedback and additional usecases for the export feature. Please reply in the comments in the following format:

## Sample Feedback
- Feature: Support exporting assets 
- Use cases: GitBook, Typora, etc.

buxel:
- Feature: control what is exported, similar to publishing
- Use cases: i'd like to publish parts of my vault to our team's DevOps Wiki

PR #827, opened 2021-06-16
feat:export markdown pod functionality

jonathanyeung:
Improving functionality around markdown pod export. 

Related Links
- https://github.com/dendronhq/dendron/issues/541#issue-831219413
- https://wiki.dendron.so/notes/ccab31c4-e9ca-4ee8-b754-f786a3f3be6e.html

### DIFF
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -589,7 +589,7 @@ export class RemarkUtils {
 
   // --- conversion
 
-  static convertObsidianLinks(note: NoteProps, changes: NoteChangeEntry[]) {
+  static convertLinksToDotNotation(note: NoteProps, changes: NoteChangeEntry[]) {
     return function (this: Processor) {
       return (tree: Node, _vfile: VFile) => {
         let root = tree as DendronASTRoot;
@@ -610,6 +610,39 @@ export class RemarkUtils {
             });
           }
         });
+      };
+    };
+  }
+
+  static convertLinksFromDotNotation(note: NoteProps, changes: NoteChangeEntry[]) {
+    return function (this: Processor) {
+      return (tree: Node, _vfile: VFile) => {
+        let root = tree as DendronASTRoot;
+        let wikiLinks: WikiLinkNoteV4[] = selectAll(
+          DendronASTTypes.WIKI_LINK,
+          root
+        ) as WikiLinkNoteV4[];
+
+        let dirty = false;
+
+        wikiLinks.forEach((linkNode) => {
+          if (linkNode.value.indexOf(".") >= 0) {
+            const newValue = _.replace(linkNode.value, /\./g, "/");
+            if (linkNode.data.alias === linkNode.value) {
+              linkNode.data.alias = newValue;
+            }
+            linkNode.value = newValue;
+            dirty = true;
+          }
+        });
+        //TODO: Add support for Ref Notes and Block Links
+
+        if (dirty) {
+          changes.push({
+              note: note,
+              status: "update",
+            });
+        }
       };
     };
   }
--- a/packages/pods-core/src/builtin/MarkdownPod.ts
+++ b/packages/pods-core/src/builtin/MarkdownPod.ts
@@ -10,7 +10,6 @@ import {
 import { cleanFileName, readMD, vault2Path } from "@dendronhq/common-server";
 import {
   DendronASTDest,
-  MDUtilsV4,
   MDUtilsV5,
   RemarkUtils,
 } from "@dendronhq/engine-server";
@@ -260,7 +259,7 @@ export class MarkdownImportPod extends ImportPod<MarkdownImportPodConfig> {
             dest: DendronASTDest.MD_DENDRON,
             vault: n.vault,
           })
-            .use(RemarkUtils.convertObsidianLinks(n, []))
+            .use(RemarkUtils.convertLinksToDotNotation(n, []))
             .process(n.body);
           n.body = cBody.toString();
           if (config.frontmatter) {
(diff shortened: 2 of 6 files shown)


packages/pods-core/src/builtin/MarkdownPod.ts
```diff
@@ -336,6 +333,21 @@ export class MarkdownExportPod extends ExportPod {
         return fs.writeFile(fpath, body);
       })
     );
+
+    // Export Assets
+    vaults.forEach(async (vault) => {
```
kevinslin, line 338: this is an async function and will return an array of promises. this means this can finish after the function returns. we need to add an `await Promise.all(vaults.forEach(...)`

packages/pods-core/src/builtin/MarkdownPod.ts
```diff
@@ -336,6 +333,21 @@ export class MarkdownExportPod extends ExportPod {
...
+      //TODO: Avoid hardcoding of assets directory, or else extract to global const
+      const destPath = path.join(dest.fsPath, vault.fsPath, "assets");
+      const srcPath = path.join(wsRoot, vault.fsPath, "assets");
```
kevinslin, line 341: we have multiple types of [vaults](https://wiki.dendron.so/notes/6682fca0-65ed-402c-8634-94cd51463cc4.html#vault-types). depending on the type, the path might be different. you should use [`getRelPath`](https://github.com/dendronhq/dendron/blob/md-pod-export/packages/common-all/src/vault.ts#L32:L32) to get relative path to the vault from the workspace

kevinslin (changes requested):
1.  it looks like you have a pretty old version of the code checked out. can you rebase this against what is in master? you'll want to rebuild your workspace otherwise because of the changes -> [[Building from clean repo|dendron.dev.cook#building-from-clean-repo]]

2. about the old tests failing -> `MDUtilsV5` is not yet feature parity with `MDUtilsV4`. for example, the publish pod test is failing because [MDUtilsV4](https://github.com/dendronhq/dendron/blob/md-pod-export/packages/engine-server/src/markdown/utils.ts#L426:L426) passes in `insertTitle: config.useFMTitle` which `MDUtilsV5` does not. this is because in v5, we're doing all configuration from `DendronConfig`, see [[2021 07 Unified Refactoring|proj.2021-07-unified-refactoring#^RNbCpbpD]].

to fix, [dendronPub](https://github.com/dendronhq/dendron/blob/md-pod-export/packages/engine-server/src/markdown/remark/dendronPub.ts#L40:L40) will need to read `insertTitle` from the config

you'll want to do this for all the properties that are currently being passed in to the plugins in [MDUtilsV4](https://github.com/dendronhq/dendron/blob/md-pod-export/packages/engine-server/src/markdown/utils.ts#L405:L405)

(note: this might become quite involved so let me know if you want to pair on this or switch to use MDUtilsv4 for now)

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -614,6 +614,39 @@ export class RemarkUtils {
...
+        let wikiLinks: WikiLinkNoteV4[] = selectAll(
+          DendronASTTypes.WIKI_LINK,
+          root
+        ) as WikiLinkNoteV4[];
+
+        let dirty = false;
+
+        wikiLinks.forEach((linkNode) => {
+          if (linkNode.value.indexOf(".") >= 0) {
+            const newValue = _.replace(linkNode.value, /\./g, "/");
```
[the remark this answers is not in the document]
jonathanyeung: I actually did that initially - but reverted after doing a test against Obsidian, which only observes forward slashes and not back slashes (even though I'm on Windows). Seems like markdown generally consolidates to forward slashes only in links?

packages/pods-core/src/builtin/MarkdownPod.ts
jonathanyeung: good catch

jonathanyeung:
Thanks for the review, some comments:
1) Rebased
2) Gotcha - will revert to V4 for now.  I had upgraded to V5 when trying to add the Ref Notes capabilities, but if we're pushing that back for now it makes sense to revert to V4 for the time being.

packages/pods-core/src/builtin/MarkdownPod.ts
jonathanyeung: Hmm, I tried making this change, but then it's causing some issues in my tests.

What is getRelPath intended to return?  Maybe the same question, but when is vault.workspace meant to be set? Is it supposed to be set when a vault is added to a workspace? I notice that this value is not set in the test environment.  Is this intentional, or is this a bug? 

If I do set a value for vault.workspace in the test, then this conflicts with the following logic in DNote Utils, which will return a double prefix (something like `C:\mydir\C:\mydir\vault1`)
```typescript
  static getFullPath(opts: {
    wsRoot: string;
    vault: DVault;
    basename: string;
  }) {
    const root = path.isAbsolute(opts.vault.fsPath)
      ? VaultUtils.getRelPath(opts.vault)
      : path.join(opts.wsRoot, VaultUtils.getRelPath(opts.vault));
    return path.join(root, opts.basename);
  }
```

packages/engine-server/src/markdown/remark/utils.ts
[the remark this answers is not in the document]
kevinslin: ah, got it. i think your right

packages/pods-core/src/builtin/MarkdownPod.ts
kevinslin: discussed offline, let me know if you need more context here

kevinslin (approved):
good job. don't forget to update the documentation for export pods -> dendron.topic.pod.builtin.markdown.export.md

PR #828, opened 2021-06-16
feat: filter graph by string

hfellerhoff:
Added:
- Whitelist/blacklist support for filtering the graph
- Filters based on label and filename
- Supports comma-separated and space-separated lists lists as input


packages/dendron-next-server/components/graph-filter-view.tsx
```diff
@@ -112,6 +122,59 @@ const GraphFilterView = ({ config, setConfig }: FilterProps) => {
...
+      }, delay)
+    );
+  };
+
+  return (
+    <Space direction="vertical" style={{ marginBottom: "1rem" }}>
+      <Typography>{label}</Typography>
+      <Input
+        defaultValue={entry.value}
+        onChange={(newValue) => handleChange(newValue.target.value)}
```
kevinslin, line 161: Looks like c/p isn't working in the webview because of the workaround we have: [[Copy and Paste|pkg.dendron-next-server.arch#copy-and-paste]]
Can we make sure that keyboard events are propagated upwards to vscode? You can see an working example of this is `Dendron: Configure` in the webview there.

packages/dendron-next-server/components/graph.tsx
```diff
@@ -1,12 +1,7 @@
...
 import React, { useEffect, useRef, useState } from "react";
-import cytoscape, {
-  Core,
-  EdgeDefinition,
-  ElementsDefinition,
-  EventHandler,
-} from "cytoscape";
+import cytoscape, { Core, EdgeDefinition, EventHandler } from "cytoscape";
 // @ts-ignore
 import euler from "cytoscape-euler";
```
kevinslin, line 6: there's no `@types/` package for this?

kevinslin (changes requested):

packages/dendron-next-server/components/graph.tsx
hfellerhoff: There's not. I can look into adding a `d.ts` file to be able to get rid of the `//@ts-ignore` if you'd like

kevinslin: yeah, that would be preferred :)

packages/dendron-next-server/components/graph-filter-view.tsx
hfellerhoff: The keyboard events are now being propagated correctly, but despite that they still aren't triggering commands like Copy and Paste. Because of that this is a bit blocked at the moment - every other component of the PR should be good to go

packages/dendron-next-server/pages/vscode/graph-note.tsx
```diff
@@ -31,6 +31,7 @@ import {
 } from "../../lib/graph";
 import GraphFilterView from "../../components/graph-filter-view";
 import useGraphElements from "../../hooks/useGraphElements";
+import usePropegateKeystrokes from "../../hooks/usePropegateKeystrokes";
```
kevinslin, line 34: remove

packages/dendron-next-server/pages/vscode/graph-note.tsx
```diff
@@ -40,6 +41,7 @@ export default function FullNoteGraph({
   const [config, setConfig] = useState<GraphConfig>(graphConfig.note);
 
   const elements = useGraphElements({ type: "note", engine });
+  usePropegateKeystrokes();
```
kevinslin, line 44: remove

packages/dendron-next-server/pages/vscode/graph-schema.tsx
```diff
@@ -15,6 +15,7 @@ import {
   GraphViewMessage,
   GraphViewMessageType,
 } from "@dendronhq/common-all";
+import usePropegateKeystrokes from "../../hooks/usePropegateKeystrokes";
```
kevinslin, line 18: see above comments

kevinslin (changes requested):
played around with this. to enable c/p, you'll need to pass in the following option when calling `createWebviewPanel`

```
        enableFindWidget: true,
```

hfellerhoff:
> played around with this. to enable c/p, you'll need to pass in the following option when calling `createWebviewPanel`
> 
> 
> 
> ```
> 
>         enableFindWidget: true,
> 
> ```

Glad it's simple - I'll make that change

packages/dendron-next-server/hooks/usePropegateKeystrokes.tsx
```diff
@@ -0,0 +1,37 @@
+import { useEffect } from "react";
+
+const usePropegateKeystrokes = () => {
```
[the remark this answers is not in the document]
hfellerhoff: Does useVSCodeMessage handle sending messages as well? At first glance it looks like it's designed for listening to messages, not sending them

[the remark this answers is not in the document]
kevinslin: yep. it listens to messages and then posts them to the parent. this is what we're relying on for `ConfigureWithUI`

kevinslin (approved):

PR #831, opened 2021-06-18
feat/seed-enhance

kevinslin:
bag of improvements for seed bank:

- better cli output
- support custom `site` object
- copy note url is seed specific
- convert remove vaults into seeds


packages/common-all/src/types/seed.ts
```diff
@@ -6,6 +6,15 @@ export type SeedConfig = {
...
   contact?: SeedContact;
+  /**
+   * Url for seed
+   */
+  site?: SeedSite;
+};
+
+export type SeedSite = {
+  url: string;
+  index?: string;
```
jonathanyeung, line 17: What does index here mean/do?

packages/engine-server/src/seed/service.ts
```diff
@@ -99,8 +104,24 @@ export class SeedService {
...
+    const spath = SeedUtils.seed2Path({ wsRoot, id });
+    fs.ensureDirSync(path.dirname(spath));
+    const git = simpleGit({ baseDir: wsRoot });
+    const options: {
+      "--track"?: string;
+    } = {};
+    if (branch) {
+      options["--track"] = branch;
+    }
+    await git.clone(seed.repository.url, spath, options);
```
jonathanyeung, line 124: Any possibilities of name collisions here? i.e. if folder w/ name already exists.

packages/dendron-cli/src/commands/base.ts
```diff
@@ -47,11 +49,13 @@ export abstract class CLICommand<TOpts, TOut> extends BaseCommand<TOpts, TOut> {
...
-    // add wsRoot if not exist
     if (!args.wsRoot) {
       const cwd = process.cwd();
-      if (!fs.existsSync(path.join(cwd, "dendron.yml"))) {
-        console.log("no workspace deted. --wsRoot must be set");
+      if (
+        !fs.existsSync(path.join(cwd, "dendron.yml")) &&
+        !this.wsRootOptional
+      ) {
+        console.log("no workspace detecd. --wsRoot must be set");
```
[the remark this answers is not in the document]
kevinslin: you are finding all my typos :)

packages/common-all/src/types/seed.ts
kevinslin: equivalent of `siteIndex`, see https://wiki.dendron.so/notes/f2ed8639-a604-4a9d-b76c-41e205fb8713.html#siteindex-optional

packages/engine-server/src/seed/service.ts
kevinslin: we are going with the npm approach where the name has to be globally unique. a seed name is `{publisher}.{name}`

jonathanyeung (approved):

PR #833, opened 2021-06-18
Removing unnecessary ext installs and recommendations

jonathanyeung:


packages/engine-server/src/workspace/vscode.ts
```diff
@@ -77,8 +77,7 @@ const _EXTENSIONS: ConfigUpdateEntry[] = [
   { default: "dendron.dendron-paste-image" },
   { default: "dendron.dendron-markdown-shortcuts" },
   { default: "dendron.dendron-markdown-preview-enhanced" },
-  { default: "dendron.dendron-markdown-links" },
-  { default: "github.github-vscode-theme" },
+  { default: "dendron.dendron-markdown-links", action: "REMOVE" },
```
jonathanyeung, line 80: @kevinslin let me know if REMOVE is the correct action here. I can also revert the github-theme change if you'd like.

kevinslin (approved):

PR #834, opened 2021-06-18
spike: implement extract blocks

SeriousBug:


packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -751,4 +785,116 @@ export class RemarkUtils {
...
+    const proc = MDUtilsV5.procRemarkFull({
+      engine,
+      vault: note.vault,
+      fname: note.fname,
+      dest: DendronASTDest.MD_DENDRON,
+    });
+    const slugger = getSlugger();
+
+    // Read and parse the note
+    // TODO: It might be better to get the text from the editor
```
kevinslin, line 813: this is okay. we want to decouple, as much as makes sense, dendron fucntionality from vscode specific methods

kevinslin (approved):

PR #836, opened 2021-06-18
Dev joshi crop

Harshita-mindfire:


packages/engine-server/src/config.ts
```diff
@@ -45,6 +45,7 @@ export class DConfig {
       noXVaultWikiLink: true,
       lookupConfirmVaultOnCreate: false,
       dayOfWeek: 1,
+      defaultDailyJournalVault: "vault",
```
kevinslin, line 48: should not be set by default. just delete this line

kevinslin (changes requested):
we can leave the defaultDailyJournalVault command blank. also, looks like the tests are failing due to the snapshot output being different. 

see instructions for updating snapshot here -> https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.qa.md#updating-test-snapshots


that being said, once you get rid of the default, tests should pass again so consider this for future reference

kevinslin (changes requested):
this looks good. one more thing: can we add tests in `src/test/suite-integ/CreateDailyJournalNote.test.ts` ?

Harshita-mindfire:
> this looks good. one more thing: can we add tests in `src/test/suite-integ/CreateDailyJournalNote.test.ts` ?

yes Kevin, I'm on it.

packages/plugin-core/src/test/suite-integ/CreateDailyJournalNote.test.ts
```diff
@@ -1,21 +1,86 @@
...
+      onInit: async ({ wsRoot, vaults }) => {
+        withConfig(
+          (config) => {
+            config.lookupConfirmVaultOnCreate = false;
+            config.defaultDailyJournalVault = vaults[0].fsPath;
+            return config;
+          },
+          { wsRoot }
+        );
+        stubVaultPick(vaults);
```
kevinslin, line 55: we shouldn't need to stub the vault picker here since default journal means we don't need to pick

packages/plugin-core/src/test/suite-integ/CreateDailyJournalNote.test.ts
```diff
@@ -1,21 +1,86 @@
...
-  test("basic", (done) => {
+  test("default journal vault", (done) => {
     runLegacyMultiWorkspaceTest({
       ctx,
-      onInit: async ({}) => {
+      onInit: async ({ wsRoot, vaults }) => {
+        withConfig(
+          (config) => {
+            config.lookupConfirmVaultOnCreate = false;
+            config.defaultDailyJournalVault = vaults[0].fsPath;
```
kevinslin, line 50: while this works because the fsPath is the same as the vaultName in this case, we should be using `VaultUtils.getName` to set this value

packages/plugin-core/src/test/suite-integ/CreateDailyJournalNote.test.ts
```diff
@@ -1,21 +1,86 @@
...
+      onInit: async ({ wsRoot, vaults }) => {
+        withConfig(
+          (config) => {
+            config.lookupConfirmVaultOnCreate = true;
+            config.defaultDailyJournalVault = vaults[0].fsPath;
+            return config;
+          },
+          { wsRoot }
+        );
+        await new CreateDailyJournalCommand().run();
```
kevinslin, line 79: this should require `stubVaultPick`. the reason this is probably working is because we don't have a `sinon.restore` call. see https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.qa.md#stubs

kevinslin (changes requested):

kevinslin (changes requested):
it looks like the first test is failing due to a timeout. can you make sure 

```ts
  test("basic", (done) => {
    runLegacyMultiWorkspaceTest({
      ctx,
      onInit: async ({}) => {
        await new CreateDailyJournalCommand().run();
        expect(
          getActiveEditorBasename().startsWith("daily.journal")
        ).toBeTruthy();
        done();
      },
    });
  });
```

is in the same block as the other tests? you can run a test individually by following instructions here: [[Run a specific test for Dendron Plugin|dendron.dev.qa#run-a-specific-test-for-dendron-plugin]]

kevinslin (approved):

PR #845, opened 2021-06-21
feat/workspace-trust-for-hooks

jonathanyeung:
Configuration for Dendron plug-in extension to support VS Code's new workspace trust controls. 

Hooks need to be disabled if the workspace is not trusted, as it can run custom code dynamically.  To disable this functionality, this change adds in a new parameter to EngineWriteOptsV2, which can tell the engine write method to not run any hooks as part of the write.  Plugin-Core's `EngineAPIService` can control whether or not this param is set based on workspaceTrust settings within the plug-in environment. 

This deviates from the original design when we had discussed exposing a workspaceTrust setting directly in the Engine API - I think that the engine should remain agnostic to the concept of 'workspace trust', as it is a setting that is _only_ relevant to the VS Code Client, and not the CLI, or other future clients. 

Out of scope in this change - controlling pod behavior; that functionality can come in a subsequent PR.


packages/plugin-core/src/services/EngineAPIService.ts
```diff
@@ -1,12 +1,65 @@
...
+  EngineWriteOptsV2,
+  NoteProps,
+  WriteNoteResp,
+} from "@dendronhq/common-all";
 import { DendronEngineClient, HistoryService } from "@dendronhq/engine-server";
 import path from "path";
 import { DendronWorkspace } from "../workspace";
 
 export class EngineAPIService extends DendronEngineClient {
-  static create({ port }: { port: number | string }) {
```
jonathanyeung, line 6: @kevinslin - there's a minor issue here with this static create Factory method.  Because EngineAPIService extends DendronEngineClient, I can't create a similar override `Create` that returns an `EngineAPIService` as the function signature will no longer match.  That's why I've changed the factory method to `createEngine` instead (a bit hacky).  

The static methods don't gel too well with inheritance; any thoughts on reworking the `DendronEngineClient` to use parameterized constructor overloads instead of a static Factory constructor?

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -709,43 +709,51 @@ export class FileStorage implements DStore {
...
-          severity: ERROR_SEVERITY.MINOR,
-          message: "error with hook",
-          payload: stringifyError(err),
-        })
-    );
-    if (resp instanceof DendronError) {
-      error = resp;
-      this.logger.error({ ctx, error: stringifyError(error) });
+    if (opts?.runHooks === false) {
+      debugger;
```
kevinslin, line 713: don't forget to remove this

packages/engine-test-utils/src/__tests__/engine-server/topics/hooks.spec.ts
```diff
@@ -251,6 +251,52 @@ describe("engine", async () => {
...
+        const vault = _.find(vaults, { fsPath: "vault1" })!;
+        const note = NoteUtils.create({
+          id: "hooked",
+          fname: "hooked",
+          body: "hooked body",
+          vault,
+        });
+        await engine.writeNote(note, { newNode: true, runHooks: false });
+        const ent = engine.notes["hooked"];
+        debugger;
```
kevinslin, line 267: debugger statement

packages/plugin-core/src/services/EngineAPIService.ts
kevinslin: we've been using `create`to try to give a common interface for different clients. we could parameterize the `create` method as well, having it take an optional generic 

```ts
type CreateOpts<T = any> {
  port: number
  extra?: T
}
```

the idea is that `creating` the client should be fairly specific depending on the client which is why we didn't use inheritance

kevinslin (changes requested):
looks good, just minor issue with cleaning up debugger statements. 

would also add a section about workspace trust in the docs -> [[Workbench|dendron.topic.workbench]]

packages/plugin-core/src/services/EngineAPIService.ts
jonathanyeung: I see - the issue is less w/ the params into the `create`, but the fact that the create must return a `DendronEngineClient` and not an `EngineAPIService`. This causes the overridden version of `writeNote` to not get called. 

It still works ok now though because `engineClient` has a public constructor.

kevinslin (approved):

PR #850, opened 2021-06-22
fix: Ensure `runEngineTestV5` always restores mocks

SeriousBug:
Fixes the issue with a failing test causing future tests to fail with sinon errors.


jonathanyeung:
@SeriousBug I think the test failures are due to my change. Let me try to resolve.

kevinslin (changes requested):
instead of a try catch, can we do a try/finally?

kevinslin (approved):

PR #851, opened 2021-06-22
feat: show/hide stubs in graph view

hfellerhoff:
Adds the ability to show/hide stub notes when in graph view. The option can be found under `Filter > Show Stubs`. This only applies to the note graph, and does not appear in the schema graph.

NOTE: This branch is based off of `feature-branch-regex-filter`. This can be changed if need be, but the goal was to limit conflicts between the two branches as some structural things were changed in that branch.


kevinslin (changes requested):
it looks like this has some overlap with the `feat: filter graph by string` pr. can you make sure this is based of the HEAD of that branch?

kevinslin (approved):

PR #854, opened 2021-06-22
fix: updated time should only change if note content changes

kevinslin:
Tired of merge conflicts with syncing? Me too :)


packages/common-server/src/filesv2.ts
```diff
@@ -188,17 +188,22 @@ export function file2NoteWithCache({
...
       // vault can change without note changing so we need to add this
-      note = { ...cache.notes[name].data, body, vault };
+      // add `contentHash` to this signature because its not saved with note
+      note = { ...cache.notes[name].data, body, vault, contentHash: sig };
       return { note, matchHash, noteHash: sig };
     }
   }
   note = string2Note({ content, fname, vault });
+  note.contentHash = sig;
   return { note, matchHash, noteHash: sig };
```
jonathanyeung, line 207: since `note.contentHash` now contains the hash, should we remove the redundant `noteHash` return value?

jonathanyeung (approved):

kevinslin:
Your right. Will remove it

Issue #841, opened 2021-06-18
Broken frontmatter prevents Dendron from initializing


sagarbehere:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x] Misc (Choose this if your not sure)

## Describe the bug

If a file in the vault has broken frontmatter, Dendron does not initialize successfully

## To Reproduce

1. Open an existing file in the vault
2. Introduce an extra : in the id block. If the line is `id: 829639dd-d027-46d3-bf59-fad32ab8b0a7` then change it to `id: : 829639dd-d027-46d3-bf59-fad32ab8b0a7`
3. Save the file. Quit VSCode. Start VSCode again.
4. See error "Issue initializing Dendron" (see attached screenshot)

## Expected behavior

Dendron initializes even if frontmatter is broken. Dendron then notifies the user that file(s) in question seem to have broken frontmatter. In general, unsanitized/incorrect user input in the form of poorly formatted files should not prevent Dendron from initializing.

## Screenshots

<img width="455" alt="Screen Shot 2021-06-18 at 4 38 48 PM" src="https://user-images.githubusercontent.com/1622959/122624382-c2dc9b80-d054-11eb-961f-b7ad4b170fb9.png">

## Desktop (please complete the following information)

- OS: macOS 10.15.7 Catalina
- Browser: n/a
- Version: n/a

## Additional context

N/A

## Dendron Log file
[dendron.log](https://github.com/dendronhq/dendron/files/6679674/dendron.log)

sidwellr:
I had a similar issue today. For me the file dendron.topic.lookup.schemas.md had an extra ">" character in a link: "[[> Dendron: Reload Index|dendron.topic.commands]]"; removing the ">" allowed Dendron to initialize once again. (I don't know how the character got there.)

SeriousBug:
With version `0.48`, both these issues should no longer crash Dendron's initialization.

PR #855, opened 2021-06-22
fix: notes failing parsing shouldn't crash initialization

SeriousBug:
Wrapped parts of note parsing that might throw errors. We collect errors from all broken notes, which then get combined into a composite error if there is more than 1 error. This allows initialization to continue even if some notes are broken.

#841

### DIFF
--- a/packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
+++ b/packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
@@ -1,10 +1,11 @@
 import {
   DVault,
   NotesCacheEntryMap,
+  NoteUtils,
   VaultUtils,
   WorkspaceOpts,
 } from "@dendronhq/common-all";
-import { createLogger, vault2Path } from "@dendronhq/common-server";
+import { createLogger, tmpDir, vault2Path } from "@dendronhq/common-server";
 import {
   FileTestUtils,
   getLogFilePath,
@@ -13,6 +14,7 @@ import {
 import { DendronEngineV2, readNotesFromCache } from "@dendronhq/engine-server";
 import fs from "fs-extra";
 import _ from "lodash";
+import path from "path";
 import { runEngineTestV5 } from "../../engine";
 import {
   ENGINE_CONFIG_PRESETS,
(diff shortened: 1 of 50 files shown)


packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
```diff
@@ -189,3 +185,65 @@ describe("engine, config/", () => {
...
 });
+
+describe("engine, init", () => {
+  test("handles broken frontmatter gracefully", async () => {
+    const dir = tmpDir().name;
+    const fname = "broken-note.md";
+    // Create a workspace with a note that has a broken frontmatter
+    await runEngineTestV5(
+      async ({ wsRoot, vaults }) => {
+        await NoteTestUtilsV4.createNote({
```
kevinslin, line 196: you can write this in `preSetupHook`. that is for doing any setup ahead of vault initialization

packages/engine-test-utils/src/engine.ts
```diff
@@ -116,8 +116,9 @@ export async function setupWS(opts: {
   vaults: DVault[];
   workspaces?: DWorkspace[];
   asRemote?: boolean;
+  wsRoot?: string;
 }) {
-  const wsRoot = tmpDir().name;
+  const wsRoot = opts.wsRoot || tmpDir().name;
```
kevinslin, line 121: `preSetupHook` gets passed `wsRoot` so this won't be necessary

packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
```diff
@@ -189,3 +185,65 @@ describe("engine, config/", () => {
     });
   });
 });
+
+describe("engine, init", () => {
+  test("handles broken frontmatter gracefully", async () => {
```
kevinslin, line 190: instead of writing this directly here, can you add it to ENGINE_PRESETS under INIT? this way, it will test both the engine response and the engine API response in `engine.spec.ts`

kevinslin (changes requested):

packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
SeriousBug: There are actually tests already in `INIT` for notes with broken frontmatters and broken schemas. I deleted this test, and extended the existing ones to check that the initialization completed despite the caught error.

SeriousBug: And thanks for the suggestion, there was actually a bug that occurred with the API response but not engine!

kevinslin: awesome. then our tests are doing their jobs :)

kevinslin (approved):

PR #858, opened 2021-06-23
feat/ajv-to-validate-pods

Harshita-mindfire:
config for Email Publish Pod not implemented, podConfig have additional parameters than EmailPublishConfig (to,from etc).

### DIFF
--- a/packages/common-all/package.json
+++ b/packages/common-all/package.json
@@ -46,7 +46,6 @@
     "@types/title": "^1.0.5",
     "@types/unist": "^2.0.3",
     "@types/yamljs": "^0.2.31",
-    "ajv": "^8.6.0",
     "dropbox": "^4.0.30",
     "github-slugger": "^1.3.0",
     "gray-matter": "^4.0.2",
--- a/packages/common-all/package.json
+++ b/packages/common-all/package.json
@@ -46,6 +46,7 @@
     "@types/title": "^1.0.5",
     "@types/unist": "^2.0.3",
     "@types/yamljs": "^0.2.31",
+    "ajv": "^8.6.0",
     "dropbox": "^4.0.30",
     "github-slugger": "^1.3.0",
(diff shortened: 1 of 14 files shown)


kevinslin (changes requested):
Can you write a helper function for the export config?
You can put it [PodUtils](https://github.com/dendronhq/dendron/blob/feat/ajv-to-validate-pods/packages/pods-core/src/utils.ts#L18:L18)

```ts
class PodUtils {
	...
	static createExportConfig<T extends ExportPodConfig>(opts: {required: string[], properties: any}) => {
		return {
			type: "object",
			additionalProperties: false,
			required: [
				"dest",
				...opts.required,
			],
			properties: {
				dest: {
					type: "string",
					description: "Where to export to",
					default: "Airtable-Workspace",
				},
				includeBody: {
					type: "boolean",
					default: true,
					description: "should body be included",
					nullable: true,
				},
				includeStubs: {
					type: "boolean",
					description: "should stubs be included",
					nullable: true,
				},
				ignore: { type: "array", items: { type: "string" }, nullable: true },
				...opts.properties,
			},
		};
	}
}
```

You can then use it like below

```ts
  get config(): JSONSchemaType<AirtableExportConfig> {
    return createExportConfig<AirtableExportConfig>({required: [...], properties: {...}})
	}
```

Before you do `Publish` and `Import`, can we make sure to finish `Export` and then have the other two as separate PRs once we are good with Export?

Harshita-mindfire:
> Before you do `Publish` and `Import`, can we make sure to finish `Export` and then have the other two as separate PRs once we are good with Export?

Okay, we can revert changes for Import and Publish and create a check of podkind in util.js, since the config of all kinds of pod gets created in `genConfigFile`.

Harshita-mindfire:
created new PR with only changes for `export`

PR #864, opened 2021-06-24
Fix/import pod error reporting

jonathanyeung:
Improving robustness / user experience in Import Pod:
- if the import configuration has invalid YAML, it shows a helpful error message to the info box.
- If any notes have issues with importation, then it won't block the rest of the import - other notes will get imported, and then an error message will show exactly which notes had issues.


jonathanyeung:
oops, forked this branch off unrelated changes.  will rebranch and reopen a new PR.

PR #865, opened 2021-06-24
fix/import pod YML errors

jonathanyeung:
Improving robustness / user experience in Import Pod:

- If the import configuration has invalid YAML, it shows a helpful error message to the info box.
- If any notes have issues with importation, then it won't block the rest of the import - other notes will get imported, and then an error message will show exactly which notes had issues.

### DIFF
--- a/packages/plugin-core/src/commands/ImportPod.ts
+++ b/packages/plugin-core/src/commands/ImportPod.ts
@@ -65,6 +65,9 @@ export class ImportPodCommand extends BaseCommand<
         window.showErrorMessage(
           "The configuration is invalid YAML. Please fix and run this command again."
         );
+      else {
+        throw e;
+      }
       return;
     }
   }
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/commands/ImportPod.ts
```diff
@@ -47,16 +47,26 @@ export class ImportPodCommand extends BaseCommand<
...
+        );
+        return;
+      }
+      return { podChoice, config: maybeConfig };
+    } catch (e) {
+      // The user's import configuration has YAML syntax errors:
+      if (e.name === "YAMLException")
+        window.showErrorMessage(
+          "The configuration is invalid YAML. Please fix and run this command again."
+        );
```
kevinslin, line 67: if error is not a YAML error, we should let it propagate through. commands will log errors and display in UI if uncaught from the child command

kevinslin (changes requested):

packages/plugin-core/src/commands/ImportPod.ts
jonathanyeung: good catch, updated. let me know if there's better methodology on the re-throw; coming from Java/C# I'm finding exception handling to be a quite different in JS haha

kevinslin: re-throw is the typical response. we have two patterns for error handling: throwing and catching in a base class or returning an error object

kevinslin (approved):

PR #866, opened 2021-06-24
chore: including vscode-paste-image in extension pack

jonathanyeung:
Making the paste-image extension a part of the Dendron extension pack so that it will automatically install with Dendron

### DIFF
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -27,7 +27,7 @@
   "main": "./out/src/extension.js",
   "extensionPack": [
     "dendron.dendron-markdown-preview-enhanced",
-    "mushan.vscode-paste-image"
+    "dendron.dendron-paste-image"
   ],
   "contributes": {
     "languages": [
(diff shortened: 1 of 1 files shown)


packages/plugin-core/package.json
```diff
@@ -26,7 +26,8 @@
   ],
   "main": "./out/src/extension.js",
   "extensionPack": [
-    "dendron.dendron-markdown-preview-enhanced"
+    "dendron.dendron-markdown-preview-enhanced",
+    "mushan.vscode-paste-image"
```
kevinslin, line 30: we actually have a fork of this because the original extension isn't published in vscodium registry. change to dendron.dendron-paste-image

kevinslin (changes requested):

packages/plugin-core/package.json
jonathanyeung: Interesting - let me change.

kevinslin (approved):

Issue #800, opened 2021-06-06
Position cursor to end of file when a new note is created.


hikchoi:
## Please Select if your Request is Either something new or an Enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##
Currently when you create a new note, the cursor is placed at the start of the active document. Frontmatter is already populated, so if you start typing right away this will break the frontmatter.

## Describe the solution you'd like ##
Place cursor at the end of file when a new note is created.

## Describe alternatives you've considered ##
The current workflow is to always hold <kbd>Page Down</kbd> after creating a new note. This is often forgotten and will break the flow.

## Additional context ##
I notice this happening a lot to me when I am creating a new journal entry and then using `Dendron: Insert Note` to carry over the last day's entry.

SeriousBug:
This has been added in version `0.48`.

Issue #489, opened 2021-02-16
Option to fold frontmatter automatically


kevinslin:
## Context

Frontmatter takes up space and some people don't use them

## Proposal

Option to make frontmatter fold automatically

wjbuys:
(I still want this feature, bad stalebot!)

kevins8:
Sorry Jacob, stale bot is only supposed to apply to issues that are
unclear. This is still part of our backlog!

xypnox:
Additionally, When the editor opens with a new file created via lookup, the cursor remains in the start of the frontmatter and I have to navigate at the bottom to start writing. It would be better if the cursor was placed after the frontmatter when a new file was opened. I think this should also be doable for existing files but is more relevant for new files.

PR #870, opened 2021-06-25
feat: Focus after the frontmatter when opening a note & option to auto-fold frontmatter

SeriousBug:
1. Added infrastructure to detect when a note is opened for the first time (e.g. creating a new note, or opening an existing note, but not when switching between already open notes). This should assist with any similar features in the future.
2. We now move the selection to just after the frontmatter when opening a note for the first time in this session.
3. Added an option to automatically fold the frontmatter when opening a note for the first time in this session. This is *not* enabled by default.


SeriousBug:
#800 

#489

SeriousBug:
One quirk is that if you had a note that was already open the last time you used VSCode, after you restart VSCode, when you switch to that note we treat it as if you just opened that note for the first time. This is unfortunately a little hard to avoid because VSCode itself treats them as files that are being opened for the first time (e.g. it generates a "file opened" event when switching to these notes).

If we do want to avoid this, we could get around it by saving the set of open files, and skipping ones that were open before VSCode was closed.

packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -100,4 +106,33 @@ export class WorkspaceWatcher {
...
+    this._openedDocuments.set(document.uri.fsPath, document);
+  }
+
+  /** The documents that have been opened during this session that have not been viewed yet in the editor.
+   *
+   * Had to add undefined because
+   */
+  private _openedDocuments: Map<string, TextDocument> = new Map();
+
+  /** Do not use this function, please go to `WindowWatcher.onFirstOpen() instead.`
```
jonathanyeung, line 120: Actually do we need the `workspace.onDidOpenTextDocument` event? Can we just move this openedDocument tracking into windowWatcher and only rely on the `window.onDidChangeActiveTextEditor` event?

SeriousBug: Unfortunately not. `window.onDidChangeActiveTextEditor` activates when switching between editors, so it would fire whenever the user switched to an already open note.

Worse, when `window.onDidChangeActiveTextEditor` activates `window.selection` is not set to the actual user selection in that window yet, so we can't even check if the user had selected something and switched out. 

And when `workspace.onDidOpenTextDocument` activates, the active editor has not switched yet, so we can't just grab the active editor at that point either.

packages/plugin-core/src/utils.ts
```diff
@@ -447,6 +448,31 @@ export class VSCodeUtils {
...
+      point.column - 1 + (offset?.column || 0)
+    );
+  }
+
+  /** Fold the foldable region at the given position for the active editor.
+   *
+   * This is equivalent to selecting that point, and using the "Fold" command in the editor.
+   */
+  static foldActiveEditorAtPosition(position: vscode.Position) {
+    return vscode.commands.executeCommand("editor.fold", position);
```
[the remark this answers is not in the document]
SeriousBug: This was because the folding provider was also folding the line after the frontmatter (off by 1 error). Fixed it.

jonathanyeung (approved):

packages/common-all/src/types/workspace.ts
```diff
@@ -238,6 +238,9 @@ export type DendronConfig = {
    * Configuration for Random Note Lookup Command
    */
   randomNote?: RandomNoteConfig;
+
+  /** Automatically fold frontmatter when opening a new note. False by default. */
+  autoFoldFrontmatter?: boolean;
```
kevinslin, line 243: awesome job with this. only nitpick is to default this to `true`. when possible, we opt for [sensible defaults](https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.principles.md#sensible-defaults)

kevinslin (changes requested):

kevinslin (approved):

PR #872, opened 2021-06-25
feat : ajv to validate export conf

Harshita-mindfire:
Added changes for validating export config.

### DIFF
--- a/packages/pods-core/src/basev3.ts
+++ b/packages/pods-core/src/basev3.ts
@@ -237,11 +237,12 @@ export abstract class ExportPod<
 
   validate(config: Partial<T>) {
     const validateConfig = ajv.compile(this.config);
-    validateConfig(config);
-    if (!validateConfig(config)) {
+    const valid = validateConfig(config);
+    if (!valid) {
+      const errors = ajv.errorsText(validateConfig.errors);
       throw new DendronError({
-        message: "validation errors",
-        payload: "error",
+        message: `validation errors: ${errors}`,
+        payload: `error: ${errors}`,
       });
     }
   }
--- a/packages/pods-core/src/basev3.ts
+++ b/packages/pods-core/src/basev3.ts
@@ -13,6 +13,8 @@ import { Item } from "klaw";
 import _ from "lodash";
 import { URI } from "vscode-uri";
 import { PodKind } from "./types";
+import Ajv, { JSONSchemaType } from "ajv";
+const ajv = new Ajv();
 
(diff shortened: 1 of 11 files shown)


packages/pods-core/src/basev3.ts
```diff
@@ -225,37 +229,19 @@ export abstract class ExportPod<
...
-    const { dest } = config;
-    const configJSON = JSON.stringify(config);
-    if (_.isUndefined(dest)) {
+    const validateConfig = ajv.compile(this.config);
+    validateConfig(config);
+    if (!validateConfig(config)) {
       throw new DendronError({
-        message: `no dest specified. config: ${configJSON}`,
+        message: "validation errors",
+        payload: "error",
```
kevinslin, line 244: can we return the list of errors here as well in `payload`?

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -34,40 +36,36 @@ export class AirtableExportPod extends ExportPod<AirtableExportConfig> {
...
-      },
-      {
-        key: "srcFieldMapping",
-        description:
-          "mapping of airtable fields with the note eg: {Created On: created, Notes: body}",
-        type: "object",
-        required: true,
-      },
-    ]);
+    }) as JSONSchemaType<AirtableExportConfig>;
```
kevinslin, line 68: instead of doing a cast, can `createExportPodConfig` take a generic `T extends ExportPodConfig`?

kevinslin (changes requested):

packages/pods-core/src/builtin/AirtablePod.ts
Harshita-mindfire: Actually no, this results in two errors : incompatible "type" error and value of T never used.
Also, to resolve the " value of T never used" error, if we keep the return type of  createExportPodConfig to be : JSONSchemaType<T>, ajv doesn't allow a generic schema to have a type as "object".

```js
static createExportConfig<T extends ExportPodConfig>(opts: { required: string[]; properties: any }) : JSONSchemaType<T>
 {
    return {
      type: "object",
      ...
}
}
```

packages/pods-core/src/basev3.ts
Harshita-mindfire: Yes. Adding in message as well, since only message is visible to user.

kevinslin (approved):

PR #843, opened 2021-06-21
feat: nextjs based note preview

namjul:
<a href="https://www.loom.com/share/279bf309459143c781fb514712dc17ff">  </a>

PR #880, opened 2021-06-27
Fix contentHas is undefined

kevinslin:


packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -82,9 +81,9 @@ export class WorkspaceWatcher {
       return { changes: [] };
     }
     const match = NoteUtils.RE_FM_UPDATED.exec(content);
-    const noteHash = genHash(content);
     let changes: TextEdit[] = [];
-    if (match && note.contentHash && note.contentHash !== noteHash) {
+
+    if (match && WorkspaceUtils.noteContentChanged({ content, note })) {
```
namjul, line 86: I tried this branch in #843 and after a save `note.contentHash` is still `undefined`. When I run `Dendron: Reload Index` it works only the first time

kevinslin: you're right, there was an issue on saving hash with new content. just fixed with new push

namjul: awesome, this works now!

namjul (approved):

PR #882, opened 2021-06-28
Refactor Dendron Configure command

Corei13:
Demo: https://www.loom.com/share/60fea05f0e0a44538e525a0d0ca2047d


packages/dendron-next-server/pages/vscode/configure.tsx
```diff
@@ -69,24 +117,125 @@ export default function Config({
...
+          return { site: errors };
+        }}
+        validateOnChange={true}
+      >
+        {({ values, errors }) => (
+          <Form {...formItemLayout}>
+            <Typography style={{ textAlign: "center" }}>
+              <Title>Dendron Configuration </Title>
+            </Typography>
+            <Form.Item
```
kevinslin, line 169: one of the reasons to switch to ajv is to generate this instead of manually specifying. 
can you create a task to generate the json schema for `DendronConfig`? 
then use one of nextjs build time features (eg. `getStaticProps`) to read the json schema and generate the form during build time

packages/plugin-core/src/commands/ConfigureWithUI.ts
```diff
@@ -43,7 +45,7 @@ function getWebviewContent2(opts: { title: string }) {
     </style>
   </head>
   <body>
-    <iframe width="100%" height="100%" src="http://localhost:${port}/workspace/config.html"></iframe>
+    <iframe width="100%" height="100%" src="http://localhost:${port}/vscode/configure?ws=${wsRoot}&port=${port}"></iframe>
```
kevinslin, line 48: you shouldn't need to modify this since only the below block will execute when you are running this in dev mode. when this is ready, we'll remove the entire if/else statement

```
    if (getStage() === "dev") {
      resp = WebViewUtils.genHTMLForWebView({
        title: "Dendron Config",
        view: DendronWebViewKey.CONFIGURE,
      });
    } else {
```

kevinslin (changes requested):
looks like its crashing on input right now, see [loom](https://www.loom.com/share/e4150d9c57744d4c930331b39d58f90f)

kevinslin (changes requested):
1. looks like tree view isn't working anymore. can you rebase from master and see if it works? 

2. can we group configuration into the following buckets for the sidebar?

- Workspace (sort - custom, use this sorting)
  - vaults: DVault[];
  - workspaces?: { [key: string]: DWorkspaceEntry | undefined };
  - seeds?: { [key: string]: SeedEntry | undefined };
  - workspaceVaultSync?: DVaultSync;
  - hooks?: DHookDict;
  - initializeRemoteVaults?: boolean;

- Features (sort  - custom)
  - journal: JournalConfig;
  - mermaid?: boolean;
  - randomNote?: RandomNoteConfig;
	- usePrettyRefs?: boolean;
	- useKatex?: boolean;

- Advanced (sort - alphabetical)
  - lookupConfirmVaultOnCreate?: boolean;
  - noAutoCreateOnDefinition?: boolean;
  - noXVaultWikiLink?: boolean;
  - useFMTitle?: boolean;
  - useNoteTitleForLink?: boolean;
  - defaultInsertHierarchy?: string;
  - dev?: DendronDevConfig;
  - noTelemetry
  - noCaching

- Publishing (sort - alphabetical)
	- hierarchyDisplay?: boolean;
  - hierarchyDisplayTitle?: string;
	- site: DendronSiteConfig;

- Don't Show in Form
  - version: number;
  - dendronVersion?: string;
  - useNunjucks?: boolean;
  - noLegacyNoteRef?: boolean;
  - feedback?: boolean;
  - apiEndpoint?: string;

3. for object config like `journal`, there's no indication that we're editing a journal. we should show a header for these object properties
	- https://www.loom.com/i/12e1c828323b43179f649ce2d502dfd6

4. for enums, can we show detail text for what the enum means?
	- https://www.loom.com/i/12e1c828323b43179f649ce2d502dfd6


5. LookupConfirmVaultOnCreate, description text, can we update the docs to be HTML docstring? currently its `Pick vault when creating new note. Docs: https://dendron.so/notes/24b176f1-685d-44e1-a1b0-1704b1a92ca0.html#specify-vault-location-when-creating-a-note`

6. when i hit save changes, there's no progress indicator and the config ui becomes frozen

PR #892, opened 2021-06-29
Native Dendron Preview

kevinslin:


packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -29,8 +30,9 @@ export class WorkspaceWatcher {
       this,
       context.subscriptions
     );
+    //   watcher.onDidChange(_.debounce(this.onDidChange, 500), this)
```
jonathanyeung, line 33: nit: remove commented code.

jonathanyeung (approved):

packages/plugin-core/src/WorkspaceWatcher.ts
kevinslin: done

PR #894, opened 2021-06-30
feat: auto completion for anchors and headers

SeriousBug:
Typing any of the following will now trigger a completion:

```
Completes headers:
[[#

Completes existing block anchors:
[[#^

Completes all blocks:
[[^
```

This is working only for same file links for now (for example, `[[#header]]`. I'll expand it to work with other files while waiting for feedback, and add "complete all files `[[^^`" support in the coming days.


packages/plugin-core/src/test/suite-integ/CompletionProvider.test.ts
```diff
@@ -0,0 +1,264 @@
...
+} from "../../features/completionProvider";
+import { VSCodeUtils } from "../../utils";
+import { CompletionItem, Position } from "vscode";
+import { NoteUtils } from "@dendronhq/common-all";
+import { expect } from "../testUtilsv2";
+import { NoteTestUtilsV4 } from "@dendronhq/common-test-utils";
+
+suite("completionProvider", function () {
+  let ctx = setupBeforeAfter(this, {
+    beforeHook: () => {},
```
kevinslin, line 17: if you don't run anything before, you don't actually need the `beforeHook` method

kevinslin (changes requested):
awesome work. you took this one step further with header completion :)

just one nitpick about extra function and we should be good to go

jonathanyeung (approved):

kevinslin:
This looks good. let me know when you feel ready to move this out of draft status

SeriousBug:
> This looks good. let me know when you feel ready to move this out of draft status

I'm trying to track down an issue where typing `[[#^` completely breaks the autocomplete. It gets stuck saying `Loading...`, and autocomplete stops working completely (for everything, not just blocks), and there are no logs or anything to show what might be wrong. I'll un-draft this and let you know once I fix that.

kevinslin:
roger, sounds good

SeriousBug:
@kevins8 Fixed the issue! Turns out using `document.getWordRangeAtPosition` with a complex regex effectively grinds VSCode to a halt.

SeriousBug:
We now also have support for completing anchors and headers in other files, e.g. `[[note#]]` for headers and `[[note#^]]` for anchors. Unfortunately I don't think we can support completing all blocks in other files or all files due to VSCode limitations, I wrote down my thoughts at: [[Vscode Limitations|proj.block-references.vscode-limitations]]

kevinslin:
> We now also have support for completing anchors and headers in other files, e.g. `[[note#]]` for headers and `[[note#^]]` for anchors. Unfortunately I don't think we can support completing all blocks in other files or all files due to VSCode limitations, I wrote down my thoughts at: [[Vscode Limitations|proj.block-references.vscode-limitations]]

i think you made the right call. to close off this task, can you add global auto complete as a separate task in your weekly journal so we can triage it and put it in the backlog?

kevinslin (changes requested):
a few issues I found after manually testing

1. autocompleting on same block multiple times creates a new anchor each time

```
This is some text ^Rdatb52As7x6QbqwH2Xhk ^tb_l3wdQ0dH38_y5R3ITx

[[#^tb_l3wdQ0dH38_y5R3ITx]]
[[#^Rdatb52As7x6QbqwH2Xhk]]
```

2. goto header in same note doesn't go to the right position

- repro:
	- add the following to [[dendron.ref.links]] in `test-workspace`
	- ctrl+enter
	- also doesn't work after reloaindg owrkspace
```
[[#unpublished-page]]
```

3. goto anchor in same note doesn't go to right position


- repro:
	- add anchor in [[dendron.ref.links]] in `test-workspace`
	- ctrl+enter
	- also doesn't work after reloaindg owrkspace
```
[[#^GvByNb8f]]

## Wiki Link

- [[dendron.ref.figure]]


## XVault Link

- vault1 bar: [[dendron://vault/bar]] ^GvByNb8f
- vault2 bar: [[dendron://vault2/bar]]

```

SeriousBug:
@kevinslin I investigated the issues you reported:

1 and 2: Issues with Goto note with header/anchor I think were related to anchor parsing issues, and seem to be fixed now.

3: The issue with anchors getting generated twice is because we get the blocks from the engine, which has the saved version of the note. If the user gets blocks without saving the note, the blocks will be out-of-date. Some possible solutions:
* For same-file completions, we can perform the get blocks operation on the client side, parsing the text of the open document instead.
* We can force save the note after generating an anchor.
    * This doesn't solve another aspect of the out-of-date contents issue: if the user added a new block and hasn't saved yet, that block will be missing from the autocomplete
* If the `TextDocument.isDirty`, force a save before asking the engine for completions

I think the first option is sufficient. Since it's just for same-file completions, it isn't a very expensive operation.

kevinslin:
- for the duplicate header, a few comments
	- we recently made a [change](https://github.com/dendronhq/dendron/blob/feat/anchor-auto-complete/packages/plugin-core/src/WorkspaceWatcher.ts#L45:L45) to update the note based on text change instead of on save
		- note update happens [here](https://github.com/dendronhq/dendron/blob/feat/anchor-auto-complete/packages/plugin-core/src/services/NoteSyncService.ts#L84:L84)
		- issue is that the current `updateNote` call does not update the note body (even though we send the whole note to the server)
		- we can add an additional prop to [updateNote](https://github.com/dendronhq/dendron/blob/feat/anchor-auto-complete/packages/engine-server/src/engineClient.ts#L350:L350) to take an `updateBody?` parameter that defaults to `false`
	- note that for the above solution, we might still end up with a duplicate anchor if the completion is triggered before the update cycle takes place
		- there's ways to fix this but we should leave that for a phase 2 task  (add to your triage for weekly planning and we can put this on the backlog)
- looks like if i create a header `[[#]]` without completion, it causes an error for `getBlocks` -> https://www.loom.com/share/23536fc01c664418b174ab94377507af
	- subsequent completions result in the following error  `{"ctx":"provideBlockCompletionItems","error":{"isComposite":false,"status":"unknown","payload":"{}","name":"DendronError"},"msg":"Unable to get blocks for autocomplete"}`

SeriousBug:
>         * issue is that the current `updateNote` call does not update the note body (even though we send the whole note to the server)

That doesn't seem to be the case, I can see that the note body is also updated. I switched both `findAnchors` and the completions to use the cached note, and it does get the updated note body.

>     * looks like if i create a header `[[#]]` without completion, it causes an error for `getBlocks` -> https://www.loom.com/share/23536fc01c664418b174ab94377507af

This is because that's a broken wikilink, and the wikilink parser throws an error rather than not parsing it as a wikilink. I noticed this issue and tried a few fixes to reject broken wikilinks during parsing, but the changes I attempted were breaking other features. I'd suggest merging this feature and tackling that separately, which I'd be happy to take on.

kevinslin:
> > ```
> >     * issue is that the current `updateNote` call does not update the note body (even though we send the whole note to the server)
> > ```
> 
> That doesn't seem to be the case, I can see that the note body is also updated. I switched both `findAnchors` and the completions to use the cached note, and it does get the updated note body.
> 
> > ```
> > * looks like if i create a header `[[#]]` without completion, it causes an error for `getBlocks` -> https://www.loom.com/share/23536fc01c664418b174ab94377507af
> > ```
> 
> This is because that's a broken wikilink, and the wikilink parser throws an error rather than not parsing it as a wikilink. I noticed this issue and tried a few fixes to reject broken wikilinks during parsing, but the changes I attempted were breaking other features. I'd suggest merging this feature and tackling that separately, which I'd be happy to take on.

got it. lets take that on as a separate task then. feel free to schedule for the following week. 

as for updating the body, what i mean is that it updates the note on disk but not in memory. using serialize works for now though so lets stick with that for now

kevinslin (approved):

PR #896, opened 2021-06-30
feat/upload telemetry on command invocation

jonathanyeung:
Adds a single telemetry event upon invocation of any command.

### DIFF
--- a/packages/plugin-core/src/commands/base.ts
+++ b/packages/plugin-core/src/commands/base.ts
@@ -3,6 +3,7 @@ import { DLogger } from "@dendronhq/common-server";
 import _ from "lodash";
 import { window } from "vscode";
 import { Logger } from "../logger";
+import { AnalyticsUtils } from "../utils/analytics";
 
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/commands/base.ts
```diff
@@ -56,6 +57,9 @@ export abstract class BaseCommand<
   async run(args?: Partial<TRunOpts>): Promise<TOut | undefined> {
     // @ts-ignore
     const ctx = `${this.__proto__.constructor.name}:run`;
+
```
kevinslin, line 60: to make this slightly more useful, can we track two things for commands?

1. duration for invocation. see [here](https://github.com/dendronhq/dendron/blob/feat/anchor-auto-complete/packages/plugin-core/src/_extension.ts#L393:L393) for example
2. error event. basically add a finally clause and set error to true if there was an error

kevinslin (changes requested):

kevinslin (approved):

PR #902, opened 2021-07-01
fix/correct command name in prod telemetry

jonathanyeung:
Because of JS minifying, the telemetry on command events would be minimized and report incorrectly.


kevinslin (changes requested):
looks like LaunchTutorialCommand is missing the key property :)

The tests are failing

jonathanyeung:
> looks like LaunchTutorialCommand is missing the key property :)
> 
> The tests are failing

derp haha. will fix shortly

kevinslin (approved):

PR #903, opened 2021-07-01
feat: ajv to validate publish pod

Harshita-mindfire:


packages/pods-core/src/builtin/EmailPod.ts
```diff
@@ -15,49 +17,46 @@ export class EmailPublishPod extends PublishPod<EmailPublishConfig> {
...
-        type: "boolean",
-        default: false,
-        example: "you <username@outlook.com>",
+  get config(): JSONSchemaType<EmailPublishConfig> {
+    return PodUtils.createPublishConfig({
+      required: [],
+      properties: {
+        from: {
+          description: "from address",
+          type: "boolean",
```
jonathanyeung, line 26: I know this is copy /paste from previous, but shouldn't these types on to/from address be strings and not booleans?

jonathanyeung (changes requested):

packages/pods-core/src/builtin/EmailPod.ts
Harshita-mindfire: well yeah, that makes sense. Will update it, although this pod might need other changes as well since it is not in action currently.

packages/pods-core/src/builtin/EmailPod.ts
```diff
@@ -15,49 +17,46 @@ export class EmailPublishPod extends PublishPod<EmailPublishConfig> {
...
-        default: false,
-        example: "you <username@outlook.com>",
+  get config(): JSONSchemaType<EmailPublishConfig> {
+    return PodUtils.createPublishConfig({
+      required: [],
+      properties: {
+        from: {
+          description: "from address",
+          type: "string",
+          default: false,
```
jonathanyeung, line 27: Sorry one more small thing - is `false` a valid default value for a string? or does it need to be an empty string or something like that?

Harshita-mindfire: `false` should not be a default value for string. My bad I didn't see it while changing the type from boolean to string.

jonathanyeung (approved):

PR #909, opened 2021-07-02
feat: show unreferenced links in backlink panel

hikchoi:
This PR:
- Adds functionality to show unreferenced links in the backlink panel.

Tasks:
- [x] Find and add unreferenced links to note property.
- [x] Update unreferenced links on store init.
- [x] Update unreferenced links on when note content changes.
- [x] Show unreferenced links in backlinks panel.
- [x] Add onClick behavior to convert unreferenced links into wikilinks.
- [x] Write tests
- [x] Write release notes and documents

### DIFF
--- a/packages/common-all/src/types/foundation.ts
+++ b/packages/common-all/src/types/foundation.ts
@@ -48,7 +48,7 @@ export type DLoc = {
  @deprecated use {@link DNoteLink}
  */
 export type DLink = {
-  type: "ref" | "wiki" | "md" | "backlink" | "unreferenced";
+  type: "ref" | "wiki" | "md" | "backlink" | "linkCandidate";
   value: string;
   alias?: string;
   position: Position;
--- a/packages/common-all/src/types/foundation.ts
+++ b/packages/common-all/src/types/foundation.ts
@@ -48,7 +48,7 @@ export type DLoc = {
  @deprecated use {@link DNoteLink}
  */
 export type DLink = {
-  type: "ref" | "wiki" | "md" | "backlink";
+  type: "ref" | "wiki" | "md" | "backlink" | "unreferenced";
   value: string;
   alias?: string;
   position: Position;
--- a/packages/common-all/src/types/typesv2.ts
+++ b/packages/common-all/src/types/typesv2.ts
@@ -63,7 +63,7 @@ export type DNoteLinkData = {
   xvault?: boolean;
 };
 export type DNoteLink<TData extends DNoteLinkData = DNoteLinkData> = {
-  type: "ref" | "wiki" | "md" | "backlink";
+  type: "ref" | "wiki" | "md" | "backlink" | "unreferenced";
   position?: Position;
   from: DNoteLoc;
   to?: DNoteLoc;
--- a/packages/common-all/src/types/typesv2.ts
+++ b/packages/common-all/src/types/typesv2.ts
@@ -63,7 +63,7 @@ export type DNoteLinkData = {
   xvault?: boolean;
 };
 export type DNoteLink<TData extends DNoteLinkData = DNoteLinkData> = {
-  type: "ref" | "wiki" | "md" | "backlink" | "unreferenced";
+  type: "ref" | "wiki" | "md" | "backlink" | "linkCandidate";
   position?: Position;
   from: DNoteLoc;
   to?: DNoteLoc;
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -273,9 +273,9 @@ export type DendronDevConfig = {
    */
   enableWebUI?: boolean;
   /**
-   * Enable displaying and indexing unreferenced links. Default is false
+   * Enable displaying and indexing link candidates. Default is false
    */
-  enableUnrefLinks?: boolean;
+  enableLinkCandidates?: boolean;
 };
 
 export type DendronSiteConfig = {
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -272,6 +272,10 @@ export type DendronDevConfig = {
    * Enable experimental web ui. Default is false
    */
   enableWebUI?: boolean;
+  /**
+   * Enable displaying and indexing unreferenced links. Default is false
+   */
+  enableUnrefLinks?: boolean;
 };
 
 export type DendronSiteConfig = {
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -333,7 +333,7 @@ export class FileStorage implements DStore {
     );
     const allNotes = _.flatten(out);
     this._addBacklinks({ notesWithLinks, allNotes, allNotesCache });
-    if (this.engine.config.dev?.enableUnrefLinks) {
+    if (this.engine.config.dev?.enableLinkCandidates) {
       const ctx = "_addUnrefLinks";
       const start = process.hrtime();
       this._addUnrefLinks(allNotes);
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -333,6 +333,13 @@ export class FileStorage implements DStore {
     );
     const allNotes = _.flatten(out);
     this._addBacklinks({ notesWithLinks, allNotes, allNotesCache });
+    if (this.engine.config.dev?.enableUnrefLinks) {
+      const ctx = "_addUnrefLinks";
+      const start = process.hrtime();
+      this._addUnrefLinks(allNotes);
+      const duration = getDurationMilliseconds(start);
+      this.logger.info({ ctx, duration });
+    }
     return { notes: allNotes, errors };
   }
 
@@ -371,6 +378,23 @@ export class FileStorage implements DStore {
         return;
       }
     });
+  }
+  
+  _addUnrefLinks(allNotes: NoteProps[] ) {
+    return _.map(allNotes, (noteFrom) => {
+      try {
+        const unreferencedLinks = LinkUtils.findUnreferencedLinks({
+          note: noteFrom,
+          notes: allNotes,
+          engine: this.engine,
+        });
+        noteFrom.links = noteFrom.links.concat(unreferencedLinks);
+      } catch (err) {
+        const error = error2PlainObject(err);
+        this.logger.error({ error, noteFrom, message: "issue with unref link" });
+        return;
+      }
+    })
   }
 
   async _initNotes(vault: DVault): Promise<{
@@ -447,34 +471,7 @@ export class FileStorage implements DStore {
             errors.push(err);
             return;
           }
-          if (this.engine.config.dev?.enableUnrefLinks) {
-            try {
-              let ctx = "findUnreferencedLinks";
-              let duration;
-              const start = process.hrtime();
 
-              const unreferencedLinks = LinkUtils.findUnreferencedLinks({
-                note: n,
-                notes: notes,
-                engine: this.engine,
-              });
-              cacheUpdates[n.fname].data.links =
-                cacheUpdates[n.fname].data.links.concat(unreferencedLinks);
-              n.links = n.links.concat(unreferencedLinks);
-
-              duration = getDurationMilliseconds(start);
-              this.logger.info({ ctx, duration });
-            } catch (err) {
-              if (!(err instanceof DendronError)) {
-                err = new DendronError({
-                  message: `Failed to read unreferenced links in note ${n.fname}`,
-                  payload: err,
-                });
-              }
-              errors.push(err);
-              return;
-            }
-          }
         } else {
           n.links = cache.notes[n.fname].data.links;
         }
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -334,9 +334,9 @@ export class FileStorage implements DStore {
     const allNotes = _.flatten(out);
     this._addBacklinks({ notesWithLinks, allNotes, allNotesCache });
     if (this.engine.config.dev?.enableLinkCandidates) {
-      const ctx = "_addUnrefLinks";
+      const ctx = "_addLinkCandidates";
       const start = process.hrtime();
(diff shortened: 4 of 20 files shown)


hikchoi:
Still a lot to work on, but drafting this PR for timely communication and visibility.
Will add more tests and also rebase so the conflict goes away.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -411,9 +411,14 @@ export class FileStorage implements DStore {
               fname: n.fname,
             }
           );
-          cacheUpdates[n.fname].data.links = links;
+          const unreferencedLinks = await LinkUtils.findUnreferencedLinks({
```
kevinslin, line 414: would be great if we could do this asynchronously and not need to `await` on this.

ideally, we would setup a websocket connection between engine server and the plugin so engine can notify the plugin when its finished long running tasks (eg. indexing all links)
not in scope for this issue. for now, I would put unreferenced links behind a feature flag (eg. `beta.enableUnreferedLinks`)

kevinslin: would be good to have profiling information around this. see profile instructions here -> [[Profiling|dendron.dev.qa#profiling]]

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -179,6 +179,47 @@ const getLinks = ({
...
+  visit(ast, [DendronASTTypes.TEXT], (node) => {
+    switch (node.type) {
+      case DendronASTTypes.TEXT:
+        textNodes.push(node);
+    }
+  });
+  const unreferencedLinks: DLink[] = [];
+
+  _.forEach(notes, (n) => {
+    const fname = n.fname;
```
kevinslin, line 201: doesn't need to be implemented now but we might be able to optimize in the future by creating a massive regex 

```ts
// pseudo code
// take 100 notes
notes.batch(100).map(batch => {
	re = RegExp(batch.notes.map(n => n.fname).join("|"))
	re.matchAll(text)
	...
})
```

kevinslin (commented):
left some comments, mostly optimizations and thoughts of how this can be integrated

packages/engine-server/src/markdown/remark/utils.ts
SeriousBug: Another option would be creating a prefix-tree data structure with the `fname`s: https://en.wikipedia.org/wiki/Trie

SeriousBug: Also, here's a simple but more efficient way to implement this:

1. Create a map of `fname`s to notes
2. After getting all the `Text` nodes from a note, split the text into words, and test if each word appears in the map.

This would have `O(n.logn)` running time, as opposed to the quadratic `O(n^2)` running time of this implementation. The only drawback is that this assumes `fname`s can't contain spaces.

It is possible to allow this to work with `fname`s containing spaces too: use the first word of the `fname` as the map key, and the list of notes with that word as the first word of their `fname` as the map value. When parsing for unlinked references, if a lookup in this map is true, then search through the list in the value for matches. This way you don't have to search through every existing `fname`, only a few ones that have the same word as the first word of their `fname`.

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -179,6 +179,47 @@ const getLinks = ({
...
+      const match = value.indexOf(fname);
+      if (match !== -1) {
+        unreferencedLinks.push({
+          type: "unreferenced",
+          from: NoteUtils.toNoteLoc(note),
+          value: value,
+          position: textNode.position as Position,
+          to: {
+            fname: fname,
+            vaultName: n.vault.fsPath,
```
SeriousBug, line 214: Should be `VaultUtils.getName(n.vault)`

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -179,6 +179,47 @@ const getLinks = ({
...
+const getUnreferencedLinks = ({
+  ast,
+  note,
+  notes,
+}: {
+  ast: DendronASTNode;
+  note: NoteProps;
+  notes: NoteProps[];
+}) => {
+  const textNodes: Node[] = [];
```
SeriousBug, line 191: Should use type `Text` from `mdast`

SeriousBug (changes requested):
The `vaultName` change is important, the others are just suggestions.

packages/engine-server/src/markdown/remark/utils.ts
hikchoi: Thanks for the feedback. I went with @SeriousBug 's second suggestion.

I definitely need to break out of my sloppy habits and embrace FP 😅

packages/engine-server/src/drivers/file/storev2.ts
hikchoi: > for now, I would put unreferenced links behind a feature flag (eg. beta.enableUnreferedLinks)

Should this be a `DendronDevConfig` ?

> would be good to have profiling information around this.

I'll add some profiling around the existing implementation and move on to your suggested async implementation right after I get the unref links to display on the backlink panel.

hikchoi: actually, that function didn't even need the `await`. It's a remnant of an older code that I had that I forgot to remove. 😅 

I'll put this behind a flag for now and after the minimum viable feature ships I'll work on fixing it to your suggestion.

jonathanyeung:
One question - what if a note that contains unreferenced links gets modified? Do we have a mechanism to invalidate the link cache of the referenced note?

hikchoi:
> * What if a note that contains unreferenced links gets modified? Do we have a mechanism to invalidate the link cache of the referenced note?

Yes. It's not pushed here yet but it will be updating the unreferenced links on change.

> * Any issues if someone with a lot of notes creates a note with a very generic name, like "the" or "a"?  Any potential overflow or perf concerns with that?

That is definitely an issue. I have made some improvements to the current logic that is not pushed yet but like Kevin suggested [here](https://github.com/dendronhq/dendron/pull/909#discussion_r663120576) there are still a bit more to improve in terms of performance. I'll test out the case you mentioned as-is and report back.

hikchoi:
@jonathanyeung suggested that we change the name of the panel now that it's displaying unreferenced links. Also pointed out that the wording `unreferenced links` is a bit unintuitive. Obsidian calls them `mentions`.

I agree that it's a bit confusing. Open for suggestions.

kevinslin:
> @jonathanyeung suggested that we change the name of the panel now that it's displaying unreferenced links. Also pointed out that the wording `unreferenced links` is a bit unintuitive. Obsidian calls them `mentions`.
> 
> I agree that it's a bit confusing. Open for suggestions.

Hmm, would "unlinked backlinks" not make sense?

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
```diff
@@ -18,6 +20,129 @@ class Backlink extends vscode.TreeItem {
...
+
+  if (!pathsSorted.length) {
+    return [];
+  }
+
+  const collapsibleState = true
+    ? vscode.TreeItemCollapsibleState.Collapsed
+    : vscode.TreeItemCollapsibleState.Expanded;
+
+  return pathsSorted.map((pathParam) => {
```
kevinslin, line 50: when backlinks aren't enabled, we'll show links(0) in the backlinks panel if only unreferenced links exist. we should show no result instead

packages/plugin-core/src/test/suite-integ/BacklinksTreeDataProvider.test.ts
```diff
@@ -97,6 +101,53 @@ suite("BacklinksTreeDataProvider", function () {
...
+          },
+          { wsRoot }
+        );
+        const isUnrefLinkEnabled = TestConfigUtils.getConfig({ wsRoot }).dev
+          ?.enableUnrefLinks;
+        expect(isUnrefLinkEnabled).toBeTruthy();
+
+        // currently unref links are not getting populated on the first engine init for some reason.
+        // FIX: once above problem is fixed, fix this test.
+        await NOTE_PRESETS_V4.NOTE_WITH_UNREF_TARGET.create({
```
kevinslin, line 131: might be because of the race condition mentioned earlier as we're checking for unlinked references while notes are still being indexed

kevinslin (changes requested):

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -446,6 +447,34 @@ export class FileStorage implements DStore {
             errors.push(err);
             return;
           }
+          if (this.engine.config.dev?.enableUnrefLinks) {
+            try {
+              let ctx = "findUnreferencedLinks";
+              let duration;
+              const start = process.hrtime();
+
+              const unreferencedLinks = LinkUtils.findUnreferencedLinks({
```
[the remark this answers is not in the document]
hikchoi: the result of `findLinks` is set to `cacheUpdates[n.fname].data.links` before this and I'm adding the unref links to that. So it will be added to the cache every time, but only after it was overwritten with the result of `findLinks`

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -446,6 +447,34 @@ export class FileStorage implements DStore {
             errors.push(err);
             return;
           }
+          if (this.engine.config.dev?.enableUnrefLinks) {
```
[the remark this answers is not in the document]
hikchoi: ah, this might be why I was having problem with the unref links not being added

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
hikchoi: I just noticed this too while using the panel. I'll fix this.

hikchoi:
- took findUnreferencedLinks out so that it runs after note index
- changed it so that the links subtree doesn't display when links don't exist
- fixed a few tests

packages/plugin-core/src/test/suite-integ/BacklinksTreeDataProvider.test.ts
```diff
@@ -126,13 +130,6 @@ suite("BacklinksTreeDataProvider", function () {
...
         expect(isUnrefLinkEnabled).toBeTruthy();
 
-        // currently unref links are not getting populated on the first engine init for some reason.
-        // FIX: once above problem is fixed, fix this test.
-        await NOTE_PRESETS_V4.NOTE_WITH_UNREF_TARGET.create({
-          wsRoot,
-          vault: vaults[0],
-        });
-        // TEMP: need to re-init.
         await new ReloadIndexCommand().execute();
```
kevinslin, line 133: do we still need a reloadindex command here?

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -199,6 +200,55 @@ const getLinks = ({
...
+    ast,
+    [DendronASTTypes.TEXT],
+    (node: Text, _index: number, parent: Parent | undefined) => {
+      if (parent?.type === "paragraph" || parent?.type === "tableCell") {
+        textNodes.push(node);
+      }
+    }
+  );
+
+  const fnameNoteMap = new Map<string, NoteProps>();
```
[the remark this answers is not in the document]
kevinslin: since this is in the critical path for initialization, can we just pass in the list of fnames f its not a big change?

kevinslin (changes requested):
For naming, unless we have a better suggestion, lets go with the following:
- Backlinks:Linked
- Backlinks:Candidates

packages/plugin-core/src/test/suite-integ/BacklinksTreeDataProvider.test.ts
hikchoi: yes, unlink refs would not have been added because we waited to set the `enableUnlinkRef` flag to true until this point. We need to reload the index so it does it for the first time.

packages/engine-server/src/markdown/remark/utils.ts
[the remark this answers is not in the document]
hikchoi: sure. instead of making the map once elsewhere and passing it into `getUnreferencedLinks`?

hikchoi:
Just pushed: 
- A bit of optimization to `getUnreferencedLinks`
- Fixed `links` to `Linked`, and `unreferenced` to `candidates`
- A few minor bug fixes

docs draft can be found in `[[backlinks panel docs draft|scratch.2021.07.14.161023.backlinks-panel-docs-draft]]`

packages/common-all/src/types/foundation.ts
```diff
@@ -48,7 +48,7 @@ export type DLoc = {
  @deprecated use {@link DNoteLink}
  */
 export type DLink = {
-  type: "ref" | "wiki" | "md" | "backlink";
+  type: "ref" | "wiki" | "md" | "backlink" | "unreferenced";
```
kevinslin, line 51: change this to `linkCandidate`

packages/common-all/src/types/typesv2.ts
```diff
@@ -63,7 +63,7 @@ export type DNoteLinkData = {
   xvault?: boolean;
 };
 export type DNoteLink<TData extends DNoteLinkData = DNoteLinkData> = {
-  type: "ref" | "wiki" | "md" | "backlink";
+  type: "ref" | "wiki" | "md" | "backlink" | "unreferenced";
```
kevinslin, line 66: see above

packages/common-all/src/types/workspace.ts
```diff
@@ -272,6 +272,10 @@ export type DendronDevConfig = {
    * Enable experimental web ui. Default is false
    */
   enableWebUI?: boolean;
+  /**
+   * Enable displaying and indexing unreferenced links. Default is false
+   */
+  enableUnrefLinks?: boolean;
```
kevinslin, line 278: change to `enableLinkCandidates`

kevinslin (changes requested):
this looks great. just a few minor naming changes and we should be good to go

hikchoi:
There were a lot more mentions of `unref`s sprinkled around. Renamed all of those appropriately as well.

kevinslin (approved):

PR #911, opened 2021-07-02
fix: block anchors inside links are parsed as anchors

SeriousBug:
Migrated block anchor parsing to V5 parsing which seems to have solved the issue.

### DIFF
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -430,16 +430,10 @@ export class FileStorage implements DStore {
             return;
           }
           try {
-            const anchors = await AnchorUtils.findAnchors(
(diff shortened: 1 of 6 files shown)


kevinslin (approved):
can we add a test case with an anchor inside a link and show how it is not parsed as an anchor?

SeriousBug:
> can we add a test case with an anchor inside a link and show how it is not parsed as an anchor?

There is actually already a test for this (unless the checks inside the test weren't enough, I'll look at it to make sure). I think what happened is that the tests were using a proc different from what `findAnchors` was using, so they didn't catch some change to `procParse` that broke the parsing. The tests use the same proc as `findAnchors` now, and I'll also go ahead and add some tests for `findAnchors` if there aren't any.

PR #912, opened 2021-07-02
chore: windowwatcher tests failing on macos github actions

SeriousBug:


SeriousBug:
The failing test is the one where it tests switching from and back to an already open note, to test that it will not move the cursor. The specific failure is that the cursor does move, as if the note was opened for the first time.

I re-read the code, and the only reason why this test would fail is if a `onDidOpenTextDocument` event gets created when switching back to the note. Perhaps there could be some cases where VSCode decides to close the previous document (say, if the container/VM running the action is low on memory), in which case this test would fail. I tried re-running the github action several times, but I was unable to trigger the failure.

My final thought would be that we can skip this test on macos, which is where the failure seem to be always happening. Otherwise, I added some logging which would confirm if a `onDidOpenTextDocument` event gets generated. I'm not sure if those logs would show up in the Github action output, but if it does that might clarify the issue. Other than that, I am out of ideas, and will be dropping any further investigation for now.

kevinslin:
the reason the test is not failing on re-run of the workflow is because i [skipped it](https://github.com/dendronhq/dendron/blob/release/0.48.1/packages/plugin-core/src/test/suite-integ/WorkspaceWatcher.test.ts#L41:L41) in master. going to unskip with your logs and see if run into anything new

kevinslin (approved):

PR #918, opened 2021-07-04
chore: add `flavor` argument to `ProcOpts`

kevinslin:
Tells processor if we are targeting a specific output (eg. publishing vs preview). Depending on the flavor, additional processing rules may apply. 

You can see the details on [[Flavors|proj.2021-07-unified-refactoring#flavors]]


packages/engine-server/src/markdown/utilsv5.ts
```diff
@@ -66,6 +84,7 @@ export type ProcDataFullOptsV5 = {
   engine?: DEngineClient;
   vault?: DVault;
   fname?: string;
+  wsRoot?: string;
```
jonathanyeung, line 87: Why is this param necessary? I find it a bit confusing - can this not be inferred from either DEngineClient? 

What does it mean if wsRoot != engine.wsRoot?

kevinslin: this is  for handling overrides and using processor when the full engine isn't available. see [[duplicate properties between|proj.2021-07-unified-refactoring#duplicate-properties-between]]

jonathanyeung: I see - can we add some comments to the type that details some of the information in [[duplicate properties between|proj.2021-07-unified-refactoring#duplicate-properties-between]]? 

But also - do we actually need a unified Type `ProcDataFullOptsV5` to contain the superset of options across all four types of procs? Why can't each one have a tailored set opts with what's required for each proc type? 

Right now, it looks like this is an exception driven workflow (we'll throw if a required data param is missing). Is there anything preventing us from making it a compile-time check instead?

kevinslin: I see - can we add some comments to the type that details some of the information in [[duplicate properties between|proj.2021-07-unified-refactoring#duplicate-properties-between]]?

> yep, will do

But also - do we actually need a unified Type ProcDataFullOptsV5 to contain the superset of options across all four types of procs? Why can't each one have a tailored set of opts with what's required for each proc type?

> so currently there are just two permutations: no data or all data. `ProcDataFullOptsV5` has a bunch of optional properties to specify overrides. when accessing these properties from within the processor, `wsRoot` is preferred over `engine.wsRoot`. the properties currently are actually used in all four types of procs

Right now, it looks like this is an exception driven workflow (we'll throw if a required data param is missing). Is there anything preventing us from making it a compile-time check instead?

> it is. we could make this a compile time workflow by creating a custom function for each mode.
```ts
procRemarkParseNoData
procRemarkParseAllData
```

kevinslin:
since this is already merged into master for the weekly release, i've made the changes in this [pr](https://github.com/dendronhq/dendron/pull/931)

SeriousBug (approved):

PR #919, opened 2021-07-04
feat(preview): better image support

kevinslin:
In order to render images, we need to re-write image urls to calls to localhost and also make it so the server can return images. 

Full details written here -> [[Images|pkg.dendron-next-server.arch#images]]


packages/common-all/src/api.ts
```diff
@@ -257,13 +266,34 @@ abstract class API {
     }
     return payload;
   }
+
+  async _makeRequestRaw(args: IDoRequestArgs) {
```
jonathanyeung, line 270: Why is this method necessary? Is this not roughly the same logic as `_makeRequest()` without `payloadData` specified?

packages/api-server/src/modules/assets.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  static instance() {
+    if (!AssetsController.singleton) {
+      AssetsController.singleton = new AssetsController();
+    }
+    return AssetsController.singleton;
+  }
+
+  async get({ fpath, ws }: AssetGetRequest): Promise<RespV2<string>> {
+    const engine = await getWS({ ws });
```
jonathanyeung, line 21: not related to this review but - should we rename `getWS()` to `getDEngine()`?  I find it a bit confusing that this method returns a `DEngine` and not a `DWorkspace`

Also - getWS() takes a workspace root as a param. Do we expect the api-server to potentially have multiple `DEngine` instances? Currently, it seems 1-1, given that `WorkspaceController` is a singleton.

If the answer is no, then should we remove `ws` from being a parameter to this function?

packages/api-server/src/modules/assets.ts
```diff
@@ -0,0 +1,38 @@
...
+    const { wsRoot, vaults } = engine;
+    if (!WorkspaceUtils.isPathInWorkspace({ wsRoot, vaults, fpath })) {
+      return {
+        error: DendronError.createFromStatus({
+          status: ERROR_STATUS.INVALID_CONFIG,
```
jonathanyeung, line 26: Is `INVALID_CONFIG` the correct error here? Seems either DNE or something related to missing asset is more appropriate?

packages/engine-server/src/markdown/remark/dendronPreview.ts
```diff
@@ -0,0 +1,51 @@
...
+import visit from "unist-util-visit";
+import { VFile } from "vfile";
+import { EngineUtils } from "../../utils";
+import { DendronASTTypes } from "../types";
+import { MDUtilsV5 } from "../utilsv5";
+import { NoteRefsOpts } from "./noteRefs";
+
+type PluginOpts = NoteRefsOpts & {};
+
+function handleImage(proc: Unified.Processor, node: Image) {
```
jonathanyeung, line 17: Not related to this code here, but do we have metrics implemented around the time it takes to finish rendering a preview?

packages/common-all/src/api.ts
kevinslin: the return signature is different. payload is an object of {error: Error, data: T}. when returning a raw request, we are returning either Error|Buffer

packages/api-server/src/modules/assets.ts
kevinslin: 1. your right about the rename. it is confusing. will implement that in a separate review
2. the reason it takes a workspace is some pre-mature optimziation. the idea is in the future, instead of starting the workspace engine with vscode, the dendron engine will run as a daemon when you start up the computer. this will make it possible to integrate it with any running process (eg. chrome extension, alfred, other apps) as well as do indexing ahead of time instead of when launchig

packages/api-server/src/modules/assets.ts
kevinslin: this is protecting against bad access. one could you the `getAsset` api to retrieve the contents of any file on your system - this limits it to only paths underneath your workspace

packages/engine-server/src/markdown/remark/dendronPreview.ts
kevinslin: nope, but will add that as a next task

packages/api-server/src/modules/assets.ts
jonathanyeung: can someone call this API with an fpath that doesn't point to a valid location? That seems to be a possibility right?

jonathanyeung (approved):
One minor follow up comment/question on the `ERROR_STATUS.INVALID_CONFIG`

packages/api-server/src/modules/assets.ts
kevinslin: yep, that's a possibility as well. will add an additional `DNE` check

kevinslin: see this [pr](https://github.com/dendronhq/dendron/pull/931) for changes

PR #920, opened 2021-07-04
Multiple Preview Fixes

kevinslin:
Adds the following missing features to the preview

- better block quote support
- add mermaid support
- add base css 
- support katex rendering
- respect fmTitle heading

### DIFF
--- a/packages/dendron-next-server/styles/scss/content.scss
+++ b/packages/dendron-next-server/styles/scss/content.scss
@@ -1,0 +1,196 @@
+@charset "UTF-8";
+
+//
+// Styles for rendered markdown in the .main-content container
+//
+// stylelint-disable selector-no-type, max-nesting-depth, selector-max-compound-selectors, selector-max-type
+
+.main-content {
+  line-height: $content-line-height;
+
+  ol,
+  ul,
+  dl,
+  pre,
+  address,
+  blockquote,
+  .table-wrapper {
+    margin-top: 0.5em;
+  }
+
+  a {
+    overflow: hidden;
+    text-overflow: ellipsis;
+    // white-space: nowrap;
+  }
+
+  ul,
+  ol {
+    padding-left: 1.5em;
+  }
+
+  li {
+    .highlight {
+      margin-top: $sp-1;
+    }
+  }
+
+  ol {
+    list-style-type: none;
+    counter-reset: step-counter;
+
+    > li {
+      position: relative;
+
+      &::before {
+        position: absolute;
+        top: 0.2em;
+        left: -1.6em;
+        color: $grey-dk-000;
+        content: counter(step-counter);
+        counter-increment: step-counter;
+        @include fs-3;
+
+        @include mq(sm) {
+          top: 0.11em;
+        }
+      }
+
+      ol {
+        counter-reset: sub-counter;
+
+        li {
+          &::before {
+            content: counter(sub-counter, lower-alpha);
+            counter-increment: sub-counter;
+          }
+        }
+      }
+    }
+  }
+
+  ul {
+    list-style: none;
+
+    > li {
+      &::before {
+        position: absolute;
+        margin-left: -1.4em;
+        color: $grey-dk-000;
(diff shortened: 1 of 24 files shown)


jonathanyeung (approved):
Changes looking good!  I didn't scrutinize the .scss files too closely though.

packages/dendron-next-server/styles/scss/content.scss
```diff
@@ -0,0 +1,215 @@
...
+
+  cite {
+    font-style: italic;
+
+    &:before {
+      content: "\2014";
+      padding-right: 5px;
+    }
+  }
+}
```
namjul, line 215: Many hardwired style tokens in this file. I am guessing that is ok for now and at some point we want to have a more coherent system across at least the markdown preview and the published site. And maybe but I think less important also unified with the components that make up the vscode UI.

namjul: Maybe a good starting place would be to have a single source for style tokens by partially using https://github.com/system-ui/theme-specification inside `_variables.scss `

packages/dendron-next-server/styles/scss/support/_variables.scss
```diff
@@ -0,0 +1,135 @@
+//
```
namjul, line 1: I am confused about this second `variables.scss` file. What does the `support` folder imply?

packages/dendron-next-server/styles/scss/content.scss
namjul: I am just seing that we are doing this already in this PR. Keeping it for record.

packages/dendron-next-server/styles/scss/content.scss
```diff
@@ -0,0 +1,215 @@
...
+  h3,
+  h4,
+  h5,
+  h6 {
+    position: relative;
+    margin-top: 1.5em;
+    margin-bottom: 0.25em;
+
+    &:first-child {
+      margin-top: $sp-2;
```
namjul, line 182: Some places we use tokens other not.

packages/dendron-next-server/styles/scss/support/_variables.scss
```diff
@@ -0,0 +1,135 @@
...
+  sp-1: $spacing-unit * 0.25,
+  sp-2: $spacing-unit * 0.5,
+  sp-3: $spacing-unit * 0.75,
+  sp-4: $spacing-unit,
+  sp-5: $spacing-unit * 1.5,
+  sp-6: $spacing-unit * 2,
+  sp-7: $spacing-unit * 2.5,
+  sp-8: $spacing-unit * 3,
+  sp-9: $spacing-unit * 3.5,
+  sp-10: $spacing-unit * 4,
```
namjul, line 87: I like the values choosen.

namjul (commented):
I am wondering what the intent is with how this PR adds new css. It has its own flavor or does it reference an existing standard (eg tailwind)? Are there rules to follow when adding css? It looks to me we sattle on scss for the markdown-preview and publishing and antd for UI components.

packages/dendron-next-server/styles/scss/content.scss
kevinslin: our css is... a mess. it is currently a frankenstein of the following two projects:
- https://github.com/pmarsceill/just-the-docs
- https://github.com/mmistakes/minimal-mistakes

the css you see in this review is what was selectively migrated from [dendron-11ty](https://wiki.dendron.so/notes/rIUY5NLx0inCdmRLJOFaq.html), our current publishing template. 
if you at some point want to do a refactoring/unification of our styling, all I can say is yes, please!

namjul: Yes, I would enjoy that. Maybe a good next step would be to create an RFC.

kevinslin: that would be great. will leave it to then for the rfc :)

PR #933, opened 2021-07-06
feat: improved block anchor in publishing

SeriousBug:
Block anchors are now attached to the actual block they reference, and appear similar to the anchors for headers. See below:

https://www.loom.com/share/bcc872131d3e4aaca92e82f176d9dffa


kevinslin (approved):
awesome ux improvement!

PR #937, opened 2021-07-07
feat: de-emphasize block anchors in the editor

SeriousBug:
Parses the document to decorate block anchors. By parsing the note to provide these decorations, we ensure that only well-formed anchors will be highlighted, and the decorations will update automatically if we change anything about block anchors in parsing. Works well with both dark and light mode:

This also sets up an opportunity to decorate other things in the editor: for example we can highlight unlinked wikilinks, or color the alias of links differently etc. These highlights give the user immediate feedback, so I think it would be very useful.


kevinslin (changes requested):
great change! is it possible to get decorators programmatically in vscode? if so, we should use that mechanism to add a test case in `src/test/suite-integ/WindowWatcher.test.ts`

if vscode doesn't make that easy, we can use the pattern [here](https://github.com/dendronhq/dendron/blob/feat/syntax-highlight-editor/packages/plugin-core/src/WorkspaceWatcher.ts#L119:L119) to return changes applied

packages/plugin-core/src/test/suite-integ/WindowDecorations.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import _ from "lodash";
+
+/** Check if the ranges decorated by `decorations` contains `text` */
+function assertTextDecorated(
+  text: string,
+  decorations: vscode.DecorationOptions[],
+  document: vscode.TextDocument
+) {
+  for (const decoration of decorations) {
+    if (document.getText(decoration.range) === text) return;
```
kevinslin, line 18: nitpick. can we use the expect(...).toEqual(...) syntax?
this way, we get proper reporting on number of tests run

kevinslin (changes requested):

kevinslin (approved):

Issue #917, opened 2021-07-04
Direct Child filter should ignore stubs


kpathakota:
## Please Select if your Request is Either something new or an Enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

When filtering out direct child notes, stubs should be ignored since they clutter up lookup.

SeriousBug:
This is now released for 0.50.

PR #940, opened 2021-07-07
feat: Direct child filter ignores stubs

SeriousBug:
#917


kevinslin (approved):
you are on a streak 💯  thanks for picking up the crop this week :)

PR #943, opened 2021-07-08
Feat/GitHub import pod

Harshita-mindfire:
https://github.com/dendronhq/dendron-site/pull/140


packages/pods-core/src/builtin/GithubPod.ts
```diff
@@ -0,0 +1,318 @@
...
+              }
+              body
+            }
+          }
+        }
+      }
+    }`;
+    try {
+      result = await graphql(query, {
+        headers: { authorization: `token ${token}` },
```
jonathanyeung, line 122: Is an auth token required even for a public repository?  If it's not required, maybe better to make it an optional param.

packages/pods-core/src/builtin/GithubPod.ts
```diff
@@ -0,0 +1,318 @@
...
+  getDataFromGithub = async (opts: Partial<GithubAPIOpts>) => {
+    let result;
+    const { owner, repository, status, created, afterCursor, token } = opts;
+    let queryVal = `repo:${owner}/${repository} is:issue is:${status}`;
+    if (!_.isUndefined(created)) {
+      queryVal = queryVal.concat(` ${created}`);
+    }
+
+    const query = `query search($val: String!, $after: String)
+    {search(type: ISSUE, first: 100, query: $val, after: $after) {
```
jonathanyeung, line 95: nit: extract 100 to const with name PAGE_SIZE or something similiar (I assume this is the page size?)

packages/engine-test-utils/src/__tests__/pods-core/GithubPod.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+              },
+            },
+          },
+        ],
+      },
+    };
+    fname = "github.issues.902-Test Issue";
+  });
+
+  test("Bad Credentials", async () => {
```
jonathanyeung, line 39: Does this test make an actual API call to Github? If so, it's better to tag this test as test.skip in the main test pass, or else it may cause some test reliability issues during the test pass.

jonathanyeung (changes requested):
Changes looking good! Just some minor comments, thanks.

packages/pods-core/src/builtin/GithubPod.ts
Harshita-mindfire: Yes the token is required since unlike the rest API, GitHub only allows requests at GraphQL API when it's has a token for authentication.

packages/pods-core/src/builtin/GithubPod.ts
Harshita-mindfire: okay. Page size is actually 30. 100 is the max no of data that can be fetched at a time.

packages/engine-test-utils/src/__tests__/pods-core/GithubPod.spec.ts
Harshita-mindfire: Yes it would try to make an API call to github. I'll add test.skip to avoid reliability issues. Thanks for catching this

packages/pods-core/src/builtin/GithubPod.ts
```diff
@@ -0,0 +1,318 @@
...
+  Time,
+  DEngineClient,
+} from "@dendronhq/common-all";
+
+const ID = "dendron.github";
+
+type GithubImportPodCustomOpts = {
+  owner: string;
+  repository: string;
+  to?: string;
```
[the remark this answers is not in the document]
Harshita-mindfire: Okay. Would it be better if I rename it to startDate and endDate?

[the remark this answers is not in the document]
jonathanyeung: o - sure thing, yes that'd be better if it's YYYY-MM-DD.

packages/pods-core/src/builtin/GithubPod.ts
jonathanyeung: o I see - does 100 represent the max number of issues to pull in? like `TOP 100` in a SQL query? 

No need to amend the page size to 100 from 30 in that case.

Harshita-mindfire: Yep- the first 100 from the stack.

packages/engine-test-utils/src/__tests__/pods-core/GithubPod.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+              },
+            },
+          },
+        ],
+      },
+    };
+    fname = "github.issues.902-Test Issue";
+  });
+
+  test.skip("Bad Credentials", async () => {
```
kevinslin, line 39: is there a reason why this test is skipped?

kevinslin (commented):
just a question about a skipped test

packages/engine-test-utils/src/__tests__/pods-core/GithubPod.spec.ts
Hj2207: @kevinslin : please find the reason to skip the testcase in above comments.

Harshita-mindfire: @kevinslin Please find the reason to skip the testcase in above comments.

jonathanyeung (approved):

Issue #876, opened 2021-06-27
Local image not rendered in peek preview


SR--:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ x] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

A peek preview is shown but the image is not rendered

![image](https://user-images.githubusercontent.com/1008059/123539293-552f1e80-d77c-11eb-9b82-84728a6d826e.png)

## To Reproduce

1. Make a note and add an image.
2. Make another note and make a link to the first note.
3. Point to the link in the new note to see the peek preview.
4. The text is visible but the image is not.

## Expected behavior

Expect to see the locally stored image.

The path is: ![](/assets/images/2021-06-27-16-15-58.png)

## Desktop (please complete the following information)

- OS: Win
- Version [e.g. 47.1]

## Additional

Discussed with @SeriousBug on Discord.

SeriousBug:
Local images within notes should now display in the hover preview with version 0.50.

SR--:
It does not work for me in v50. There is not broken icon but I cannot see the image either. There is just an empty like. It looks OK in the preview.

SR--:
And, actually, I see the broken icon in another note...

![image](https://user-images.githubusercontent.com/1008059/125391440-e0005200-e3e7-11eb-80c1-205491a13bb2.png)

SeriousBug:
> And, actually, I see the broken icon in another note...

Could you show what that part of the previewed note looks like? (`![](/assets/...` part)

If you could also write the full path to the image file, that would help diagnose the issue. VSCode is picky about what it will show in the preview so we have to calculate the full path to the image, that might still be broken.

SR--:
This is in the note:

`![](/assets/images/2021-06-16-15-53-43.png)`

This is in OS (Windows):

`C:\_Local Only\dennotes\brain\assets\images\2021-06-16-15-53-43.png`

It is visible in both standard and Dendrone preview.

SeriousBug:
Could you also share the output from "Dendron:Dev: Diagnostic Report"? Please feel free to send it to SeriousBug on the Dendron discord via private message if you have any privacy concerns.

[...]

PR #946, opened 2021-07-09
fix: local images in hover preview are now displayed

SeriousBug:
Local images within notes were broken in the hover preview because VSCode's hover preview was not able to find the relative image paths.

Switched the hover preview MD generation to `MDUtilsV5`, and tried `PREVIEW` flavor, but the hover preview refuses to display non-HTTPS image links so links to API do not work. I instead added a `HOVER_PREVIEW` flavor which transforms the local image URLs to full paths, which works for the hover.

#876


kevinslin (changes requested):
looks good. two changes:
- can you add a test to utilsv5 here -> https://github.com/dendronhq/dendron/blob/dev/packages/engine-test-utils/src/__tests__/engine-server/markdown/utilsv5.spec.ts#L31:L31 ?
- it looks like the integration tests are failing. might require a rebase from master

kevinslin (approved):

PR #948, opened 2021-07-09
Feat/note templates

jonathanyeung:
Adding infra around supporting initializing a workspace with a seed. This can enable us to provide users with a vault containing templates that can be used for notes.


packages/engine-server/src/workspace/service.ts
```diff
@@ -135,6 +135,10 @@ export class WorkspaceService {
     return path.join(this.wsRoot, "dendron");
   }
 
+  get seedService(): SeedService {
```
jonathanyeung, line 138: @kevinslin let me know if this is an ok way to expose SeedService, or if you had something else in mind.

jonathanyeung:
There are a few inadvertent artifacts that are meant to be removed.

packages/engine-server/src/workspace/service.ts
kevinslin: yeah, looks good

packages/engine-server/src/workspace/service.ts
```diff
@@ -469,6 +473,37 @@ export class WorkspaceService {
...
   }
 
+  static async createEmptyWorkspace(opts: WorkspaceServiceCreateOpts) {
+    const { wsRoot, vaults } = opts;
+    const ws = new WorkspaceService({ wsRoot });
+    fs.ensureDirSync(wsRoot);
+    // this creates `dendron.yml`
+    ws.createConfig();
+    // add gitignore
+    const gitIgnore = path.join(wsRoot, ".gitignore");
```
kevinslin, line 483: - can we extract writing the ignore contents into a separate function and add `seeds` to one of the ignored files?
- we can also use this in `createWorkspace`

packages/engine-server/src/workspace/service.ts
```diff
@@ -469,6 +473,37 @@ export class WorkspaceService {
...
+    await _.reduce(
+      vaults,
+      async (prev, vault) => {
+        await prev;
+        return ws.createVault({ vault });
+      },
+      Promise.resolve()
+    );
+    // check if this is the first workspace created
+    if (_.isUndefined(MetadataService.instance().getMeta().firstWsInitialize)) {
```
kevinslin, line 501: since this is used in `createWorkspace` can we move this into a `postWorkspaceCreateHook` method? in the future, we could expose this to plugins to add their own hooks

packages/plugin-core/src/workspace/templateInitializer.ts
```diff
@@ -0,0 +1,26 @@
...
+import { WorkspaceInitializer } from "./workspaceInitializer";
+
+/**
+ * Template Workspace Initializer - add the templates seed to the workspace:
+ */
+export class TemplateInitializer
+  extends BlankInitializer
+  implements WorkspaceInitializer
+{
+  async onWorkspaceCreation(opts: {
```
kevinslin, line 13: for the template initializer, did we also owant to open the preview to a particular note on initial open? is it worthwile creating a `onWorkspaceFirstOpen` hook?

packages/engine-server/src/workspace/service.ts
jonathanyeung: Apologies - this function `createEmptyWorkspace()` was just a temporary artifact I put here while prototyping; I forgot to delete before the review and it's not being used anywhere.

packages/plugin-core/src/workspace/templateInitializer.ts
jonathanyeung: I had a hook, but I removed it for now - the user experience I was thinking about for this was to have templateInitializer be the default new workspace config.  It puts an empty vault (with just root.md) + this templates seed vault. So my thinking, at least at the moment, is to have the launch onWorkspaceFirstOpen experience be the same as it is now; but we can always amend this later.

packages/engine-server/src/workspace/service.ts
jonathanyeung: same comment above; this function will be removed.

kevinslin:
for the es linit fixes, could you send a PR for it in a separate branch? i think a bunch of people are running into this and its also good practice to limit the scope of any indiviidual PR

jonathanyeung:
> for the es linit fixes, could you send a PR for it in a separate branch? i think a bunch of people are running into this and its also good practice to limit the scope of any indiviidual PR

Yep, I was thinking about pulling it into a separate PR but then I got lazy :P

kevinslin:
Looks good. Can you add a test to initialize with a template [here](https://github.com/dendronhq/dendron/blob/feat/note-templates/packages/plugin-core/src/test/suite-integ/Extension.test.ts#L80:L80)?

Issue #887, opened 2021-06-29
Make autocomplete aware of vaults when suggesting notes for linking


SR--:
## Please Select if your Request is Either something new or an Enhancement ##

- [x ] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ x] Retrieve. Backlinks, references, graph view
- [x ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

When creating a wiki link the autocomplete does not differentiate between notes from different vaults, this can lead to confusion and incorrect entries.

## Describe the solution you'd like ##

Autocomplete should alert the user of the source vault, with the link from the current vault listed first.

## Additional ##

Adding this as suggested by @SeriousBug on Discord

kpathakota:
@SeriousBug - could you add what other context you'd like for this and then we can decide whether to work on it immediately/backlog it etc. 

Is this something already in your wheelhouse since working on autocomplete? We've also moved away from vault-specific links (i.e. [[dendron://vault/note]] and are trying to keep things consistent by just linking [[note]].

SR--:
The simplest would be to at least match the way it looks in Lookup where the name of the vault is in the brackets after the suggestion (from memory, I am away from a computer with dendron for a few days).

SeriousBug:
Thanks for the feedback. I'll be working on this enhancement this month.

SeriousBug:
Visual Studio Code 1.58 adds support for [displaying extra information with completion items](https://code.visualstudio.com/api/references/vscode-api#CompletionItemLabel), which is perfect to display the vault information. I'll suggest waiting until we start using that version for Dendron, and then we can take advantage of that feature to display vault names.

SeriousBug:
All features relating to this are now live. Vault names should appear next to notes in autocomplete, and cross-vault links should get generated when needed. I wasn't able to get note previews during autocomplete working, but we will look into adding that feature later.

SR--:
Great. Thanks
________________________________

SeriousBug:
We should investigate if we can display the vault name next to all notes, not just the selection.

SR--:
[...]

PR #949, opened 2021-07-09
feat: wikilink autocomplete sorts current vault notes before other notes

SeriousBug:
This pull request changes it so that notes from the current vault will appear before notes from other vaults. This order is only initially however, VSCode may change the order as the user types to narrow the selection.

This partially helps #887. I have more features planned to improve it further.

### DIFF
--- a/packages/plugin-core/src/features/completionProvider.ts
+++ b/packages/plugin-core/src/features/completionProvider.ts
@@ -50,8 +50,8 @@ export const provideCompletionItems = (
     .text.substr(0, position.character);
   Logger.debug({ ctx, linePrefix, position, msg: "enter" });
 
-  const isResourceAutocomplete = linePrefix.match(/\!\[\[\w*$/); // eslint-disable-line no-useless-escape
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/features/completionProvider.ts
```diff
@@ -109,6 +110,10 @@ export const provideCompletionItems = (
     item.insertText = longRef; //insertText;
 
     item.sortText = padWithZero(index);
+    // Sort notes from current vault before other vaults
+    if (currentVault && !VaultUtils.isEqual(currentVault, note.vault, wsRoot)) {
+      item.sortText = 'x' + item.sortText;
```
kevinslin, line 115: if i'm reading this correctly, are we updating the sort by adding an `x` in front of the zero padded index? if so, a comment with some details would be great

kevinslin (changes requested):

packages/plugin-core/src/features/completionProvider.ts
SeriousBug: Is this what you had in mind?

kevinslin: yep!

kevinslin (approved):

Issue #649, opened 2021-04-20
Multi-Vault:  Default certain hierarchies to a specific vault


da5is:
## Please Select if your Request is Either something new or an Enhancement ##

- [ ] Enhancement of an existing Feature.
- [X] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

As a dendron user, I use Multi-Vault functionality to segregate data that I can't push publicly, or want to protect differently.  I wish that I could cause hierarchies to default to a specific vault - as in, all hierarchies that begin with finance.* would default to a vault that's on a local only filesystem.

## Describe the solution you'd like ##

It seems that the most appropriate place to configure this is within the hierarchies configuration.

## Describe alternatives you've considered ##

I'm doing this currently through manual moves and a .gitignore for specific hierarchies but it'd be great to be able to default this.

flammehawk:
I think this would fit into the `dendron.yml` of the Multi-Vault Workspace under the `Vaults` key entries, but this should also include Schemas that are defined.

da5is:
That makes sense too - wasn't thinking about the `dendron.yml` file at that moment.

jonathanyeung:
I have an alternate proposed solution for this issue.  One of the things I'd like to avoid is Dendron having _too_ many configs, which puts a burden on the user to have to tweak a lot of settings in order to get their desired behavior.

Instead, I think we can be a little intelligent with the vault suggestion.

### Current Behavior
If `lookupConfirmVaultOnCreate: false` (or not set), then the note will get created in the same vault as the currently focused note.
else `lookupConfirmVaultOnCreate: true`, then you will always get prompted to pick among all your vaults.

### Proposed Behavior
We'll examine the hierarchy of the new note and see if it matches any existing hierarchies (or schemas, potentially) across your vaults. If it does, then we'll try suggest that vault as the destination for the new note.

The suggested list of vaults to place the note would be listed in order based on the following criteria:

1) If the vault of the current note context ALSO has a matching hierarchy, then suggest this vault
2) Any vault(s) with a matching hierarchy
[...]

PR #955, opened 2021-07-12
Feat/multi vault note creation

jonathanyeung:
Related to https://github.com/dendronhq/dendron/issues/649

On Note Creation in a multi-vault environment, this change will try to pick the correct vault location for the new note based on two factors - hierarchy match and note context.


packages/plugin-core/src/test/suite-integ/LookupUtils.test.ts
```diff
@@ -0,0 +1,210 @@
...
+        genRandomId: true,
+      });
+    });
+  }
+}
+
+/**
+ * Tests the Vault Recommendation For New Notes Functionality in Utils
+ */
+suite.only("Lookup Utils Test", function runSuite() {
```
kevinslin, line 69: remove `.only`. also, feel free to add it to the [husky](https://github.com/dendronhq/dendron/blob/dev/hooks/pre-push.js#L20:L20) scripit

packages/plugin-core/src/components/lookup/utils.ts
```diff
@@ -390,16 +407,180 @@ export class PickerUtilsV2 {
...
-    >;
-    return resp;
+    const resp = await VSCodeUtils.showQuickPick(items, {
+      title: "Select Vault",
+    });
+
+    return resp ? resp.vault : undefined;
+  }
+
+  // Determine which are the most appropriate vault(s) to create this note in.
```
kevinslin, line 462: use 
```
/**
 *
 **/
```

format comments so they show up in docstrings

packages/plugin-core/src/components/lookup/utils.ts
```diff
@@ -390,16 +407,180 @@ export class PickerUtilsV2 {
...
+    let vaultSuggestions: VaultPickerItem[] = [];
+
+    const engine = getWS().getEngine();
+
+    // Only 1 vault, no other options to choose from:
+    if (engine.vaults.length <= 1) {
+      return Array.of({ vault });
+    }
+
+    const hierarchies = fname.split(".").slice(0, -1);
```
kevinslin, line 483: we refer to the top part of a hierarchy as the `domain`. there's actually a utility method to get it [here](https://github.com/dendronhq/dendron/blob/feat/multi-vault-note-creation/packages/common-all/src/dnode.ts#L243:L243). its not used anywhere so you are free to use it.

it currently throws an error if the node you pass in is the root node but feel free to change it to return the root instead

packages/plugin-core/src/test/suite-integ/LookupUtils.test.ts
```diff
@@ -0,0 +1,210 @@
...
+  const ctx: vscode.ExtensionContext = setupBeforeAfter(this, {});
+  this.timeout(TIMEOUT);
```
kevinslin, line 71: this is already done in `setupBeforeAfter` so you can remove

kevinslin (changes requested):

packages/plugin-core/src/test/suite-integ/LookupUtils.test.ts
jonathanyeung: good catch :) let me add

packages/plugin-core/src/components/lookup/utils.ts
jonathanyeung: I see - I've diverged a bit here though; whereas the `DNodeUtils` uses `DNodeProps` and `DNodePropsDict`, I stripped down the args on this method to just have fname of the proposed new note.

packages/plugin-core/src/test/suite-integ/LookupUtils.test.ts
jonathanyeung: will add husky change as part of a separate review.

kevinslin (changes requested):
looks like tests are failing due to unused variable

kevinslin:
Looks like tests are now failing on scratch notes

jonathanyeung:
> Looks like tests are now failing on scratch notes

Apologies haha - I actually have some issues running tests locally as I hit some transient issues, so I end up relying on the Github test pass...dunno if it's a Windows problem.  Let me try to dig in

kevinslin:
looks like tests are still failing

```
 @dendronhq/plugin-core:   1) Lookup, notesv2
@dendronhq/plugin-core:        onAccept with modifiers
@dendronhq/plugin-core:          with lookupPrompt on other vault:
@dendronhq/plugin-core:      Error: Timeout of 300000ms exceeded. For
```

jonathanyeung:
> looks like tests are still failing
> 
> ```
>  @dendronhq/plugin-core:   1) Lookup, notesv2
> @dendronhq/plugin-core:        onAccept with modifiers
> @dendronhq/plugin-core:          with lookupPrompt on other vault:
> @dendronhq/plugin-core:      Error: Timeout of 300000ms exceeded. For
> ```

This spiraled into me finally getting around to this: https://github.com/dendronhq/dendron/pull/980

Basically it was impossible for me to cleanly run tests locally haha.  I might just rebase on 980 so that I can see what other tests are getting affected by these changes.

kevinslin (approved):

PR #960, opened 2021-07-12
feat: editor highlighting for wikilinks & references

SeriousBug:
Wikilinks are now highlighted, with the aliases being italicized. If the note is not found, then these are highlighted differently to make them stand out.

I'm not sure if these are a good way to highlight wikilinks, so I'd like some feedback on what people think.

Dark theme: 
dark theme

Light theme:
light theme

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,19 @@ module.exports = {
     "max-len": "off",
     "prefer-template": "off",
     "consistent-return": "off",
+    // less restrictive airbnb
+    'no-restricted-syntax': [
+      'error',
+      {
+        selector: 'ForInStatement',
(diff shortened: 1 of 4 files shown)


kevinslin:
For invalid states (eg. vault or link doesn't exist), I would associate a warning color (eg. orange or red). when in doubt, the bootstrap [color codes](https://getbootstrap.com/docs/4.0/utilities/colors/) are useful

SeriousBug:
> warning color (eg. orange or red)

This is actually using a theme-dependent warning color (the color comes from the theme). Should I keep that, or set a custom color?

ddbrierton:
*Please* use theme-dependent colours.

kevinslin (approved):

PR #961, opened 2021-07-12
fix: last modiified timestamp not updating

kevinslin:


packages/plugin-core/src/services/NoteSyncService.ts
```diff
@@ -80,6 +118,8 @@ export class NoteSyncService {
       wsRoot: eclient.wsRoot,
     });
     note.anchors = anchors;
+    note.updated = NoteUtils.genUpdateTime();
```
jonathanyeung, line 121: reuse the `now` variable from before?

packages/plugin-core/src/services/NoteSyncService.ts
```diff
@@ -63,6 +63,15 @@ export class NoteSyncService {
       notes: eclient.notes,
       wsRoot: DendronWorkspace.wsRoot(),
     }) as NoteProps;
+
+    // NOTE: it might be worthwile to only do this after checking that the current note is still active
+    // 
+    // we have this logic currently and it doesn't seem to be causing issues
+    // this could lead to thrashing if user makes a change and quickly changes to a dififerent active window
+    // in practice, this has never been reported 
+    const editor = await VSCodeUtils.openNote(noteHydrated);
```
jonathanyeung, line 72: for my understanding - why is this call necessary? Is it possible to edit the file without explicitly opening it? 

Additionally - is the desired behavior to mark a note as updated even prior to saving, now that we're triggering off a different event?

jonathanyeung (approved):

packages/plugin-core/src/services/NoteSyncService.ts
kevinslin: not that I know of. you can get a text edtor via workbench but that is read-only. will need to do more investigation to confirm but unfortunately don't have bandwidth right now

for marking a note as updated - its currently driven more by implementation rather than purity. changing it at `onWillSave` is tough because that would triigger `onDidChange`. we'd have to add a check there to ignore `updated` changes otherwise

PR #975, opened 2021-07-14
feat: lapsed user prompt

jonathanyeung:
Prompt users with a modal dialog to get started with Dendron under the following conditions:

1) Have not initialized a workspace before
2A) >1 week since install
2B) >1 week since last reminder prompt.

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -2,6 +2,7 @@ import {
   CONSTANTS,
   DendronError,
   getStage,
+  Time,
   VaultUtils,
   VSCodeEvents,
 } from "@dendronhq/common-all";
@@ -19,6 +20,7 @@ import {
   WorkspaceService,
 } from "@dendronhq/engine-server";
 import _ from "lodash";
+import { Duration } from "luxon";
 import path from "path";
 import semver from "semver";
 import * as vscode from "vscode";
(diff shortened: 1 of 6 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -522,4 +524,55 @@ async function showWelcomeOrWhatsNew({
...
+function showLapsedUserMessage() {
+  const START_TITLE = "Get Started";
+
+  AnalyticsUtils.track(VSCodeEvents.ShowLapsedUserMessage);
+  MetadataService.instance().setLapsedUserMsgSendTime();
+
+  vscode.window
+    .showInformationMessage(
+      "Get started with Dendron.",
+      { modal: true },
```
kevinslin, line 542: what happens when a user doesn't have 1.58 of vscode?

packages/plugin-core/src/_extension.ts
```diff
@@ -522,4 +524,55 @@ async function showWelcomeOrWhatsNew({
...
+function shouldDisplayLapsedUserMsg(): boolean {
+  const ONE_WEEK = Duration.fromObject({ weeks: 1 });
+  const CUR_TIME = Duration.fromObject({ seconds: Time.now().toSeconds() });
+  const metaData = MetadataService.instance().getMeta();
+
+  // If we haven't prompted the user yet and it's been a week since their
+  // initial install OR if it's been one week since we last prompted the user
+  const refreshMsg =
+    (!metaData.lapsedUserMsgSendTime &&
+      ONE_WEEK <=
```
kevinslin, line 566: for first install i would shorten this to 1 day

packages/plugin-core/src/_extension.ts
```diff
@@ -522,4 +524,55 @@ async function showWelcomeOrWhatsNew({
...
+        const ws = getWS();
+        return ws.showWelcome();
+      } else {
+        AnalyticsUtils.track(VSCodeEvents.LapsedUserMessageRejected);
+        return;
+      }
+    });
+}
+
+function shouldDisplayLapsedUserMsg(): boolean {
```
kevinslin, line 557: test please :)

kevinslin (changes requested):

packages/plugin-core/src/_extension.ts
jonathanyeung: Actually, it turns out this feature came out long before 1.58.  The only feature added in 1.58 is the ability to add a `detail` prop in the options, which I decided not to utilize.

packages/plugin-core/src/_extension.ts
jonathanyeung: tests added!

kevinslin (approved):

PR #983, opened 2021-07-15
enhance: prefill `replace prefix` input box with output of `match text` input box

hikchoi:
This PR:
- Passes output of `match text` input to prefill `replace prefix` input box.


hikchoi:
If only I knew it would have been a one-liner back when I opened the issue 😅

jonathanyeung (approved):

kevinslin (approved):

Issue #944, opened 2021-07-08
Journal format is still using the settings Dendron: Default Journal Date Format even though it states as DEPRECATED. 


HopperTech:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [X] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

The Journal format is still reading from the settings Dendron: Default Journal Date Format even though this states "DEPRECATED. Use journal settings in dendron.yml"

## To Reproduce

1. In the dendron.yml file , change the journal: -> dateFormat:  - my desired format is:   'q'q.y.MM.'w'W.dd
2. Run the "Developer: Reload Window" command
3. Execute the "Dendron: Create Daily Journal Note"
4. **Notice** that the file name is still using the the format:   y.MM.dd
5. Open the extension's settings
6. Change the value of "Dendron: Default Journal Date Format" to the desired format
7. Run the "Developer: Reload Window" command
8. Execute the "Dendron: Create Daily Journal Note"
9. **Notice** that the file name is in the desired format

## Expected behavior

The Journal file name should be in the desired date format

## Screenshots

N/A

## Desktop (please complete the following information)

- OS: Windows
- Browser: N/A using VSCode
- Version: 0.49.0

## Additional context

In troubleshooting this issue I discovered that the genNoteName in the utils.js file is pulling in the deprecated constant value. 

## Dendron Log file

N/A

kpathakota:
@HopperTech - I'm not able to repro this. I see that it's currently using the value in my dendron.yml file. 

Do you have the formatting correct? Adding this at the bottom of your dendron.yml file should help. 

```md
journal:
    dailyDomain: name
    name: journal
    dateFormat: y.MM.dd
    addBehavior: childOfDomainNamespace
    firstDayOfWeek: 1
```

aleksey-rowan:
I get the same behaviour:

```yml
journal:
    dailyDomain: daily
    name: journal
    dateFormat: dd
    addBehavior: childOfDomainNamespace
    firstDayOfWeek: 1
```

Still generates notes with the format of `y.MM.dd`.

Dendron version: 0.50

HopperTech:
[...]

PR #984, opened 2021-07-15
fix: Use new config when creating special notes

hikchoi:
This PR:
- Fixes issue with journal and scratch note creation configuration still using values from deprecated config location.

related to: #944


kevinslin (changes requested):
tests please :)

hikchoi:
- Added test for correct behavior

kevinslin (approved):

kevinslin (approved):

PR #985, opened 2021-07-15
feat: hashtags

SeriousBug:
Adds support for hashtags. These are a shorthand for adding tags to notes. Hashtags look like:

```
#my-tag

#hierarchical.tag
```

These are identical to writing `[[tags.my-tag]]` and `[[tags.hierarchical.tag]]`.

This pull request adds support for parsing, compiling to markdown, publishing, GotoNote, Ctrl+Click, hover preview, and Move/Rename/Refactor notes support for hashtags.

It also adds support for displaying colors for hashtags in the editor. Here is an example:

The colors are randomly generated based on the tag. They are consistent across runs, same tag will always get the same color.

Another feature is that when tags are moved to (using refactor or move note commands) under `tags.` hierarchy, any wikilinks that link to them will get converted to hashtags. This can ease adoption by existing users as they can switch to hashtags by simply refactoring their tags. Hashtags are converted back to wikilinks if a tag is moved out of the tags hierarchy.


SeriousBug:
To make reviewing easier: the hashtags get parsed to a `HashTag` type, but they get converted to `WikiLinks` in many places like adding links and backlinks to notes, or rendering to HTML. A few places (like `getReferenceAtPosition`) were modified to recognize hashtags in text.

packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -159,7 +162,7 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
               value: alias,
             },
           ],
-        };
+        } as any /* type checker otherwise complains about hName */;
```
kevinslin, line 165: instead of casting to `any`, can we create a custom type `RehypeLinkData` and add it [here](https://github.com/dendronhq/dendron/blob/dev/packages/engine-server/src/markdown/types.ts#L80:L80)?

packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -181,7 +184,7 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
                 value: alias,
               },
             ],
-          };
+          } as any /* type checker otherwise complains about hName */;
```
kevinslin, line 187: see above

packages/plugin-core/src/test/suite-integ/GotoNote.test.ts
```diff
@@ -21,9 +21,7 @@ import { runLegacyMultiWorkspaceTest, setupBeforeAfter } from "../testUtilsV3";
 
 const { ANCHOR_WITH_SPECIAL_CHARS, ANCHOR } = GOTO_NOTE_PRESETS;
 suite("GotoNote", function () {
-  let ctx: vscode.ExtensionContext;
-
-  ctx = setupBeforeAfter(this, {
+  const ctx = setupBeforeAfter(this, {
     beforeHook: () => {},
```
kevinslin, line 25: not necessary to have `beforeHook` block if we're not running anything

kevinslin (changes requested):
for testing, can you also add `dendron.ref.tags` into `test-workspace`? 

any new language features should be added there so that its easy to both manually test and verify in ci/cd (we'll add a snapshot to the entire workspace as part of the build process iin the future to look for publishing based regressions)

kevinslin (changes requested):
this looks great. last thing is just to add an example to `test-workspace`. see prior comment


```
for testing, can you also add dendron.ref.tags into test-workspace?

any new language features should be added there so that its easy to both manually test and verify in ci/cd (we'll add a snapshot to the entire workspace as part of the build process iin the future to look for publishing based regressions)
```

SeriousBug:
I'll fix the merge conflict and re-request a review, just a second.

kevinslin (approved):

PR #988, opened 2021-07-15
Feat/gdoc import pod

Harshita-mindfire:
PR : Documentation for Google Docs Import Pod
https://github.com/dendronhq/dendron-site/pull/145


packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+   */
+  documentId: string;
+
+  /**
+   * name of hierarchy to import into
+   */
+  fname: string;
+
+   /**
+   * import comments from the doc in text or json format
```
jonathanyeung, line 35: Please add a note on the default behavior when this is not set (and also to corresponding the dendron-site docs)

packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+
+   /**
+   * document Id of doc to import
+   */
+  documentId: string;
+
+  /**
+   * name of hierarchy to import into
+   */
+  fname: string;
```
jonathanyeung, line 32: Maybe rename to `hierarchyDestination` or `hierarchyDest`.

packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+      Authorization: `Bearer ${token}`,
+      "Content-Type": "application/json",
+    };
+    try{
+      const result = await axios.get(
+        `https://docs.googleapis.com/v1/documents/${documentId}`,
+        { headers }
+      )
+      const markdown = googleDocsToMarkdown(result.data);
+      const body= markdown.split("---")
```
jonathanyeung, line 110: What if the doc contents naturally has a `---`?

packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+    return note;
+  }
+
+  createNote= async(
+    note: NoteProps,
+    engine: DEngineClient,
+    wsRoot: string,
+    vault: DVault
+  ) => {
+      const n = NoteUtils.getNoteByFnameV5({
```
jonathanyeung, line 231: nit: maybe rename n to `existingNote` or something like that.

packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+      fname: note.fname,
+      notes: engine.notes,
+      vault,
+      wsRoot,
+    });
+    if (!_.isUndefined(n)) {
+      if(n.custom.revisionId && n.custom.revisionId !== note.custom.revisionId) {
```
jonathanyeung, line 238: what's the intended behavior if revisionId doesn't exist or if it's the same?  Should we not go ahead and overwrite the note contents too?

packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+        n.custom.revisionId = note.custom.revisionId;
+        n.body = note.body;
+        await engine.writeNote(n, { newNode: true });
+        }   
+       }
+    else {
+      await engine.writeNote(note, { newNode: true });
```
jonathanyeung, line 245: I wonder if we should have a `confirmOverwrite` option so that users can have an option to get prompted on whether or not they want to overwrite their existing doc? Maybe in the next iteration.

packages/pods-core/src/builtin/GDocPod.ts
```diff
@@ -0,0 +1,273 @@
...
+type GDocImportPodCustomOpts = {
+  /**
+   * google docs personal access token
+   */
+  token: string;
+
+   /**
+   * document Id of doc to import
+   */
+  documentId: string;
```
jonathanyeung, line 27: for my understanding: What's the right documentId to specify here? Is it something like `meet.planning.triage` or is something like `1t9IoUjKs9sDa6vR3BvwXX1leABiTnfT-f1aPupBgpyI` (as seen from the URL when viewing the gdoc https://docs.google.com/document/d/1t9IoUjKs9sDa6vR3BvwXX1leABiTnfT-f1aPupBgpyI/edit)

Maybe we should add some additional notes either here and/or on the dendron-site wiki so that it's clearer what should be used here.

jonathanyeung (changes requested):

packages/pods-core/src/builtin/GDocPod.ts
Hj2207: Every google doc has it's revision id. Each time we edit the docs, the revision id gets updated. If the revision id is same that means the doc hasn't been updated, we intended to not update the note in that case

packages/pods-core/src/builtin/GDocPod.ts
Hj2207: The document id should be `1t9IoUjKs9sDa6vR3BvwXX1leABiTnfT-f1aPupBgpyI` from the url of doc. I should update about this field's specifications in the documentation

packages/pods-core/src/builtin/GDocPod.ts
Hj2207: Fname is the **required** field to specify the hierarchy for note

packages/pods-core/src/builtin/GDocPod.ts
Hj2207: Hm I must look for this scenario, since the package that we have used to convert docs to .md returns a string with some metadata wrapped around `---`

packages/pods-core/src/builtin/GDocPod.ts
Harshita-mindfire: Fname is the **required** field to specify the hierarchy for note

packages/pods-core/src/builtin/GDocPod.ts
Harshita-mindfire: Every google doc has it's revision id. Each time we edit the google doc, the revision id gets updated. If the revision id is same that means the doc hasn't been updated, we intended to not update the note in that case

packages/pods-core/src/builtin/GDocPod.ts
Harshita-mindfire: The document id should be 1t9IoUjKs9sDa6vR3BvwXX1leABiTnfT-f1aPupBgpyI from the url of doc. I should update about this field's specifications in the documentation

packages/pods-core/src/builtin/GDocPod.ts
Harshita-mindfire: Yes I should consider this case, will fix this in new commit.

packages/pods-core/src/builtin/GDocPod.ts
kevinslin: that would be great. @Harshita-mindfire can you add this to the [[Next|handbook.sop.weekly-journal#next]] section of your weekly journal so we can triage it?

jonathanyeung (approved):

PR #993, opened 2021-07-18
Capitalize command

golergka:


jonathanyeung:
Hi @golergka - is this functionality not the same as the built in editor.action.transformToUppercase in VS Code? 

See https://stackoverflow.com/questions/35184509/make-selected-block-of-text-uppercase

Issue #959, opened 2021-07-12
"Copy Note Link" cuts off the header text at a wikilink


af4jm:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

I have a header that includes wiki-link(s), e.g.

``` markdown
## Outline of [[1|bible.1_kings]] & [[2|bible.2_kings]] Kings
```

If I select part of that link & trigger "Dendron: Copy Note Link", the link I get ends `#outline-of`, cutting off at the wikilink instead of making a GitHub wiki compatible link

Possibly related to issue #958 

## Desktop (please complete the following information)

- OS: Windows
- VSCode 1.58.0
- Dendron 0.49.0

SeriousBug:
This issue has been fixed with Dendron 0.52. Please let us know if you are still having issues.

Issue #958, opened 2021-07-12
"Copy Note Link" adds hash text to the header


af4jm:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

I have a header that is also a wiki-link, e.g.

``` markdown
## [[DANIEL|bible.daniel]]
```

If I select part of that link & trigger "Dendron: Copy Note Link", it adds ` ^` then several random characters to the end of the header (& the link text in what is copied is the document Title instead of the header text)

Possibly related to #959 

## Desktop (please complete the following information)

- OS: Windows
- VSCode 1.58.0
- Dendron 0.49.0

teauxfu:
Not sure if the behavior is intentional or not but I think the `^` and random characters are a block anchor for linking to that header specifically 
https://wiki.dendron.so/notes/f1af56bb-db27-47ae-8406-61a98de6c78c.html#block-anchor

af4jm:
yes it is... I assume that means Dendron's preview & publish strip those out? As someone using Dendron strictly as a PKM, meaning I'll never publish (for me, the P in PKM is personal & not public), I'm frustrated that the only places I render are in previews in VSCode, iOS, & Android (and Dendron preview v1 is so buggy I had to uninstall it in favor of the built in... might switch back if v2 gets CSS support)... or occasionally in git in my browser, and as much as I like the backlinks feature & Ctrl+L to jump to a link... having wikilinks instead of MD links & having 2000 files in a single folder both causes issues everywhere except VSCode (not to mention that autocomplete only works for [[filename]] and not for [[some text|filename]], which has to be all of my wikilinks because the single folder thing requires `prefix.` at the front of file names & I don't want to see those in preview... it's a trade-off either way & I know Dendron's got some some good stuff coming down the pipe

SeriousBug:
This issue has been fixed with Dendron 0.52. Please let us know if you are still having issues.

PR #1004, opened 2021-07-20
fix: wikilink aliases with apostrophes & headers containing wikilinks

SeriousBug:
Wikilinks containing apostrophes were failing to parse, that's now fixed.

Also #958 and #959. The correct links should be generated now, both in the editor and in publishing, and "Copy Note Link" will recognize the link and avoid adding a block anchor.

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/CopyNoteLink.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/CopyNoteLink.test.ts
@@ -61,6 +61,32 @@ suite("CopyNoteLink", function () {
       })
     });
 
+    test("with a header link containing unicode characters", (done) => {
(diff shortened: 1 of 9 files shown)


packages/plugin-core/src/test/suite-integ/CopyNoteLink.test.ts
```diff
@@ -37,6 +35,32 @@ suite("CopyNoteLink", function () {
       });
     });
 
+    test("with a header that's a link", (done) => {
```
kevinslin, line 38: can we add a test case where a header link contains an emoji?

kevinslin (changes requested):

kevinslin (approved):

PR #1005, opened 2021-07-21
feat: warn against missing or bad frontmatter

SeriousBug:
Warns against:
* Frontmatter missing
* Frontmatter has no id
* Frontmatter id is bad (starts or ends with underscore or hypen)

It also provides automatic code action to fix the issue, which runs the doctor command to fix the frontmatter.

https://www.loom.com/share/ab0c253442814bc2b78f03f08bd7d342

### DIFF
--- a/packages/plugin-core/src/commands/Doctor.ts
+++ b/packages/plugin-core/src/commands/Doctor.ts
@@ -159,6 +159,9 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
     if (ws.fileWatcher) {
       ws.fileWatcher.pause = true;
     }
+    // Make sure to save any changes in the file because Doctor reads them from
+    // disk, and won't see changes that haven't been saved.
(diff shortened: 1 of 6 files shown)


kevinslin (changes requested):
1.  Noticed some snapshot tests are failing. Can you make sure to update snapshots for tests?  [[Case Specific|dendron.dev.qa.sop#case-specific]]

2. Seems to be an issue when the file is `dirty` but not saved. Doctor won't be able to fix the issue because the file hasn't been updated. See [here](https://www.loom.com/share/3c4adb31df8c45e5ac1af50b45b578e3)

SeriousBug:
I fixed the issue with unsaved files, but I don't see the failing snapshots locally. I tried updating the snapshots but found no changes on my machine. Looks like it's only happening on macos, any thoughts on how to resolve that?

kevinslin (approved):

kevinslin:
Looks good. Jest has a 5s timeout, the gitpod issue seems transient so we can mark it if it shows up again

PR #1008, opened 2021-07-21
feat: launch engine in separate process

kevinslin:
Makes Dendron no longer block VSCode before initializing engine. Also makes preview v2 updates not block VSCode 

This also has a few other smaller changes:
- makes `watch.sh` script work in windows
- removes the `tutorial.md` missing warning when running plugin tests
- adds `sinon.restore` to run automatically after `setupBeforeAndAfter`
- various formatting fixes


packages/plugin-core/src/_extension.ts
```diff
@@ -138,43 +136,30 @@ async function postReloadWorkspace() {
...
-        const port = DendronWorkspace.serverConfiguration().serverPort;
-        const engine = WSUtils.updateEngineAPI(port);
-        await engine.init();
-        Logger.info({ ctx, msg: "fin init Engine" });
-        await reloadWorkspace();
-      }
-    }
-  );
+  const logPath = DendronWorkspace.instance().context.logPath;
+  const out = await ServerUtils.execServerNode({
```
jonathanyeung, line 155: I'm guessing that setting debugger statements in the engine will no longer break if this is a separate process? If so, maybe we need to build out a developer mode vs production mode.

Some random ideas from previous projects: in dev mode, we can detect if an existing server is running. If not, then instantiate one. If it does exist, then connect to it.

We can instantiate the debug server via another instance of VS Code. In that case, breakpoints should hit in the server too?

kevinslin: good point. we already do that for the integration tests because it takes too long to spawn a separate process otherwise. will update

packages/plugin-core/src/test/testUtilsV3.ts
```diff
@@ -300,6 +338,7 @@ export function setupBeforeAfter(
     if (opts?.afterHook) {
       await opts.afterHook();
     }
+    sinon.restore();
```
jonathanyeung, line 341: Not sure if this will properly restore on Windows; I can investigate this one later.

packages/plugin-core/src/server.ts
```diff
@@ -0,0 +1,14 @@
+import { ServerUtils } from "@dendronhq/api-server";
+import { stringifyError } from "@dendronhq/common-all";
+
+(async () => {
```
jonathanyeung, line 4: nit: since this function/js file isn't explicitly referenced, can we add a comment around how this is intended to be used and/or that it's invoked from execa in _extension.ts

packages/plugin-core/src/_extension.ts
```diff
@@ -379,7 +364,24 @@ export async function _activate(
...
+      WSUtils.handleServerProcess({
+        subprocess,
+        context,
+        onExit: () => {
+          const txt = "Restart Dendron";
+          vscode.window
+            .showErrorMessage("Dendron engine encountered an error", txt)
+            .then(async (resp) => {
+              if (resp === txt) {
+                AnalyticsUtils.track(VSCodeEvents.ServerCrashed);
```
jonathanyeung, line 378: any short error message context that we can add to the analytics event? 

Longer term I'll investigate the backend error reporting strategy

jonathanyeung (changes requested):
Looks good - main point of feedback is the earlier comment about debuggability

packages/plugin-core/src/test/suite-integ/StartServer.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      await new ResetConfigCommand().execute({ scope: "all" });
+      homeDirStub = TestEngineUtils.mockHomeDir();
+    },
+    afterHook: async () => {
+      homeDirStub.restore();
+    },
+    noStubExecServerNode: true
+  });
+
+	describe("basic", function() {
```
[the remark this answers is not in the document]
kevinslin: yeah, i realize prettier is not running anymore automatically. will have a separate pr for formatting

packages/plugin-core/src/test/testUtilsV3.ts
kevinslin: hmm, is there an outstanding issue with sinon on windows?

packages/plugin-core/src/_extension.ts
kevinslin: updated

kevinslin:
Responded to comments in new PR to avoid messy rebase here -> https://github.com/dendronhq/dendron/pull/1025/files

packages/plugin-core/src/test/testUtilsV3.ts
jonathanyeung: Seems like it...the `sinon.restore()` sandbox style restores doesn't seem to clear the stubs.  I've had to explicitly call restore on the individual stubs objects.

PR #1010, opened 2021-07-22
feat: scratch modifier for lookup V3

hikchoi:
This PR:
- Adds scratch modifier to lookup V3
- Adds tests for scratch modifier
- Adds tests for interaction between the scratch and journal modifier

fyi: this branch is already rebased to be on top of `feat/journal-lookupv3`. See #1009

### DIFF
--- a/packages/plugin-core/src/components/lookup/buttons.ts
+++ b/packages/plugin-core/src/components/lookup/buttons.ts
@@ -31,7 +31,8 @@ export type ButtonCategory =
   | "note"
   | "split"
   | "filter"
-  | "effect";
+  | "effect"
+  | "v2"; //TODO: better category name?
 
 export type ButtonHandleOpts = { quickPick: DendronQuickPickerV2 };
 
@@ -50,6 +51,9 @@ export function getButtonCategory(button: DendronBtn): ButtonCategory {
   }
   if (isEffectButton(button)) {
     return "effect";
+  }
+  if (button.type === "v2") {
+    return "v2"; //TODO: better type name
   }
   throw Error(`unknown btn type ${button}`);
 }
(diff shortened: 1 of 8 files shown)


packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -253,26 +267,168 @@ suite("NoteLookupCommand", function () {
           const out = (await cmd.run({
             noteType: LookupNoteTypeEnum.journal,
             noConfirm: true,
-          }));
+          })) as CommandOutput;
           
           expect(engine.config.journal.dateFormat).toEqual("y.MM.dd");
           // quickpick value should be `foo.journal.yyyy.mm.dd`
           const today = ((new Date()).toISOString().split("T")[0]).split("-").join(".");
```
kevinslin, line 274: this will fail depending on local timezone. we generate journals based on local time but this is GMT time

kevinslin (changes requested):
see comment. also, can you rebase from master? I rebased master into the journal note pull request that will affect this

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
hikchoi: ah missed that. will fix.

packages/plugin-core/src/components/lookup/buttons.ts
```diff
@@ -31,7 +31,8 @@ export type ButtonCategory =
   | "note"
   | "split"
   | "filter"
-  | "effect";
+  | "effect"
+  | "v2"; //TODO: better category name?
```
kevinslin, line 35: can we do `other`? a little more descriptive then v2 and indicates we don't have a per-defined behavior yet for this

kevinslin (changes requested):

packages/plugin-core/src/components/lookup/buttons.ts
hikchoi: fixed :)

PR #1012, opened 2021-07-22
feat: add selection modifier buttons to lookup V3

hikchoi:
This PR:
- Adds modifier buttons for `selection2link` and `selectionExtract`
***
- [x] Implement button
- [x] Implement onTrigger behaviors
- [x] Add tests

### DIFF
--- a/packages/plugin-core/src/components/lookup/buttons.ts
+++ b/packages/plugin-core/src/components/lookup/buttons.ts
@@ -2,7 +2,10 @@ import {
   DNodePropsQuickInputV2,
   NoteQuickInput,
   NoteUtils,
+  NoteProps,
+  getSlugger,
 } from "@dendronhq/common-all";
+import { getWS } from "../../workspace";
 import _ from "lodash";
 import * as vscode from "vscode";
 import { QuickInputButton, ThemeIcon } from "vscode";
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/components/lookup/buttons.ts
```diff
@@ -80,6 +83,48 @@ function isSplitButton(button: DendronBtn) {
...
+const selectionToNoteProps = async (opts: {
+  selectionType: string;
+  note: NoteProps;
+}) => {
+  const ws = getWS();
+  const resp = await VSCodeUtils.extractRangeFromActiveEditor();
+  const { document, range } = resp || {};
+  const { selectionType, note } = opts;
+  const { selection, text } = VSCodeUtils.getSelection();
+  //TODO: split these up?
```
kevinslin, line 96: i think this is fine since we only plan on doing one selection item at a time

packages/plugin-core/src/components/lookup/buttons.ts
```diff
@@ -80,6 +83,48 @@ function isSplitButton(button: DendronBtn) {
   return _.includes(["horizontal", "vertical"], button.type);
 }
 
+// TODO: make it into a utils method?
+const selectionToNoteProps = async (opts: {
```
kevinslin, line 87: heuristic for making this into utils method is if this is used in at least 2 places

kevinslin (approved):

PR #1024, opened 2021-07-23
feat: add a variety of graph indicators

hfellerhoff:
Adds/changes:
- A loading indicator on graph startup
- A loading indicator when changing between local/full note graphs
- A message when local graph has no active note, with a button to toggle to view the full graph
- Better graph config view hiding/showing


kevinslin (changes requested):
there is an error when toggling from the full graph to local and back again

https://www.loom.com/share/a3887e1455c1420eb2d9e9514f180580

hfellerhoff:
> there is an error when toggling from the full graph to local and back again
> 
> 
> 
> https://www.loom.com/share/a3887e1455c1420eb2d9e9514f180580

Yeah, this is something I touched on I think in another PR - this error is thrown (I believe) when the graph gets a request to re-render and hasn't finished its previous render. It doesn't affect the final output, and actually won't show up at all in a production build or affect the rendered graph.

kevinslin:
Can we block the request while a render is in flight?

.eslintrc.js
```diff
@@ -105,6 +105,8 @@ module.exports = {
         ],
       },
     ],
+    // fix for .tsx
+    "react/jsx-filename-extension": "off",
```
namjul, line 109: What is the aim of turning this rule off? I think it should only apply for `jsx` files and we don't use them. For typescript files we want to force using `tsx` extension. Do we want the same for js. In any case can you extend the comment to explicate it more.

About the word explication I find the following article giving good insights https://mindlevelup.wordpress.com/2017/07/15/explication/. Sharing resources ;)

kevinslin:
looks like there's some conflicts with master here

kevinslin (approved):

PR #1025, opened 2021-07-23
chore: address pr comments from pr 1008

kevinslin:
https://github.com/dendronhq/dendron/pull/1008

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -1,23 +1,23 @@
-import { ServerUtils } from "@dendronhq/api-server";
+import { launchv2, ServerUtils, SubProcessExitType } from "@dendronhq/api-server";
 import {
   CONSTANTS,
   DendronError,
   getStage,
   Time,
   VaultUtils,
-  VSCodeEvents,
+  VSCodeEvents
 } from "@dendronhq/common-all";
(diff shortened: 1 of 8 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -157,6 +157,16 @@ async function startServerProcess(): Promise<{
...
+  if (getStage() !== "prod") {
+    const out = await launchv2({
+      logPath: path.join(__dirname, "..", "..", "dendron.server.log"),
+    });
+    return {port: out.port}
+  }
+
+  // start server is separate process
   const logPath = DendronWorkspace.instance().context.logPath;
   const out = await ServerUtils.execServerNode({
```
jonathanyeung, line 171: do we need to similarly put this inside an `else{` block; i.e. only launch the background process if we're prod?

kevinslin: nope. we should never hit this in development

jonathanyeung (approved):
Approval for posterity sake; changes have already been merged though.

packages/plugin-core/src/_extension.ts
jonathanyeung: Ah, misread the code logic.

PR #1027, opened 2021-07-23
feat(community issue): graph zoom speed option

hfellerhoff:
**Adds:**
Ability to configure zoom speed for the graph.

**To test:**
Add the following property to your `dendron.yml`, reload the window, and open the note or schema graph. The default zoomSpeed is 1.
```
graph:
    zoomSpeed: 1
```

### DIFF
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -184,6 +184,11 @@ export type DendronConfig = {
   hierarchyDisplayTitle?: string;
 
   /**
+   * Configuration for note and schema graphs
+   */
+  graph?: DendronGraphConfig;
+
+  /**
    * Don't automatically create note when looking up definition
    */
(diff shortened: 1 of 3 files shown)


jonathanyeung:
Is zoom speed == 1 the current default value (prior to implementing this change)? If so, does it make sense to set the default value to < 1 if people are finding it too sensitive right now?

jonathanyeung (approved):

hfellerhoff:
> Is zoom speed == 1 the current default value (prior to implementing this change)? If so, does it make sense to set the default value to < 1 if people are finding it too sensitive right now? 

The only thing that worries me is that the experience changes drastically depending on the input device. A speed of 1 will seem fast for an external mouse, but will be perfectly fine for a trackpad for instance. That might be a perfectly fine thing to tweak, it's just something to consider

PR #1034, opened 2021-07-25
feat: set preview v2 as default

kevinslin:


packages/plugin-core/src/utils/md.ts
```diff
@@ -87,34 +89,44 @@ export const containsOtherKnownExts = (pathParam: string): boolean =>
...
+      if (resp === "Install Instructions") {
+        VSCodeUtils.openLink(
+          "https://wiki.dendron.so/notes/8de4209d-84d3-45f8-96a4-34282e34507d.html"
+        );
+      }
+    });
+
+  }
+  static async openPreview() {
     if (!getWS().config.dev?.enablePreviewV2) {
```
jonathanyeung, line 116: See earlier comment - should we remove this setting?

packages/engine-server/src/config.ts
```diff
@@ -48,6 +48,9 @@ export class DConfig {
       mermaid: true,
       useKatex: true,
       autoFoldFrontmatter: true,
+      dev: {
+        enablePreviewV2: true,
```
jonathanyeung, line 52: should we just remove this setting altogether?

packages/plugin-core/src/test/suite-integ/ShowPreview.test.ts
```diff
@@ -0,0 +1,90 @@
...
+						config.dev = {
+							enablePreviewV2: false
+						}
+						return config;
+					},
+					{ wsRoot }
+				);
+				const note = engine.notes["foo"];
+				await VSCodeUtils.openNote(note)
+				sinon.stub(MarkdownUtils, "hasLegacyPreview").returns(false)
```
jonathanyeung, line 52: I still haven't been able to figure out why, but I don't think `sinon.restore()` (default sandbox restore) will work on the Windows test pass and may cause issues since there are multiple instances of the same stub in this file.  Might be good to explicitly restore the stub object, i.e.

```typescript
const blah = sinon.stub(...)
...
blah.restore()
```

packages/engine-server/src/migrations/migrations.ts
```diff
@@ -39,6 +42,21 @@ export const ALL_MIGRATIONS: Migrations[] = [
       },
     ],
   },
+  {
+    version: "0.51.4",
+    changes: [
+      {
+        name: "don't switch to legacy preview if not currently on it",
```
jonathanyeung, line 49: Hmm I'm a little confused with the logic here. It looks like we're **not** converting existing users who haven't tried out preview v2 yet to start using preview v2?

packages/engine-server/src/config.ts
kevinslin: i was going to give preview v1 at least another two weeks before removing it altogether as an option

packages/engine-server/src/migrations/migrations.ts
kevinslin: yeah, the plan is to try out preview v2 with new users and have it bake for a bit before migrating existing users

jonathanyeung (approved):

PR #1045, opened 2021-07-27
feat: Add horizontal split modifier to lookup v3

hikchoi:
This PR:
- Adds horizontal split modifier to `NoteLookupCommand`
- Adds test for said modifier


hikchoi:
Merging this into `feat/mod-lookupv3` just so I have a catch-all branch for the remaining branches that add modifiers to `NoteLookup` to branch off / merge to. Chaining branch after branch is fine but I found rebasing each successive branches to one it branched off of a bit cumbersome 😅 

I'll have another branch for effects type modifiers that will merge into `feat/mod-lookupv3`.

kevinslin (approved):

PR #1047, opened 2021-07-27
fix: platform agnostic vault paths in yml config

jonathanyeung:


kevinslin:
Think this is going to involve a bigger change. Wrote some comments here -> [[Platform Agnostic Vaults|scratch.2021.07.27.082216.platform-agnostic-vaults]]

PR #1053, opened 2021-07-28
fix: sinon issues

kevinslin:


.github/workflows/ci.yml
```diff
@@ -71,7 +67,7 @@ jobs:
             node_modules
             packages/*/node_modules
             packages/*/.vscode-test
-          key: ${{ runner.os }}-${{ matrix.node-version }}-node-modules-${{ hashFiles('yarn.lock')}}
+          key: ${{ runner.os }}-${{ matrix.node-version }}-node-modules-1-${{ hashFiles('yarn.lock')}}
```
jonathanyeung, line 70: Why was it necessary to rev this?

jonathanyeung (commented):
Unfortunately it still looks like there are some failures in windows plugin.  Still appears to be mocking issues in `Telemetry.test.ts`

.github/workflows/ci.yml
kevinslin: windows cache wasn't pulling in new dependencies so manually invalidated

Issue #1048, opened 2021-07-27
Notes created with 'doctor' from '#' (tags), include punctuation chars that follow them in the name


MaDDmin:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

Notes created with 'doctor' from '#' (tags), include punctuation chars that follow them in the name.

## To Reproduce

1. Create a note with some text like, for example: 
"I just love #Dendron."
2. Run "Dendron:Doctor" command with the 'createMissingLinkedNotes' option.

## Expected behavior
The note I would expect to be created would be 'tags.Dendron' . Instead, it creates 'tags.Dendron."' .
Please note the dot and the double quote characters.

SeriousBug:
Thanks for the report, I'll create a fix for this. As a workaround for now, please feel free to use a wikilink like `[[#Dendron|tags.Dendron]]`.

SeriousBug:
This has now been fixed in version 0.53. Please let us know if there are any further issues.

PR #1054, opened 2021-07-28
fix: hashtags include punctuation and quotation marks

SeriousBug:
Hashtags are now not allowed to contain any punctuation or quotation marks. This allows them to be more easily mixed into text, for example:

```
This issue is #important, and should be prioritized.

This issue is #important. It should be prioritized.

They said "This issue is #important", so it should be prioritized.
```

In all these examples, the tag is `#important` without the following marks.

#1048

### DIFF
--- a/packages/engine-server/src/markdown/remark/hashtag.ts
+++ b/packages/engine-server/src/markdown/remark/hashtag.ts
@@ -6,16 +6,54 @@ import { DendronASTDest, DendronASTTypes, HashTag } from "../types";
 import { MDUtilsV4 } from "../utils";
 import { Element } from "hast";
 
+/** All sorts of punctuation marks and quotation marks from different languages. Please add any that may be missing.
+ * 
+ * Be warned that this excludes period (.) as it has a special meaning in Dendron. Make sure to handle it appropriately depending on the context.
+ * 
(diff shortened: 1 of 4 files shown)


kevinslin (changes requested):
for hashtags, we actually want to include `.` since `#foo.bar` is an acceptable tag in dendron land

SeriousBug:
> for hashtags, we actually want to include `.` since `#foo.bar` is an acceptable tag in dendron land

Periods are handled as a special case, only trailing periods are ignored. For example `#foo.bar` is "foo.bar" but `#foo.` is only "foo". 

I'll check out the failing tests in a bit, they looked fine on my end.

SeriousBug:
I checked and we already had a test that includes a hierarchy (`#my.hash.tag0`).

It's very interesting that the Windows tests are failing the way they do. It suggests that the regular expressions are not working in the same way in Windows and other versions. I'll see what I can find about this.

kevinslin:
you're right, missed that. for the windows regex issue, feel free to leave that to me. i'll have some time tomorrow to debug and it's a lot easier with a windows machine. in the future, we'll also have a windows vm for the team to ssh into to debug these sorts of issues

kevinslin (approved):

PR #1046, opened 2021-07-27
feat: add effects modifier to lookup v3

hikchoi:
This PR:
- adds copyNoteLink modifier to `NoteLookup`
- adds a test for said modifier

PR #1056, opened 2021-07-29
feat: add remaining modifiers to NoteLookup

hikchoi:
This PR:
- Adds horizontal split modifier (#1045)
- Adds copy note link modifier (#1046)
- Adds tests for various modifier interactions

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
@@ -1311,7 +1311,7 @@ suite("NoteLookupCommand", function () {
           });
           
           expect(content).toEqual([
-            "[[Ch1foo.ch1]]",
+            "[[Ch1|foo.ch1]]",
             "[[Bar|bar]]",
             "[[Lorem|lorem]]",
             "[[Ipsum|ipsum]]",
--- a/packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
@@ -394,7 +394,6 @@ suite("NoteLookupCommand", function () {
             noteType: LookupNoteTypeEnum.journal
           });
 
-          controller.quickpick.show();
           let { journalBtn, scratchBtn } = getNoteTypeButtons(
             controller.quickpick.buttons
           )
@@ -433,7 +432,6 @@ suite("NoteLookupCommand", function () {
             noteType: LookupNoteTypeEnum.scratch
           });
 
-          controller.quickpick.show();
           let { journalBtn, scratchBtn } = getNoteTypeButtons(
             controller.quickpick.buttons
           )
@@ -472,7 +470,6 @@ suite("NoteLookupCommand", function () {
             noteType: LookupNoteTypeEnum.journal
           });
 
-          controller.quickpick.show();
           let { journalBtn, scratchBtn } = getNoteTypeButtons(
             controller.quickpick.buttons
           )
@@ -558,7 +555,6 @@ suite("NoteLookupCommand", function () {
             selectionType: "selection2link",
           });
 
-          controller.quickpick.show();
 
           let { selection2linkBtn, selectionExtractBtn } = getSelectionTypeButtons(
             controller.quickpick.buttons
@@ -638,7 +634,6 @@ suite("NoteLookupCommand", function () {
             selectionType: "selectionExtract",
           });
 
-          controller.quickpick.show();
 
           let { selection2linkBtn, selectionExtractBtn } = getSelectionTypeButtons(
             controller.quickpick.buttons
@@ -680,7 +675,6 @@ suite("NoteLookupCommand", function () {
             selectionType: "selection2link",
           });
 
-          controller.quickpick.show();
 
           let { selection2linkBtn, selectionExtractBtn } = getSelectionTypeButtons(
             controller.quickpick.buttons
@@ -764,7 +758,6 @@ suite("NoteLookupCommand", function () {
             splitType: "horizontal"
           });
 
-          controller.quickpick.show();
 
           let { horizontalSplitBtn } = getSplitTypeButtons(
             controller.quickpick.buttons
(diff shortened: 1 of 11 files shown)


packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -671,5 +713,611 @@ suite("NoteLookupCommand", function () {
...
+          });
+          controller.quickpick.show();
+
+          const { copyNoteLinkBtn } = getEffectTypeButtons(
+            controller.quickpick.buttons
+          );
+
+          // hack: need to wait here a bit.
+          // TODO: A more elegant way to do this.
+          await new Promise(resolve => setTimeout(resolve, 500));
```
hikchoi, line 808: example of test that needs waiting @kevinslin

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -671,5 +713,611 @@ suite("NoteLookupCommand", function () {
...
+          const { cmd } = await prepareCommandFunc({
+            wsRoot,
+            vaults,
+            engine,
+            opts: { split: true }
+          });
+
+          // if we don't wait here, the test runs too fast and misses one note.
+          // TODO: find a better way to get past this.
+          await new Promise((resolve) => setTimeout(resolve, 500));
```
hikchoi, line 1266: another one @kevinslin

hikchoi:
Submitting a draft for timely communication.

pending tasks:
- [ ] fix note type modifier so that failing tests pass
- [ ] make multi select + copyNoteLink test work

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -671,5 +713,611 @@ suite("NoteLookupCommand", function () {
...
+            fname: "ipsum",
+            vault: vaults[0],
+            wsRoot
+          });
+        },
+        onInit: async ({ wsRoot, vaults, engine }) => {
+          // make clean slate.
+          VSCodeUtils.closeAllEditors();
+
+          VSCodeUtils.openNote(engine.notes["foo"]);
```
kevinslin, line 1256: if you add an `await` here, you should be able to remove the `setTimeout`

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
kevinslin: copying should only happen after `onTriggerButton`. shouldn't await come after that? also removed it locally and this worked fine

kevinslin (changes requested):

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
hikchoi: ooooh I feel dumb now hahaha

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
hikchoi: For some reason `await clipboard.readText()` kept grabbing whatever was in my clipboard before, not the note links. Maybe it's not a flaky test and it just doesn't work properly 😅  I'll look into it.

hikchoi: It's really weird because it worked last night and now it doesn't regardless of the `setTimeout`.

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -671,5 +713,596 @@ suite("NoteLookupCommand", function () {
...
+          ));
+
+          expect(horizontalSplitBtn.pressed).toBeFalsy();
+
+          done();
+        }
+      });
+    });
+
+    test("copyNoteLink basic", (done) => {
```
kevinslin, line 787: the reason `copyNoteLink basic` test doesn't work is because button effects get triggered on `prepareQuickPick`. at this point, the picker has no items attached to it. that only happens during `showQuickPick`. this doesn't affect other modifiers because they don't rely on the `picker` having items attached to it. 

what I would do in this case is just remove the `CopyNoteLink` tests for now - we might modify it later but right now its not worth the effort

kevinslin (changes requested):
high level comment:

you're doing too much with the tests right now. while my feedback for most people is to add more test, I'm going to give you the opposite feedback here. too many tests aren't great since we're still changing the behavior rather frequently which means we need more tests to be updated. think of our tests as guardrails vs blueprints  - they're here to make sure big features work, not to execute all code paths 

right now, speed of delivery is much more important then getting full coverage

kevinslin (changes requested):
also heads up I think there's a bug in the following where button actions aren't being awaited. see `-/+` signs for diff

```ts
  static async refreshPickerBehavior(opts: {
    quickpick: DendronQuickPickerV2;
    buttons: DendronBtn[];
  }) {
    const buttonsEnabled = _.filter(opts.buttons, { pressed: true });
    const buttonsDisabled = _.filter(opts.buttons, { pressed: false });
    // call onDisable first so that
    // they don't modify state of the quickpick after onEnable.
    await Promise.all(
      buttonsDisabled.map((bt) => {
        - bt.onDisable({ quickPick: opts.quickpick });
         + bt.onDisable({ quickPick: opts.quickpick });
      })
    );
    await Promise.all(
      buttonsEnabled.map((bt) => {
        - bt.onEnable({ quickPick: opts.quickpick });
        + return bt.onEnable({ quickPick: opts.quickpick });
      })
    );
  }
```

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
hikchoi: I'll revert it back so the test skips (pushed after enabling again last night to demonstrate where it failed).

I won't spend time fixing it further, but wouldn't `controller.quickpick.show()` below attach the items to the quickpick?

hikchoi:
>  think of our tests as guardrails vs blueprints - they're here to make sure big features work, not to execute all code paths
> right now, speed of delivery is much more important then getting full coverage

noted 👍🏼

hikchoi:
changing this into a draft. Found an issue with a weird `onHide()` behavior.

hikchoi:
onHide behavior is fixed and hopefully this is ready to be merged 😅

kevinslin (approved):

PR #1057, opened 2021-07-29
Feat/GitHub publish pod

Harshita-mindfire:


.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
             node_modules
             packages\*\node_modules
             packages\*\.vscode-test
-          key: ${{ runner.os }}-${{ matrix.node-version }}-node-modules-1-${{ hashFiles('yarn.lock')}}
+          key: ${{ runner.os }}-${{ matrix.node-version }}-node-modules-2-${{ hashFiles('yarn.lock')}}
```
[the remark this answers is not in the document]
Harshita-mindfire: 69d010b

packages/plugin-core/src/commands/PublishPod.ts
```diff
@@ -90,7 +90,12 @@ export class PublishPodCommand extends BaseCommand<
     }
   }
 
-  async showResponse() {
-    window.showInformationMessage("contents copied to clipboard");
+  async showResponse(resp: any) {
```
jonathanyeung, line 93: Where does this get invoked? Is it possible to move the github specific behavior into an override method within `GithubPod`?

Harshita-mindfire: It gets invoked from [here](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/commands/base.ts#L85).

jonathanyeung: Ok for to leave it for now since it's such a small functionality, but in general we should find ways to avoid putting child class logic into base class logic, even if it means refactoring how the existing code is laid out.

jonathanyeung (approved):

PR #1063, opened 2021-07-30
feat: frontmatter tags

SeriousBug:
Tags now can be added to the frontmatter. Tags are added as

```yaml
tags:
 - foo
 - bar.baz
```
Or, for a single tag there is a shorthand:

```yaml
tags: foo
```

This includes support for:
- backlinks for frontmatter tags
- Goto Note with a frontmatter tag selected
- hover and "Ctrl+click" on a frontmatter to preview/goto the tag note
- Rename/Move/Refactor renames the frontmatter tags
This also includes a preliminary attempt at adding publishing support, which doesn't seem to be functioning. I'm currently having trouble diagnosing this, so I will be handling it later in another PR. Since it doesn't seem to break anything, I opted to keep it in.


packages/common-all/src/dnode.ts
```diff
@@ -326,6 +329,17 @@ export class NoteUtils {
...
+   * 
+   * Captured groups:
+   * 
+   * * `singleTag`: Text is a tag like `tags: foo`, and this is the text (`foo`) of that tag.
+   * * `multiTag`: Text is a list of tags, and this is that list including all list syntax. Use `RE_FM_LIST_ITEM` to extract individual items.
+   * * `lastMultiTag`: Text is a list of tags, and this is the text (`foo`) of the last tag in the list.
+   */
+  static RE_FM_TAGS = new RegExp(`^(?<beforeTags>tags: *)((?<singleTag>${LINK_NAME})|(?<multiTag>(\n+ *- *(?<lastMultiTag>${LINK_NAME}))+))`, "m");
+  /** Extracts strings from a YAML list. */
+  static RE_FM_LIST_ITEM = new RegExp(`^(?<beforeListItem> *- *)(?<listItem>${LINK_NAME})`);
```
kevinslin, line 342: heads up that tags can also be in inline list form. eg:  [one, two, three]

test-workspace/vault/dendron.ref.frontmatter-tags.md
```diff
@@ -0,0 +1,10 @@
+---
+id: HyqmR36gQFN45GCI
+title: Frontmatter Tags
+desc: ''
+updated: 1627071084405
+created: 1627070098539
+tags: test
```
kevinslin, line 7: can we create multiple children of this to test different cases? `fm-tags.single`, `fm-tags.multi`, ...

packages/common-all/src/dnode.ts
```diff
@@ -793,7 +807,17 @@ export class NoteUtils {
   }
 
   static serializeMeta(props: NoteProps) {
-    const builtinProps = _.pick(props, [
+    // Remove all undefined values, because they cause `matter` to fail serializing them
+    // @ts-ignore
```
kevinslin, line 811: instead of doing a ts-ignore, can we declare it a partial here and then create a validation type narrowing function `validateNoteProps(noteprops: Partial<NoteProps>)` that validates we have the necessary fields?

packages/common-all/src/utils.ts
```diff
@@ -128,3 +129,24 @@ export class DefaultMap<K, V> {
...
+  static replaceTag({note, oldTag, newTag}: {note: NoteProps, oldTag: string, newTag?: string}) {
+    if (_.isUndefined(note.tags) || _.isString(note.tags)) {
+      note.tags = newTag;
+    } else {
+      const index = _.findIndex(note.tags, (tag) => tag === oldTag);
+      if (newTag) {
+        if (index >= 0) {
+          note.tags[index] = newTag;
+        } else {
+          // Weird, can't find the old tag. Add the new one anyway.
```
kevinslin, line 144: does this happen in practice?

packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -69,12 +71,27 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
...
+      if (insertTitle) {
+        const idx = _.findIndex(root.children, (ent) => ent.type !== "yaml");
+        root.children.splice(
+          idx,
+          0,
+          u(DendronASTTypes.HEADING, { depth: 1 }, [u("text", note.title)])
+        );
+      }
+      // Add frontmatter tags, if any, ahead of time. This way wikilink compiler will pick them up and render them.
+      if (note?.tags && note.tags.length > 0) {
```
kevinslin, line 83: this should be configurable. eg: `showFrontMatterTags`

packages/engine-test-utils/src/presets/engine-server/rename.ts
```diff
@@ -864,6 +864,211 @@ const NOTES = {
...
+    async ({ wsRoot, vaults, engine }) => {
+      await engine.renameNote({
+        oldLoc: {
+          fname: "tags.foo",
+          vaultName: VaultUtils.getName(vaults[0]),
+        },
+        newLoc: { fname: "bar", vaultName: VaultUtils.getName(vaults[0]) },
+      });
+      const note = NoteUtils.getNoteByFnameV5({fname: "primary", notes: engine.notes, wsRoot, vault: vaults[0]});
+      const noteFileContents = await readFile(NoteUtils.getFullPath({note: note!, wsRoot}), "utf-8");
```
kevinslin, line 1040: we have a shortcut function, [checkFile](https://github.com/dendronhq/dendron/blob/feat/frontmatter-tags/packages/engine-test-utils/src/utils/index.ts#L49:L49) that makes this more concise. apply to previous/subsequent cases

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -93,6 +94,30 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+            };
+          }
+          // Otherwise, check the text up to the selection for frontmatter tags
+          // list. The last tag in that is either the selection, or something
+          // before the selection. If it is the selection, then the user has the
+          // tag selected. This is not perfect since the user could be selecting
+          // something past the frontmatter that just happens to match the last
+          // tag, but that's an edge case that's not disruptive.
+          const frontmatterText = editor?.document.getText(new Range(0, 0, selection.start.line, currentLine.length));
+          const selectedItem = currentLine.match(NoteUtils.RE_FM_LIST_ITEM)?.groups?.listItem;
```
kevinslin, line 113: as mentioned above, this regex will mix inline list case. wondering if it would be less error prone to just grab the text and run [string2note](https://github.com/dendronhq/dendron/blob/feat/frontmatter-tags/packages/common-server/src/filesv2.ts#L138:L138) to parse for tags? as an optimization, we could just parse the frontmatter instead of the entire note

kevinslin: we could refactor [this](https://github.com/dendronhq/dendron/blob/feat/frontmatter-tags/packages/plugin-core/src/services/NoteSyncService.ts#L21:L21) into a general utility

kevinslin (changes requested):

packages/plugin-core/src/commands/GotoNote.ts
SeriousBug: The problem is, a parsed note has no information about what line of text each frontmatter property came from. The `matter` library we use doesn't provide that information, just the parsed YAML. I considered adding a unist YAML parser to parse the contents of the frontmatter into a tree, but opted to just parse with the regex for now since it's simpler.

But I'm thinking now that it might actually be a better choice to use the actual parser, since the YAML is more complex that I initially thought and my implementation doesn't do a great job of parsing it.

packages/common-all/src/utils.ts
SeriousBug: No, the old tag always gets removed in the tests. I decided to cover that case just in case.

kevinslin (changes requested):
1. notice that inline tags are still giving me problem. for the following

```
tags: [one]
```
running `GoTo Note` results in `tags. .md` being created and the following error

```js
stack trace: YAMLException: unexpected end of the stream within a flow collection at line 8, column 1:
    
    ^
	at generateError (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:167:10)
	at throwError (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:173:9)
	at readFlowCollection (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:770:3)
	at composeNode (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1360:11)
	at readBlockMapping (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1089:11)
	at composeNode (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1359:12)
	at readDocument (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1525:3)
	at loadDocuments (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1588:5)
	at load (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1614:19)
	at Object.safeLoad (/Users/kevinlin/code/dendron/node_modules/js-yaml/lib/js-yaml/loader.js:1637:10)
	at Object.parse (/Users/kevinlin/code/dendron/packages/common-server/lib/filesv2.js:116:49)
	at /Users/kevinlin/code/dendron/node_modules/gray-matter/lib/parse.js:12:17
	at parseMatter (/Users/kevinlin/code/dendron/node_modules/gray-matter/index.js:109:17)
	at Object.matter [as default] (/Users/kevinlin/code/dendron/node_modules/gray-matter/index.js:50:10)
	at Object.string2Note (/Users/kevinlin/code/dendron/packages/common-server/lib/filesv2.js:121:58)
	at NoteSyncService.onDidChange (/Users/kevinlin/code/dendron/packages/plugin-core/out/src/services/NoteSyncService.js:107:36)
	at runNextTicks (internal/process/task_queues.js:58:5)
	at processTimers (internal/timers.js:494:9)
/Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-fork.js:5
rejected promise not handled within 1 second: YAMLException: unexpected end of the stream within a flow collection at line 8, column 1:
    


```


2. notice that its easy to write a tag like the following
```
tags:
  - `one `
```
can we strip whitespace at the end?

SeriousBug:
Thanks for catching the whitespace issue, I also noticed that tags with quotes were not handled correctly. Fixed both.

I am unable to reproduce the other error you mention. It's coming from `NoteSyncService` so I don't think it's related to `GotoNote`.
It could it be that when you are typing the new frontmatter tag, as your edits sync it leaves the frontmatter in a broken state at some point (e.g. when you just typed `tags: ` and haven't typed the rest yet).

kevinslin:
hmm, working for me now, i think i might have not had the yaml dependency installed properly. this looks good. can you resolve the merge conflicts? feel free to merge afterwards

kevinslin (approved):

PR #1067, opened 2021-07-30
enhance: performance improvement for block/header autocomplete

SeriousBug:
When doing an existing blocks only autocomplete (`[[#` or `[[#^`) the blocks are filtered on the server side rather than the client side. This improves responsiveness in the editor for long notes since less work is done within the editor.

### DIFF
--- a/test-workspace/scripts/randomNote.js
+++ b/test-workspace/scripts/randomNote.js
@@ -1,0 +1,130 @@
+#!/usr/bin/env node
+/** Creates a randomly generated note.
+ *
+ * The generated note will contain paragraphs, lists, tables, wikilinks, block
+ * anchors etc., trying to make use of Dendron features.
+ * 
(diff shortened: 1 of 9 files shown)


test-workspace/scripts/randomNote.js
```diff
@@ -0,0 +1,130 @@
...
+ * The generated note will contain paragraphs, lists, tables, wikilinks, block
+ * anchors etc., trying to make use of Dendron features.
+ * 
+ * There is currently no options for this script, but you can edit the following variables to edit what gets generated.
+ */
+/* eslint-disable no-console */
+/* eslint-disable no-lone-blocks */
+/* eslint-disable no-plusplus */
+
+/** How many markdown elements to generate. */
```
kevinslin, line 13: could you add some docs in `[[dendron.dev.qa.perf]]` on using this? (doesn't exist yet)

kevinslin (approved):

PR #1069, opened 2021-07-31
feat: custom tag coloring

SeriousBug:
Allows users to set custom colors for tags. The colors are displayed in the editor highlighting, and in publishing.


packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -116,18 +126,31 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
...
+            (colorType === "configured" ||
+              opts?.noRandomlyColoredTags !== false)
+          ) {
+            color = makeColorTranslucent(maybeColor, TAG_BG_TRANSLUCENCY);
+          }
         }
 
         const copts = opts?.wikiLinkOpts;
         if (opts?.transformNoPublish) {
-          const notes = NoteUtils.getNotesByFname({
```
kevinslin, line 125: 1.  Heads up that a processor can now run in `IMPORT` mode which means notes don't exist in the engine. I will make sure to create a test case for this.
What this means is that we shouldn't try to get the note unless we need it which is why the calls to `getNotesByFname` are gated by the `if` statements
https://github.com/dendronhq/dendron-site/blob/dev/vault/pkg.dendron-markdown.arch.md#procmode

kevinslin (changes requested):

packages/engine-server/src/markdown/remark/dendronPub.ts
SeriousBug: Thanks for bringing that up. I'm slightly confused though:
1. In what situation would `dendronPub` get run while in `IMPORT` mode?
2. Is the existing code properly gated for this? I see [this line](https://github.com/dendronhq/dendron/blob/master/packages/engine-server/src/markdown/remark/dendronPub.ts#L72) which is gated behind `!insideNoteRef && root.children`, but otherwise it will throw an error if there is no note.

kevinslin: today it happens for the `MarkdownImportPod`. see [here](https://github.com/dendronhq/dendron/blob/20ed719af4835bc844dfd7012896d711f282895d/packages/pods-core/src/builtin/MarkdownPod.ts#L298-L298). the places where its not gated are bugs.

SeriousBug: Thanks, got it. I'll update `dendronPub` to cover that then. I can also add a test case for it if you haven't worked on that yet.

kevinslin:
That would be awesome. Thanks!

kevinslin (approved):

PR #1073, opened 2021-08-02
feat: obsidian import pod

hfellerhoff:
Changes:

- [x] Add import pod
- [x] Add basic Obsidian to Dendron parsing
- [x] Optimize parsing (limit duplicate selectors, bugfixes)
- [x] Add field to `dendron.yml` specifying which style file to apply to graph
- [x] Change `Configure Graph Styles` command to open whatever file is specified in `dendron.yml` under `graph: stylePath`


kevinslin (changes requested):
1. can you rebase from master: seems like theirs a conflict. 
2. can we add some tests for this?

Issue #421, opened 2020-12-22
Self-reference not updated on refactor/rename


rasmuse:
Thanks for addressing #406 so swiftly!

All the error messages are now gone, but I still have a couple of unexpected behaviors, for which I am now opening two separate issues. This is the first.

## My system

- Ubuntu 20.04.1 LTS
- VS Code v1.52
- Dendron 0.21.0

## How to reproduce

Create a note named `some-note` (using lookup -> create new, if that matters). Enter the following content (below the frontmatter):

```markdown
[[some-note]]
```

Save (and optionally close) the file.

Rename it using refactor or rename, e.g., to `some-note-renamed`.

## Expected result

A note `some-note-renamed` with the following content:

```markdown
[[some-note-renamed]]
```

## Actual result

The note is renamed to `some-note-renamed`, but its content is still

```markdown
[[some-note]]
```

kpathakota:
Hmm... I'm struggling to see the value of self-referenced links. Is this for publishing 🧐? Nonetheless ... it is a valid bug that still repros.

rasmuse:
Thanks @kpathakota for working on this.

No, I don't have an important use case for self-references. The reason I brought it up was that for me it is very important that rename/refactor is dependable and well defined. I did have some issues previously with rename/refactor (see #406) and then I tried to break down very carefully what behavior I expect from it. This issue as well as #422 were the remaining open questions after the error messages in #406 were resolved.

I leave it up to someone else to decide whether this issue is actually worth solving. But _in principle_ I insist that self-references should be handled as suggested above.

aleksey-rowan:
@rasmuse, I believe there a duplicate issue addressing the same problem: https://github.com/dendronhq/dendron/issues/1207

I'll close this one as the linked issue is better triaged. Please upvote it or leave a comment there.

Issue #370, opened 2020-11-22
Rename header command


tsnieman:
**Is your feature request related to a problem? Please describe.**
if using `[[foo#header]]` syntax, i have to manually update all header-link references.

**Describe the solution you'd like**
Using the header of the current section of the note for editor cursor position, rename the heading and perform an update to all relevant header-link references.

Behavior outline:
1. have `[[foo#header]]` in various files
2. place editor cursor (i.e. focused input cursor) in section of `foo.md` that has a heading above it.
3. trigger the `Dendron: Refactor Header` command
4. prompt asks for new heading (prefilled with current heading)
5. user types and submits `bar`
6. the heading of their current editor section updated to the user's input `bar`
7. all references to `[[foo#header]]` across all hierarchies get changed to `[[foo#bar]]`
  - this should also include `[[Formatted links|foo#header]]`-like references as well.

**Describe alternatives you've considered**
- Search-and-replace for `foo#header]]` ? :man_shrugging: 

**Additional context**
- Thought of this after seeing [the 0.17.0 release](https://www.dendron.so/notes/f09162ec-6b20-4b53-bd5c-68aaa4698ed8.html)
- [Discord reference](https://discord.com/channels/717965437182410783/735365126227493004/780210944793640970)

Issue #189, opened 2020-09-11
Rename Header Command


kevinslin:
## Context

> Renaming notes changes their references automatically and headers probably need the same kind of love now). Or at least some visual indicator that this header is referenced somewhere and editing it would break things.
- Concerned Dendronite

## Proposal
In the future, just like you can rename a note, you should be able to rename a header. This will cause any references pointing to that header to also rename.

aleksey-rowan:
[Rename Header](https://wiki.dendron.so/notes/srajljj10V2dl19nCSFiC/#rename-header) command is now available.

PR #1081, opened 2021-08-04
feat: rename header command

SeriousBug:
Adds the rename header command. The command works with both wikilinks and references, including range references (`![[foo#start:#end]]`). It also correctly handles headers that include block anchors or wikilinks (`## foo [[bar|my.bar]] baz`).


SeriousBug:
#421 #370 #189

kevinslin (changes requested):
this looks good. one addition: currently our heads usually have an alias associated with it because that's the default behavior of `CopyNoteLink`. when renaming a header link, if the alias is the same as the header, lets also rename the alias to the new header

kevinslin:
Looks like there's an error from the alias being undefined

SeriousBug:
Sorry I was still working on it, I'll let you know when it's ready for review.

kevinslin (approved):

kevinslin:
Looks good. Can you resolve the merge conflict? feel free to merge afterwards

PR #1083, opened 2021-08-04
fix: highlighting is not displayed

SeriousBug:
In some cases, highlighting would not get displayed. The issue was caused by a bug where unused decorators were being discarded rather than emptied. This would silently break highlighting as any type of highlighting that becomes unused is discarded and no longer can be displayed.


kevinslin (approved):

SeriousBug:
#1093

PR #1065, opened 2021-07-30
chore: replace moment with luxon in calendar-view

namjul:
This PR contains:

- remove `moment` imports in `calendar-view.tsx`
- quickfix [`9744b24` (#1065)](https://github.com/dendronhq/dendron/pull/1065/commits/9744b2435f9157d913ebebf108e47728c196a029)
- makes `firstDayOfWeek` a noop [[Start of Week|scratch.2021.07.31.080905.start-of-week]]

[webpack-analyze-bundle before](https://user-images.githubusercontent.com/328836/127860298-dfc0b7f8-1522-4157-a377-b8409bf7aa07.png)
[webpack-analyze-bundle after](https://user-images.githubusercontent.com/328836/127860293-8dd95c39-baf2-437a-903b-e1bb8605f9e4.png)

PR #1088, opened 2021-08-04
feat(calendar-view): allow journal settings deviating from defaults

namjul:
This is a follow up issue from https://github.com/dendronhq/dendron/pull/1065#pullrequestreview-721482799

- It includes:
  - [x] journal.dateFormat: change year, month, and day component (tokens from https://github.com/moment/luxon/blob/master/docs/formatting.md#table-of-tokens should work)
  - [x] journal.dailyDomain: change to non default value
  - [x] journal.name: change to non default value 
  - ~~[ ] journal.addBehavior: change to other add behavior and see if it works~~
  - adds `redux-logger` to have redux actions visible inside vscode (redux devtools don't work there)
- fixes:
  - issue with `syncNote` which gets called with `note` being `undefined`.
  - calendar-view's memo `isEqual` function to rather consider `engine.notes.length`

### DIFF
--- a/packages/dendron-next-server/pages/vscode/calendar-view.tsx
+++ b/packages/dendron-next-server/pages/vscode/calendar-view.tsx
@@ -58,20 +58,6 @@ function getMaybeDatePortion({ fname }: NoteProps, journalName: string) {
 
 const today = luxonGenerateConfig.getNow();
 const { EngineSliceUtils } = engineSlice;
-const {
-  journal: {
-    dailyDomain: DEFAULT_JOURNAL_DAILY_DOMAIN,
-    dateFormat: DEFAULT_JOURNAL_FORMAT,
-    name: DEFAULT_JOURNAL_NAME,
-  },
-} = {
-  // TODO replace with `DConfig.getDefaultConfig() from @dendronhq/engine-server`
-  journal: {
-    dailyDomain: "daily",
-    name: "journal",
-    dateFormat: "y.MM.dd",
-  },
-};
 
 function CalendarView({ engine, ide }: DendronProps) {
   // --- init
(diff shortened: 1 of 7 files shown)


packages/dendron-next-server/pages/vscode/calendar-view.tsx
```diff
@@ -81,19 +81,20 @@ function CalendarView({ engine, ide }: DendronProps) {
 
   const maxDots: number = 5;
   const wordsPerDot: number = 250;
-  const dailyJournalDomain = config?.journal.dailyDomain || "daily";
-  const defaultJournalName = config?.journal.name || "journal";
+  const journalDailyDomain = config?.journal.dailyDomain || "daily";
+  const journalName = config?.journal.name || "journal";
 
   // luxon token format lookup https://github.com/moment/luxon/blob/master/docs/formatting.md#table-of-tokens
-  let defaultJournalDateFormat = config?.journal.dateFormat || "y.MM.dd";
```
kevinslin, line 88: instead of hardcoding, can we get the default value from `Dconfig.genDefaultConfig().journal.dateFormat`? same comments to other fields

kevinslin (changes requested):

packages/dendron-next-server/pages/vscode/calendar-view.tsx
namjul: [`fa20956` (#1088)](https://github.com/dendronhq/dendron/pull/1088/commits/fa20956b1b0a078589402d3a7f561a646f954ba1)

I tried using `DConfig.genDefaultConfig()` from `@dendronhq/engine-server` but errord implying that the imported code fails on the frontend. We would need to move `getDefaultConfig` to `@dendronhq/common-all` or something similar.

The reason we need default in `calendar-view.tsx` is because in the first call `config` is `undefiend` and certain functions would fail because of it. It currently does accidentally not happen because `notes` and `noteActive` are at that time also `undefined` and therefore the functions won't be called.

namjul: Solved it this way [`9a3fec3` (#1088)](https://github.com/dendronhq/dendron/pull/1088/commits/9a3fec3a25f114726ef5752af3459b7714ba0d01)

kevinslin (approved):

namjul:
I cleanup up the branch. There are still some things not working after a quick QA test. Will get to it on Monday.

kevinslin (approved):

PR #1089, opened 2021-08-04
Enhance/markdown publish to resolve wikilink

Harshita-mindfire:
https://www.loom.com/share/af61c957794d4b60a70e6837f084d67c

### DIFF
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -983,23 +983,18 @@ export class RemarkUtils {
         let dirty = false;
 
         wikiLinks.forEach((linkNode) => {
-          const existingNote = _.find(engine.notes, { fname : linkNode.value });
-          if(existingNote) {
-            const urlRoot = dendronConfig.site?.siteUrl || '';
-            const { vault } = existingNote;  
+          const existingNote = _.find(engine.notes, { fname: linkNode.value });
+          if (existingNote) {
+            const urlRoot = dendronConfig.site?.siteUrl || "";
+            const { vault } = existingNote;
             linkNode.value = WorkspaceUtils.getNoteUrl({
               config: dendronConfig,
               note: existingNote,
               vault,
               urlRoot,
-            })
-              const newValue = _.replace(linkNode.value, /\./g, "/");
-              if (linkNode.data.alias === linkNode.value) {
-                linkNode.data.alias = newValue;
-              }
-              dirty = true;
+            });
+            dirty = true;
           }
-          
         });
         //TODO: Add support for Ref Notes and Block Links
 
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -986,13 +986,12 @@ export class RemarkUtils {
           const existingNote = _.find(engine.notes, { fname : linkNode.value });
           if(existingNote) {
             const urlRoot = dendronConfig.site?.siteUrl || '';
-            const vault = existingNote.vault;  
+            const { vault } = existingNote;  
             linkNode.value = WorkspaceUtils.getNoteUrl({
               config: dendronConfig,
               note: existingNote,
               vault,
               urlRoot,
-              maybeNote: existingNote
             })
               const newValue = _.replace(linkNode.value, /\./g, "/");
               if (linkNode.data.alias === linkNode.value) {
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -964,6 +964,44 @@ export class RemarkUtils {
     };
   }
 
+  static convertWikiLinkToUrl(
+    note: NoteProps,
+    changes: NoteChangeEntry[],
+    engine: DEngineClient,
+  ) {
+    return function (this: Processor) {
+      return (tree: Node, _vfile: VFile) => {
+        const root = tree as DendronASTRoot;
+        const wikiLinks: WikiLinkNoteV4[] = selectAll(
+          DendronASTTypes.WIKI_LINK,
+          root
+        ) as WikiLinkNoteV4[];
+
+        let dirty = false;
+
+        wikiLinks.forEach((linkNode) => {
+          const existingNote = _.find(engine.notes, { fname : linkNode.value });
+          if (linkNode.value.indexOf(".") >= 0) {
+            const newValue = _.replace(linkNode.value, /\./g, "/");
+            if (linkNode.data.alias === linkNode.value) {
+              linkNode.data.alias = newValue;
+            }
+            linkNode.value = `https://dendron/notes/${existingNote?.id}`;
+            dirty = true;
+          }
+        });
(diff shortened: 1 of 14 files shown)


packages/engine-test-utils/src/__tests__/pods-core/MarkdownPod.spec.ts
```diff
@@ -58,6 +79,46 @@ describe("markdown publish pod", () => {
       { expect, preSetupHook: ENGINE_HOOKS.setupRefs }
     );
   });
+
+  test("wikiLinktoURL with siteUrl", async () => {
```
kevinslin, line 83: can you add additional test cases here?

- test with xvault link to same vault
- test with xvault link to a seed vault with different url
- test with a regular link in a seed vault that has a different url

See [here](https://github.com/dendronhq/dendron/blob/master/packages/engine-test-utils/src/__tests__/dendron-cli/commands/seedCLICommand.spec.ts#L1:L1) for an example about testing with seeds

packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -878,6 +881,52 @@ export class RemarkUtils {
...
+          if (existingNote) {
+            const urlRoot = dendronConfig.site?.siteUrl || "";
+            const { vault } = existingNote;
+            linkNode.value = WorkspaceUtils.getNoteUrl({
+              config: dendronConfig,
+              note: existingNote,
+              vault,
+              urlRoot,
+            });
+            const newValue = _.replace(linkNode.value, /\./g, "/");
```
kevinslin, line 911: this shouldn't be necessary

packages/pods-core/src/builtin/MarkdownPod.ts
```diff
@@ -320,20 +320,30 @@ export class MarkdownImportPod extends ImportPod<MarkdownImportPodConfig> {
...
   static description: string = "publish markdown";
 
-  get config(): JSONSchemaType<PublishPodConfig> {
+  get config(): JSONSchemaType<MarkdownPublishPodConfig> {
     return PodUtils.createPublishConfig({
       required: [],
-      properties: {},
-    }) as JSONSchemaType<PublishPodConfig>;
+      properties: {
+        wikiLinkToURL: {
```
kevinslin, line 335: the default here should be false

packages/pods-core/src/utils.ts
```diff
@@ -211,7 +198,10 @@ export class PodUtils {
   }
 
   static hasRequiredOpts(_pClassEntry: PodClassEntryV4): boolean {
```
kevinslin, line 200: can you explain the logic here? we should not show a config prompt if the pod has no required ops. instead of doing a special case, lets change this to return true if any of the fields in the pod are nullable and don't have a default

kevinslin (changes requested):

kevinslin:
also, can you rebase off of master? there's a bunch of formatting changes there that are now mixed in this diff

packages/engine-server/src/workspace/utils.ts
```diff
@@ -57,4 +68,97 @@ export class WorkspaceUtils {
...
+  /**
+   * Generate url for given note or return `undefined` if no url is specified
+   * @param opts
+   *
+   */
+  static getNoteUrl(opts: {
+    config: DendronConfig;
+    note: NoteProps;
+    vault: DVault;
+    urlRoot?: string;
```
jonathanyeung, line 81: Seems like urlRoot is determined from within DendronConfig.  Should we remove this parameter and move the `const urlRoot = dendronConfig.site?.siteUrl || "";` logic into this function?

packages/engine-server/src/workspace/utils.ts
```diff
@@ -57,4 +68,97 @@ export class WorkspaceUtils {
...
+          }
+        }
+      }
+    }
+    let root = "";
+    if (!_.isUndefined(urlRoot)) {
+      root = urlRoot;
+    } else {
+      // assume github
+      throw Error("not implemented");
```
jonathanyeung, line 115: convert to use `DendronError`.  What does "assume github" mean here?  If we decide to keep the `urlRoot` parameter, then it should be a mandatory parameter.

packages/engine-server/src/workspace/utils.ts
```diff
@@ -57,4 +68,97 @@ export class WorkspaceUtils {
...
+   * Generate url for given note or return `undefined` if no url is specified
+   * @param opts
+   *
+   */
+  static getNoteUrl(opts: {
+    config: DendronConfig;
+    note: NoteProps;
+    vault: DVault;
+    urlRoot?: string;
+    maybeNote?: NoteProps;
```
jonathanyeung, line 82: What's this parameter for?

packages/engine-server/src/workspace/utils.ts
Harshita-mindfire: We could, but in `CopyNoteUrl` it is also determined from `DendronWorkspace`
```ts
const urlRoot =
      config?.site?.siteUrl ||
      DendronWorkspace.configuration().get<string>(
        CONFIG.COPY_NOTE_URL_ROOT.key
      );
```

packages/engine-server/src/workspace/utils.ts
Harshita-mindfire: the open document in vscode provided by `CopyNoteUrl` command

packages/engine-server/src/workspace/utils.ts
Harshita-mindfire: After Refactoring copyNoteUrl command, the urlRoot can be undefined.
```ts
const urlRoot =
      config?.site?.siteUrl ||
      DendronWorkspace.configuration().get<string>(
        CONFIG.COPY_NOTE_URL_ROOT.key
      );
```

packages/engine-server/src/workspace/utils.ts
jonathanyeung: Ah ok - I see now, this is just a refactor of the code that was inside `CopyNoteUrl`. Instead of having this as a param, in `CopyNoteUrl` why don't we do a check - if maybeNote is undefined, just return and stop the command execution there (and maybe show a warning/msg dialog to the user). Then we can avoid having this parameter in this function - if we're not viewing a note, then there should be no URL to copy.

packages/engine-server/src/workspace/utils.ts
jonathanyeung: In this case - if urlRoot is not defined, then let's go ahead and stop the command within copyNoteUrl itself.  There's no reason to continue if there's no base URL.

Harshita-mindfire: If urlRoot is not defined here, we move forward and check if the vault is seed. If it is a seed, we determine the value of urlRoot from 
```ts
if (vault.seed) {
      if (config.seeds && config.seeds[vault.seed]) {
        const maybeSite = config.seeds[vault.seed]?.site;
        if (maybeSite) {
          urlRoot = maybeSite.url;
}
```

Else we stop and throw error.

jonathanyeung (approved):

kevinslin (approved):

Issue #1070, opened 2021-07-31
cli vault creation is added to dendron.yml but not dendron.code-workspace


ScriptAutomate:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

Creating a vault via CLI is added to `dendron.yml` but not `dendron.code-workspace`. This means that the vault won't show up in the VSCode workspace/editor unless added properly.

This looks like some changes would need to happen potentially in:

- [packages/dendron-cli/src/commands/vaultCLICommand.ts](https://github.com/dendronhq/dendron/blob/master/packages/dendron-cli/src/commands/vaultCLICommand.ts)
- [packages/engine-server/src/workspace/service.ts: async createVault](https://github.com/dendronhq/dendron/blob/8afa5bf07853d98f3c9939e3b8490c45cf2a5e1a/packages/engine-server/src/workspace/service.ts#L241)
- [packages/engine-server/src/workspace/service.ts: async addVault](https://github.com/dendronhq/dendron/blob/8afa5bf07853d98f3c9939e3b8490c45cf2a5e1a/packages/engine-server/src/workspace/service.ts#L193)

## To Reproduce

```bash
# Create vault via CLI
# npx dendron-cli vault --vault my-new-vault --vaultPath 'repos/my-new-vault'
dendron-cli vault --vault my-new-vault --vaultPath 'repos/my-new-vault'
```

The above command creates a vault, and adds it properly to `dendron.yml`, but does not add it to `dendron.code-workspace` as a listed vault. Opening the workspace in VSCode will now show the vault, unless manually adding it to `dendron.code-workspace`.

## Expected behavior

The new vault should be added also to the `dendron.code-workspace` file:

### Example before creating vault

```json
...
  "folders": [
      {
          "path": "repos/my-old-vault",
          "name": "my-old-vault"
      }
  ],
...
```

### Example after creating vault

```json
...
  "folders": [
      {
          "path": "repos/my-new-vault",
          "name": "my-new-vault"
      },
      {
          "path": "repos/my-old-vault",
          "name": "my-old-vault"
      }
  ],
...
```

[...]

PR #1094, opened 2021-08-04
fix: add new vaults from CLI to code workspace

kevinslin:
Addresses #1070

### DIFF
--- a/packages/dendron-cli/src/commands/vaultCLICommand.ts
+++ b/packages/dendron-cli/src/commands/vaultCLICommand.ts
@@ -63,7 +63,11 @@ export class VaultCLICommand extends CLICommand<CommandOpts, CommandOutput> {
             fsPath: vaultPath,
           };
           const wsService = new WorkspaceService({ wsRoot });
-          const resp = await wsService.createVault({ vault, noAddToConfig });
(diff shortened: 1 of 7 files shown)


packages/dendron-cli/src/commands/vaultCLICommand.ts
```diff
@@ -63,7 +63,11 @@ export class VaultCLICommand extends CLICommand<CommandOpts, CommandOutput> {
             fsPath: vaultPath,
           };
           const wsService = new WorkspaceService({ wsRoot });
-          const resp = await wsService.createVault({ vault, noAddToConfig });
+          const resp = await wsService.createVault({
+            vault,
+            noAddToConfig,
+            addToCodeWorkspace: true,
+          });
           this.print(`${vaultPath} crated`);
```
jonathanyeung, line 71: nit on existing issue: sp on "created"

jonathanyeung (approved):

packages/dendron-cli/src/commands/vaultCLICommand.ts
kevinslin: fixed!

PR #1098, opened 2021-08-05
chore: bad `expect` in plugin tests

SeriousBug:
Tests in `engine-server/rename.ts` were being run in the plugin, which was then failing because the `expect` is not defined. The failed tests would print out errors, but would still pass.

I worked around the failure, another alternative would be to completely strip out the `expect` or avoid the `checkFile` functions in `rename.ts` tests. I haven't investigated why the tests were passing despite the error.

### DIFF
--- a/packages/common-test-utils/src/utils.ts
+++ b/packages/common-test-utils/src/utils.ts
@@ -197,14 +197,10 @@ export async function runJestHarness<TOpts>(
 
 export async function runJestHarnessV2(results: any, expect: any) {
   return _.map(await results, (ent) => {
-    try {
-      if (_.isBoolean(ent.expected) && ent.expected === true) {
-        expect(ent.actual).toBeTruthy();
-      } else {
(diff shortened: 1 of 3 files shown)


kevinslin (changes requested):
can you verify if this is still the case?

checked https://github.com/dendronhq/dendron/blob/master/packages/engine-test-utils/src/presets/engine-server/rename.ts#L22:L22

and don't see any calls to `check*` functions in the current version of master

SeriousBug:
It's the [`checkFile`](https://github.com/dendronhq/dendron/blob/master/packages/engine-test-utils/src/presets/engine-server/rename.ts#L884) function, which then calls `checkString` and `checkNotInString`.

kevinslin:
Ah, I see. Instead of swallowing the error, can we create a new function `checkFileNoExpect` that returns true/false if the check passes?

kevinslin (approved):

PR #1100, opened 2021-08-05
feat: generate json schema from config

kevinslin:
usage: dendron dev generate_json_schema_from_config


SeriousBug (approved):
Looks good, just 2 small things:

1. Should the validator file be in `common-all` instead?
2. We should document the process to add new config options and note that the validator needs to be regenerated.

kevinslin:
1. made change to add it also to `common-all/data`
2. added docs about the process [here](https://github.com/dendronhq/dendron-site/blob/dev/vault/pkg.common-all.dev.md#L36:L36)

PR #1103, opened 2021-08-06
feat: resolve vim keybinding conflict on initial install

hikchoi:
This PR:
- Adds the ability to resolve vim keybinding conflict on initial install ( cmd/ctrl + l )
- Adds tests
- Adds `Extension Integ Tests - File (with extension)`

Note that one of the tests only works if you don't pass in `--disable-extensions` for the integ test. I have added a separate launch configuration for this for now.

### DIFF
--- a/packages/plugin-core/.vscode/launch.json
+++ b/packages/plugin-core/.vscode/launch.json
@@ -144,6 +144,26 @@
       "outFiles": ["${workspaceFolder}/out/src/test/**/*.js"]
     },
     {
+      "name": "Extension Integ Tests - File (with extensions)",
+      "type": "extensionHost",
+      "request": "launch",
+      "runtimeExecutable": "${execPath}",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceFolder}",
+        "--extensionTestsPath=${workspaceFolder}/out/src/test/suite-integ/index",
+      ],
+      "env": {
+        "STAGE": "test",
+        "TEST_TO_RUN": "${fileBasenameNoExtension}"
+      },
+      "skipFiles": [
+        "<node_internals>/**/*.js",
+        "${workspaceFolder}/**/node_modules/mocha/**/*.js",
+        "**/extensionHostProcess.js"
+      ],
+      "outFiles": ["${workspaceFolder}/out/src/test/**/*.js"]
+    },
+    {
       "name": "Extension Integ Tests - File Remote",
       "type": "extensionHost",
       "request": "launch",
--- a/packages/plugin-core/.vscode/launch.json
+++ b/packages/plugin-core/.vscode/launch.json
@@ -159,26 +159,6 @@
       "skipFiles": [
         "<node_internals>/**/*.js",
         "${workspaceFolder}/**/node_modules/mocha/**/*.js",
-        "**/extensionHostProcess.js"
-      ],
-      "outFiles": ["${workspaceFolder}/out/src/test/**/*.js"]
-    },
-    {
-      "name": "Extension Integ Tests - File Remote",
-      "type": "extensionHost",
-      "request": "launch",
-      "runtimeExecutable": "${execPath}",
-      "args": [
-        "--extensionDevelopmentPath=${workspaceFolder}",
-        "--extensionTestsPath=${workspaceFolder}/out/src/test/suite-integ/index",
-      ],
-      "env": {
-        "STAGE": "test",
-        "TEST_TO_RUN": "${fileBasenameNoExtension}"
-      },
-      "skipFiles": [
-        "<node_internals>/**/*.js",
-        "${workspaceFolder}/**/node_modules/mocha/**/*.js",
       ],
       "outFiles": ["${workspaceFolder}/out/src/test/**/*.js"]
     }
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -28,6 +28,7 @@ import {
   WorkspaceService,
 } from "@dendronhq/engine-server";
 import { ExecaChildProcess } from "execa";
+import fs from "fs-extra";
 import _ from "lodash";
 import { Duration } from "luxon";
 import path from "path";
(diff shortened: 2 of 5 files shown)


hikchoi:
looks like the keybinding config path in the CI runner is wrong so the tests are failing. will investigate.

hikchoi:
Keeping this as a draft because the window tests are failing, but otherwise it's ready for review.

packages/plugin-core/.vscode/launch.json
```diff
@@ -143,6 +143,26 @@
       ],
       "outFiles": ["${workspaceFolder}/out/src/test/**/*.js"]
     },
+    {
```
kevinslin, line 146: if you look at the task `      "name": "Extension Integ Tests - File Remote",`  - this is actually running the integ test with extension. feel free to rename it instead of creating a new one

packages/plugin-core/src/_extension.ts
```diff
@@ -425,6 +429,21 @@ export async function _activate(
     toggleViews(false);
   }
 
+  if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
+    const VimInstalled = VSCodeUtils.isExtensionInstalled("vscodevim.vim");
```
kevinslin, line 433: camelcase please for variable names

packages/plugin-core/src/_extension.ts
```diff
@@ -586,3 +605,37 @@ export function shouldDisplayLapsedUserMsg(): boolean {
     refreshMsg
   );
 }
+
+export function getVimExtensionResolvedKeybinding() {
```
kevinslin, line 609: can we rename to make this a little clearer? something like `checkAndApplyVimKeybindingOverrideIfNotExist`(a mouthful but makes it really clear that we are doing something

packages/plugin-core/src/_extension.ts
```diff
@@ -425,6 +429,21 @@ export async function _activate(
     toggleViews(false);
   }
 
+  if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
+    const VimInstalled = VSCodeUtils.isExtensionInstalled("vscodevim.vim");
+    if (VimInstalled) {
+      AnalyticsUtils.track(ExtensionEvents.VimExtensionInstalledInitial);
+      const [keybindingConfigPath, resolvedKeybindings] = getVimExtensionResolvedKeybinding();
```
kevinslin, line 436: prefer object returns vs tuples. see [[prefer returning objects instead of tuples|dendron.dev.style#prefer-returning-objects-instead-of-tuples]]

packages/plugin-core/src/utils.ts
```diff
@@ -492,6 +492,47 @@ export class VSCodeUtils {
...
+        userConfigDir = process.env.APPDATA + "\\" + name + "\\User\\";
+        delimiter = "\\";
+        break;
+      }
+      default: {
+        userConfigDir = process.env.HOME + "/.config/" + name + "/User/";
+        break;
+      }
+    }
+    return [userConfigDir, delimiter, osName];
```
kevinslin, line 528: [[prefer returning objects instead of tuples|dendron.dev.style#prefer-returning-objects-instead-of-tuples]]

packages/plugin-core/src/test/suite-integ/Extension.test.ts
```diff
@@ -68,15 +71,38 @@ function lapsedMessageTest({
...
+    userConfigDir,
+    "keybindings.json"
+  ].join("");
 
   const ctx: ExtensionContext = setupBeforeAfter(this, {
     beforeHook: async () => {
       await resetCodeWorkspace();
       await new ResetConfigCommand().execute({ scope: "all" });
       homeDirStub = TestEngineUtils.mockHomeDir();
+      if (!fs.existsSync(keybindingConfigPath)) {
```
kevinslin, line 87: instead of overwriting system keybindings, lets stub this into `tmpDir`. see [[Stubbing Mutating Disk Changes|dendron.dev.qa#stubbing-mutating-disk-changes]]

packages/plugin-core/src/_extension.ts
```diff
@@ -425,6 +429,21 @@ export async function _activate(
...
+      if (!_.isUndefined(resolvedKeybindings)) {
+        if (!fs.existsSync(keybindingConfigPath)) {
+          fs.ensureFileSync(keybindingConfigPath);
+          fs.writeFileSync(keybindingConfigPath, "[]");
+        }
+        writeJSONWithComments(keybindingConfigPath, resolvedKeybindings);
```
kevinslin, line 442: can we also add another event here? eg.  track(VimExtensionInstalled, fixApplied: true)

packages/plugin-core/src/_extension.ts
```diff
@@ -425,6 +429,21 @@ export async function _activate(
     toggleViews(false);
   }
 
+  if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
+    const VimInstalled = VSCodeUtils.isExtensionInstalled("vscodevim.vim");
+    if (VimInstalled) {
+      AnalyticsUtils.track(ExtensionEvents.VimExtensionInstalledInitial);
```
kevinslin, line 435: we can change this to just `VimExtensionInstalled`

packages/plugin-core/src/_extension.ts
```diff
@@ -586,3 +605,37 @@ export function shouldDisplayLapsedUserMsg(): boolean {
...
   );
 }
+
+export function getVimExtensionResolvedKeybinding() {
+  // check where the keyboard shortcut is configured
+  const [userConfigDir, _delimiter, osName] = VSCodeUtils.getCodeUserConfigDir();
+  const keybindingConfigPath = [userConfigDir, "keybindings.json"].join("");
+
+  // read keybindings.json
+  const keybindings = readJSONWithCommentsSync(keybindingConfigPath);
```
kevinslin, line 615: can we do a check if the file exists? if not, dendron should create it

kevinslin (changes requested):

packages/plugin-core/src/_extension.ts
hikchoi: Will this be a separate count from the one just above? I don't see how much insight we can get from this unless we try to resolve keybinding conflicts on every `InstallStatus.UPGRADED` instead of `InstallStatus.INITIAL_INSTALL`.

kevinslin: it could be that the user installed the extension but something went wrong with the fix. 

now that I think about it, this might change this to an error metric instead (jonathan is currently working on error reporting) in a future change

kevinslin (approved):

PR #1106, opened 2021-08-06
feat: tag colors in parents cascade to children

SeriousBug:
Setting a tag color in a parent tag will cascade to children. For example, setting a color in `tags.foo` means tags `tags.foo.bar` and `tags.foo.bar.baz` will all use that color. Setting the color in a child tag will override the cascading color.

Importantly, this functionality works even when the tag notes are missing, for example if `tags.foo` exists and has a color set but `tags.foo.bar` and `tags.foo.bar.baz` doesn't, then `#foo.bar` and `#foo.bar.baz` will still get the color set in `tags.foo`.


packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -146,8 +146,12 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
         }
 
         let color: string | undefined;
-        if (note && value.startsWith(TAGS_HIERARCHY)) {
-          const [maybeColor, colorType] = NoteUtils.color({ note });
+        if (mode !== ProcMode.IMPORT && value.startsWith(TAGS_HIERARCHY)) {
+          const [maybeColor, colorType] = NoteUtils.color({
```
kevinslin, line 150: see [[prefer returning objects instead of tuples|dendron.dev.style#prefer-returning-objects-instead-of-tuples]]

kevinslin (approved):
looks good, just a nitpick about the return style. after you fix, feel free to rebase and merge

PR #1108, opened 2021-08-07
feat: provide YAML validator & suggest YAML extension

SeriousBug:
Provides a validator when editing `dendron.yml`. The validator only takes effect when `redhat.vscode-yaml` extension is installed, so the extension is now recommended for workspaces. The validator provides autocomplete, documentation & warnings when editing `dendron.yml`. Here's a few examples of what it looks like:

When merging this PR, please check that the validator works correctly in the packaged extension. I'm not sure if the validator schema is actually getting included in the extension itself. I can also check this myself if there are some instructions on how to create a local packaged version of the extension.


kevinslin (commented):
Nice timing - I just merged changes that automates building the plugin locally. See [[dendron.sop.weekly-release.build.v2#summary]]

You can follow the instructions to [[here|dendron.sop.weekly-release.build.v2#^eWBG9TEW8tnx]] at which point you should have a `vsix` package in your `plugin-core` folder

PR #1110, opened 2021-08-09
chore: add note lookupv3 default configs

hikchoi:
This PR:
- Adds new default configuration for NoteLookupCommand.
- Change `NoteLookupCommand` to respect default configuration.
- Fixes breaking tests that are affected by the new configuration.
- Adds migration (placeholder until release version is confirmed) for new configuration.

### DIFF
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -94,6 +94,20 @@ export enum NoteAddBehavior {
   "asOwnDomain" = "asOwnDomain",
 }
 
+export enum LookupSelectionType {
+  "selection2link" = "selection2link",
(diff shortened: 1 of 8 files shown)


kevinslin:
This looks good. Just a heads up that we should normally avoid leaving commented parts in the code and instead do a separate task for the migration. Can you add a backlog item to do the config migration?

kevinslin (approved):

hikchoi:
> This looks good. Just a heads up that we should normally avoid leaving commented parts in the code and instead do a separate task for the migration. Can you add a backlog item to do the config migration?

My bad. I'll add a backlog to clean it up / do the actual config migration.

PR #1111, opened 2021-08-09
fix: properly debounce picker update

hikchoi:
This PR:
- fixes an issue with the debounced picker value updating.
- see [[investigate debounce issue|scratch.2021.08.09.201548.investigate-debounce-issue]]


kevinslin (approved):
Thanks for investigating and great writeup.

Can you move it to the public record into `pkg.dendron-plugin.design.lookup-debounce` and leave it under [[Design Decisions|pkg.dendron-plugin.arch.note-lookup#design-decisions]] in our docs?

PR #1112, opened 2021-08-09
chore: change note lookup related keybindings to use v3

hikchoi:
This PR:
  - changes `cmd/ctrl + l`, `cmd/ctrl + shift + j`, `cmd/ctrl + shift + s` to use lookup v3 instead of lookup v2.
  - lookup v2 is still accessible, but now does not have a keybinding.

_This is a pre-emptive PR in case we want to transition to lookup v3 on this week's release, and is left as a draft for that reason._


hikchoi:
Closing this PR as the changes here will be included in #1141

PR #1114, opened 2021-08-09
chore: nextjs pr updates

kevinslin:


kevinslin:
Addressing @namjul comments from #1074 

> Would be great to have a how to test description

Test description here -> [[Build|pkg.nextjs-template.dev#build]]

> I assume the styles sections was copied from dendron-next-server. We slowly start to accumulate more tech dept on the styling side and I am a bit worried how we are going to "cleanup".

This is temporary. The plan is to merge all styles together into `common-frontend` package and use symlinks during development to make it work. also open to suggestions if you have any to amke this better

> the global eslint rules are currently not enabled for this new packages, equally for the rest apart from plugin-core

Good catch. Just removed custom .eslint

namjul:
> This is temporary. The plan is to merge all styles together into common-frontend package and use symlinks during development to make it work. also open to suggestions if you have any to amke this better

I would say let's go along with your plans until we really feel we need to make bigger adjustments. To have styles all in one spot `common-frontend` is definitely a good next move. 

My suggestions would rather go in the direction of building up the design-system package. In that than to introduce some primitive components (`Box`, `Text`, `Input`, etc.) with which everything else can be built upon on. Later then also start building pattern components (`Calendar`, `Tree`, etc.)

namjul (approved):

namjul:
> Test description here -> [[Build|pkg.nextjs-template.dev#build]]

They seem to be missing in the `dendron-site` seed.

PR #1119, opened 2021-08-10
feat: colored tags in tree view & tags at bottom

SeriousBug:
Added a special symbol next to tags. All tags under `tags.` will have their symbol colored using the auto-generated color or user set color. The `tags` entry is also always moved to the bottom of the tree to keep it at a consistent position.


hikchoi (approved):
Just tried it out in the test workspace. nitpick: the gap between the colored tag and the note title could be a little wider, but otherwise looks great. 🙂

kevinslin (approved):

PR #1123, opened 2021-08-10
fix: issue with direct child filter partially omitting values in quickpick

hikchoi:
This PR:
- Fixes an issue in Lookup (v2) where the direct child filter would be applied _after_ pagination has happened, which allowed some values to be omitted from the selectable items.
- Adds example notes that demonstrate the fix.

### DIFF
--- a/packages/plugin-core/src/components/lookup/LookupControllerV3.ts
+++ b/packages/plugin-core/src/components/lookup/LookupControllerV3.ts
@@ -50,6 +50,7 @@ export class LookupControllerV3 {
   protected _cancelTokenSource?: CancellationTokenSource;
   public _quickpick?: DendronQuickPickerV2;
   public fuzzThreshold: number;
+  public _provider?: ILookupProviderV3;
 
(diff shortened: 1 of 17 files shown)


kevinslin (changes requested):
we should also fix this issue in LookupProviderV3 since that will be the default soon. you'll need to do the check [here](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/components/lookup/utils.ts#L703:L703)

jonathanyeung:
lgtm for v2 changes

hikchoi:
This PR also:
- fixes a bug where triggering the direct child filter button would revert the picker value change to initial value
- changes picker refresh behavior to 
  - only refresh buttons that needs to be refreshed
    - this change broke the behavior of note type modifiers, which has also been fixed.
  - properly refresh picker after direct child filter is triggered

hikchoi:
fyi: #1135 is branched off of this PR so this one should go in first.

kevinslin (approved):

PR #1128, opened 2021-08-11
enhance/pod-type-in-event

jonathanyeung:
Enhance the pod analytics event to include whether the config has been set and what type of pod is being run.


kevinslin (approved):
looks good, don't forget to update the  [[Telemetry|dendron.ref.telemetry]] docs

PR #1130, opened 2021-08-11
fix: lookupv3 selection issue

kevinslin:


packages/plugin-core/src/components/lookup/LookupControllerV3.ts
```diff
@@ -175,6 +175,7 @@ export class LookupControllerV3 {
   }
 
   onHide() {
+    this._quickpick?.dispose();
```
hikchoi, line 178: ah, for the quickpick's own `onDidHide` I was disposing it before it hit this point. was the quickpick not getting disposed in other places where `lc.onHide()` was called causing the problem?

kevinslin: no, this was just something i added in case it wasn't disposed. the reason it was failing was because we're explicitly passing in `selectedItems` to `execute` so we need to use the passed in `selectedItems` instead of `quickpick.selectedItems`.

hikchoi (approved):

PR #1141, opened 2021-08-13
chore: transition lookup to v3

hikchoi:
This PR:
- Moves exports from v2 related files to respective v3 component modules
- Cleans up all stale imports that reference parts of lookup v2
- Refactor GoDownCommand to use lookup v3
- Refactor LookupJournal / LookupScratch tests to use lookup v3
- Deregister lookup v2 command and related keybindings

NOTE: Test suites related to `LookupCommand`, and the v2 lookup command components are still left but commented out. Once all the test suites are re-written to test respective parts of lookup v3, the components and tests that are commented out will be removed

### DIFF
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -159,14 +159,6 @@
         "desc": "Rename a header, and update all links to it."
       },
       {
-        "command": "dendron.lookup",
-        "title": "Dendron: Lookup",
-        "desc": "Initiate note lookup",
-        "docLink": "dendron.topic.lookup.md",
-        "docPreview": "",
-        "when": "dendron:pluginActive"
-      },
-      {
         "command": "dendron.lookupNote",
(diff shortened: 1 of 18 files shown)


hikchoi:
fyi: @jonathanyeung 
- I've moved the `VaultSelectionMode` that is used in `Go To Note` to `src/components/lookup/types`. 
- Lookup v3 already implements an explicit `VaultSelection` button that can be used toggle the vault selection behavior so I haven't changed lookup v3 to use `getOrPromptVaultForNewNote`. 
- I've checked that the go to note works after the transition
- Let me know if I missed anything with this

hikchoi:
update:
- Rebased after the quickpick subscriber changes

kevinslin (approved):
This looks good. Some things we should make sure before next weeks release:
- migrate over lookupv2 test case
- migrate keybindings for folks that are using `lookup` to `noteLookup`

PR #1142, opened 2021-08-13
feat: Insert Note Index command

hikchoi:
This PR:
- Adds a new command `InsertNoteIndexCommand` that will let users to insert a list of wikilinks to direct children of current note.
- Adds basic tests for this command.
- The core logic will be refactored to be exposed to the cli in the later phase to allow advanced use cases.
- This is phase 1 of [RFC 18 -  Add Note Indexes](https://wiki.dendron.so/notes/6TOh3VApIUfap7c3.html).
  - Discussion: https://github.com/dendronhq/dendron/discussions/1076
- PR for docs: https://github.com/dendronhq/dendron-site/pull/162


packages/plugin-core/src/commands/InsertNoteIndexCommand.ts
```diff
@@ -0,0 +1,78 @@
...
+  CommandOutput
+> {
+  key = DENDRON_COMMANDS.INSERT_NOTE_INDEX.key;
+
+  // TODO: make this into a util once the cli version is implemented.
+  // NOTE: the marker flag is not exposed to the plugin yet.
+  genNoteIndex(notes: NoteProps[], opts: {
+    marker?: boolean
+  }) {
+    const listItems = notes.map((note) => `- [[${note.title}\|${note.fname}]]`)
```
kevinslin, line 29: for the index, what about the case with the same note in multiple vaults?
also, should [noXvaultWikiLink](https://wiki.dendron.so/notes/f83c1d87-eac0-48f3-a5cf-8a69989d8ec1.html#noxvaultwikilink) impact how links are generated?

kevinslin (changes requested):

packages/plugin-core/src/commands/InsertNoteIndexCommand.ts
```diff
@@ -0,0 +1,78 @@
...
+export class InsertNoteIndexCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput
+> {
+  key = DENDRON_COMMANDS.INSERT_NOTE_INDEX.key;
+
+  // TODO: make this into a util once the cli version is implemented.
+  // NOTE: the marker flag is not exposed to the plugin yet.
+  genNoteIndex(notes: NoteProps[], opts: {
+    marker?: boolean
```
jonathanyeung, line 27: Should we make this default to true?

jonathanyeung (commented):
lgtm

packages/plugin-core/src/commands/InsertNoteIndexCommand.ts
hikchoi: My reasoning to default it to be false is because if the user is triggering the command themselves in a note, they are adding the index block as the _content_ of the note. This means the plugin should not look at the note and try to update the index so I felt the markers would be irrelevant here.

I'm of the opinion that the markers should only be in the notes that we are going to implement as best practice, and for truly autogenerated ones like the ones that would be added in bulk via export pod or cli. If the user wants to have the markers added when using this command themselves, it should be passed in as an argument instead of it being the default.

Does adding a new flag in `dendron.yml` sound like a good middle ground?

packages/plugin-core/src/commands/InsertNoteIndexCommand.ts
hikchoi: You're right. I changed it so that it uses `createWikiLink` so that it can add the vault prefix as necessary.

hikchoi:
Updates:
- Handled multi-vault cases.
- Added new config to enable markers.
- Added tests for said updates.
- Revised docs PR to mention these updates.

kevinslin (approved):

PR #1151, opened 2021-08-16
fix: race conditions in note lookup

kevinslin:
- user creates accept before picker items are finished updating. in this case, we fetch results and create a new note   
- user creates quickpick for the first time, currently results in two updates which causes quickpick list to re-order

### DIFF
--- a/packages/common-all/src/fuse.ts
+++ b/packages/common-all/src/fuse.ts
@@ -10,8 +10,8 @@ import {
   SchemaModuleProps,
   NoteUtils,
   DNodeUtils,
+  DEngineClient,
 } from ".";
-import { DEngineClient } from "../lib";
 import { DendronConfig, DVault } from "./types/workspace";
 
 export type NoteIndexProps = {
--- a/packages/common-all/src/fuse.ts
+++ b/packages/common-all/src/fuse.ts
@@ -9,7 +9,9 @@ import {
   NotePropsDict,
   SchemaModuleProps,
   NoteUtils,
+  DNodeUtils,
 } from ".";
+import { DEngineClient } from "../lib";
 import { DendronConfig, DVault } from "./types/workspace";
 
 export type NoteIndexProps = {
@@ -140,6 +142,7 @@ export class FuseEngine {
   }
 }
 
+const PAGINATE_LIMIT = 50;
 export class NoteLookupUtils {
   /**
    * Get qs for current level of the hierarchy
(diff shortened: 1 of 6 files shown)


packages/plugin-core/src/components/lookup/LookupControllerV3.ts
```diff
@@ -141,27 +144,33 @@ export class LookupControllerV3 {
...
+    Logger.info({ ctx, msg: "onUpdatePickerItems:pre" });
     // initial call of update
     await provider.onUpdatePickerItems({
       picker: quickpick,
       token: cancelToken.token,
       fuzzThreshold: this.fuzzThreshold,
     });
+    Logger.info({ ctx, msg: "onUpdatePickerItems:post" });
     if (!nonInteractive) {
+      provider.provide(this);
```
jonathanyeung, line 167: Does moving this line add any perceivable delay between typing "Ctrl+L" and focus moving to the input box?

packages/plugin-core/src/components/lookup/utils.ts
```diff
@@ -707,6 +712,34 @@ export class NotePickerUtils {
     });
   };
 
+  /**
+   * Get picker results without input from the user
```
jonathanyeung, line 716: nit: can we update this description? It's a little confusing - b/c there is actually "input" from the user, but it's just that b/c of limitations with how the quick pick control works, the callback didn't provide adequate info (If I'm understanding the issue correctly)

jonathanyeung (commented):
Great to see these issues fixed!

packages/plugin-core/src/components/lookup/LookupControllerV3.ts
hikchoi: I just did a quick profile and this line takes 0ms in our workspace 😄  `provider.provide` is mostly just binding listener methods so I imagine there won't be any perceivable delay.

packages/plugin-core/src/components/lookup/utils.ts
hikchoi: How about

```suggestion
   * Get picker results with current picker value
```
and also rename the util to `fetchPickerResultsWithCurrentValue`?

packages/common-all/src/fuse.ts
```diff
@@ -165,4 +168,40 @@ export class NoteLookupUtils {
     const nodes = _.map(childrenOfRoot, (ent) => notes[ent]).concat(roots);
     return nodes;
   };
+
+  static async lookup({
```
hikchoi, line 172: can we call this something more descriptive?
I'm thinking something like `fetchNodesFromEngine`.

The regular `fetchPickerResults` should probably also be refactored to use this util in the future

hikchoi (commented):
Looks good other than the small nitpicks.

packages/common-all/src/fuse.ts
kevinslin: wouldn't use `fetchNodesFromEngine` because `queryNotes` does that already. this method wraps `fetchRootResults`, `queryNotes`, and takes into account the direct stub filter. this is why I ended up using `lookup` - in the future, we can imagine the web client and other clients of Dendron using this directly to do `lookup`

jonathanyeung (approved):

PR #1157, opened 2021-08-17
fix: don't insert title when rendering note refs in preview

hikchoi:
This PR:
- Fixes issue with note ref rendering in preview where the note ref's title was unnecessarily inserted.


kevinslin (approved):
can we also have a non-snapshot based test

PR #1159, opened 2021-08-17
fix: cleanup subscribers when quickpick is hidden

kevinslin:
Currently, hiding the quickpick using `ESC` key does not clean up the `lookup` subscriber listening to quickpick events. this leads to multiple `lookup` events executing if the quickpick has been hidden using `ESC` key

### DIFF
--- a/packages/plugin-core/src/commands/NoteLookupCommand.ts
+++ b/packages/plugin-core/src/commands/NoteLookupCommand.ts
@@ -211,76 +211,76 @@ export class NoteLookupCommand extends BaseCommand<
     opts: CommandGatherOutput
   ): Promise<CommandOpts | undefined> {
     const ctx = "NoteLookupCommand:enrichInputs";
-    return new Promise((resolve) => {
-      const start = process.hrtime();
-      HistoryService.instance().subscribev2("lookupProvider", {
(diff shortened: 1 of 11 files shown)


packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -206,69 +210,100 @@ export class NoteLookupCommand extends BaseCommand<
...
+    const promise = new Promise<CommandOpts | undefined>((resolve) => {
+      promiseResolve = resolve as typeof promiseResolve;
       opts.controller.showQuickPick({
         provider: opts.provider,
         quickpick: opts.quickpick,
         nonInteractive: opts.noConfirm,
         fuzzThreshold: opts.fuzzThreshold,
       });
-      const profile = getDurationMilliseconds(start);
-      AnalyticsUtils.track(VSCodeEvents.NoteLookup_Show, {
```
hikchoi, line 248: Did you intend to remove this?

hikchoi (approved):
Looks good other than the removed `NoteLookup_Show` analytics

kevinslin:
yeah, we’re currently not using this metric so decided to remove it

Issue #1107, opened 2021-08-06
Extension doesn't start, Code Helper (Renderer) 99%+ cpu


christiand:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x ] Misc (Choose this if your not sure)

## Describe the bug

This process "Code Helper (Renderer)" has my CPU pegged at 99%+ when Dendron workspace is open in VSCode. 
I've upgraded VSCode and Dendron to the latest.
If I close the dendron workspace, and open my own coding project workspace - all is fine.

Extension doesn't seem to fully load, and hangs at the progress indicator (see screenshot).
Even if cancelling the extension load, that process still pegs one of my cores.
Dendron shortcuts don't work.

## To Reproduce

1. Open my Dendron workspace in VSCode
2. Do nothing (idle)

## Expected behavior

Extension starts.
One core not maxed out.

## Screenshots
Hangs here:
![image](https://user-images.githubusercontent.com/22415/128579709-8c7d9990-b666-4395-9d26-1312cff16dc3.png)

![image](https://user-images.githubusercontent.com/22415/128577187-04269eb2-c577-4bb1-8100-a179dead800e.png)

## Desktop (please complete the following information)

- OS: Mac OS 10.15.7 (19H1323)
- VSCode:
Version: 1.59.0
Commit: 379476f0e13988d90fab105c5c19e7abc8b1dea8
Date: 2021-08-04T23:14:40.191Z (1 day ago)
Electron: 13.1.7
Chrome: 91.0.4472.124
Node.js: 14.16.0
V8: 9.1.269.36-electron.0
OS: Darwin x64 19.6.0

- Dendron v0.53.0
- Dendron Markdown Preview v0.10.57
- Dendron Snippet Maker v0.1.6

## Dendron Log file

{"level":30,"time":1628291832488,"pid":26243,"hostname":"3c22fb0e9ff2.ant.amazon.com","name":"dendron","ctx":"Logger:configure","msg":"exit","logLevel":"info"}
[...]

PR #1160, opened 2021-08-17
enhance: avoid parsing links & anchors on very long notes

SeriousBug:
When a note is too large (over 1MiB), Dendron will avoid generating backlinks and parsing anchors. This means:

1. Any outgoing links from the large note are ignored, and will not appear in backlinks
2. Any links into the large note that target an anchor will instead go to the start of the note

Other features will continue to function, the large note can be accessed with "Lookup Note", links to the large note will function, and all other notes are unaffected. The user will be warned during startup about the large note.

### DIFF
--- a/packages/common-all/src/constants.ts
+++ b/packages/common-all/src/constants.ts
@@ -29,6 +29,10 @@ export const CONSTANTS = {
    * Initial version for first installaion
    */
   DENDRON_INIT_VERSION: "0.0.0",
+  /** Files larger than this size (in characters) don't get parsed for links,
+   * link candidates, and anchors. Currently about 1MiB, although may be larger
+   * due to unicode since it's counted in characters. */
+  DENDRON_FILE_TOO_BIG: 2 ** 20,
 };
 
 export enum ERROR_STATUS {
--- a/packages/common-all/src/constants.ts
+++ b/packages/common-all/src/constants.ts
@@ -29,10 +29,8 @@ export const CONSTANTS = {
    * Initial version for first installaion
    */
   DENDRON_INIT_VERSION: "0.0.0",
-  /** Files larger than this size (in characters) don't get parsed for links,
-   * link candidates, and anchors. Currently about 1MiB, although may be larger
-   * due to unicode since it's counted in characters. */
-  DENDRON_FILE_TOO_BIG: 2 ** 20,
+  /** Default for the `maxNoteLength` config. */
+  DENDRON_DEFAULT_MAX_NOTE_LENGTH: 204800,
 };
 
 export enum ERROR_STATUS {
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -1,6 +1,7 @@
 import {
   assert,
   BulkAddNoteOpts,
+  CONSTANTS,
   DendronCompositeError,
   DendronConfig,
   DendronError,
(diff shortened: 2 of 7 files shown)


packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -447,6 +450,25 @@ export class FileStorage implements DStore {
...
+            length: n.body.length,
+          });
+          errors.push(
+            new DendronError({
+              message: `Note "${n.fname}" in vault "${VaultUtils.getName(
+                n.vault
+              )}" is too large, some features like backlinks may not work correctly for it.`,
+              severity: ERROR_SEVERITY.MINOR,
+            })
+          );
```
hikchoi, line 467: nitpick: Should this be an error? We are explicitly limiting part of the feature and otherwise the plugin is working as intended so I think a warning might be sufficient. With an error the user may think they did something wrong and it needs to be fixed.

Also, IIRC a lot of the people who encounter issues with big files were curious what _too big_ meant. It may be a nice touch to add a button in the toast notification that links to a doc that tells you what the limit is.

packages/common-all/src/constants.ts
```diff
@@ -29,6 +29,10 @@ export const CONSTANTS = {
    * Initial version for first installaion
    */
   DENDRON_INIT_VERSION: "0.0.0",
+  /** Files larger than this size (in characters) don't get parsed for links,
+   * link candidates, and anchors. Currently about 1MiB, although may be larger
+   * due to unicode since it's counted in characters. */
+  DENDRON_FILE_TOO_BIG: 2 ** 20,
```
hikchoi, line 35: Can we reliably say that 1MiB is too big for any environment? I think some people may want to change this number depending on how beefy their machines are.

something like this in `dendron.yml`:

```yaml
  ...
  file:
    maximumSizeInChars: 2 ** 20
  ...
```

packages/engine-server/src/drivers/file/storev2.ts
SeriousBug: The error is to pass the warning from the engine to the editor and have it be displayed. I didn't want to make a bigger change to pass warnings seperately, but we can do it if there's a benefit.

Agree with communicating how big is too big. Maybe I can add that to the warning message?

hikchoi (approved):
Left some questions but they are minor concerns and everything else looks good to me 👍🏼

packages/engine-server/src/drivers/file/storev2.ts
hikchoi: We do have the severity argument so maybe we can do something where the toast notification is actually made depending on the severity that gets passed, but this could change the behavior of the notification in other places so it probably should be more thought out before we do this 😅  Did we have an SOP that goes over error message conventions or something?

> Agree with communicating how big is too big. Maybe I can add that to the warning message?
Sounds good

kevinslin (changes requested):
This looks good. Two changes:

If you look at this existing [issue](https://github.com/dendronhq/dendron/issues/1107), our user had a problem with a note that was 200kb in size. Can we start the cutoff here and also make it configurable in `dendron.yml` as per @hikchoi suggestion?

SeriousBug:
Made the maximum length configurable, and added a test for it.

Also improved composite errors:
* Composite errors have fatal severity if any of their errors is fatal.
* Composite errors have minor severity if all of their errors are minor.
* Minor composite errors say "Multiple warnings: ..."
* Composite errors combine together the errors messages of all their errors.

This allows the warnings the maximum file size warnings to work even if there are multiple files that are too long.

kevinslin (approved):

PR #1164, opened 2021-08-18
feat: make breadcrumbs clickable

hikchoi:
This PR:
- Makes breadcrumbs in nextjs publishing clickable.

### DIFF
--- a/packages/nextjs-template/components/DendronBreadCrumb.tsx
+++ b/packages/nextjs-template/components/DendronBreadCrumb.tsx
@@ -1,8 +1,9 @@
 import { Breadcrumb } from "antd";
 import _ from "lodash";
 import React from "react";
-import { NoteUtils } from "@dendronhq/common-all";
+import { NoteUtils, NoteProps } from "@dendronhq/common-all";
 import { useNoteActive } from "../utils/hooks";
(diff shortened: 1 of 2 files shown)


packages/nextjs-template/components/DendronBreadCrumb.tsx
```diff
@@ -27,7 +28,14 @@ export function DendronBreadCrumb(props: DendronCommonProps) {
   return (
     <Breadcrumb style={{ margin: "16px 0" }}>
       {_.map(noteParents, (note) => {
-        return <Breadcrumb.Item key={note.id}>{note.title}</Breadcrumb.Item>;
+        const dest = note.id === props.noteIndex.id ? "/" : `/notes/${note.id}`;
```
kevinslin, line 31: can we wrap this in a utility function? `eg. getNoteUrl({note, noteIndex)`

kevinslin (changes requested):

packages/nextjs-template/components/DendronBreadCrumb.tsx
hikchoi: done!

kevinslin (approved):

PR #1166, opened 2021-08-18
feat: seed browser initial revision

jonathanyeung:
Initial Revision of Seed Browser. 

Haven't implemented tests yet, so still in draft.


packages/dendron-next-server/pages/vscode/seed-browser.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  return (
+    <>
+      <Layout className={seedStyles.layout}>
+        <Content style={{ padding: "0 50px" }}>
+          <div className={seedStyles.contentDiv}>
+            <PageHeader
+              className={seedStyles.siteHeader}
+              title="Dendron Seed Registry"
+              subTitle="Add Knowledge Bases to your Workspace"
```
kevinslin, line 61: change the subtitle here to something shorter/more catchy. eg. "A Better way to Know"

packages/plugin-core/src/commands/SeedCommandBase.ts
```diff
@@ -38,4 +45,29 @@ export abstract class SeedCommandBase<
...
+      data: {
+        msg: this.getSeedSvc().getSeedsInWorkspace(),
+      },
+      source: "vscode",
+    });
+  }
+
+  protected async onUpdatingWorkspace() {
+    const ws = getWS();
+    await ws.updateGlobalState(
```
kevinslin, line 64: for working with vscode state, can we wrap it using the state service? https://wiki.dendron.so/notes/3VvuxP7T9en9EzChUCY9D.html

packages/plugin-core/src/test/suite-integ/SeedCommand.test.ts
```diff
@@ -1,12 +1,31 @@
...
 import { SeedAddCommand } from "../../commands/SeedAddCommand";
 import { SeedRemoveCommand } from "../../commands/SeedRemoveCommand";
 import { DENDRON_COMMANDS } from "../../constants";
 import { expect } from "../testUtilsv2";
 import { runLegacyMultiWorkspaceTest, setupBeforeAfter } from "../testUtilsV3";
 
+function getFakedAddCommand(svc: SeedService) {
+  const cmd = new SeedAddCommand(svc);
+
+  sinon.replace(cmd, <any>"onUpdatingWorkspace", sinon.fake.resolves(null));
```
kevinslin, line 14: can we return the fakes and then check that they are called? `expect( $fake.calledOnce ).toBeTruthy()`

packages/plugin-core/src/test/suite-integ/SeedCommand.test.ts
```diff
@@ -1,12 +1,31 @@
...
+  sinon.replace(cmd, <any>"onUpdatedWorkspace", sinon.fake.resolves(null));
+
+  return cmd;
+}
+
+function getFakedRemoveCommand(svc: SeedService) {
+  const cmd = new SeedRemoveCommand(svc);
+
+  sinon.replace(cmd, <any>"onUpdatingWorkspace", sinon.fake.resolves(null));
```
kevinslin, line 23: see above comment

packages/plugin-core/src/workspace/seedBrowserInitializer.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Launch Seed Browser Webview
+   * @param opts
+   */
+  async onWorkspaceOpen(opts: { ws: DendronWorkspace }): Promise<void> {
+    const cmd = new SeedBrowseCommand();
+    await cmd.execute();
+
+    await opts.ws.updateGlobalState(
```
kevinslin, line 39: use `StateService` to do this

packages/plugin-core/src/workspace/workspaceInitializer.ts
```diff
@@ -49,4 +52,13 @@ export class WorkspaceInitFactory {
     );
     return state === WORKSPACE_ACTIVATION_CONTEXT.TUTORIAL.toString();
   }
+
+  static isSeedBrowserWorkspaceLaunch(
+    context: vscode.ExtensionContext
+  ): boolean {
+    const state = context.globalState.get<string | undefined>(
```
kevinslin, line 59: use `StateService`

packages/dendron-next-server/pages/vscode/seed-browser.tsx
```diff
@@ -0,0 +1,137 @@
...
+                      />
+                    </Card>
+                  </List.Item>
+                )}
+              />
+            </div>
+          </div>
+        </Content>
+        <Footer>
+          <h4>Current theme: {currentTheme}</h4>
```
kevinslin, line 115: can we remove this? since we are going to support custom theming, this would introduce extra complexity in terms of keeping track of per view themes vs global themes

kevinslin (changes requested):

packages/dendron-next-server/pages/vscode/seed-browser.tsx
jonathanyeung: Yes, good catch - this was meant to be removed.  There are a few more UI elements that I need to clean up.

packages/dendron-next-server/pages/vscode/seed-browser.tsx
jonathanyeung: I think that might be good for a browser version of this view, but for the VS Code version I think something a little more informative of what the seeds are for would be better for users? My intention here was to give a hint as to what clicking on 'Add Seed' button would do.

packages/plugin-core/src/commands/SeedCommandBase.ts
jonathanyeung: Ok - did this and then got carried away and ended up refactoring some stuff :)

kevinslin:
let me know when you think this is ready to review again. also reminder to resolve conversations as they get addressed

jonathanyeung:
> let me know when you think this is ready to review again. also reminder to resolve conversations as they get addressed

Ready for review again - ty.  Typically I let the commenter resolve the comment thread as a way to 'sign-off' on the thread, but I'm fine with either pattern.

kevinslin (approved):

kevinslin:
cool, this looks good. can you rebase from master, resolve conflicts and merge?

PR #1167, opened 2021-08-19
enhance: better error message when renaming note without id

SeriousBug:
This was an [error reported on Twitter](https://twitter.com/wisdomcitynews/status/1420200694921924615), I'm not sure if there is a corresponding issue.

I believe I reproduced the issue: it occurs if some of the renamed notes don't have a frontmatter. I improved the error message to be more informative, and to also tip the user what the issue may be.

### DIFF
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -156,12 +156,18 @@ export class FileStorage implements DStore {
       });
     }
     const noteToDelete = this.notes[id];
+    this.logger.info({ ctx, noteToDelete, opts, id });
+    if (_.isUndefined(noteToDelete))
+      throw new DendronError({
+        message: `Unable to find node ${id}`,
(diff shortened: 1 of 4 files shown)


SeriousBug:
To reproduce the issue:

1. Without Dendron running, create files "tag.foo.md" and "tag.bar.md" in a vault. Make sure these files are empty.
2. Start Dendron.
3. Use the "Refactor Hierarchy" command to refactor from `^tag[.]` to `tags.`.

jonathanyeung:
Hmm - seems like maybe a different repro than what the user reported? At least he indicated that all the notes have frontmatter.

SeriousBug:
> Hmm - seems like maybe a different repro than what the user reported? At least he indicated that all the notes have frontmatter.

Ah sorry, must have missed that. Feel free to merge this anyway, and I'll continue looking into this issue.

kevinslin (changes requested):
can we add a test for this?

kevinslin (approved):

PR #1173, opened 2021-08-20
chore: expose copyNoteLink argument for NoteLookupCommand

hikchoi:
This PR:
- Adds copyNoteLink argument for NoteLookupCommand and makes it possible to pass it through custom keybindings
- Fixes copyNoteLink button effect to trigger after onAccept so that it works for new note creation as well 
- Adds tests for new behaviors

### DIFF
--- a/packages/plugin-core/src/commands/NoteLookupCommand.ts
+++ b/packages/plugin-core/src/commands/NoteLookupCommand.ts
@@ -32,6 +32,8 @@ import {
 import {
   DendronQuickPickerV2,
   DendronQuickPickState,
+  LookupEffectType,
+  LookupEffectTypeEnum,
   LookupFilterType,
   LookupNoteType,
   LookupNoteTypeEnum,
@@ -60,6 +62,7 @@ export type CommandRunOpts = {
   noteType?: LookupNoteType;
   selectionType?: LookupSelectionType;
   splitType?: LookupSplitType;
+  effectType?: LookupEffectType;
   /**
    * NOTE: currently, only one filter is supported
    */
@@ -156,7 +159,7 @@ export class NoteLookupCommand extends BaseCommand<
       ),
       extraButtons: [
         MultiSelectBtn.create(copts.multiSelect),
-        CopyNoteLinkBtn.create(),
+        CopyNoteLinkBtn.create(copts.effectType === LookupEffectTypeEnum.copyNoteLink),
         DirectChildFilterBtn.create(
           copts.filterMiddleware?.includes("directChildOnly")
         ),
@@ -309,6 +312,9 @@ export class NoteLookupCommand extends BaseCommand<
       const outClean = out.filter(
         (ent) => !_.isUndefined(ent)
       ) as OnDidAcceptReturn[];
+      if (!_.isUndefined(quickpick.copyNoteLinkFunc)) {
+        await quickpick.copyNoteLinkFunc!(outClean.map((item) => item.node));
+      }
       await _.reduce(
         outClean,
         async (acc, item) => {
--- a/packages/plugin-core/src/commands/NoteLookupCommand.ts
+++ b/packages/plugin-core/src/commands/NoteLookupCommand.ts
@@ -32,8 +32,6 @@ import {
 import {
   DendronQuickPickerV2,
   DendronQuickPickState,
-  LookupEffectType,
-  LookupEffectTypeEnum,
   LookupFilterType,
   LookupNoteType,
   LookupNoteTypeEnum,
@@ -59,10 +57,10 @@ export type CommandRunOpts = {
   noConfirm?: boolean;
   fuzzThreshold?: number;
   multiSelect?: boolean;
+  copyNoteLink?: boolean;
   noteType?: LookupNoteType;
   selectionType?: LookupSelectionType;
   splitType?: LookupSplitType;
-  effectType?: LookupEffectType;
   /**
    * NOTE: currently, only one filter is supported
    */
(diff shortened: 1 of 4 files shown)


kevinslin (changes requested):
currently, the copy note link command will copy the link and then toggle itself back off. it looks like the new copy note link requires that you click ok to do this instead. this is okay if we wanted to change the behavior to make it more in line with other modifier buttons.

but it looks like there's different behavior when pressing "ok" vs hitting "enter" - pressing ok will copy the link or links whereas hitting "enter" will only copy the first link (if multiple are selected) as well as open the note

would recommend copy note link not open any note but just copy links and make that behavior consistent for "ok" and "enter"

hikchoi:
> currently, the copy note link command will copy the link and then toggle itself back off. 

As-is, it doesn't toggle it back off. the button stays _on_ , and the user proceeds with the remaining lookup steps.
This is confusing for a few reasons:
1. it's a togglable _modifier_, but it currently invokes immediately and acts as a proxy to an already existing command (`Dendron Copy Note Link`)
2. as-is, it will only work for notes that already exist
3. the button being toggled on gives the impression that once the lookup has completed, the links of the resulting selection of notes will be copied to the clipboard.
    - this is not true. what if I selected two notes, pressed the copy note link button, then selected two more notes? you will have to press the button again the copy the remaining selected notes. we need to defer the copy note link behavior to after lookup to meet this expectation.
    - it is also not true outside the multi-select context. nothing stops the users to toggle the button when their intention is to lookup and create a new note. here, the copy note link button will not work because it only works for existing notes. here we also need to defer the copy note link behavior to happen after lookup to meet the expectation.
 
> but it looks like there's different behavior when pressing "ok" vs hitting "enter" - pressing ok will copy the link or links whereas hitting "enter" will only copy the first link (if multiple are selected) as well as open the note

I can't repro this. I just tried with both the ok button and enter and both will copy all the selected note's links and open all the selected notes. perhaps you had the note already open and/or have `workbench.editor.enablePreview` set to true so subsequently opened notes didn't show up in the tabs ()?

> would recommend copy note link not open any note but just copy links and make that behavior consistent for "ok" and "enter"

My reasoning here is that if the user only wants the copy note link function, they would use `Dendron: Copy Note Link`.
The copy note link here is, in my opinion a _modifier_ to the _lookup_ behavior. That is, it will copy the note links of the lookup result. 

Is your opinion that users should open up lookup, then select the notes they want to copy, toggle copy note link, and just exit the lookup? 
In that case, I think the `Dendron: Copy Note Link` command should be enhanced to use a lookup component and let the users multi select what note links they want to copy, instead of fitting the main lookup's modifier to accomodate for that use case. Using the main lookup's feature only to multi select the notes you want the links to copy and exiting doesn't sound too straightforward to me personally. I think it's important to separate the concerns / intentions here.

To reiterate: 

main lookup's copy note link modifier
- let the users do a lookup, and as a side effect (hence, the _modifier_) have the link of the selected item(s) copied to the clipboard
- as a user, you want the notes to open and the links of the notes copied as well.

copy note link command
- let the users do a lookup of the note(s) that they want to have links copied to the clipboard.
- as a user, you don't want the notes to open, but want to lookup the notes you want links to be copied.

For the sake of convenience, if we decide to do this, we should have `Dendron: Copy Current Note Link` behave the same as the current `Dendron: Copy Note Link` (just copies currently opened note's link), and add a new command that you we extend so that we prompt users to select what they want to copy.

How does this sound? @kevinslin 
If you think the use case you described should be preserved, I suggest that we not make the button toggleable and re-align the expectations for this behavior. 
That is: we should make sure that the docs and description / label of this button explicitly say that **_this button will copy the currently selected notes' links and _only_ them into your clipbaord. Links of any subsequent selection of notes will not be copied as this button will invoke immediately. It also doesn't work for newly looked up notes. Use `Dendron: Copy Note Link` after note creation instead in that case_**

hikchoi:
I realize that this difference in the opinion may stem from the way we perceive what _lookup_ is.
For me _lookup_ is an overarching concept, and the _Note Lookup Command_ is an instance that implements the concept in the context of opening and creating notes. With this understanding, my intentions to how the copy note link modifier should behave makes sense.

I am just guessing here, but I think you are also considering _lookup_ as a bigger concept, but considering the `Note Lookup Command` to be the entry point of everything that is _lookup_. So in the use case of wanting to lookup multiple notes, then just copy the links of their note without opening them also fits into the usage pattern of this command.

Am I understanding this correctly?
If my understanding of your angle is correct, I can see why you want to keep the as-is behavior, as it fits into the line of thought. 

My concern is simply user experience: currently the implementation of the copy note link behavior gives a wrong affordance to what it does and may confuse users. The changes I made (and the suggestion to enhance the copy note link command) addresses that by separating the concerns of looking up notes and copying note links.

I am suspecting the overhaul of lookup is the facilitate applying the larger concept of lookup to various places by modularizing the core lookup components and using bits of them to implement isolated instances of the lookup concept in various places (in this case, Copy Note Link), so I am inclined to think that my suggestions and changes are going towards the right way.

Correct me if my line of thought is wrong.

kevinslin:
I am with your line of thought. It is a more coherent UX - the current implementation was just me getting out a v1 :)

Can you make sure to rebase this from master since we had a bunch of new changes from the early seed and I can check if the issue about `ok/enter` still repros? Otherwise, this looks good and we should just make a note of the behavior change in the release notes

hikchoi:
@kevinslin Rebased and confirmed that ok/enter behavior is still consistent.

kevinslin:
doesn't look like copying multiple links is working: https://www.loom.com/share/72d86572cda24c06966994bedb97580c?from_recorder=1&focus_title=1

kevinslin (approved):

PR #1175, opened 2021-08-20
chore: add config and keybinding migration for lookup v3

hikchoi:
This PR:
- Adds migration for new lookup config namespace.
- Adds the ability to update v2 keybindings to v3 keybindings on extension activation.
- Refactors keybinding related utilities
- Adds tests for migrations

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/migration.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/migration.test.ts
@@ -1,4 +1,4 @@
-import { InstallStatus } from "@dendronhq/common-all";
+import { InstallStatus, LookupSelectionType } from "@dendronhq/common-all";
 import {
   ALL_MIGRATIONS,
   DConfig,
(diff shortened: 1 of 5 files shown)


hikchoi:
note: keybinding upgrade will respect new additions to the arguments for `NoteLookupCommand` introduced in #1173 and #1174, so if the upgrade code needs to be fixed if there are changes to those two PRs.

packages/plugin-core/src/test/suite-integ/migration.test.ts
```diff
@@ -252,5 +252,72 @@ suite("Migration", function () {
         },
       });
     });
+
```
kevinslin, line 255: can we add a test for migrating keyboard shortcuts? reminder to mock the keyboard shortcuts dir/file (like how we currently mock the home directory) so the tests don't override our actual shortcuts

kevinslin (approved):

PR #1184, opened 2021-08-20
fix: pass fname to new note pickker item

kevinslin:


kevinslin:
Found an issue where we weren't actually passing the fname to the current vault. Made a fix for it here -> https://github.com/dendronhq/dendron/pull/1184
Feel free to merge into your branch after review

hikchoi:
Nice catch. thanks for the fix 🙏🏼

hikchoi (approved):

PR #1191, opened 2021-08-23
fix: preserve keybinding comments and back up old file

hikchoi:
This PR:
- Fixes issue with lookup v3 keybinding migration so that it preserves comments and also backs up old config
- Also makes vim keybinding resolution preserve comments.
- Adds tests for this behavior

I have spent some time trying to preserve every single comment in the config, but specifically comments that users write in the `args` block of the keybinding is non-trivial to preserve, and is probably not worth the effort. I have opted for backing up the original keybinding and also showing an information message with a button to open both new and old configs instead.

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -453,6 +453,18 @@ export async function _activate(
   if (!_.isUndefined(migratedKeybindings)) {
     fs.copyFileSync(keybindingConfigPath, `${keybindingConfigPath}.old`);
     writeJSONWithComments(keybindingConfigPath, migratedKeybindings);
+    vscode.window.showInformationMessage("Keybindings for lookup has been updated.", ...["Open keybindings.json"]).then((selection) => {
+      if (selection) {
+        const uri = vscode.Uri.file(keybindingConfigPath);
+        VSCodeUtils.openFileInEditor(uri);
+      }
+    })
+    vscode.window.showInformationMessage("Old config has been backed up.", ...["Open backup"]).then((selection) => {
+      if (selection) {
+        const uri = vscode.Uri.file(`${keybindingConfigPath}.old`);
+        VSCodeUtils.openFileInEditor(uri);
+      }
+    });
   }
 
   return showWelcomeOrWhatsNew({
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -451,8 +451,20 @@ export async function _activate(
 
   const { keybindingConfigPath, migratedKeybindings } = KeybindingUtils.checkAndMigrateLookupKeybindingIfExists();
   if (!_.isUndefined(migratedKeybindings)) {
-    fs.ensureFileSync(keybindingConfigPath);
-    fs.writeFileSync(keybindingConfigPath, JSON.stringify(migratedKeybindings));
+    fs.copyFileSync(keybindingConfigPath, `${keybindingConfigPath}.old`);
+    writeJSONWithComments(keybindingConfigPath, migratedKeybindings);
+    vscode.window.showInformationMessage("Keybindings for lookup has been updated.", ...["Open keybindings.json"]).then((selection) => {
```
kevinslin, line 456: notice sometimes vscode will not show two information boxes back to back. can we combine this into one message? `Keybindings for lookip has been updated. ["Open keybindings.json", "Open "backup"]

I think you can also adjust the timeout on the dialogue. Lets have it not fade unless the user clears it

packages/plugin-core/src/_extension.ts
```diff
@@ -451,8 +451,20 @@ export async function _activate(
 
   const { keybindingConfigPath, migratedKeybindings } = KeybindingUtils.checkAndMigrateLookupKeybindingIfExists();
```
kevinslin, line 452: currently we run this at startup everytime. can we put this logic into the migration so it only runs during upgrade?

kevinslin (changes requested):

packages/plugin-core/src/_extension.ts
hikchoi: Looks like the message won't time out until an action is selected (a button is clicked), but can't make it wait for both buttons to be clicked.

I'll consolidate the two buttons into one so that both files are open in a horizontal split (since you want to look at both of them at the same time anyways)

packages/plugin-core/src/_extension.ts
hikchoi: keybinding migration / conflict resolution is a pretty plugin-y thing to do, do we want the engine-server to run that? We'll have to import vscode and some plugin-core utils to engine-server if we want this as a migration.

I think it would be fine to keep it here, but add another check for `InstallStatus.UPGRADED`. So it only checks during upgrade.

PR #1193, opened 2021-08-23
Nextjs template style improvements

namjul:


kevinslin (changes requested):
great work on this and like the change to menu items. comments below:

when in small viewports:
  - menu in collapsed state
    - menu bar will sometimes get stuck and not collapse again:
      - https://www.loom.com/share/03f4b0ac2024406da4865c094f7d5aca?from_recorder=1&focus_title=1
    - some menu bar items won't display children as tooltip when expanding
    - is it possible to expand the menu without going to the page?  https://www.loom.com/share/b527d335e911455ba130e394395d2cd1
when in large view
  - collapsing the blog view will show the tooltip flash: https://www.loom.com/share/fc4ca4ae35b14c088ac068c52512788b

other:
  - can we have some padding between the end of the menu and the beginning of the content?  https://www.loom.com/i/61e6341052094a659ee5a86ced154d26
  - can we update the color of the menu expand item to be the same as the dendron accent color? https://www.loom.com/i/26a7358caff04a0bbd66f55907c9e962
    - this will be different depending on the theme. see [here](https://wiki.dendron.so/notes/1698aef0-efd4-4c92-a3b0-5ca702b228a3.html#updating-the-theme) for theme updates

namjul:
> some menu bar items won't display children as tooltip when expanding

is this different than the issue you point to in https://www.loom.com/share/03f4b0ac2024406da4865c094f7d5aca?from_recorder=1&focus_title=1 ?

> is it possible to expand the menu without going to the page?

We would need a way to let the user deside if he wants to open children or go to the page for touch based devices since otherwise mouse gestures are available. 
- If we add an arrow (to only open children) the area will need to grow in size and might still be easily missed.
- Another option would be to only open page on double tab. Not sure how easily antd Menu lets us do this.

PR #1194, opened 2021-08-23
Enhance/vsix build

jonathanyeung:
2 parts to this change: 
1. Create a new pipeline that can produce a .vsix artifact.
2. Insert Oauth client ID and secret via a Github secret. This is to unblock https://github.com/dendronhq/dendron/pull/1155

### DIFF
--- a/.github/workflows/create-vsix-artifact.yml
+++ b/.github/workflows/create-vsix-artifact.yml
@@ -24,38 +24,35 @@ jobs:
           yarn --version
 
       - name: Configure Git user
-        env: 
-          GOOGLE_OAUTH_CLIENT_ID: ${{ secrets.GOOGLE_OAUTH_CLIENT_ID }}
         run: |
           git config --global user.name CI
           git config --global user.email ci@dendron.so
-          echo "Client ID: $GOOGLE_OAUTH_CLIENT_ID"
 
-      # - name: Checkout source
-      #   uses: actions/checkout@v2
+      - name: Checkout source
+        uses: actions/checkout@v2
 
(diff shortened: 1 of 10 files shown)


packages/plugin-core/src/commands/ShowHelp.ts
```diff
@@ -14,6 +15,8 @@ export class ShowHelpCommand extends BasicCommand<CommandOpts, CommandOutput> {
     return {};
   }
   async execute() {
+    console.log("This is the client id: " + GOOGLE_OAUTH_ID);
```
kevinslin, line 18: remove

packages/plugin-core/src/types/global.ts
```diff
@@ -0,0 +1,8 @@
+// These strings are inserted during webpack build via definePlugin
+declare const GOOGLE_OAUTH_CLIENT_ID: string;
+const GOOGLE_OAUTH_ID = GOOGLE_OAUTH_CLIENT_ID;
```
kevinslin, line 3: should this be `process.env.GOOGLE_OAUTH_CLIENT_ID`

.github/workflows/create-vsix-artifact.yml
```diff
@@ -0,0 +1,58 @@
+name: Create VSIX Artifact
+
+on:
+  push:
+    branches:
+      - enhance/auth-build
+      - prerelease-*
```
kevinslin, line 7: can we change this to `prerelease/*`?

kevinslin (changes requested):

PR #1212, opened 2021-08-25
enhance: Adding caching optimization to improve startup/workspace initialization time.

nickolay-kondratyev:
Change to cut down initialization time to about 40% of what it is before this change. Timings with org-workspace can be found at [[Start up Optimization|user.nickolay.tasks.start-up-optimization]]

[x] - Ran unit tests successfully with jest in dendron/packages/engine-test-utils
[x] - Manually started a workspace using local server , checked that existing backlinks are operational.
[x] - Wrote a test script that grabs the payload from the /initialize API call and "standardizes" to be able to be used for comparison the following is performed during "standardization"
All the 'stub' notes are removed as well as all the ids that reference stub notes (parent & child ids),
Schema updated and created times are removed. Then after standardatization does a JSON compare with previously saved standardized output. Tested the payload from org-workspace to be the same before and after this change (standardized payload)

Unit test cases added:
```
 PASS  src/__tests__/engine-server/util/inMemoryNoteCache.spec.ts
  inMemoryNoteCache.spec.ts
    getNotesByFileNameIgnoreCase tests:
      GIVEN cache with valid note props
        ✓ WHEN calling for file name that has different cases in notes THEN get both (2 ms)
        ✓ WHEN calling for file name that matches single note THEN get the note
        ✓ WHEN calling for file name that does not exist THEN empty list
        ✓ WHEN calling undefined file name THEN throw (4 ms)
        ✓ WHEN calling with empty file name THEN throw
 ```

### CODE REVIEW

File: packages/engine-server/src/util/impl/inMemoryNoteCacheImpl.ts [hunk 1]
```diff
@@ -0,0 +1,47 @@
+import { DendronError, NoteProps } from "@dendronhq/common-all";
+import { InMemoryNoteCache } from "../inMemoryNoteCache";
+
+export class InMemoryNoteCacheImpl implements InMemoryNoteCache {
+  /** Note map which maps lowercase file name to a list {@link NoteProps}
+   *  that have the matching file name */
+  private readonly mapFNameToNotes: Map<string, NoteProps[]>;
+
+  static createCache(notes: NoteProps[]) {
```
jonathanyeung, line 9: Why static here and not just an instance method + public constructor?  Same with `initializeFileNameMap()` method below.

packages/common-server/src/timingUtil.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/** Converts the given nano seconds to milliseconds */
+export const nanosToMillis = (nanos: number) => {
+  return nanos / NANOS_IN_MILLI_SEC;
+};
+
+/**
+ * Returns a number representing the milliseconds elapsed
+ * between 1 January 1970 00:00:00 UTC and the given date */
+export const milliseconds = () => {
```
jonathanyeung, line 18: Not sure this util method is necessary - `Date().getTime()` is quite concise.

packages/engine-server/src/util/inMemoryNoteCache.ts
```diff
@@ -0,0 +1,19 @@
+import { NoteProps } from "@dendronhq/common-all";
+import { InMemoryNoteCacheImpl } from "./impl/inMemoryNoteCacheImpl";
+
+/**
+ * In memory cache for {@link NoteProps}.
+ *
+ * Use {@link InMemoryNoteCacheFactory} to get an instance of this interface.*/
+export interface InMemoryNoteCache {
```
jonathanyeung, line 8: I would actually call this something more generic - nothing about the interface methods require it to be either In-memory or a cache. Then, we could rename `InMemoryNoteCacheImpl` -> `InMemoryNoteCache`, which is the in-memory cache implementation of the `NoteRetriever` or whatever name we settle on.

packages/engine-server/src/util/nanosecUtil.ts
```diff
@@ -0,0 +1,13 @@
+const NANOS_IN_SECOND = 1000000000;
```
jonathanyeung, line 1: Isn't this duplicated logic from `packages/common-server/src/timingUtil.ts`?

packages/engine-test-utils/src/__tests__/engine-server/util/inMemoryNoteCache.spec.ts
```diff
@@ -0,0 +1,83 @@
+import { InMemoryNoteCacheFactory } from "@dendronhq/engine-server/lib/util/inMemoryNoteCache";
+import { NoteProps } from "@dendronhq/common-all";
+
+/** For tests ONLY.
```
jonathanyeung, line 4: nit: fine to leave, but also not necessary to emphasize this is for tests only as this is in the engine-test-utils package.

packages/engine-server/src/util/impl/inMemoryNoteCacheImpl.ts
```diff
@@ -0,0 +1,47 @@
...
+      InMemoryNoteCacheImpl.initializeFileNameMap(notes)
+    );
+  }
+
+  private constructor(noteMap: Map<string, NoteProps[]>) {
+    this.mapFNameToNotes = noteMap;
+  }
+
+  private static initializeFileNameMap(notes: NoteProps[]) {
+    const map = new Map<string, NoteProps[]>();
```
jonathanyeung, line 20: Is the reason this is `NoteProps[]` and not `NoteProps` as the value because multiple vaults may have the same note name?

packages/engine-server/src/util/inMemoryNoteCache.ts
```diff
@@ -0,0 +1,19 @@
...
+  /** Returns list of {@link NoteProps} that have matching file name (ignoring
+   *  the file name case). Will return empty list if no notes match.  */
+  getNotesByFileNameIgnoreCase(fileName: string): NoteProps[];
+}
+
+/** Factory for {@link InMemoryNoteCache} */
+export class InMemoryNoteCacheFactory {
```
jonathanyeung, line 15: Perhaps a Factory isn't necessary here - are we actually going to be able to return a different implementation based on the argument (`NoteProps`)?  Or are we injecting the factory into a class such that the class can control object lifecycle?  Seems like no in either case.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -357,39 +359,48 @@ export class FileStorage implements DStore {
...
+    const ctx = "_addBacklinks:ext";
+    const start = process.hrtime();
+
+    this._addBacklinksImpl(allNotes, notesWithLinks);
+
+    const duration = getDurationMilliseconds(start);
+    this.logger.info({ ctx, duration });
+  }
+
+  private _addBacklinksImpl(
```
jonathanyeung, line 379: since these are both private, probably fine to just merge into a single `_addBacklinks()` method - but I'll leave it up to you.

jonathanyeung (changes requested):
Overall looking good!

jonathanyeung (approved):

PR #1219, opened 2021-08-26
enhance: Add multi-select for move note command

nickolay-kondratyev:
Added integration test. Performed the following manual testing:

```
GIVEN [note.note-1, note.note-2] in vault-A
    WHEN note-1 is moved without multi select to vault-B
        THEN note.note-1 is in vault-B
    WHEN [note.note-1, note.note-2] is moved with multi select to vault-B
        THEN [note.note-1, note.note-2] are in vault-B

GIVEN [note.note-1, note.note-2] are in vault-A 
    AND [note.note-3] are in vault-C
        WHEN [note.note-1, note.note-2, note.note-3] are multi select moved to vault-B
            THEN [note.note-1, note.note-2, note.note-3] are in vault-B
```


hikchoi (changes requested):
Any reason the manual testing couldn't be in integ test?
 
 NK> Added additional integration testing. 

 Also, this is not strictly related to multi-select, but there is a new history event action (`changeState`) added when the lookup picker's onHide is triggered. The move note command was written before this addition so it's not handling for this case in `gatherInputs` and throwing an error. This should properly be resolved here.
 
 NK> Added an if branch for changeState/onHide.

packages/common-all/src/arrays.ts
```diff
@@ -0,0 +1,14 @@
...
+  interface Array<T> {
+    /** Extension method returns true when array has elements, false otherwise. */
+    hasElements(): boolean;
+  }
+}
+
+Array.prototype.hasElements = function <T>(): boolean {
+  const arr = this as Array<T>;
+  return arr.length > 0;
+};
```
hikchoi, line 14: I'm not sure if we should be extending native objects for array empty checks. 

We are already using lodash extensively, which provides `_.isEmpty` that could be used instead of extending the native array.

In any case, I think a simple `if (someArray.length)` should suffice.
`if (Array.isArray(someArray) && someArray.length)` if you want to be extra careful.

packages/common-test-utils/src/index.ts
```diff
@@ -810,3 +810,104 @@ export class NoteTestPresetsV2 {
     ];
   }
 }
+
+/**
+ * Utility class to make the tests against vault to be concise while having descriptive
+ * failure messages. */
+export class VaultAsserter {
```
hikchoi, line 817: I think these utility functions could be part of `VaultUtils`. Or if you want to isolate utilities strictly for testing, I suggest we keep the naming convention consistent with the others and call it say, `VaultTestUtils`.

packages/common-test-utils/src/index.ts
```diff
@@ -810,3 +810,104 @@ export class NoteTestPresetsV2 {
...
+  }
+
+  /** Asserts that no match is found  */
+  async noMatch(noMatchList: string[]) {
+    const hasNoneMatched = await EngineTestUtilsV4.checkVault({
+      ...this.workspaceVault,
+      nomatch: noMatchList,
+    });
+
+    assert(
```
hikchoi, line 831: Plugin tests are written in `mocha` to work with VSCode. We are using `jest` to test our plugin, so we wrap assertions to be jest-like in `testUtils`. We should keep things consistent and switch these into jest like `expect`s.

packages/common-test-utils/src/index.ts
```diff
@@ -810,3 +810,104 @@ export class NoteTestPresetsV2 {
     ];
   }
 }
+
+/**
+ * Utility class to make the tests against vault to be concise while having descriptive
+ * failure messages. */
+export class VaultAsserter {
+  private readonly workspaceVault: WorkspaceVault;
```
hikchoi, line 818: `WorkspaceVault` is a [vault that belongs to another workspace](https://wiki.dendron.so/notes/6682fca0-65ed-402c-8634-94cd51463cc4.html#vault-types).

packages/common-test-utils/src/index.ts
```diff
@@ -810,3 +810,104 @@ export class NoteTestPresetsV2 {
...
+    );
+  }
+
+  private vaultToString() {
+    return `Vault fsPath:'${this.workspaceVault.vault.fsPath}' vaultName: '${this.workspaceVault.vault.name}'`;
+  }
+}
+
+/** Asserter utility to have more concise tests, while having descriptive failure messages.
+ *  Supports fluent API. */
```
hikchoi, line 860: I'm a bit torn with the decision of introducing a fluent interface here.

As I understand, this allows methods to be chained and read like plain english which makes code more readable.
The downside I see here is that the classes can get very large, and would get larger with further attempts to extend what it does. (which in turn can slow us down).

I am aware some of the utility methods we currently have doesn't quite roll off the tongue, but I think we are generally doing a good job with having a descriptive method name (at least the ones that are being added recently), which does a fairly good job at making code readable.

My question is, is the overhead worth the potential readability gain? IMO, code can be readable regardless of how it is structured as long as the developers agree with a consistent convention. 

My opinion is that if we want more clean code, we should spend time discussing and standardizing conventions and cleaning up existing code that doesn't fit what we agreed upon. My concern is that we are exchanging a slight readability increase on the client side (which is currently only us, as this code isn't a library that is exposed for general use)and sacrificing the ease of incremental improvement and timely iteration.

Would love other's opinion on this. Correct me if I'm not understanding this correctly.

packages/plugin-core/src/test/suite-integ/MoveNoteCommand.test.ts
```diff
@@ -516,41 +530,174 @@ suite("MoveNoteCommand", function () {
...
+          wsRoot,
+        }) as NoteProps;
+
+        await VSCodeUtils.openNote(fooNote);
+        const cmd = new MoveNoteCommand();
+
+        // @ts-ignore
+        VSCodeUtils.showQuickPick = async () => {
+          return "proceed";
+        };
```
hikchoi, line 582: this could be mocked with `sinon`.

`sinon.stub(VSCodeUtils, "showQuickPick").resolves("proceed");`

packages/plugin-core/src/test/suite-integ/MoveNoteCommand.test.ts
```diff
@@ -516,41 +530,174 @@ suite("MoveNoteCommand", function () {
...
-              wsRoot,
-            })
+          VSCodeUtils.getActiveTextEditor()?.document.fileName.endsWith(
+            path.join("vault3", "foo.md")
           )
         ).toBeTruthy();
+
+        // Check that the files are not in old vaults anymore
+        await vaultAsserter(vault1).noMatch(["foo.md"]);
+        await vaultAsserter(vault2).noMatch(["bar"]);
```
hikchoi, line 615: Left a separate comment regarding the fluent interface, so strictly commenting about readability here.

I think this could be more readable (and be consistent with what we are currently doing) if it looks something like:

```
const fooExists = await VaultTestUtils.noteExists({ vault: vault1, notes: ["foo"] });
expect(fooExists).toBeTruthy();
```

packages/common-test-utils/src/index.ts
hikchoi: I realize this may be an effort to introduce DSL like tests (i.e. `vault().should.not.have.note("foo")`)
Is this the goal here? I am not aware of any discussions about this happening so I may be commenting on things that have already been talked about 😅  Sorry if this is something already discussed

packages/common-all/src/arrays.ts
nickolay-kondratyev: Sure, for now I can just use `_.isEmpty(arrayValue)` although [imho]: `arrayValue.hasElements()` reads nicer.

On using length in an if statement as above `if (someArray.length)` I would disagree with such style since it is a bit on the implicit side meaning takes advantage of typescripts/javascripts subtlety about truthiness evaluation of 0 being false. And would rather write `if (someArray.length === 0)` to be more explicit.

kevinslin: Lets avoid monkey patching javascript built-ins. This makes code harder to follow since its no longer standard javascript. 

As for negations - generally agree that negative conditionals should be avoided when possible but a more important principle we follow is to reduce complexity in the code base. ideally, there should be one way of doing things. so rather have `!_.isEmpty` be the way for checking array fullness instead of adding a new method

packages/common-test-utils/src/index.ts
nickolay-kondratyev: Let's remove the comment about the fluent API to avoid this turning into something bigger than it is (some wrappers to have descriptive failure messages). While having the client code that uses these wrappers be fairly concise.

kevinslin: To follow with @hikchoi, we currently don't use fluent style when writing tests so lets stick with existing conventions

kevinslin (changes requested):

packages/common-test-utils/src/index.ts
nickolay-kondratyev: Renamed to VaultTestUtils

packages/common-test-utils/src/index.ts
nickolay-kondratyev: jest's expects do not seem to allow custom error messages to be added to the calls, which prompted the creation of these wrapper classes. Given lack of messages, and that when the tests fail, they give line numbers in JS output instead of TS source code. Adding error messages adds quite a bit of clarity of what is failing. 

There are way to add messages to jest, but they require custom matchers or adding additional dependency hence thought just using assert here would be a simpler approach, since its hidden from the user of tests anyway and will throw an error similar how jest will throw an error. The main difference being of JestAssertionError being thrown vs AssertionError being thrown.

packages/common-test-utils/src/index.ts
nickolay-kondratyev: Will rename to just vault.

packages/common-test-utils/src/index.ts
nickolay-kondratyev: > My opinion is that if we want more clean code, we should spend time discussing and standardizing conventions and cleaning up existing code that doesn't fit what we agreed upon. My concern is that we are exchanging a slight readability increase on the client side (which is currently only us, as this code isn't a library that is exposed for general use)and sacrificing the ease of incremental improvement and timely iteration.

Let's push away from any `fluent apis` (which I shouldn't have added in the comment, I was just planning on returning `this` of the asserter not going for any complex full on fluent api) and just discuss using more clear wrappers. Which can actually speed us up rather slow down our iterations. Given that, we end up with less code in tests, clear failure messages which speed up troubleshooting times. 

Maybe it's just me, but it looks like quite a bit of readability improvement of what to assert is trying to achieve:  

```
vaultNoteAsserter(vaultFrom).noMatchByFName("foo");
```

VS

```
expect(
          _.isUndefined(
            NoteUtils.getNoteByFnameV5({
              fname: "foo",
              notes,
              vault: vaultFrom,
              wsRoot,
            })
          )
        ).toBeTruthy();
```

packages/plugin-core/src/test/suite-integ/MoveNoteCommand.test.ts
nickolay-kondratyev:
```
const fooExists = await VaultTestUtils.noteExists({ vault: vault1, notes: ["foo"] });
expect(fooExists).toBeTruthy();
```

Unfortunately, this won't get us a descriptive failure message. Per other comments, just refactored to assertions to look as other tests in this file.

packages/plugin-core/src/test/suite-integ/MoveNoteCommand.test.ts
nickolay-kondratyev: done.

packages/common-test-utils/src/index.ts
nickolay-kondratyev: As per later discussion refactoring to use the current way.

packages/common-test-utils/src/index.ts
nickolay-kondratyev: removed

packages/common-test-utils/src/index.ts
nickolay-kondratyev: removed

packages/common-all/src/arrays.ts
nickolay-kondratyev: Since we also have some usages of arr.length (Eg. https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/external/memo/utils/utils.ts#L206) and it avoids a double negative using that for now.

packages/common-test-utils/src/index.ts
hikchoi: Thanks for the clarifications. Let's make some time to discuss this :) I agree that test errors could have better messages than what we have right now.

packages/engine-server/src/engineClient.ts
```diff
@@ -322,6 +322,12 @@ export class DendronEngineClient implements DEngineClient {
     });
   }
 
+  /** Renames the note.
+   *
+   *  WARNING: When doing bulk operations. Do not invoke multiple requests to this
+   *  command in parallel, wait for a single call to finish before requesting another call.
+   *  Otherwise some race condition starts to cause intermittent failures.
```
hikchoi, line 329: Is this something that will affect end users? If so, it should be noted in the docs when we ship it. or ideally, prevent it from happening internally.

hikchoi (commented):
Changes look good 👍🏼 
As for the discussion about test wrappers, it's probably a bigger scope than what this PR is addressing, but I agree that it deserves some attention.

Let's make some time and discuss about it soon.

packages/engine-server/src/engineClient.ts
nickolay-kondratyev: As long as we call this in sequence it should work as expected. I was running into intermittent issue when trying to issue multiple calls at the same time and then wait for all of their completion

Waiting for a single change to fully complete (as this method is currently used in this PR) should work all the time without this intermittent issue:

```
 for (const move of necessaryMoves) {
      // We need to wait for a rename to finish before triggering another rename
      // eslint-disable-next-line no-await-in-loop
      const changes = await engine.renameNote(move);

      allChanges.push(...(changes.data as NoteChangeEntry[]));
    }
```

The warning is added so we ourselves use this renameNote() only in sequence and not in parallel.

hikchoi: got it!

kevinslin (approved):

hikchoi (approved):

PR #1221, opened 2021-08-26
spike: groundwork for standalone vaults

kevinslin:
Does some of the ground work for standalone vaults.

- check if we are in a dendron workspace
    - workspace can be a code workspace (a .code-workspace file) or a native workspace (folder with dendron.yml)
- depending on what sort of workspace we're in, Dendron will instantiate either a Dendron[Native|Code]Workspace that adheres to the interface `DWorkspaceV2`
- instead of using `DendronWorkspace` directly to access workspace properties, we will used the `DWorkspaceV2` client

Need to clean up all the instances where we call `Dendron.instance` and `getWs` to use `getWSV2`. Since this is a bigger change, going to merge it into mainline in chunks. This is the first chunk


packages/engine-server/src/workspace/service.ts
```diff
@@ -740,6 +742,56 @@ export class WorkspaceService {
...
+    currentVersion: string;
+    previousVersion: string;
+    dendronConfig: DendronConfig;
+  }) {
+    // check if we need to force a migration
+    try {
+      const maybeRaw = DConfig.getRaw(this.wsRoot);
+      if (
+        _.isUndefined(maybeRaw.journal) &&
+        workspaceInstallStatus === InstallStatus.INITIAL_INSTALL
```
hikchoi, line 766: wouldn't all the workspaces with `InstallStatus.INITIAL_INSTALL` have the `journal` config in there by default?

packages/plugin-core/src/workspace.ts
```diff
@@ -351,6 +379,18 @@ export class DendronWorkspace {
     this.L.info({ ctx, msg: "initialized" });
   }
 
+  getOrThrowWorkspaceImpl(): DWorkspaceV2 {
```
hikchoi, line 382: nitpick: to be consistent with the other get-or-throwers in the codebase

```suggestion
  getWorkspaceImplOrThrow(): DWorkspaceV2 {
```

packages/engine-server/src/workspace/service.ts
kevinslin: you're right. updated

packages/plugin-core/src/workspace.ts
kevinslin: updated

hikchoi (approved):
LGTM

PR #1224, opened 2021-08-27
Enhance/UI progress indicators for long running cmds

Harshita-mindfire:
- [x] Reload Index
- [x] Workspace Sync
- [x] Refractor Hierarchy

### DIFF
--- a/packages/plugin-core/src/commands/Sync.ts
+++ b/packages/plugin-core/src/commands/Sync.ts
@@ -55,15 +55,16 @@ export class SyncCommand extends BasicCommand<CommandOpts, CommandReturns> {
         cancellable: false,
       },
       async (progress) => {
-        progress.report({ message: "committing repos" });
+        progress.report({ increment: 0, message: "committing repos" });
         const committed = await workspaceService.commitAndAddAll();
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/commands/Sync.ts
```diff
@@ -48,14 +48,27 @@ export class SyncCommand extends BasicCommand<CommandOpts, CommandReturns> {
...
-    L.info(committed);
+    const { committed, pulled, pushed } = await window.withProgress(
+      {
+        location: ProgressLocation.Notification,
+        title: "Syncing Workspace",
+        cancellable: false,
+      },
+      async (progress) => {
+        progress.report({ message: "committing repos" });
+        const committed = await workspaceService.commitAndAddAll();
```
jonathanyeung, line 59: nit: can you also call `progress.increment()` - I think this gives a nice progress bar progression for users to see.  Maybe 25% for committed, 50% for pull, final 25% for push?

Harshita-mindfire: a66b2de

kevinslin (approved):

PR #1238, opened 2021-08-30
enhance: cache rendered previews using LRU cache.

nickolay-kondratyev:
Using Least recently used cache implementation for caching note previews. [lru-cache](https://www.npmjs.com/package/lru-cache) has 40Million weekly downloads.  Has ISC license, ISC="It is functionally equivalent to the simplified BSD and MIT licenses" [ISC wiki](https://en.wikipedia.org/wiki/ISC_license).

LRU unit cache test output:
```
  lruCache.spec.ts
    GIVEN lruCache configured for 5 elements
      WHEN added 5 elements
        ✓ THEN all 5 elements are present in the cache (3ms)
        AND added 6th element
          ✓ THEN last 5 elements are present in the cache (1ms)
          ✓ THEN first element added is NOT in the cache (1ms)
        AND asked for element #1
          AND added 6th element
            ✓ THEN element #1 is IN the cache
            ✓ THEN element #2 is NOT in the cache (1ms)
            ✓ THEN elements 3-6 are in the cache
    Creation error cases:
      ✓ WHEN trying to init cache with 0 max item THEN throw
      ✓ WHEN trying to init cache with negative max items THEN throw
```

API tests:
```
  api/note/render tests:
    ✓ WHEN calling /render on non existent file THEN get payload with error. (217 ms)
    ✓ WHEN calling /render on valid file THEN get rendered preview. (61 ms)
```

Tested against local instance to make sure cache is operating as expected 
```
WHEN noCaching is set in config THEN we do not cache preview

WHEN config is default
   THEN we cache the preview
       WHEN 10+ previews are opened 
          THEN originally cached preview is dropped from cache (re-render is called when its pulled up)
   WHEN note is updated 
      THEN preview is re-rendered 
 ```

### DIFF
--- a/packages/engine-server/src/enginev2.ts
+++ b/packages/engine-server/src/enginev2.ts
@@ -36,6 +36,7 @@ import {
   RenameNotePayload,
   RenderNoteOpts,
   RenderNotePayload,
+  ResponseUtil,
   RespV2,
   SchemaModuleDict,
   SchemaModuleProps,
@@ -57,6 +58,9 @@ import { FileStorage } from "./drivers/file/storev2";
 import { LinkUtils, MDUtilsV5, ProcFlavor } from "./markdown";
 import { AnchorUtils, RemarkUtils } from "./markdown/remark/utils";
 import { HookUtils } from "./topics/hooks";
+import { Cache, NullCache } from "./util/cache";
+import { LruCache } from "./util/lruCache";
+import { milliseconds } from "@dendronhq/common-all/lib/timing";
 
 type CreateStoreFunc = (engine: DEngineClient) => DStore;
 type DendronEngineOptsV2 = {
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: packages/engine-server/src/enginev2.ts [hunk 1]
```diff
@@ -57,6 +58,9 @@ import { FileStorage } from "./drivers/file/storev2";
 import { LinkUtils, MDUtilsV5, ProcFlavor } from "./markdown";
 import { AnchorUtils, RemarkUtils } from "./markdown/remark/utils";
 import { HookUtils } from "./topics/hooks";
+import { Cache, NullCache } from "./util/cache";
+import { LruCache } from "./util/lruCache";
+import { milliseconds } from "@dendronhq/common-all/lib/timing";
```
kevinslin, line 63: this should be an absolute import `@dendronhq/common-all`. as a followup task, can you add this check to our husky hooks on the pre-commit stage? see [here](https://wiki.dendron.so/notes/a80f36d9-01d0-4c5a-a228-867b093a4913.html#adding-a-new-husky-hook)

packages/engine-server/src/util/cache.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  /** Set the given data, potentially removing a different item depending
+   *  on cache implementation and cache state. */
+  set(key: K, data: T): void;
+}
+
+/** Null object implementation of {@link Cache} to be used when we
+ *  dont want the code to actually use the cache. */
+export class NullCache<K, T> implements Cache<K, T> {
+  // @ts-ignore
```
kevinslin, line 14: Unless there is a special circumstance, avoid using this statement since it skips typechecking. Alternatives to some common use cases:
- unused variables: prefix the variable with a `_`  (eg. `_foo`) to have typescript ignore it

packages/engine-server/src/util/cache.ts
```diff
@@ -0,0 +1,23 @@
...
+  get(key: K): T | undefined {
+    return undefined;
+  }
+
+  // @ts-ignore
```
kevinslin, line 19: see above

kevinslin (changes requested):
can you also add a test that incorporates calling the API? instructions for writing tests with API endpoints [here](https://wiki.dendron.so/notes/X6NIty0XA4k6SPNONiW8Q.html#^goOLosrw7a1W)

packages/engine-server/src/util/cache.ts
nickolay-kondratyev: Sounds good.

packages/engine-server/src/enginev2.ts
nickolay-kondratyev: Added tests for the API. (In the updated pull request)

Added husky pre commit check https://github.com/dendronhq/dendron/pull/1243

jonathanyeung (approved):

PR #1239, opened 2021-08-30
feat: enable usePrettyRefs for nextJS publishing and preview

hikchoi:
This PR:
- enables `config.usePrettyRefs` and `config.site.userPrettyRefs` that overrides how note references are rendered in preview and publishing respectively.
- adds tests for both cases
- updates snapshots

### DIFF
--- a/packages/engine-server/src/markdown/remark/noteRefsV2.ts
+++ b/packages/engine-server/src/markdown/remark/noteRefsV2.ts
@@ -16,7 +16,6 @@ import {
 import { file2Note } from "@dendronhq/common-server";
 import _ from "lodash";
 import { brk, html, paragraph, root } from "mdast-builder";
-import { ConfigUtils } from "../../config";
 import { Eat } from "remark-parse";
 import Unified, { Plugin } from "unified";
 import { Node, Parent } from "unist";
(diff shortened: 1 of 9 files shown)


packages/engine-server/src/markdown/remark/noteRefsV2.ts
```diff
@@ -306,39 +306,45 @@ export function convertNoteRefASTV2(
...
-
   if (!vault) {
     return {
       error: new DendronError({ message: "no vault specified" }),
       data: [],
     };
   }
-  let { prettyRefs, wikiLinkOpts } = compilerOpts;
+  const { wikiLinkOpts } = compilerOpts;
+  const sitePrettyRefConfig = _.isUndefined(procOpts.config?.site?.usePrettyRefs)
```
kevinslin, line 325: can we use [DConfig.getProp](https://github.com/dendronhq/dendron/blob/512b476e8ac986887f25bc9a21bf7b189ea19ce9/packages/engine-server/src/config.ts#L114-L114) here?

packages/engine-server/src/markdown/remark/noteRefsV2.ts
```diff
@@ -306,39 +306,45 @@ export function convertNoteRefASTV2(
...
+    ? true
+    : procOpts.config.site.usePrettyRefs;
+  const prettyRefConfig = _.isUndefined(procOpts.config?.usePrettyRefs)
```
kevinslin, line 328: same comment as above

kevinslin (changes requested):

packages/engine-server/src/markdown/remark/noteRefsV2.ts
hikchoi: done!

kevinslin:
can you take a look at the integration test failure?

kevinslin (approved):

PR #1243, opened 2021-08-31
chore: add pre-commit husky check to avoid direct imports from packages

nickolay-kondratyev:
### Test 1: Should not allow commit:
Test attempt to commit 
```
import { milliseconds } from "@dendronhq/common-all/lib/timing";

console.log(milliseconds());
```

Error message looks as:
```
The following files contains 'Direct import from a package.
This check catches direct imports such as:
import { milliseconds } from "@dendronhq/common-all/lib/timing";
While the above import should look as:
import { milliseconds } from "@dendronhq/common-all";' in them:
packages/engine-test-utils/src/__tests__/api-server/assets.spec.ts
```

---- 
### Test 2 : Should allow commit
Was able to commit:
```
import { milliseconds } from "@dendronhq/common-all";

console.log(milliseconds());
```

### DIFF
--- a/hooks/pre-commit.js
+++ b/hooks/pre-commit.js
@@ -1,6 +1,16 @@
 const {checkToken} = require("./common");
 const {exec} = require("./exec");
 
+const ErrorMessages = {
+  AVOID_DIRECT_IMPORT_FROM_PACKAGES:[
+    `Direct import from a package.`,
+    `This check catches direct imports such as:`,
+    `import { milliseconds } from "@dendronhq/common-all/lib/timing";`,
+    `While the above import should look as:`,
+    `import { milliseconds } from "@dendronhq/common-all";`
+  ].join('\n')
+}
+
 function main() {
   const gitCommand = `git diff --staged --name-only`;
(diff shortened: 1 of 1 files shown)


hooks/pre-commit.js
```diff
@@ -1,6 +1,16 @@
 const {checkToken} = require("./common");
 const {exec} = require("./exec");
 
+const ErrorMessages = {
+  AVOID_DIRECT_IMPORT_FROM_PACKAGES:[
+    `Direct import from a package.`,
+    `This check catches direct imports such as:`,
+    `import { milliseconds } from "@dendronhq/common-all/lib/timing";`,
+    `While the above import should look as:`,
+    `import { milliseconds } from "@dendronhq/common-all";`
```
hikchoi, line 10: nitpick: could be more generic so as to not be confused as an actual match that blocked the commit.

hikchoi (approved):
Tested it out and looks good other than a small nitpick.

hooks/pre-commit.js
nickolay-kondratyev: Noted, made the message more generic in pull request v2

PR #1264, opened 2021-09-01
chore: reduce vscode bundle size

kevinslin:
Some initial work on reducing our bundle size. removes 10mb off final package


jonathanyeung (approved):
Curious - how much size does this save us?

Issue #1237, opened 2021-08-30
Markdown Import pod does not update links referencing assets


Harshita-mindfire:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [x] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

The assets are imported in assets folder, but the links referencing assets remains the same after importing
with markdown pod. 

## To Reproduce

assets folder structure of source to import 

    |--images
          |---2021-07-25-12-25-06.png

md file referencing asset has the following link
```
![](/images/2021-07-25-12-25-06.png)
```

assets Folder structure of vault after running markdown import

 ```
assets
     |---2021-07-25-12-25-06-wV4ld9vIX4pL.png
```
md file referencing it remains
```
![](/images/2021-07-25-12-25-06.png)
```
## Expected behavior

md file referencing it should update the link as
```
![](/assets/2021-07-25-12-25-06-wV4ld9vIX4pL.png)
```
## Dendron Log file

Please attach the output of `>Dev:Dendron: Open Logs` here

**Comment by hikchoi** [comment]
Thanks for the report @Harshita-mindfire :) Are you going to be taking this item yourself?

**Comment by Harshita-mindfire** [comment]
Sure @hikchoi, assigning to self.

**Comment by Harshita-mindfire** [comment]
This should be fixed in v0.59 of Dendron. Please post here if not, otherwise this issue will auto close in 4 days

**Comment by sotn3m** [comment]
There is one additional scenario that does not work. Two examples:
``` 
folder
    |--- media
    |     |--- image.png
    |     |--- same_level.md
    |--- something.md
```
The contents of `same_level.md`: 
```markdown
![](image.jpg)
![](./image.jpg)
```
The contents of `something.md`:
```markdown
![](media/image.jpg)
![](./media/image.jpg)
```
None of the 4 links are updated.

In essence, relative links are not updated at the moment.

tgrosinger:
I'm still experiencing issues with this. I am importing a markdown Vault from Obsidian. I have a structure like this:

```
folder
  |---assets
  |------|---image.JPEG
  |---something.md
```
The contents of `something.md` before import were:

```
![[image.JPEG]]
```

[...]

PR #1265, opened 2021-09-01
Fix/markdown import to update asset links

Harshita-mindfire:
Fix for: [#1237](https://github.com/dendronhq/dendron/issues/1237)

 Loom: https://www.loom.com/share/5f101648d7dc46d59d9283267216ba1c

### DIFF
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -44,7 +44,6 @@ import type {
   Text,
 } from "mdast";
 import * as mdastBuilder from "mdast-builder";
-import path from "path";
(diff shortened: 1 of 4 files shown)


jonathanyeung (changes requested):
See prev. comment about being consistent on `/` slashes. Thanks!

Harshita-mindfire:
> I notice in the loom video that in the imported note the path has a mix of back and forward slashes. Can we normalize to all forward `/` slashes?

This might be the case for windows only, should I replace all `\` with '/' then?

SeriousBug (commented):
Small nitpicks, looks good.

jonathanyeung (approved):

PR #1296, opened 2021-09-06
fix: disallow toggling of vault selection behavior in move note

nickolay-kondratyev:
# fix: disallow toggling of vault selection behavior in move note
Disallows toggling of vault selection behavior while doing MoveNote command. 

Note: the vault mode button is still shown since it has behavior for prompting the vault selection, but it is now not toggleable and always stays in alwaysPrompt mode. 
Screen Shot 2021-09-06 at 8 17 01 PM

# Pull Request Checklist

### Quality Assurance


hikchoi (approved):
LGTM

kevinslin (changes requested):
can you add a test for this? might want to create a `button.test.ts`

kevinslin:
also - this should be split off into a separate discussion: is this something we want to dis-allow? another option is to remove the "rename note" command and have move note act as both. as a reminder, here are our [product principles](https://wiki.dendron.so/notes/7fcebd7d-6411-4c9d-8baf-65629dc018a1.html#product). 

currently, we don't have a guiding principle to answer this question. 

I like python's philosophy "There should be one– and preferably only one –obvious way to do it".
Dendron's take on this - "there should be one obvious way to do it and that way can change over time"

So getting back to this, we can have a "move note" that does both "rename" and "move" or we can have a separate 'move command' and 'rename command'. in case of the latter, move note should just show you the vault selection menu and not give you an option to rename.

@hikchoi @jonathanyeung -> thoughts?

hikchoi:
> also - this should be split off into a separate discussion: is this something we want to dis-allow? another option is to remove the "rename note" command and have move note act as both. as a reminder, here are our [product principles](https://wiki.dendron.so/notes/7fcebd7d-6411-4c9d-8baf-65629dc018a1.html#product).
> 
> currently, we don't have a guiding principle to answer this question.
> 
> I like python's philosophy "There should be one– and preferably only one –obvious way to do it".
> Dendron's take on this - "there should be one obvious way to do it and that way can change over time"
> 
> So getting back to this, we can have a "move note" that does both "rename" and "move" or we can have a separate 'move command' and 'rename command'. in case of the latter, move note should just show you the vault selection menu and not give you an option to rename.
> 
> @hikchoi @jonathanyeung -> thoughts?

I agree on the idea of splitting those two. It communicate clearer what the command(s) do. Things that have a lot of moving parts can easily break anyways. The downside is that we could end up with a million different commands, but I personally much prefer to have a headache of having too many commands than maintaining a swiss army knife that is slowly falling apart.

jonathanyeung:
Stepping back a bit - I think maybe we should separate the functionality like this: 

1) Move Note(s) - meant to move a note or notes from one vault to another.
2) Remove the Rename Command altogether in favor of having someone rename a note the same way they would rename any file in VS Code as in the Contextual UI spec (either right click rename, or use refactor hierarchy).

Putting the vault move functionality aside, I feel like the original motivation behind creating a 'Move Note' command in the first place was because we had cache and engine problems in picking up the changes.  That seems more of a workaround though?

kevinslin:
> Putting the vault move functionality aside, I feel like the original motivation behind creating a 'Move Note' command in the first place was because we had cache and engine problems in picking up the changes. That seems more of a workaround though?

Move note is used to move notes between vaults. Super useful in multi-vault when deciding between making certain notes public/private or for organizing existing material between different vaults.

> Remove the Rename Command altogether in favor of having someone rename a note the same way they would rename any file in VS Code as in the Contextual UI spec (either right click rename, or use refactor hierarchy).

Only issue is that VSCode currently doesn't provide a way to rename with the command prompt (that I know of). Goes against Dendron being dev-friendly (keyboard driven)

jonathanyeung:
Haha - just thought of a different approach:

Why don't we instead
1) Move Note 'Renaming' Functionality to a different command (which is an equivalent command to just renaming in explorer/file system, but have a dev friendly command)
2) Merge the Vault Moving functionality of the current `Move Note` with `Refactor Hierarchy`?  The idea is actually the same - we're moving a set of notes from one place to another, whether it's in the same vault but different hierarchy, or same hierarchy but changing the vaults (or both).  The picker options should be consistent between the two types of operations I think.  I bet the request to have multi-select on `Move Note` came about because someone wanted to move an entire hierarchy to a different vault? 

The UI may need to be tweaked a bit to make it flow better, but semantically I think this makes sense?

kevinslin (approved):

PR #1308, opened 2021-09-08
enhance: conditionally bubble up create new & allow queries longer than fname when special chars are used.

nickolay-kondratyev:
Screen Shot 2021-09-08 at 5 50 10 PM

# Conditionally bubble up create new and allow longer queries than file name searched when special chars are used.
When special characters are not used bubble up create new entry, and allow queries that are longer than file names searched when special characters are used. For this change allows 'hi world !bye' to match 'hi.world'

# Addresses
* [[ scratch.2021.09.08.121701.lookup-bubble-up-create-new-when-no-exact-match-and-no-special-characters]]
* [[scratch.2021.09.07.213159.improve-look-up-do-not-count-special-chars-for-hiding-results]]

# Note 
As mentioned in the comments our current & previous behavior allows creation of notes that can be confusing for FuseJS to look up. For example we can create a note 'hi$world' but if we query for just dollar sign '$' FuseJS will not look up 'hi$world' since '$' will be treated as special character according to https://fusejs.io/examples.html#extended-search rules. 'hi$world' can still be matched by 'hi world'. Given that we don't know if our users have such characters in their files not disabling creation of such files in this PR but it is something for us to think about. 

# Pull Request Checklist

### Quality Assurance

### DIFF
--- a/packages/common-all/src/fuse.ts
+++ b/packages/common-all/src/fuse.ts
@@ -97,6 +97,24 @@ type FuseEngineOpts = {
 };
 
 export class FuseEngine {
+  /**
+   * Characters that are specially treated by FuseJS search
+   * Reference https://fusejs.io/examples.html#extended-search
+   *
+   * Includes '*' which is not specially treated by FuseJS but we currently
+   * map '*' to ' ' which specially treated by FuseJS.
+   */
+  private static readonly SPECIAL_QUERY_CHARACTERS = [
+    "*",
+    " ",
+    "|",
+    "^",
+    "$",
+    "!",
+    "=",
+    "'",
+  ];
+
   public notesIndex: Fuse<NoteIndexProps>;
   public schemaIndex: Fuse<SchemaProps>;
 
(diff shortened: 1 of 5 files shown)


kevinslin (changes requested):
Changes look good, can you also update the docs here -> [[Summary|dendron.topic.lookup.create#summary]] to describe this behavior?

hikchoi (approved):
lgtm. +1 on the point that we should have docs about special queries.

nickolay-kondratyev:
Updated the docs to describe bubble up behavior https://github.com/dendronhq/dendron-site/pull/191

On the point of special queries documentation we have FuseJS extended search documentation linked in Lookup note documentation https://wiki.dendron.so/notes/a7c3a810-28c8-4b47-96a6-8156b1524af3.html but we can expand upon their doc to make the special queries syntax be more visible in our documentation.

kevinslin (approved):

PR #1317, opened 2021-09-08
Updated OpenLink.ts

Stedag:
Feat: use Dendron: Open_Link to open both generic URLs and assets found in Dendron vaults.

Further, instead of needing to highlight an asset's relative path, this change allows Open_Link to open the asset path or URI found at the cursor/selection. This also amends for vs-code not handling links with different protocols than https://, http://, or vscode://, allowing users to open other programs more smoothly.

### DIFF
--- a/packages/plugin-core/src/commands/OpenLink.ts
+++ b/packages/plugin-core/src/commands/OpenLink.ts
@@ -46,7 +46,6 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
       this.L.error({ error });
       return { error };
     }
-
     let assetPath: string;
     if (text.indexOf("://") !== -1) {
       window.showInformationMessage(
@@ -80,6 +79,6 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
         return { error };
       });
     }
-    return { String: assetPath };
+    return { filepath: assetPath };
   }
 }
--- a/packages/plugin-core/src/commands/OpenLink.ts
+++ b/packages/plugin-core/src/commands/OpenLink.ts
@@ -40,11 +40,15 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
       return { error };
     }
     let assetPath: string;
-    if (text.indexOf("://") !== -1) {
+    if (
+      text.indexOf(":/") !== -1 ||
+      text.indexOf("/") === 0 ||
+      text.indexOf(":\\") !== -1
+    ) {
       window.showInformationMessage(
-        "the selection reads as a full URI so an attempt will be made to open it"
+        "the selection reads as a full URI or filepath so an attempt will be made to open it"
       );
-      env.openExternal(Uri.parse(text));
+      env.openExternal(Uri.parse(text.replace("\\", "/"))); // make sure vscode doesn't choke on "\"s
       assetPath = text;
     } else {
       const wsRoot = getDWorkspace().wsRoot;
--- a/packages/plugin-core/src/commands/OpenLink.ts
+++ b/packages/plugin-core/src/commands/OpenLink.ts
@@ -7,10 +7,11 @@ import path from "path";
 import { PickerUtilsV2 } from "../components/lookup/utils";
 import { DENDRON_COMMANDS } from "../constants";
 import { resolvePath, VSCodeUtils } from "../utils";
+import { getURLAt } from "../utils/md";
 import { isAnythingSelected } from "../utils/editor";
 import { getExtension, getDWorkspace } from "../workspace";
 import { BasicCommand } from "./base";
-import { Selection, env, Uri, window } from "vscode";
+import { env, Uri, window } from "vscode";
 
 type CommandOpts = {};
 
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/commands/OpenLink.ts
```diff
@@ -25,47 +26,106 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+    }
+
+    const editor = VSCodeUtils.getActiveTextEditor();
+
+    if (editor) {
+      const docText = editor.document.getText();
+      const offsetStart = editor.document.offsetAt(editor.selection.start);
+      const offsetEnd = editor.document.offsetAt(editor.selection.end);
+
+      const selectUri = true;
```
kevinslin, line 45: instead of parsing the link in here, can you add a method `getURLAt` in [utils/editor](https://github.com/dendronhq/dendron/blob/21689914d0c84735d243b988dcceb276df97380f/packages/plugin-core/src/utils/editor.ts#L122-L122)?
you can take a look at [CopyLinkUrl](https://github.com/dendronhq/dendron/blob/fe83a99d11feffb0c588a5ae3b23b73b6283c851/packages/plugin-core/src/commands/CopyNoteURL.ts#L62-L62) to see how we use this on a link

packages/plugin-core/src/test/suite-integ/OpenLink.test.ts
```diff
@@ -20,7 +20,7 @@ suite("OpenLink", function () {
       onInit: async ({}) => {
         const cmd = new OpenLinkCommand();
         const { error } = await cmd.execute();
-        expect(error!.message).toEqual("nothing selected");
+        expect(error!.message).toEqual("no valid path or URL selected");
```
kevinslin, line 23: can you add a test for the added functionality here? You can see how we add tests with selection enabled [here](https://github.com/dendronhq/dendron/pull/1317/files). instructions for running specific test [here](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#run-specific-test)

kevinslin (changes requested):
thanks for the PR! left some comments in terms of conventions. let me know if you have any questions

kevinslin:
Hey @Stedag - just checking if you had any questions on the comments

packages/plugin-core/src/commands/OpenLink.ts
```diff
@@ -25,47 +27,59 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+    let text = "";
+
     if (!isAnythingSelected()) {
-      const error = DendronError.createFromStatus({
-        status: ERROR_STATUS.INVALID_STATE,
-        message: `nothing selected`,
-      });
-      this.L.error({ error });
-      return { error };
+      window.showInformationMessage(
```
kevinslin, line 34: don't forget to return here

packages/plugin-core/src/commands/OpenLink.ts
```diff
@@ -25,47 +27,59 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
-    await open(assetPath).catch((err) => {
-      const error = DendronError.createFromStatus({
-        status: ERROR_STATUS.UNKNOWN,
-        error: err,
-      });
-      this.L.error({ error });
-      return { error };
-    });
-    return { filePath: assetPath };
+    return { String: assetPath };
```
kevinslin, line 83: avoid generic names like `String`. unless there's good reason, keep the existing name `filePath`

packages/plugin-core/src/test/suite-integ/OpenLink.test.ts
```diff
@@ -1,31 +1,100 @@
...
         done();
       },
     });
   });
 
+  //TODO
+  test.skip("With an invalid character in the selection.", (done) => {
+    let noteWithLink: NoteProps;
+    runSingleVaultTest({
```
kevinslin, line 34: `runSingleVaultTest` requires a different setup. you want to use `runLegacyMultiWorkspaceTest` and then tests will work

kevinslin (changes requested):

packages/plugin-core/src/commands/OpenLink.ts
Stedag: The proposed functionality would allow a user to follow a link without selecting it so having nothing selected should no longer be an error

packages/plugin-core/src/test/suite-integ/OpenLink.test.ts
```diff
@@ -54,10 +52,9 @@ suite("OpenLink", function () {
     });
   });
 
-  //TODO
   test.skip("grab a URL under the cursor.", (done) => {
```
Stedag, line 55: This test is skipped because it actually opens a link. I don't imagine maintainers want to deal with an unexpected pop-up when testing the plugin.

It may be a good Idea to test this on mac/Linux before shipping (I'm on windows).

kevinslin: For this case, you can use `stubs` as a way of checking for functionality without invoking side effects.
See our docs about doing this here -> https://wiki.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#stubs

You can see an example of how it is used [here](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/DoctorCommand.test.ts#L169:L169)

kevinslin (changes requested):
really close now. left a comment about tests

Stedag:
Stubbing seems too simple... but it seems to work! :D

packages/plugin-core/src/commands/OpenLink.ts
```diff
@@ -25,47 +26,56 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
     let assetPath: string;
-    if (text.startsWith("asset")) {
-      const vault = PickerUtilsV2.getOrPromptVaultForOpenEditor();
-      assetPath = path.join(vault2Path({ vault, wsRoot }), text);
+    if (
+      text.indexOf(":/") !== -1 ||
+      text.indexOf("/") === 0 ||
+      text.indexOf(":\\") !== -1
+    ) {
+      window.showInformationMessage(
```
Stedag, line 48: Should this info still be announced?

packages/plugin-core/src/commands/OpenLink.ts
```diff
@@ -25,47 +26,56 @@ export class OpenLinkCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
     }
-    const wsRoot = getDWorkspace().wsRoot;
     let assetPath: string;
-    if (text.startsWith("asset")) {
-      const vault = PickerUtilsV2.getOrPromptVaultForOpenEditor();
-      assetPath = path.join(vault2Path({ vault, wsRoot }), text);
+    if (
+      text.indexOf(":/") !== -1 ||
+      text.indexOf("/") === 0 ||
+      text.indexOf(":\\") !== -1
```
Stedag, line 46: There are a few other path formats that may be good to check for (like "~" or "%APPDATA%" but this should be enough for now.

kevinslin: agreed, we can deal with those in separate issues as they come up

kevinslin (approved):

PR #1334, opened 2021-09-14
feat: nextjs publishing fulltext search

SeriousBug:
Added full text search capability to nextjs publishing. The search works on both the titles and the bodies of notes. 

Currently, the search data is exported with the nextjs export, and it gets fetched and loaded the first time the user clicks on the search bar. The fusejs index is provided through a generalized React hook, and new hooks can be added easily. As an example, I added a hook to generate the index on the client side, which is not currently used.

The following images were cropped vertically, but not horizontally.


packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -0,0 +1,80 @@
...
+          >
+            <div>{LOADING_MESSAGE}</div>
+          </AutoComplete.Option>
+        ) : undefined}
+        {fuse
+          ? fuse.search(value).map((note) => (
+              <AutoComplete.Option key={note.item.id} value={note.item.fname}>
+                <div>
+                  {note.item.title}
+                  <span style={{ marginLeft: "10px", opacity: 0.7 }}>
```
kevinslin, line 70: can we add a preview for the fusejs results? so when the results are rendered, show the raw text for the search result

```
{10 chars context} {{textThatIsMatched}} {10 chars context}
```

kevinslin (changes requested):

packages/nextjs-template/components/DendronSearch.tsx
SeriousBug: fuse.js isn't giving me any info for which part of the document is matched. Do you know if there's a setting that I might be missing for that?

SeriousBug:
Added:
- Preview for the matching note text, with context, with matching part highlighted
- Matching parts in the title highlighted

Wide:


Narrow:


The "..." parts only get added if the matching part isn't at the start or end of the note.

One caveat is only the first matching part of the note text is displayed. I also implemented a version that displays all matching parts, but it ends up being very confusing in my opinion:

kevinslin (approved):
looks good. can we add an item to triage for fulltext search preview improvements (eg. either render results in html or strip out markdown symbols from results)

PR #1347, opened 2021-09-16
enhance: Added optional image and imageAlt to SEO open graph options

santosh898:
This is a small enhancement to use the image from the frontmatter in the Note's SEO open graph options.

This is my for PR. Dropping my discord handle here  - santosh898#7877


packages/common-all/src/types/foundation.ts
```diff
@@ -183,6 +183,8 @@ export type NoteSEOProps = {
   updated: number;
   created: number;
   excerpt?: string;
+  image?: string;
```
kevinslin, line 186: can you add some docstrings describing these properties?

packages/nextjs-template/components/DendronSEO.tsx
```diff
@@ -47,6 +47,9 @@ export default function DendronSEO() {
   const seoProps = NoteUtils.getSEOProps(noteActive);
   const title = seoProps.title;
   const description = seoProps.excerpt || config.site.description;
+  const images = seoProps.image
+    ? [{ url: seoProps.image, alt: seoProps.imageAlt }]
```
kevinslin, line 51: will there be a problem if `url` is set and `imageAlt` is undefined?

kevinslin (changes requested):
awesome to see your first PR in :)

some minor nitpicks and questions

packages/common-all/src/types/foundation.ts
santosh898: Will do

santosh898: Will do

packages/nextjs-template/components/DendronSEO.tsx
santosh898: Looks like it does. I followed `next-seo` type declarations and did this. But now i dug into the [OpenGraph Sepc](https://ogp.me/#structured) and the alt is required.

Should we move the image from `frontMatter.custom` to `frontMatter.image?.` and change image from `string` to `{ url: string, alt: string}`?

santosh898: Or using the note title as the alt prop's fallback is easier?

SeriousBug: Adding this as an actual property to frontmatter sounds like a better design to me. Here's some docs on how to go about adding it: https://wiki.dendron.so/notes/3c80629b-4048-48f0-bdfd-352645bda2ec.html#adding-a-new-frontmatter-property

packages/common-all/src/types/foundation.ts
santosh898: Hey @kevinslin i adeed the docString for the new image prop.

packages/nextjs-template/components/DendronSEO.tsx
santosh898: @kevinslin and @SeriousBug please take a look, i've added the `image` as an optional property to the frontmatter.

kevinslin (approved):

kevinslin (approved):

PR #1349, opened 2021-09-17
feat(command): move header command

hikchoi:
# feat(command): move header command
This PR adds a new command `Dendron: Move Header`, that lets you select a header similar to `Dendron: Rename Header`, and choose a note to move the header.
The selected header will be removed from the original note and added to the end of the selected note's body.
All wikilinks that reference this header will be updated to point to the newly moved header.

Docs PR: https://github.com/dendronhq/dendron-site/pull/204

---

# Pull Request Checklist

### Quality Assurance

#### Special Cases

### Docs

### DIFF
--- a/packages/plugin-core/src/commands/MoveHeader.ts
+++ b/packages/plugin-core/src/commands/MoveHeader.ts
@@ -31,7 +31,8 @@ import path from "path";
 import { getEngine } from "../workspace";
 import { EngineAPIService } from "../services/EngineAPIService";
 import { delayedUpdateDecorations } from "../features/windowDecorations";
-import { findReferences } from "../utils/md";
+import { findReferences, FoundRefT } from "../utils/md";
+import { Location } from "vscode";
 
 type CommandInput =
   | {
@@ -55,15 +56,15 @@ export class MoveHeaderCommand extends BasicCommand<
 > {
   key = DENDRON_COMMANDS.MOVE_HEADER.key;
 
-  headerNotSelectedError = new DendronError({
+  private headerNotSelectedError = new DendronError({
     message: "You must first select the header you want to move.",
   });
 
-  noActiveNoteError = new DendronError({
+  private noActiveNoteError = new DendronError({
     message: "No note open.",
   });
 
-  getProc = (engine: EngineAPIService, note: NoteProps) => {
+  private getProc = (engine: EngineAPIService, note: NoteProps) => {
     return MDUtilsV5.procRemarkParse(
       { mode: ProcMode.FULL },
       {
@@ -76,49 +77,34 @@ export class MoveHeaderCommand extends BasicCommand<
   };
 
   /**
-   * Recursively check if two given node is identical.
-   * At each level _position_ is omitted as this can change if
-   * you are comparing from two different trees.
-   * @param a first {@link Node} to compare
-   * @param b second {@link Node} to compare
-   * @returns boolean
+   * Helper for {@link MoveHeaderCommand.gatherInputs}
+   * Validates and processes inputs to be passed for further action
+   * @param engine 
+   * @returns {}
    */
-  hasIdenticalChildren = (a: Node, b: Node): boolean => {
-    if (_.isEqual(Object.keys(a).sort(), Object.keys(b).sort())) {
-      const aOmit = _.omit(a, ["position", "children"]);
-      const bOmit = _.omit(b, ["position", "children"]);
-      if (_.isEqual(aOmit, bOmit)) {
-        if (_.has(a, "children")) {
-          return _.every(a.children as Node[], (aChild, aIndex) => {
-            const bChild = (b.children as Node[])[aIndex];
-            return this.hasIdenticalChildren(aChild, bChild);
-          });
-        }
-        return true;
-      } else {
-        return false;
-      }
-    } else {
-      return false;
-    }
-  };
-
-  async gatherInputs(opts: CommandInput): Promise<CommandOpts | undefined> {
-    const engine = getEngine();
+  private validateAndProcessInput(engine: EngineAPIService): {
+    proc: Processor,
+    origin: NoteProps,
+    targetHeader: Heading,
+  } {
     const { editor, selection } = VSCodeUtils.getSelection();
+
+    // basic input validation
     if (!editor) throw this.noActiveNoteError;
     if (!selection) throw this.headerNotSelectedError;
 
-    // const text = editor.document.getText();
     const line = editor.document.lineAt(selection.start.line).text;
     const maybeNote = VSCodeUtils.getNoteFromDocument(editor.document);
     if (!maybeNote) {
       throw this.noActiveNoteError;
     }
 
+    // parse selection and get the target header node
     const proc = this.getProc(engine, maybeNote);
     const parsedLine = proc.parse(line);
     let targetHeader: Heading | undefined;
+    // Find the first occurring heading node in selected line.
+    // This should be our target.
     visit(parsedLine, [DendronASTTypes.HEADING], (heading: Heading) => {
       targetHeader = heading;
       return false;
@@ -126,40 +112,16 @@ export class MoveHeaderCommand extends BasicCommand<
     if (!targetHeader) {
       throw this.headerNotSelectedError;
     }
-    const tree = proc.parse(maybeNote.body);
-    let headerFound = false;
-    let foundHeaderIndex: number | undefined;
-    let nextHeaderIndex: number | undefined;
-    visit(tree, (node, index) => {
-      if (nextHeaderIndex) {
-        return;
-      }
-      const depth = node.depth as Heading["depth"];
-      if (!headerFound) {
-        if (node.type === DendronASTTypes.HEADING) {
-          if (
-            depth === targetHeader!.depth &&
-            this.hasIdenticalChildren(node, targetHeader!)
-          ) {
-            headerFound = true;
-            foundHeaderIndex = index;
-            return;
-          }
-        }
-      } else if (node.type === DendronASTTypes.HEADING) {
-        if (foundHeaderIndex) {
-          if (depth <= targetHeader!.depth) nextHeaderIndex = index;
-        }
-      }
-    });
+    return { proc, origin: maybeNote, targetHeader }
+  }
 
-    const nodesToMove = nextHeaderIndex
-      ? (tree.children as any[]).splice(
-          foundHeaderIndex!,
-          nextHeaderIndex! - foundHeaderIndex!
-        )
-      : (tree.children as any[]).splice(foundHeaderIndex!);
-
+  /**
+   * Helper for {@link MoveHeaderCommand.gatherInputs}
+   * Prompts user to do a lookup on the desired destination.
+   * @param opts 
+   * @returns 
+   */
+  private promptForDestination(opts: CommandInput) {
     const lookupController = LookupControllerV3.create({
       nodeType: "note",
       disableVaultSelection: true,
@@ -176,7 +138,21 @@ export class MoveHeaderCommand extends BasicCommand<
       initialValue: opts?.dest ? opts.dest.fname : undefined,
       nonInteractive: !_.isUndefined(opts?.dest),
     });
+    return lookupController;
+  }
 
+  async gatherInputs(opts: CommandInput): Promise<CommandOpts | undefined> {
+    // validate and process input
+    const engine = getEngine();
+    const { proc, origin, targetHeader } = this.validateAndProcessInput(engine);
+
+    // extract nodes that need to be moved
+    const originTree = proc.parse(origin.body);
+    const nodesToMove = RemarkUtils.extractHeaderBlock(originTree, targetHeader);
+
+    const lc = this.promptForDestination(opts);
+
+    // Wait for provider
     return new Promise((resolve) => {
       HistoryService.instance().subscribev2("lookupProvider", {
         id: this.key,
@@ -186,13 +162,13 @@ export class MoveHeaderCommand extends BasicCommand<
             const cdata = event.data as NoteLookupProviderSuccessResp;
             resolve({
               dest: cdata.selectedItems[0],
-              origin: maybeNote,
+              origin,
               nodesToMove,
-              modifiedOriginTree: tree,
+              modifiedOriginTree: originTree,
               originProc: proc,
               engine,
             });
-            lookupController.onHide();
+            lc.onHide();
           } else if (event.action === "error") {
             this.L.error({ msg: `error: ${event.data}` });
             resolve(undefined);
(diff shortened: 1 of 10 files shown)


packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+} & CommandOpts;
+
+export class MoveHeaderCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput
+> {
+  key = DENDRON_COMMANDS.MOVE_HEADER.key;
+
+  headerNotSelectedError = new DendronError({
+    message: "You must first select the header you want to move.",
```
jonathanyeung, line 59: Can you add severity minor to these? 

I'm thinking of using severity as a way to determine whether an error gets uploaded to error analytics, as we don't want this kind of error reported.

I'll do a write-up shortly for a error-reporting style guide / sentry docs.

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+    }
+  };
+
+  async gatherInputs(opts: CommandInput): Promise<CommandOpts | undefined> {
+    const engine = getEngine();
+    const { editor, selection } = VSCodeUtils.getSelection();
+    if (!editor) throw this.noActiveNoteError;
+    if (!selection) throw this.headerNotSelectedError;
+
+    // const text = editor.document.getText();
```
jonathanyeung, line 112: super nit: remove commented code.

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+        return false;
+      }
+    } else {
+      return false;
+    }
+  };
+
+  async gatherInputs(opts: CommandInput): Promise<CommandOpts | undefined> {
+    const engine = getEngine();
+    const { editor, selection } = VSCodeUtils.getSelection();
```
jonathanyeung, line 108: dumb question, I forgot - does at least one character need to be selected for this to work? will this return anything if I just put my cursor in the header but nothing is selected?

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+    const nodesToMove = nextHeaderIndex
+      ? (tree.children as any[]).splice(
+          foundHeaderIndex!,
+          nextHeaderIndex! - foundHeaderIndex!
+        )
+      : (tree.children as any[]).splice(foundHeaderIndex!);
+
+    const lookupController = LookupControllerV3.create({
+      nodeType: "note",
+      disableVaultSelection: true,
```
jonathanyeung, line 165: Do we actually want this disabled? Can we move the header to a note in a different vault?

jonathanyeung (changes requested):
Cool! 

Thinking out loud for a few things: 

1. Wonder if it'd be a cool UI that if no header is currently selected, we show a quick pick containing all the headers in the doc and provide the ability to move it that way.  Maybe that kind of UI could also support moving multiple headers at the same time.  (not saying we should do this now, maybe something for v-next++)

2. I'm thinking about discoverability of this feature.  Not sure how hard it'd be to detect if someone does a Cut / Paste on a header that is referenced, could we warn the user some how and point them to using this command instead?  Or perhaps ultimately, what would be really cool is if we can detect that automagically during the cut/paste, we can then give a prompt to the user if they would like for us to fix their references on their behalf.  Maybe something I'll add to the contextual UI spec.

packages/plugin-core/src/commands/MoveHeader.ts
kevinslin: +1 on this. that should be possible

packages/plugin-core/src/commands/MoveHeader.ts
kevinslin: I believe so. we have the problem right now we're putting the cursor on a line for `CopyNoteLink` grabs the header (often when I don't mean to do it)

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+    const line = editor.document.lineAt(selection.start.line).text;
+    const maybeNote = VSCodeUtils.getNoteFromDocument(editor.document);
+    if (!maybeNote) {
+      throw this.noActiveNoteError;
+    }
+
+    const proc = this.getProc(engine, maybeNote);
+    const parsedLine = proc.parse(line);
+    let targetHeader: Heading | undefined;
+    visit(parsedLine, [DendronASTTypes.HEADING], (heading: Heading) => {
```
kevinslin, line 122: you can use [isHeading](https://github.com/dendronhq/dendron/blob/e83f4945c0838221bb635ca69ce9b905c9d3f6d0/packages/engine-server/src/markdown/remark/utils.ts#L853-L853) for this

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+    lookupController.show({
+      title: "Select note to move header to",
+      placeholder: "note",
+      provider: lookupProvider,
+      initialValue: opts?.dest ? opts.dest.fname : undefined,
+      nonInteractive: !_.isUndefined(opts?.dest),
+    });
+
+    return new Promise((resolve) => {
+      HistoryService.instance().subscribev2("lookupProvider", {
```
kevinslin, line 181: as a separate item, we should abstract this away into a utility that all lookup commands can more easily adopt

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  /**
+   * Recursively check if two given node is identical.
+   * At each level _position_ is omitted as this can change if
+   * you are comparing from two different trees.
+   * @param a first {@link Node} to compare
+   * @param b second {@link Node} to compare
+   * @returns boolean
+   */
+  hasIdenticalChildren = (a: Node, b: Node): boolean => {
```
kevinslin, line 86: is this expensive for big documents? will it be worthwile to add a shallow flag where we only check 2 levels deep?

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+      return false;
+    });
+    if (!targetHeader) {
+      throw this.headerNotSelectedError;
+    }
+    const tree = proc.parse(maybeNote.body);
+    let headerFound = false;
+    let foundHeaderIndex: number | undefined;
+    let nextHeaderIndex: number | undefined;
+    visit(tree, (node, index) => {
```
kevinslin, line 133: - can we abstract this away in [RemarkUtils](https://github.com/dendronhq/dendron/blob/e83f4945c0838221bb635ca69ce9b905c9d3f6d0/packages/engine-server/src/markdown/remark/utils.ts#L822-L822)?

```ts
extractHeaderBlock(targetHeader: Heading) => {blockStartIndex: number, blockEndIndex: number }
```

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+          } else {
+            this.L.error({ msg: `unhandled error: ${event.data}` });
+            resolve(undefined);
+          }
+        },
+      });
+    });
+  }
+
+  removePosition = (nodes: Node[]) => {
```
kevinslin, line 211: is this used anywhere?

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+      // find match text in note and modify the node
+      const proc = this.getProc(engine, note);
+      const tree = proc.parse(note.body);
+      visit(tree, (node: Node) => {
+        if (node.type === DendronASTTypes.WIKI_LINK) {
+          if (
+            node.value === origin.fname &&
+            _.includes(anchorsToUpdateNames, node.data!.anchorHeader)
+          ) {
+            node.value = dest!.fname;
```
kevinslin, line 300: lets have a followup task to make it possible to move xvault. we can add it to `triage`

packages/plugin-core/src/test/suite-integ/MoveHeader.test.ts
```diff
@@ -0,0 +1,151 @@
...
+import { VSCodeUtils } from "../../utils";
+import { runLegacyMultiWorkspaceTest, setupBeforeAfter } from "../testUtilsV3";
+import { describe, beforeEach } from "mocha";
+import vscode from "vscode";
+import { expect } from "../testUtilsv2";
+import { MoveHeaderCommand } from "../../commands/MoveHeader";
+import _ from "lodash";
+
+suite("MoveHeader", function () {
+  const ctx = setupBeforeAfter(this, {});
```
kevinslin, line 15: the `{}` is optional

kevinslin (changes requested):

packages/plugin-core/src/commands/MoveHeader.ts
nickolay-kondratyev: Should we create a class such as `UserError` to distinguish between the type of errors (if this is a user error).

Could be still very beneficial to upload user errors to error analytics to have insight into whether some parts of UI are common potholes for users.

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+        engine,
+        fname: note.fname,
+        vault: note.vault,
+        dest: DendronASTDest.MD_DENDRON,
+      }
+    );
+  };
+
+  /**
+   * Recursively check if two given node is identical.
```
nickolay-kondratyev, line 79: spelling nit: Recursively check if two given nodes are identical.

And should this comment be something like: Recursively check if two given nodes have identical children?

packages/plugin-core/src/test/suite-integ/MoveHeader.test.ts
```diff
@@ -0,0 +1,151 @@
...
+            const out = await cmd.run({ dest: destNote });
+            expect(out!.origin.body.includes("## Foo header")).toBeFalsy();
+            expect(out!.dest!.body.includes("## Foo header")).toBeTruthy();
+            done();
+          }),
+        });
+      });
+
+      test("THEN reference to moved header is updated", (done) => {
+        runLegacyMultiWorkspaceTest({
```
nickolay-kondratyev, line 78: If I am not mistaken this `runLegacyMultiWorkspaceTest` is doing brand new workspace setup each time that it runs, some concern whether this will make our tests slow and whether we want to move such setup higher up and into beforeAll?

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -0,0 +1,311 @@
...
+  removePosition = (nodes: Node[]) => {
+    return _.map(nodes, (node) => {
+      visit(node, (n) => {
+        delete n["position"];
+      });
+      return node;
+    });
+  };
+
+  async execute(opts: CommandOpts): Promise<CommandOutput> {
```
nickolay-kondratyev, line 220: This isn't necessarily request to change this function, but more so kick off some discussions on the preferred code style, that we typically gravitate towards.

This in particular talking about the typical length of functions that we are ok with. Some rules of thumbs that we can consider:
* keeping the length of functions to allow entire function to be able to fit on a single screen. 
* Anytime there is a comment describing a block of code, ask whether it makes sense to split that code block into a function, name that function as the comment and get rid of the comment. 

For example as broadstrokes refactoring this function could have some splits added as:

```
 async execute(opts: CommandOpts): Promise<CommandOutput> {
    const ctx = "MoveHeaderCommand";
    this.L.info({ ctx, opts });
    const {
      dest,
      origin,
      nodesToMove,
      modifiedOriginTree,
      originProc,
      engine,
    } = opts;

    // deep copy the origin before mutating it
    const originDeepCopy = _.cloneDeep(origin);
    // remove header from origin
    const modifiedOriginContent = originProc.stringify(modifiedOriginTree);

    origin.body = modifiedOriginContent;
    await engine.writeNote(origin, {
      updateExisting: true,
    });

    // append header to destination
    await this.appendHeaderToDestination(engine, dest, nodesToMove);

    delayedUpdateDecorations();

    // update reference
    const anchorsToUpdateNames = this.findAnchorsToUpdate(originDeepCopy, modifiedOriginContent);

    const foundReferences = await findReferences(origin.fname);
    const updated: NoteProps[] = [];

    _.forEach(foundReferences, async (reference) => {
      const { location, isCandidate } = reference;
      // ignore candidate refs.
      if (isCandidate) return false;
      const note = this.getNoteByLocation(location, engine);
      if (!note) return false;

      // find match text in note and modify the node
      const proc = this.getProc(engine, note);
      const tree = proc.parse(note.body);
      visit(tree, (node: Node) => {
        if (node.type === DendronASTTypes.WIKI_LINK) {
          if (
            node.value === origin.fname &&
            _.includes(anchorsToUpdateNames, node.data!.anchorHeader)
          ) {
            node.value = dest!.fname;
          }
        }
      });
      note.body = proc.stringify(tree);
      const writeResp = await engine.writeNote(note, { updateExisting: true });
      updated.push(writeResp.data[0].note);
      return;
    });
    return { ...opts, updated };
  }

  private async appendHeaderToDestination(engine: EngineAPIService, dest: NoteProps | undefined, nodesToMove: Node[]) {
    const destProc = this.getProc(engine, dest!);
    const destTree = destProc.parse(dest!.body);
    destTree.children = (destTree.children as Node[]).concat(nodesToMove);
    const modifiedDestContent = destProc.stringify(destTree);
    dest!.body = modifiedDestContent;
    await engine.writeNote(dest!, {
      updateExisting: true,
    });
  }

  private findAnchorsToUpdate(originDeepCopy: NoteProps, modifiedOriginContent: string) {
    const anchorsBefore = RemarkUtils.findAnchors(originDeepCopy.body);
    const anchorsAfter = RemarkUtils.findAnchors(modifiedOriginContent);
    const anchorsToUpdate = _.differenceWith(
      anchorsBefore,
      anchorsAfter,
      this.hasIdenticalChildren
    );
    const anchorsToUpdateNames = _.map(anchorsToUpdate, (anchor) => {
      const slugger = getSlugger();
      const payload = AnchorUtils.anchorNode2anchor(anchor, slugger);
      return payload![0];
    });
    return anchorsToUpdateNames;
  }

  private getNoteByLocation(location: any, engine: EngineAPIService) {
    const fsPath = location.uri.fsPath;
    const {vaults, wsRoot, notes} = engine;
    const fname = NoteUtils.normalizeFname(path.basename(fsPath));
    const vault = VaultUtils.getVaultByNotePath({
      fsPath,
      wsRoot,
      vaults,
    });
    const note = NoteUtils.getNoteByFnameV5({
      fname,
      notes,
      vault,
      wsRoot,
    });
    return note;
  }
  ```

File: packages/plugin-core/src/commands/MoveHeader.ts [hunk 15]
```diff
@@ -0,0 +1,311 @@
...
+    const anchorsToUpdateNames = _.map(anchorsToUpdate, (anchor) => {
+      const slugger = getSlugger();
+      const payload = AnchorUtils.anchorNode2anchor(anchor, slugger);
+      return payload![0];
+    });
+
+    const foundReferences = await findReferences(origin.fname);
+    const { vaults, wsRoot, notes } = engine;
+    const updated: NoteProps[] = [];
+    _.forEach(foundReferences, async (reference) => {
```
nickolay-kondratyev, line 271: Also discussion point: we often use lodash methods when streaming API has an alternatives. 

As example this code could be written with streaming api such that there is less state that we are dealing with while streaming through the data, with less state at each step in theory should be much easier to grasp the logic. 

```
    foundReferences
      .filter(r => !r.isCandidate)
      .map(r => this.getNoteByLocation(r.location.uri.fsPath, wsRoot, vaults, notes))
      .filter(note => note !== undefined)
      .forEach(note => {
        // find match text in note and modify the node
        const proc = this.getProc(engine, note!);
        const tree = proc.parse(note!.body);
        visit(tree, (node: Node) => {
          if (node.type === DendronASTTypes.WIKI_LINK) {
            if (
              node.value === origin.fname &&
              _.includes(anchorsToUpdateNames, node.data!.anchorHeader)
            ) {
              node.value = dest!.fname;
            }
          }
        });
        note!.body = proc.stringify(tree);
        const writeResp = await engine.writeNote(note!, { updateExisting: true });
        updated.push(writeResp.data[0].note);
      });

```

packages/plugin-core/src/test/suite-integ/MoveHeader.test.ts
hikchoi: The problem here is that these testing utilities aren't modular. I can probably make it _slightly_ more readable by currying `runLegacyMultiWorkspaceTest` on each step of the Given-When-Then clauses, but at the end it will be invoked once all information needed to test is there, which is what I'm currently doing here already.

This is a pretty substantial friction at this point because we pretty much have to undo what `runLegacyMultiWorkspaceTest` is bundling together into individual lifecycles, and then spread those on top of the new BDD-light style structure of testing.

Not saying it couldn't be done or shouldn't be done. I'm just a bit concerned if we should focus on it _now_. Was curious if you had any thoughts on it.

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: That is a good point. After reviewing a couple of your PRs I sensed that you take a lot of effort in commenting your code which I found to be useful when reviewing. I was planning to add some coding style guide regarding this, but I have been deferring that for a bit 😅 

I definitely agree on the points you made here. A rule of thumb in terms of splitting code into separate functions right now that our team follows is to split it when it makes sense to generalize it so that it could be reused somewhere else. (but I've seen things get split up even when it wasn't reused somewhere else, for readability).

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: I think this is another coding style (or the lack thereof) issue. We try to use lodash where using it makes things easier, but we don't really have a concrete guiding principle for what you describe here. I myself go back and forth using lodash native js methods with no solid reasoning as to why I'm doing that. It's definitely a good topic to discuss. (probably not here)

packages/plugin-core/src/test/suite-integ/MoveHeader.test.ts
kevinslin: @SeriousBug is working on native workspaces. as part of that, he'll be writing a new test harness to setup the native workspaces. we should make sure the new test harness has the desired properties that make sense. created an rfc/discussion [here](https://github.com/dendronhq/dendron/discussions/1394)

packages/plugin-core/src/commands/MoveHeader.ts
jonathanyeung: That's a good point.  At this point, probably better to err on the side of more data than less data- although we have a different analytics avenue that we typically use to understand usage patterns as opposed to the sentry pathway. It may be more appropriate to divert it there.  I'll do some more thinking on this one

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: I don't think it's necessary. We're just looking for the first occurrence of a heading within a selected parsed line here.

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: This only runs when we hit a heading while traversing, which has a pretty bounded depth.

(note that any content that follows a heading is not a children of the heading. they are siblings.)

```md
# **~~deep header~~** foo **_bar_** baz [[wiki]] 

Some paragraph
```

Here the part of the heading that has nested formatting (bold, and then strikethrough) will have the worst case depth of 4. (heading > emphasis > delete > text)

A heading can have a Phrasing Content as children, I don't see anything that can have an unbounded depth here.

That, and the fact that the iteration stops and short circuits immediately once it hits a falsy item make me think it won't be too expensive.

The absolute worst case scenario I can think of is when a person has a gigantic file, with a bunch of headings where every single heading has 4 levels of depth of markdown formatting, and targets the very last heading to move.

Even then, the depth is bounded to be pretty shallow. So I don't think we need to worry about this being expensive.

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: oops 🙃

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: you are right.

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: added severity for now

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: Blocked out a short session to see if this is trivial but it doesn't seem like it. Will be adding this in my backlog as a next item.

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: added to my backlog 👍🏼

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: Went with the suggestion here as well.

kevinslin (approved):

hikchoi:
pinging @jonathanyeung for review of the recent suggested changes.

jonathanyeung (approved):

PR #1295, opened 2021-09-06
feat: consolidate dendron configs

hikchoi:
# feat: consolidate dendron configs

This is a draft PR to showcase how the configurations will be organized.

## Overview

- `packages/common-all/src/types/configsdendronConfigs.ts` is where the top most level of configuration is defined.
  - As we go narrower with our namespace, we will define each namespace in a similar fashion.

- I have split up the code so that all each namespace is modularized.
  - Each namespace will define the following:
    - A type that holds all config of this namespace
      - These will be used internally, and in `dendron.yml`
    - A constant object that holds labels and descriptions of every instance of configs in this namespace
      - These are there to provide user friendly descriptions of the configs and will be used in the configuration UI
      - The value, label, and description is bundled up in the `DendronConfigEntry` type.
    - A method that returns a default config of this namespace
      - Each namespace defines this and the parent namespace will simply import and call it
    - Any other useful enum / type definitions related to the configs in the namespace

- Using `packages/common-all/src/types/configs/commands/lookup.ts` to illustrate the above points:
  - Everything here is related to configuring the lookup _command_, so it's under `DendronCommandConfig`.
  - What is defined in `LookupConfig` is what ends up in `dendron.yml`
  - Currently we let the users select what selection behavior should be for note lookup, and the possible choices are defined in `NoteLookupSelectionBehaviorEnum`, and in turn `NoteLookupSelectionBehavior` (which is generated from the enum's keys).
  - Constants in `SELECTION_BEHAVIORS` map the keys in `NoteLookupSelectionBehavior` to respective key's `DendronConfigEntry`s.
    - These will be used to generate useful messages about the lookup command configs in the configuration UI
    - If you introduce new behavior in `NoteLookupSelectionBehaviorEnum`, the compiler will catch this and warn you that you have to add a new label/description here as well. 
      - because `SELECTION_BEHAVIORS` has a type of 
        ```js
          { [key in NoteLookupSelectionBehavior]: DendronConfigEntry<string>}
        ```
  - `genDefaultLookupConfig` will define what the default `LookupConfig` should look like.
  - This will be called by it parent: `genDefaultCommandConfig`.

---

# Pull Request Checklist

You can go to [dendron pull requests](https://wiki.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e.html) to see full details for items in this checklist.

### Quality Assurance
[...]

PR #1373, opened 2021-09-21
enhance: Leave trace on source note while using selectionExtract

Harshita-mindfire:
```markdown
# enhance(workspace) : Leave a trace on source note when using selectionExtract
```

PR for docs update: https://github.com/dendronhq/dendron-site/pulls

# Pull Request Checklist

### Quality Assurance

#### Special Cases

### Docs

Example PR Description
```markdown
# feat: capitalize all foos

This changes capitalizes all occurences of `foo` to `Foo` 

Docs PR: <URL_TO_DOCS_PR>
```

### First Time PR
- [~] sign the [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) which will be prompted by our github bot after you submit the PR
- [~] add your [discord](https://discord.gg/AE3NRw9) alias in the review so that we can give you the [horticulturalist](https://wiki.dendron.so/notes/7c00d606-7b75-4d28-b563-d75f33f8e0d7.html#horticulturalist) badge in our community

### Analytics

### DIFF
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -130,6 +130,7 @@ export enum LookupSelectionType {
 
 export type NoteLookupConfig = {
   selectionType: LookupSelectionType;
+  leaveTrace?: boolean;
 };
 
 export type LookupConfig = {
--- a/packages/common-all/src/types/workspace.ts
+++ b/packages/common-all/src/types/workspace.ts
@@ -130,7 +130,7 @@ export enum LookupSelectionType {
 
 export type NoteLookupConfig = {
   selectionType: LookupSelectionType;
-  leaveTrace?: boolean;
+  leaveTrace: boolean;
 };
 
 export type LookupConfig = {
--- a/packages/plugin-core/src/components/lookup/buttons.ts
+++ b/packages/plugin-core/src/components/lookup/buttons.ts
@@ -4,11 +4,12 @@ import {
   NoteQuickInput,
   NoteUtils,
 } from "@dendronhq/common-all";
+import { DConfig } from "@dendronhq/engine-server";
 import _ from "lodash";
 import * as vscode from "vscode";
 import { QuickInputButton, ThemeIcon } from "vscode";
 import { clipboard, DendronClientUtilsV2, VSCodeUtils } from "../../utils";
-import { getExtension } from "../../workspace";
+import { getDWorkspace, getExtension } from "../../workspace";
 import {
   DendronQuickPickerV2,
   LookupEffectType,
@@ -96,13 +97,26 @@ const selectionToNoteProps = async (opts: {
   switch (selectionType) {
     case "selectionExtract": {
       if (!_.isUndefined(document)) {
+        const ws = getDWorkspace();
+        const leaveTrace =
+          DConfig.getProp(ws.config, "lookup").note.leaveTrace || false;
         const body = "\n" + document.getText(range).trim();
         note.body = body;
         // don't delete if original file is not in workspace
         if (!ext.workspaceService?.isPathInWorkspace(document.uri.fsPath)) {
           return note;
         }
-        await VSCodeUtils.deleteRange(document, range as vscode.Range);
+        if (leaveTrace) {
+          const editor = VSCodeUtils.getActiveTextEditor();
+          await editor?.edit((builder) => {
+            const link = note.fname;
+            if (!_.isUndefined(selection) && !selection.isEmpty) {
+              builder.replace(selection, `![[${link}]]`);
+            }
+          });
+        } else {
+          await VSCodeUtils.deleteRange(document, range as vscode.Range);
+        }
       }
       return note;
     }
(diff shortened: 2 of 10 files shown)


packages/plugin-core/src/components/lookup/buttons.ts
```diff
@@ -96,13 +97,26 @@ const selectionToNoteProps = async (opts: {
...
           return note;
         }
-        await VSCodeUtils.deleteRange(document, range as vscode.Range);
+        if (leaveTrace) {
+          const editor = VSCodeUtils.getActiveTextEditor();
+          await editor?.edit((builder) => {
+            const link = note.fname;
+            if (!_.isUndefined(selection) && !selection.isEmpty) {
+              builder.replace(selection, `![[${link}]]`);
+            }
```
hikchoi, line 115: here it would be nice to use `NoteUtils.createWikiLink` so that we can handle link aliases and vault prefixes too.
Check out how `CopyNoteLink` handles creating wikilinks to see an example usage

packages/common-all/src/types/workspace.ts
```diff
@@ -130,6 +130,7 @@ export enum LookupSelectionType {
 
 export type NoteLookupConfig = {
   selectionType: LookupSelectionType;
+  leaveTrace?: boolean;
```
hikchoi, line 133: heads up that there will be significant changes to the configurations and we will be introducing a lot of changes to it incrementally for the next few weeks.

I have written a config convention guide here: https://github.com/dendronhq/dendron-site/pull/196/files

For boolean configs in particular, we want to make it a required property of the config namespace so that defaults are explicit in the configuration and we don't have to take account for assumed defaults in the implementation.
This helps the next person down the line who has to read the code have a clear understanding of what behavior is expected.

On naming, we want to try limiting our selection of verbs in the transitive verb form names of configs, but in this case it's a bit awkward to name it `enableLeaveTrace` or something... so I think this is fine for now. Welcome to suggest anything better if you can come up with it after going through the convention guide 🙂

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -1120,7 +1120,14 @@ suite("NoteLookupCommand", function () {
...
         },
-        onInit: async ({ vaults, engine }) => {
+        onInit: async ({ wsRoot, vaults, engine }) => {
+          withConfig(
+            (config) => {
+              config.lookup.note.leaveTrace = true;
+              return config;
+            },
+            { wsRoot }
+          );
```
hikchoi, line 1130: can we leave the current test as-is and make a separate test case for `leaveTrace`?

hikchoi (changes requested):
Left some comments on potential improvements 🙂

jonathanyeung:
lgtm

packages/common-all/src/types/workspace.ts
Harshita-mindfire: Agreed. 
Although what if we provide a default value in config file without making it a required? The reason behind this is if this config is not present in the dendron.yml, it should not throw an error to user on initialization. 
Somewhat similar to the `lookupConfirmVaultOnCreate`. Even though the default value is present on init, it is not a required config. What do you think?

hikchoi: My reasoning for having all configurations that have bounded choices is because it is clear what it does. This is true for configs that are enums as well.

One anecdotal case is `selectionType` of lookup. This used to be either `selectionExtract` or `selection2link`. As a user that has not seen how this is implemented, and without reading the documentation, what do you think omitting this setting will do?

Putting myself in a moderately engaged user's standpoint I still would think omitting this setting will mean the selections will not affect how lookup is done. That was not the case until I added a `none` case. Even for an experience user and a developer, I often feel like I don't know what a config is doing unless I've been using that a lot or have worked on the area related to it.

Having documentation that explicitly states what it does is great and we strive to do so, but even with a best intent we sometimes forget about documentation or wrongly assume how things work. If we don't have control over what the user will actually do here (decided to read the docs), we should have safety measures. and my argument is being explicit and letting the user know something is wrong on init is a better way of handling this.

Your reasoning is valid, but I think it complicates the behavior in a way that it's hard to expect what's going to happen if you don't have prior knowledge of how Dendron is working internally. I would lean towards properly handling errors on init if a required property is missing (this right now is done by validating the json schema, but this is something a user can choose not to do). _Providing accurate and helpful errors when things go wrong is a better UX than allowing loosely defined behavior and keeping the user happy (but oblivious)_ is my argument here.

Another concern is consistency. If we allow one config to be optional for the sake of preventing a potential error, we are opening the gates to all the configs that have bounded choices to be optional. Of those, some will consequently not be optional sometimes. the expectations don't match for all config keys. This is a problem for the user but also a problem for us (and all future contributors) developing and adding new configs.

`lookupConfirmVaultOnCreate` will be required moving forward, btw (in fact, almost all booleans will be: https://github.com/dendronhq/dendron/pull/1295#issuecomment-919140938)
image

Harshita-mindfire: Very well thought. Thanks :)

hikchoi (approved):
Changes look good 👍🏼

PR #1375, opened 2021-09-21
fix/for-preview-links-to-navigate-within-the-same-page-in-the-note

nickolay-kondratyev:
fix/for-preview-links-to-navigate-within-the-same-page-in-the-note

## General
This makes the preview links act as navigators for the note, BUT it does not currently navigate within the preview https://www.loom.com/share/24f3f11ac94442e6aacf46dea3e1593c 

### Quality Assurance

### DIFF
--- a/packages/common-all/src/types/typesv2.ts
+++ b/packages/common-all/src/types/typesv2.ts
@@ -612,6 +612,7 @@ export enum CalendarViewMessageType {
 export enum NoteViewMessageType {
   "onClick" = "onClick",
   "onGetActiveEditor" = "onGetActiveEditor",
+  "messageDispatcherReady" = "messageDispatcherReady",
 }
 
 export enum ThemeMessageType {
--- a/packages/plugin-core/src/commands/ShowPreviewV2.ts
+++ b/packages/plugin-core/src/commands/ShowPreviewV2.ts
@@ -22,6 +22,15 @@ type CommandOutput = any;
 
 const title = "Dendron Preview";
 
+export const extractAnchorIfExists = (link: string) => {
+  if (link.indexOf("#") === -1) {
+    return undefined;
+  } else {
+    const tokens = link.split("#");
+    return tokens[tokens.length - 1];
+  }
+};
+
 const tryGetNoteFromDocument = (
   document: vscode.TextDocument
 ): NoteProps | undefined => {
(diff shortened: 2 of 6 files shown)


kevinslin (changes requested):
1.  is this working for you? when I try to do this locally, I run into an exception thrown by [assertUnreachable](https://github.com/dendronhq/dendron/blob/1ec313e0d38e3353685d859de4bdcb91380217f7/packages/plugin-core/src/commands/ShowPreviewV2.ts#L219-L219)


2.  for any manual test (eg. clicking on the preview link), can you also add an example in `test-workspace`?

see [[PR Checklist|dendron://dendron.dendron-site/dendron.dev.pull-request.checklist#^9jtWc6ov340p]]

nickolay-kondratyev:
Weird that its reaching that assert for you. Looking at code diff the assert is in the outer switch statement which is being modified by this change. Can you reproduce the failure without this commit? 

Here is the change working in test workspace:
https://www.loom.com/share/3c53d2ff4adb43d590ddc4f2b4f05897

kevinslin:
> Weird that its reaching that assert for you. Looking at code diff the assert is in the outer switch statement which is not being modified by this change. Can you reproduce the failure without this commit?
> 
> Here is the change working in test workspace:
> https://www.loom.com/share/3c53d2ff4adb43d590ddc4f2b4f05897

1. What version of VSCode are you using? I'm using 1.60.2 and am getting this issue -> https://www.loom.com/share/3c53d2ff4adb43d590ddc4f2b4f05897

After some digging it looks like regular preview from master also has this issue. If you're not on 1.60.2, can you upgrade and see if you can repro?

 2. Can you make sure to create a manual test in `test-workspace` as part of this commit?
3. no worries about current commits but moving forward,  can you make sure prs  start with `[feat|fix|enhance|]/[{description-of-pr-in-kebab-case}]`
        - `eg. feat/add-thisthing`

nickolay-kondratyev:
1. The loom is a copy of my loom but I did run into assertUnreachable on the master branch as well (looking at the console output) the current commit added a case statement for the new message type so if thats the one you ran into before it should be gone now.

2. Added manual test. [[dendron.links.heading-anchors]] Also in the last commit added a change for the preview to navigate to heading anchor from a different page (that wasn't in as well). 

3. modified the current PR to start with fix/...

kevinslin:
>     1. The loom is a copy of my loom but I did run into assertUnreachable on the master branch as well as you have mentioned (looking at the console output) the current commit added a case statement for the new message type so if thats the one you ran into before it should be gone now.
> 
>     2. Added manual test. [[dendron.links.heading-anchors]]
> 
>     3. modified the current PR to start with fix/...
> 
> 
> Also in the last commit added a change for the preview to navigate to heading anchor from a different page (that wasn't a functionality before as well).
> 
> To note: preview navigation will not work correctly for notes with non standards ids such as notes in `test-workspace` that have ids that are not alphanumeric and are not guidlike since regex will not match ids like 'links.headers' due to it not being alphanumeric.

left comment in https://github.com/dendronhq/dendron/pull/1385

kevinslin (approved):

PR #1381, opened 2021-09-22
enhance(publish): support twitter card

kevinslin:


packages/common-all/src/utils.ts
```diff
@@ -1,10 +1,11 @@
-import minimatch from "minimatch";
-import semver from "semver";
 import GithubSlugger from "github-slugger";
-import querystring from "querystring";
 import _ from "lodash";
+import minimatch from "minimatch";
+import querystring from "querystring";
```
viztor, line 4: I think I got warned installing that `querystring` is [deprecated](https://www.npmjs.com/package/querystring)

jonathanyeung (approved):

PR #1382, opened 2021-09-22
chore: `npx dev` will not work

viztor:


packages/dendron-cli/src/commands/publishCLICommand.ts
```diff
@@ -211,7 +211,7 @@ export class PublishCLICommand extends CLICommand<CommandOpts, CommandOutput> {
       overrides,
     });
     this.print(
-      `to preview changes, run "cd ${getNextRoot(wsRoot)} && npx dev"`
+      `to preview changes, run "cd ${getNextRoot(wsRoot)} && npx next dev"`
```
viztor, line 214: might have to be `cd` `yarn` `yarn dev` since we do use external packages.

kevinslin (approved):

PR #1395, opened 2021-09-23
fix: add /images/ directory to logoUrl so the Dendron logo can be displayed

nwehner:
```markdown
# fix: Add /images/ directory to logoUrl so that the Dendron logo can be displayed

Very small change to the `DendronLogoOrTitle` component, to add the images directory to the logoURL base directory.

The original base directory included only `assets`. However, since all logos should be images, it makes sense to have logos stored in the `images` subdirectory.

Would be interested to know if others store their logos in the images directory, too. I'm on Discord at @nwehner. 
```

### DIFF
--- a/packages/nextjs-template/components/DendronLogoOrTitle.tsx
+++ b/packages/nextjs-template/components/DendronLogoOrTitle.tsx
@@ -25,7 +25,9 @@ export default function DendronLogoOrTitle() {
       >
         {engine.config?.site.logo ? (
           <Logo
-            logoUrl={"/assets/" + path.basename(engine.config?.site.logo)}
+            logoUrl={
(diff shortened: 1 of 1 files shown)


kevinslin (changes requested):
this is probably what we should have started off with but introducing this here might introduce backwards compatibility issues. can we do a fallback to the old behavior if no file is found?

nwehner:
Since we don't have access to `fs` in nextjs-template, we can't check for the existence of the logo. It looks like it would make sense to reject the PR, and move that functionality all into NextjsExportPod. We should add an export function from there that'll resolve the proper directory for nextjs-template. It seems the logo directory is resolved uniquely in each module at the moment.

PR #1406, opened 2021-09-24
fix: notes added outside Dendron are missed

SeriousBug:
Turns out, `sentryReportingCallback` drops the `this` value for wrapped functions. This becomes important for methods that use the `this` value, which was the case for the `FileWatcher`. This would cause these functions to fail immediately, and thus miss added or deleted notes.

I fixed the issue and added a warning for the future.


kevinslin (commented):
good catch!

can you start an entry about logging in `dendron.dev.style`? we can incrementally create our best practices for logging, starting with the callout about the sentry issue

PR #1407, opened 2021-09-24
feat(publish): mobile navigation

viztor:
a working, albeit not as ideal implementation of mobile nav simulating existing 11ty template without changing much of the laylout.

### DIFF
--- a/packages/nextjs-template/components/DendronLayout.tsx
+++ b/packages/nextjs-template/components/DendronLayout.tsx
@@ -43,7 +43,7 @@ export default function DendronLayout(
     >
       <DendronTreeMenu
         {...props}
-        collapsed={isCollapsed}
+        collapsed={isCollapsed && isResponsive}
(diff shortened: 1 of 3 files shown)


kevinslin (changes requested):
awesome new design! one issue I found at certain screen size is that the hamburger button doesn't make it to the header. see https://www.loom.com/share/99daa50bedf54123ab8e9333c70cd812

viztor:
> awesome new design! one issue I found at certain screen size is that the hamburger button doesn't make it to the header. see https://www.loom.com/share/99daa50bedf54123ab8e9333c70cd812

I did some last minute fix without realizing the side effect for the side bar. I'm having difficulty reproduce the hamburger issue though, can you pull and check if the issue still persist?

kevinslin:
that fixed it!

kevinslin (approved):

PR #1409, opened 2021-09-24
feat: initial survey

hikchoi:
# feat: initial survey
This PR:
- Adds an initial survey that will be used to further customize the onboarding experience

DOCS: https://github.com/dendronhq/dendron-site/pull/212

# Pull Request Checklist

### Quality Assurance

#### Special Cases

### Docs
- [~] If your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR

### Analytics

### DIFF
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -204,7 +204,7 @@ export class SurveyUtils {
             AnalyticsUtils.track(SurveyEvents.InitialSurveyAccepted, {
               answerCount,
             });
-            vscode.window.showInformationMessage("Survey submitted.");
+            vscode.window.showInformationMessage("Survey submitted! Thanks for helping us make Dendron better 🌱");
           } else {
             vscode.window.showInformationMessage("Survey cancelled.");
             AnalyticsUtils.track(SurveyEvents.InitialSurveyRejected);
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -182,8 +182,8 @@ export class SurveyUtils {
       AnalyticsUtils.track(SurveyEvents.InitialSurveyPrompted);
       vscode.window
         .showInformationMessage(
-          "Would you like to tell us a bit about yourself?",
-          { modal: true , detail: "This info will be used to provide a better onboarding experience. This will take less than a minute to complete."},
+          "Welcome to Dendron!",
+          { modal: true , detail: "Would you like to tell us a bit about yourself? This info will be used to provide a better onboarding experience. This will take less than a minute to complete."},
           { title: "Proceed" }
         )
         .then(async (resp) => {
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -204,7 +204,9 @@ export class SurveyUtils {
             AnalyticsUtils.track(SurveyEvents.InitialSurveyAccepted, {
               answerCount,
             });
+            vscode.window.showInformationMessage("Survey submitted.");
           } else {
+            vscode.window.showInformationMessage("Survey cancelled.");
             AnalyticsUtils.track(SurveyEvents.InitialSurveyRejected);
           }
         })
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -2,6 +2,7 @@ import { SurveyEvents } from "@dendronhq/common-all";
 import { AnalyticsUtils } from "./utils/analytics";
 import * as vscode from "vscode";
 import _ from "lodash";
+import { Logger } from "./logger";
 
 export class DendronSurvey {
   choices: readonly vscode.QuickPickItem[];
@@ -142,8 +143,8 @@ export class SurveyUtils {
       AnalyticsUtils.track(SurveyEvents.InitialSurveyPrompted);
       vscode.window
         .showInformationMessage(
-          "Would you like to tell us about yourself so that we can provide a better onboarding experience?",
-          { modal: true },
+          "Would you like to tell us a bit about yourself?",
+          { modal: true , detail: "This info will be used to provide a better onboarding experience"},
           { title: "Proceed" }
         )
         .then(async (resp) => {
@@ -167,6 +168,10 @@ export class SurveyUtils {
           } else {
             AnalyticsUtils.track(SurveyEvents.InitialSurveyRejected);
           }
+        })
+        // @ts-ignore
+        .catch((error: any) => {
+          Logger.error({msg: error});
         });
     }
   }
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -3,6 +3,8 @@ import { AnalyticsUtils } from "./utils/analytics";
 import * as vscode from "vscode";
 import _ from "lodash";
 import { Logger } from "./logger";
+import { StateService } from "./services/stateService";
+import { GLOBAL_STATE } from "./constants";
 
 export class DendronSurvey {
   choices: readonly vscode.QuickPickItem[];
@@ -204,6 +206,10 @@ export class SurveyUtils {
             AnalyticsUtils.track(SurveyEvents.InitialSurveyAccepted, {
               answerCount,
             });
+            StateService.instance().updateGlobalState(
+              GLOBAL_STATE.INITIAL_SURVEY_SUBMITTED,
+              true
+            );
             vscode.window.showInformationMessage("Survey submitted! Thanks for helping us make Dendron better 🌱");
           } else {
             vscode.window.showInformationMessage("Survey cancelled.");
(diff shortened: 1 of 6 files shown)


packages/plugin-core/src/survey.ts
```diff
@@ -0,0 +1,173 @@
...
+   * Flip a coin to randomly prompt initial survey.
+   * Asks three questions about background, use case, and prior tools used.
+   * @param forcePrompt skip flipping a coin and force prompting.
+   */
+  static async maybePromptInitialSurvey(forcePrompt?: boolean) {
+    const shouldPrompt = forcePrompt ? true : !!Math.floor(Math.random() * 2);
+    if (shouldPrompt) {
+      AnalyticsUtils.track(SurveyEvents.InitialSurveyPrompted);
+      vscode.window
+        .showInformationMessage(
```
kevinslin, line 144: instead of using an information message, can we show a full on modal after the user clicks showTutorial?

kevinslin (changes requested):
Instead of using an information message, can we show a full on modal after the user clicks showTutorial?

Flow:
- user clicks "Getting Started"
- Launch workspace
- Show survey with modal:
    1. What is your background (this helps us give you the right context for the features we’ll be talking about)
        * Choices (check all that apply): 
            * Software Developer
            * Manager
            * Student
            * Technical Writer
            * Dev Ops
            * Other
    2. Tell us what you want to use Dendron for (this helps us tailor the material to what’s relevant for you)
        * Choices (check all):
            * Team knowledge base
            * Publishing
            * Snippet manager
            * Todos and Agenda
            * Meeting notes
            * Research
            * Mobile note taking
            * Other
    3. Are you coming from an existing tool(s)? (we have pods to help you import your data)
        - Choices:
            - obsidian
            - ...
- After we get results, proceed with regular tutorial flow

packages/plugin-core/src/survey.ts
```diff
@@ -0,0 +1,173 @@
...
+      results: results.map((result) => result.label),
+    });
+  }
+
+  onReject() {
+    AnalyticsUtils.track(SurveyEvents.UseCaseRejected);
+  }
+
+  static create() {
+    const title = "Tell us what you want to use Dendron for";
```
jonathanyeung, line 92: nit: maybe for extra clarity that they can select multiple, and maybe drop 'tell us': 

"How would you like to use Dendron? Select all that apply"

Also - it'd be good to maybe add into the title or somewhere to show the progress - that they're on question 1/3, 2/3, 3/3 - that way, they know it's very short and that they'll be done in < 1 minute.

packages/plugin-core/src/survey.ts
```diff
@@ -0,0 +1,173 @@
...
+    const choices = [
+      { label: "No" },
+      { label: "Foam" },
+      { label: "Roam" },
+      { label: "Logseq" },
+      { label: "Notion" },
+      { label: "OneNote" },
+      { label: "Obsidian" },
+      { label: "Evernote" },
+      { label: "Google Keep" },
```
jonathanyeung, line 127: Add Other option here too?

packages/plugin-core/src/survey.ts
jonathanyeung: @kevinslin  - I think this is already set with `modal: true` prop below?

hikchoi:
> Instead of using an information message, can we show a full on modal after the user clicks showTutorial?

What do you mean by full on modal? Do you mean something like this? 
image

AFAIK, VSCode doesn't expose an API to the extension authors yet.

packages/plugin-core/src/survey.ts
hikchoi: Good point. will change

hikchoi: It's pretty clear that you can select multiples though.
image

packages/plugin-core/src/survey.ts
hikchoi: yeah that's a bit tricky to do with the current quickpick. especially if we are allowing `canSelectMany`. If we want to have "other", then we should have another quick input chained after this

hikchoi: I've added a check in `onAnswer` to see if the user checked `Other`, and made it prompt a quick input to manually type in an answer. Looks a bit hacky but I think this is good enough for now.

hikchoi:
UPDATE:

I've made some changes suggested, but there isn't a native way to show a custom modal dialog at this point.

That is why I opted for showing a modal information message, which will start a chain of quickselects that contain survey questions.

The only other option here would be coming up with a markdown based survey similar to the tutorial, but I don't think that's any better than the solution as-is.

hikchoi:
I'll add some screenshots to illustrate how it will look.

When the user clicks on the big green button, the following steps will show

1. Ask if they want to answer some surveys
image

- I wish we can change the icon here, because it _really_ looks like VSCode is giving you a warning instead of us.
- I have decided to have a short "welcome to dendron" message that follows with a detail message to better convey that this message is coming from us.

2. q1
image

- You can multiselect all that applies here. including _other_.
- If you select other, a `QuickInput` like the following will show
image

3. q2 and q3

4. Final notification that it was submitted through a toast.
image

hikchoi:
VSCode doesn't give you direct access to the UI DOM, but there _is_ a way to create custom dialogs.
Only catch is that you have to spawn a separate Electron app, which can then communicate with the extension using ipc. (sounds crazy but... https://marketplace.visualstudio.com/items?itemName=anseki.vscode-color)
Absolutely not worth the effort in my opinion, so we are stuck waiting for the vscode team to expose an API for the modal dialog if we want a better UI with surveys.

packages/plugin-core/src/survey.ts
jonathanyeung: Cool - in that case, no need.

jonathanyeung:
> 1. Ask if they want to answer some surveys
>    image

IMO this looks fine - only nitpick: Instead of 'cancel' do we have the option to add our own text?  Basically, to better indicate that if they opt not to do the quick survey, then they'll just be taken to the rest of the initialize workspace / getting started workflow.  Here, I feel like a user may be afraid that clicking cancel will just take them back to the previous webview screen (with the green button).

hikchoi:
> IMO this looks fine - only nitpick: Instead of 'cancel' do we have the option to add our own text? Basically, to better indicate that if they opt not to do the quick survey, then they'll just be taken to the rest of the initialize workspace / getting started workflow. Here, I feel like a user may be afraid that clicking cancel will just take them back to the previous webview screen (with the green button).

Agreed. Maybe something like `Skip Survey` and `Proceed`?

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -97,6 +99,15 @@ export class TutorialInitializer
       WORKSPACE_ACTIVATION_CONTEXT.NORMAL
     );
 
+    const extensionInstallStatus = VSCodeUtils.getInstallStatusForExtension({
+      previousGlobalVersion: StateService.instance().getGlobalVersion(),
+      currentVersion: DendronExtension.version(),
+    });
+
+    if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
```
kevinslin, line 107: this will never initialize because we set the global version in [_extension](https://github.com/dendronhq/dendron/blob/c5a98e25dd442f540eec37333e89e3df24d6b466/packages/plugin-core/src/_extension.ts#L573-L573).

can we just launch the survey everytime the tutorial is launched? we can update [globalState](https://github.com/dendronhq/dendron/blob/f57810222a8d6bedcc1054bafc887f25fbfa9cd8/packages/plugin-core/src/services/stateService.ts#L91-L91) to save if the user has finished the survey

packages/plugin-core/src/survey.ts
```diff
@@ -0,0 +1,219 @@
...
+
+            const answerCount = [
+              backgroundResults,
+              useCaseResults,
+              priorToolsResults,
+            ].filter((value) => !_.isUndefined(value)).length;
+            AnalyticsUtils.track(SurveyEvents.InitialSurveyAccepted, {
+              answerCount,
+            });
+            vscode.window.showInformationMessage("Survey submitted.");
```
kevinslin, line 207: add a bit more character in this message -> Survey submitted! Thanks for helping us make Dendron better 🌱

kevinslin (changes requested):

packages/plugin-core/src/survey.ts
hikchoi: I did clean up the message and add steps / total as you suggested.

packages/plugin-core/src/workspace/tutorialInitializer.ts
hikchoi: added a new global state key to check for this.

packages/plugin-core/src/survey.ts
hikchoi: Changed as suggested (and some more in other places)

hikchoi:
> this looks good - last thing. can we add a test for the survey changes?

Since this is triggered when a new tutorial workspace is spawned, there isn't  a native way to test if it works there.

I'll add some basic tests for the individual survey classes though.

kevinslin (approved):

PR #1410, opened 2021-09-24
fix(publish): fix links in note reference

kevinslin:
Links resolving with `.html` suffix which is no longer used in nextjs


hikchoi (approved):
LGTM

PR #1412, opened 2021-09-24
feat: Support Article Based Seo Props

nwehner:
Adds Published, Created, and Tags to each site's SEO metatags. 

I removed the Jest tests from this branch since those will need to be moved over to common-frontend, probably.

### DIFF
--- a/packages/nextjs-template/components/DendronSEO.tsx
+++ b/packages/nextjs-template/components/DendronSEO.tsx
@@ -3,7 +3,6 @@ import {
   DendronSiteConfig,
   NoteProps,
   SEOProps,
-  NoteUtils,
   Time,
   PublishUtils,
 } from "@dendronhq/common-all";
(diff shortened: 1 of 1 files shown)


kevinslin:
for the jest tests, were we not able to get them to run in `nextjs-template`?

nwehner:
There are more bugs in the way to getting testing running in `nextjs-template` than I thought, but the persistent one is that NextJS wants to transform everything through Babel, whereas TS-Jest wants to do its own transformations. I wasn't able to figure out (yet) how to get TS-Jest to use NextJS's Babel-ing. I was mostly having issues around getting the vDom setup, but I got that working. 

What I think would work best because of the Babel issues with TS-Jest is to put all our front-end tests together in a common repo (like in the `engine-test-utils` package, which has a nice selection of tests for the front-end already) where things are already working as expected with the vDom. 

Current issue is that the `nextjs-template` imports from `next/router` without ever requiring it in the dependencies :'( -- current build process only works because `common-frontend` has the module as a dependency in the package-lock.son (not actual package.json dep, either) so I have to learn how to add that module officially to the stack.

kevinslin:
got it. we'll add test as a separate pr then. for `next/router` that should be part of the next package

kevinslin (approved):

PR #1441, opened 2021-09-29
enhance/allow-creation-of-duplicate-file-name-if-there-is-a-vault-without-such-file-name

nickolay-kondratyev:
# enhance/allow-creation-of-duplicate-file-name-if-there-is-a-vault-without-such-file-name

# Note
**Note there was a weird scenario where note was created in different vault but was not indexed correctly**, reload fixed the issue. It looked like: 
Screen_Shot_2021-09-29_at_3 01 59_PM
However, not able to reproduce this right now. 

### Quality Assurance
- [ ] add a [test](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#writing-tests) for the new feature
https://www.loom.com/share/cd2b8ee8cf6e47fdafde6fee20dfd85b

### DIFF
--- a/packages/plugin-core/src/commands/NoteLookupCommand.ts
+++ b/packages/plugin-core/src/commands/NoteLookupCommand.ts
@@ -3,6 +3,7 @@ import {
   DVault,
   ERROR_STATUS,
   ErrorFactory,
+  ErrorMessages,
   NoteLookupConfig,
   NoteProps,
   NoteQuickInput,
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -429,6 +430,57 @@ export class NoteLookupCommand extends BaseCommand<
...
+        } else {
+          vault = promptedVault;
+        }
+      } else if (availVaults.length === 1) {
+        // There is only a single vault that is available so we dont have to ask the user.
+        vault = availVaults[0];
+      } else {
+        // We should never reach this as "Create New" should not be available as option
+        // to the user when there are no available vaults.
+        ErrorFactory.createInvalidStateError({
```
kevinslin, line 475: we are creating an error here but not returning it. should we log/return this error?

kevinslin (changes requested):

packages/plugin-core/src/commands/NoteLookupCommand.ts
nickolay-kondratyev: Yep, that's a miss.

Since it should never occur and represents a bug, sounds like a good case to throw the error so we can be notified of the issue quicker.

kevinslin (approved):

PR #1446, opened 2021-09-30
feat: Lapsed user survey

hikchoi:
# feat: Lapsed user survey

This PR adds a user survey that is prompted after 7 days of being inactive.

DOCS: https://github.com/dendronhq/dendron-site/pull/214

---

# Pull Request Checklist

### Quality Assurance

#### Special Cases

### Docs

### Analytics

### DIFF
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -229,7 +229,6 @@ export class LapsedUserReasonSurvey extends DendronQuickPickSurvey {
         break;
       }
       case "I've encountered a bug which stopped me from using Dendron.": {
-        // could you describe what happened?
         reason = "bug";
         extra = await vscode.window.showInputBox({
           ignoreFocusOut: true,
@@ -425,7 +424,6 @@ export class SurveyUtils {
         "Could you share some feedback to help us improve?",
         { modal: true },
         { title: "Proceed" },
-        { title: "Skip Survey"},
       ).then(async (resp) => {
         if (resp?.title === "Proceed") {
           const reasonSurvey = LapsedUserReasonSurvey.create();
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -5,14 +5,55 @@ import _ from "lodash";
 import { Logger } from "./logger";
 import { StateService } from "./services/stateService";
 import { GLOBAL_STATE } from "./constants";
+import { resolve } from "path";
 
-export class DendronSurvey {
+export class DendronQuickInputSurvey {
+  opts: {
+    title: string;
+    ignoreFocusOut: boolean;
+    placeHolder?: string;
+    prompt?: string;
+  }
+
+  constructor(opts: {
+    title: string;
+    placeHolder?: string;
+    prompt?: string;
+  }) {
+    this.opts = {...opts, ignoreFocusOut: true}; 
+  }
+
+  async onAnswer(_opts: any): Promise<void> {
+    return undefined;
+  }
+
+  onReject(_opts?: any): void {
+    return undefined;
+  }
+
+  async show(step: number, total: number) {
+    const progress = `Step ${step} of ${total}`;
+    const title = this.opts.title;
+    const showOpts = {
+      ...this.opts,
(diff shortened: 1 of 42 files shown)


packages/plugin-core/src/survey.ts
```diff
@@ -223,4 +417,64 @@ export class SurveyUtils {
...
   }
+
+  static async showLapsedUserSurvey() {
+    AnalyticsUtils.track(SurveyEvents.InitialSurveyPrompted);
+    await vscode.window
+      .showInformationMessage(
+        "Could you share some feedback to help us improve?",
+        { modal: true },
+        { title: "Proceed" },
+        { title: "Skip Survey"},
```
kevinslin, line 428: since the survey already has a cancel button, can we just use that?

packages/plugin-core/src/survey.ts
```diff
@@ -172,6 +214,158 @@ export class PriorToolsSurvey extends DendronSurvey {
...
+      { label: "I've encountered a bug which stopped me from using Dendron." },
+      { label: "I found a different tool that suits me better." },
+      { label: "Other" },
+    ];
+    return new LapsedUserReasonSurvey({ title, choices, canPickMany: false });
+  } 
+}
+
+export class LapsedUserOnboardingSurvey extends DendronQuickPickSurvey {
+  CALENDLY_URL = "https://calendly.com/d/mqtk-rf7q/onboard";
```
kevinslin, line 291: we talked about giving folks the option to go to the getting started guide as an alternative to the one on one survey. is it worhtwile adding a multi-select so users can select one or both methods here?

kevinslin (changes requested):
looks good, some minor comments

packages/plugin-core/src/survey.ts
hikchoi: Forgot to point out some decisions I made here:

So this survey shows up when the lapsed user message is displayed, and then rejected. I think it would be a bit weird if we ask them here again when they already rejected the initial prompt like a minute ago.

packages/plugin-core/src/survey.ts
hikchoi: [this](https://github.com/dendronhq/dendron/pull/1409#issuecomment-927774529) is the reason I went with this, but the original intention was to _replace_ cancel with _Skip Survey_. I don't think we can remove the cancel button though. 

I'll try to figure something out.

hikchoi: actually, just using the cancel button should be fine here.

hikchoi:
Some design decisions made:

1. We initially talked about linking the getting started page, but the flow is a bit awkward here.

When the lapsed user message is prompted, we are already asking the users if they want to get started. At that point if the user rejects the modal, we ask them if they want to do a survey. Asking if they want to get started again would be a bit weird.

I think for the users who answer _I am not sure how to get started._, we can add a follow-up prompt to show them the get started page, but for others I think not asking again sounds like a safer approach

2. follow-up prompt for _I found a different tool that suits me better_.

Here I worded the follow-up to be focused on Dendron rather than the other tool: instead of _Why did you go with the other tool?_, asking _What feature was missing in Dendron for your use case?_

3. _I've encountered a bug which stopped me from using Dendron._

We discussed that asking for an issue submission here is demanding quite a bit of activation energy. I opted for just a simple description which we can use to troubleshoot issues ourselves. If we stick to this, I think it's worth closely monitoring this on a daily (or shorter) basis.

4. Deferring all external link opens.

I figured opening external links (link to onboarding / discord etc) right when the users choose the option will break the flow of the survey, so I instead added a modal message that tells the users that it will open once the survey is finished.

hikchoi:
Some visual aid to help see what it'll look like

The original lapsed user message will be changed to this:
image

Hitting cancel here will prompt the lapsed user survey

image

First question: 
image

First two options don't have follow-up prompts.

Follow-up for third option:
image

Follow-up for fourth option:
image




Second question:
image

A modal will let you know if you say yes
image

Third question:
image

Last question:
image

After this stage, any external link that the user selected will open.

kevinslin (approved):

PR #1447, opened 2021-09-30
Feat: check for logo location

nwehner:
A few fixes to NextJsExportPod:

- Cleanup FS function imports
- Append vars to any existing file with "a" flag
- Checks `/assets`, `/assets/images`, and `<assetsPrefix>/assets` for the logo, saving path to `process.env.DENDRON_LOGO_PATH`
- **Still needs tests added** to `engine-test-utils/src/__tests__/pods-core/NextjsExportPod.spec.ts`


kevinslin (changes requested):
nice fix! could you also add tests for this at https://github.com/dendronhq/dendron/blob/enhance/airtable-pod/packages/engine-test-utils/src/__tests__/pods-core/NextjsExportPod.spec.ts#L62:L62?

Issue #440, opened 2021-01-03
Document Display Order in the Dendron Tree View


micharris42:
**Describe the bug**
In the Dendron Tree View, documents are displayed in order of creation, not by the document title or by the content of the frontmatter. This is especially noticeable for journal notes. If a set of journal notes are created out of chronological order, they will show up in the Dendron Tree View in the order of their creation, rather then ordered by date. Amending the "created" field in the frontmatter has no effect.

**To Reproduce**
1. Create a journal note.
2. Create a second journal note for an earlier date. 
3. The second note will be displayed below the first note, despite that it is for an earlier date in time.

**Expected behavior**
That documents be displayed in the Dendron Tree View ordered by either the document title or the content of the frontmatter, not by date of creation. If this is the intended behavior, that there be an option to sort the content of the Dendron Tree View.

**Screenshots**
![Capture](https://user-images.githubusercontent.com/76890677/103470410-0d38bd00-4d40-11eb-9377-30bc9e5d346e.PNG)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: NA
 - Version: Dendron 0.22.0

**Additional context**
I apologize if this has already been addressed in an earlier issue. However, after going through the Dendron documentation and the issues already on GitHub, I was unable to find a solution. Any suggestions would be appreciated. Thank you.

flammehawk:
I think a setting for the sort order would be preferable. 

I personaly would like to have my notes sorted alphabetical.

But journal notes in a way that the journal entry for the newest day is displayed first. 

Which is opinionated others may want to have the inverse from what I want so I think a setting would be the best choice.

kevinslin:
Thanks for the issue and you're the first one to bring it up so we're good. Formally including this in our backlog

aleksey-rowan:
Many note-taking apps allow for manual sorting of notes/containers and often don't have functionality for auto-sorting. I would prefer my notes auto-sorted by metadata title with an option to manually adjust their positions if needed. There could be an additional metadata property (`sort-title` for example) to override the actual title for sorting purposes, like when sorting books or movies, so all books that start with "the" don't get grouped together.

kevinslin:
[...]

PR #1459, opened 2021-10-03
fix: tree view order

SeriousBug:
Made next-server use the `common-frontend` tree sorting code,
and replaced that code to be a little simpler.

#440


kevinslin (changes requested):
can we also add some tests for this? (we actually have testing for `common-frontend` in our test package [here](https://github.com/dendronhq/dendron/blob/master/packages/engine-test-utils/src/__tests__/common-frontend/features/engine/slice.spec.tsx#L8:L8)

SeriousBug:
Thanks, I didn't notice we had tests for it. The tree view component is under `dendron-next-server` which we don't have tests for right now I think, so I instead added a test that just tests that the tree view data is constructed in the right order.

PR #1473, opened 2021-10-06
enhance: allow wiki links to be looked up

nickolay-kondratyev:
# enhance: allow wiki links to be looked up
This change allows wiki links such as [[dendron.welcome]] to be looked up within note lookup. 

# Details 
The following types of wiki links can now be looked up in note look up: 

* [[with description with header with vault|dendron://vault/dendron.welcome#header1]]
* [[with description with header|dendron.welcome#header1]]
* [[with description|dendron.welcome]]
* [[dendron.welcome]]

### Quality Assurance
```
  LookupProviderV3 utility methods:
    transformQueryString tests:
      WHEN given simple string with slashes
        ✓ THEN convert to query string to dots
        ✓ THEN wasMadeFromWikiLink is false
      WHEN given simple string with dots
        ✓ THEN value stays as is
        ✓ THEN wasMadeFromWikiLink is false
      WHEN given string with OR operator
        ✓ THEN value stays as is
        ✓ THEN wasMadeFromWikiLink is false
      WHEN given string with wiki link without description
        ✓ THEN strip out wiki link decoration
        ✓ THEN wasMadeFromWikiLink is true
      WHEN given string with wiki link with description
        ✓ THEN strip out wiki link decoration and description
        ✓ THEN wasMadeFromWikiLink is true
      WHEN given string fully qualified with wiki link with description
        ✓ THEN strip out wiki link decoration and description
        ✓ THEN wasMadeFromWikiLink is true
      WHEN given string fully qualified with wiki link with description and header
        ✓ THEN strip out wiki link decoration and description and header
        ✓ THEN wasMadeFromWikiLink is true
```
   - Added `dendron.note-lookup` file for note look up manual tests. 


nickolay-kondratyev:
Space trimming appears to be quite straightforward to add in the current code path, so added that to this review. As added benefit, if users now type `  some-note ` and use create new note, we will clean up the side space and create `some-note` without the extra spaces on the side.

> Instead of ignoring the header in the wikilink, could we go to that header if it exists, like what would happen if we used the GotoNote command?

The challenge here is that we are inside the lookup provider and it doesn't seem right to be calling the commands directly from the look up provider. So will leave adding header navigation to the future.

PR #1482, opened 2021-10-08
feat: native workspaces

SeriousBug:
This PR brings feature parity for native workspaces, and allows them to be used in production.

I tested feature parity by temporarily changing the default workspace type for tests to Native and running the tests. Some tests that have to run as Code workspaces have the configuration set now. As a result, all plugin tests (that use `runLegacy*WorkspaceTest`) can be run under a Native workspace now, but I chose to keep Code workspaces as the default.

I also did some manual testing by creating notes, adding vaults, and using Goto commands, and everything seems to function correctly. I would recommend some more manual testing by others to catch anything I may have missed.

**Important caveat**: I remember it being discussed at some point that users should be able to drop in a `dendron.yml` into a folder and start using it. This is currently only possible if the user also adds that folder as a workspace folder, otherwise they must place the `dendron.yml` at the root folder.

### DIFF
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -300,7 +300,7 @@ export class WorkspaceService {
     if (!noAddToConfig) {
       await this.addVault({ ...opts, updateWorkspace: false });
     }
-    if (opts.addToCodeWorkspace) {
+    if (opts.addToCodeWorkspace || opts.updateWorkspace) {
       await this.addVaultToCodeWorkspace(vault);
     }
     return vault;
--- a/packages/engine-server/src/workspace/utils.ts
+++ b/packages/engine-server/src/workspace/utils.ts
@@ -6,7 +6,6 @@ import {
   DVault,
   DWorkspaceV2,
   getSlugger,
-  getStage,
   isBlockAnchor,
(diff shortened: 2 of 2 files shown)


kevinslin:
Can you add  docs for this  here -> [[Workspace|dendron.ref.workspace]]?

Some additional followups (might require discussion):
- command to create a native workspace (eg. should this be added to "intialize workspace" cmd or should it be a new command)?
- how to handle multiple `dendron.yml` files (currently default to picking the first one, would be nice if we can create an uber workspace - we already have the concept of workspace vaults)
- the hypothetical use case of adding a `dendron.yml` to `/docs`, as you pointed out, won't work unless the folder is part of a workspace
    - can we initialize dendron without being in a workspace? (eg. check if a folder that was just added has a dendron.yml and if so, index it?)
        - behavior if user adds multiple folders with `dendron.yml` files
        - behavior if user removes folders with `dendron.yml` files

For any of the ambiguous items, feel free to schedule async meetings and add any pretinent findings to [[31 Native Workspace|dendron://dendron.dendron-site/dendron.rfc.31-native-workspace]]
Use your best judgement for prioritization (if some of these tasks are not #retention related so we can defer them to post-october)

Issue #1001, opened 2021-07-19
Footnotes in host and referenced notes clash counterintuitively


SR--:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

When one note is referenced within another and has footnotes, the incoming footnotes appear outside of the text box at the bottom of the note text as if they belong to the host note. This seems incorrect and if one adds references to the host note that has an embedded reference note with footnotes, the host note's footnotes override the incoming footnotes causing havoc. Needs fixing, I think.

## To Reproduce

1. Make a note A. Type some text.
2. Make another note B with some footnotes.
3. Reference note A in note B.
4. Add some footnotes to note B.
5. The new footnotes override the footnotes embedded from note A.

## Expected behavior

Note A footnotes should apprear within the embedded box.
Note B footnotes should be independent from the embedded footnotes.

SR--:
By the way, this is in Dendrone Preview Enhanced. I now see that in v51 in Preview v2 the footnotes are in the box but they are bunched into a paragraph and have empty square brackets rendered instead of numbers.

SR--:
on footnotes. in Preview v2 in v52 the superscript numbers and and the numbers in footnotes render correctly. however, in embedded/referenced notes the footnotes are now outside of the embedded box as it used to be in Preview v1. still needs fixing, please. Also, a minor issue, can the return link be rendered in a less prominent way? a pale arrow on white before was better, it is not an important piece of information and is distracting.

SR--:
Can the footnote number not act as a return link the same way the superscript number is a link to the footnote?

SeriousBug:
The return arrow is too prominent because of the symbol that's used for it. Some fonts seem to render it more prominently with the blue rectangle. We can replace it with some other symbol to fix it. Any thoughts?

I found this list of arrow symbols, but I'm not sure what to use: https://unicode-table.com/en/sets/arrow-symbols/#harpoon-arrows

It could also be any other unicode symbol.

SR--:
Hm... In this case, should not footnotes use the same format as their inline references?

**Footnotes**

one
other

[...]

PR #1520, opened 2021-10-14
fix: footnote rendering in note references

SeriousBug:
#1001 

Integrated footnote rendering into Dendron. Footnotes are now rendered alongside Children and Tags sections at the end of the note. Additionally, note references render their footnotes inside themselves.

This replicates the previous footnote rendering (uses same HTML ids), so in case a user had a link to a reference saved, those links will still work.

There's one issue that remains: If there are footnotes that are clashing between the note and the reference, the wrong footnote will get displayed. I'll create an issue to come back and fix this later.

### DIFF
--- a/packages/dendron-next-server/styles/scss/content.scss
+++ b/packages/dendron-next-server/styles/scss/content.scss
@@ -236,3 +236,14 @@ blockquote {
     }
   }
 }
+
+// Footnotes
+a.fnref {
+	vertical-align: super;
+	font-size: 0.7rem;
+}
+
+a.fn {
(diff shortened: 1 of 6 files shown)


SeriousBug:
> Can you elaborate on this? by clashing, do you mean they point to the same link?

Huh. I was talking about this example, but it looks like it's working fine. I thought this was causing issues when I started working on the feature, but I guess it works now:



Although I realize now that I need some styling on the footnote links. I'll update once I fix that.

SeriousBug:
Fixed the styling. Here's the final look:

PR #1524, opened 2021-10-15
fix: workaround for user tags & hashtags inside links

SeriousBug:
The links:

- `[@dendronhq](https://twitter.com/dendronhq)`
- `[#dendron](https://twitter.com/hashtag/dendron)`

Would get rendered as a user tag and hashtag, respectively.

This PR is a workaround to fix the rendering. These are still parsed as a usertag and hashtag, and I'm not sure how to fix root of the problem. I think the link parser is recursively parsing it's contents, which we don't actually want in this case. I don't know any way that the hashtag/usertag parser can detect that it's the child of the link and reject parsing though.

### DIFF
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -935,6 +935,7 @@ export class RemarkUtils {
 
   static isLink(node: Node): node is Link {
     return node.type === DendronASTTypes.LINK;
+  }
 
   static isFootnoteDefinition(node: Node): node is FootnoteDefinition {
     return node.type === DendronASTTypes.FOOTNOTE_DEFINITION;


kevinslin:
> The links:
> 
>     * `[@dendronhq](https://twitter.com/dendronhq)`
> 
>     * `[#dendron](https://twitter.com/hashtag/dendron)`
> 
> 
> Would get rendered as a user tag and hashtag, respectively.
> 
> This PR is a workaround to fix the rendering. These are still parsed as a usertag and hashtag, and I'm not sure how to fix root of the problem. I think the link parser is recursively parsing it's contents, which we don't actually want in this case. I don't know any way that the hashtag/usertag parser can detect that it's the child of the link and reject parsing though.

I'm not sure about this either. I don't think the parser has context about values that are ahead of it. One "hack" we could do is add the previous processed node in `procData` and check that during parsing.

PR #1556, opened 2021-10-20
fix(publish): make 11ty publishing compatible with config version 3

kevinslin:


hikchoi:
shouldn't the versions be `0.64.1`?

kevinslin:
> shouldn't the versions be `0.64.1`?

by making it the same version as the current lerna version, lerna will make sure to update the dependencies when we bump the minor. the only place where the version is hardcoded is in the plugin code

PR #1597, opened 2021-10-27
enhance: update schemas on save instead of having to reload workspace.

nickolay-kondratyev:
# Pull Request Checklist

Allow schemas to be modified & added without having to run `Reload workspace`.

Note this implementation does have a limitation: IF the user deletes a schema (within a schema file) without deleting the schema file itself THEN the schema will not be deleted until the workspace is reloaded (or the id of the deleted schema is re-used somewhere else). In the spirit of Iteration this is not seen as a blocker for this change to go forward and deliver value to the customer. Since its expected that most of the cases will be modifications and additions to the schemas which this change supports. 

[Doc changes commit on dev](https://github.com/dendronhq/dendron-site/commit/33dc4667492c44a0958013c36ef185b10cf39250) 
### Quality Assurance
  
### Docs
[Doc changes commit on dev](https://github.com/dendronhq/dendron-site/commit/33dc4667492c44a0958013c36ef185b10cf39250)


packages/plugin-core/src/utils/docutil.ts
```diff
@@ -0,0 +1,7 @@
+import { TextDocument } from "vscode";
+
+export class DocUtil {
```
kevinslin, line 3: since this is not plugin specific, can we move this logic to `common-all`? see [style](https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.dev.style.utilities.md#L25:L25) for placing utilities

```ts
  static isSchema(uri: URI) {
    return uri.fsPath.endsWith(".schema.yml");
  }
```

kevinslin (changes requested):

packages/plugin-core/src/utils/docutil.ts
nickolay-kondratyev: Was thinking that `"vscode"`
```
import { TextDocument } from "vscode";
```
Is not a dependency that exists within common-all

packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -104,10 +118,16 @@ export class WorkspaceWatcher {
     );
   }
 
+  async onDidSaveTextDocument(document: TextDocument) {
+    if (DocUtil.isSchema(document)) {
+      await SchemaSyncService.instance().onDidSave({ document });
```
jonathanyeung, line 123: We already have a vscode.showInformation() message popup when a schema updates successfully right (didn't look at the code too closely to see when exactly it pops up)?  If not, maybe it's worth it to add a notice to users that the schema has been updated when they click save?

jonathanyeung (approved):

packages/plugin-core/src/utils/docutil.ts
nickolay-kondratyev: Moved to common-all

packages/plugin-core/src/WorkspaceWatcher.ts
nickolay-kondratyev: Added the message,


Note in the latest commit also removed the older SchemaWatcher (the code looked like it attempted to do schema parsing onChange but it didn't appear to be working). Also from UX experience I would argue we don't want to try to parse schemas on change since as the users are changing the schema before they save there is going to be invalid states of schema (which would trigger error messages before the user is done with their edits, and since we currently trigger error messages with pop ups it would get a bit annoying). So approaching the schemas as code seems like the right thing to do meaning attempting to compile when the file is saved, attempting to parse when the file is saved (Unless we rework how we display errors to not be done as pop ups). Also typical VSCode configuration saves the file when the user looses focus from the file.

packages/common-all/src/util/docutil.ts
```diff
@@ -0,0 +1,7 @@
+import { TextDocument } from "vscode";
```
kevinslin, line 1: we don't depend on vscode in common-all. we depend on `vscode-url`. see https://github.com/dendronhq/dendron/blob/723d72cf3d1a117faf12296dbe03aed62be8864e/packages/common-all/package.json#L64

also, can we move this into [SchemaUtils](https://github.com/dendronhq/dendron/blob/6e27edf3d7b246e089eb0090222ae3b6a718330e/packages/common-all/src/dnode.ts#L1201) instead?

kevinslin (changes requested):

packages/common-all/src/util/docutil.ts
nickolay-kondratyev: Since this `import { TextDocument } from "vscode";` builds though wouldn't that mean that we have some transitive dependency on vscode? 

Will move to schema utils

kevinslin: the reason it works is because we are in a monorepo and node will pick up dependencies by traversing up the file tree looking for `node_modules`. when we try to build this in isolation, `common-all` will not have access to the monorepo `node_modules` and will fail

packages/plugin-core/src/WorkspaceWatcher.ts
jonathanyeung: O - sorry, I was unclear with my comment (or perhaps positioning of this comment).  I didn't mean to imply that we should parse/toast on edit; I just wanted to check that we were properly informing a user on save.  Thanks!

jonathanyeung (approved):

PR #1603, opened 2021-10-27
enhance(search): merge lookup and search field

5h1rU:
### Overview
Search now is only performed by one single input. It allows us to centralize any search, no matter if it is for lookups or regular text.

### How works?
- For lookup search, type in the input as always.
- For Text-based search, use the prefix "?" plus the words that you want to search.

### Description
- Text-based search now shows more detailed results split into two different columns. On the left side, the user can find the title and the fname of the file matched. The right side is in charge of showing any text that corresponds with the search query string.
 
- Lookup search doesn't modify the behavior still working as always.

#### Video
https://user-images.githubusercontent.com/989826/139351031-4f763a38-d8ab-4a8a-b433-896f5dab531b.mov

- Regarding small screens, search input is now part of the hamburger menu. It allows showing a full-width Autocomplete component and free space on the header.

https://user-images.githubusercontent.com/989826/139474239-0d3f57d1-7369-4e38-a767-12b9363fc715.mov

### Outstanding work
- [x] Revisit the highlighting logic when a text is matched. Currently, the text is not showing descriptions as expected.
- [x] Cleanup: Remove the DendronLookup component and all the associated logic.

### Nice to have
- [x] Better performance when the component is being rendered.


packages/nextjs-template/components/DendronLookup.tsx
```diff
@@ -1,102 +0,0 @@
-import { NoteIndexProps, NoteLookupUtils } from "@dendronhq/common-all";
```
kevinslin, line 1: heads up that this gets imported in `DendronLayout`. type error when exporting

packages/nextjs-template/components/DendronLayout.tsx
```diff
@@ -94,33 +99,26 @@ export default function DendronLayout(
       <Header
         style={{
           position: "fixed",
+          isolation: "isolate",
```
kevinslin, line 102: huh, didn't realize this was a thing. so this is to make sure the header will be on top?

kevinslin (changes requested):
1. Notice that typing in search will mutate the title:
https://www.loom.com/share/d752df5908204b21979c20a1d6b9ccc5

@SeriousBug - is`MatchTitle` mutating the results?

here is another example -> https://www.loom.com/share/ddc3d976721f48e7860e7e5f8dd5e4da

also note the lag after typing

packages/nextjs-template/components/DendronLayout.tsx
5h1rU: Yes, this creates a new stack, the idea is not to worry about a potential z-index war. Each element in isolation with `z-index: 1` will always be in front of the rest of the content.

SeriousBug:
> @SeriousBug - is`MatchTitle` mutating the results?

I was going to say no, but it does look like the bad titles persist even after removing the search query. This is also happening in the live version so the bug is not limited to this PR. I don't see MatchTitle doing anything that would mutate the title, so not sure what's going on. @5h1rU Let me know if you need help and I can take a look at this on Monday.

5h1rU:
After the last PR changes, we got:
- [x] more performance
- [x] get rid of the React duplicated key error
- [x] remove the annoying title mutation

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -49,33 +53,84 @@ function DebouncedDendronSearchComponent(props: DendronPageWithNoteDataProps) {
...
   useEffect(() => {
-    if (search) search(value, setResults);
+    if (search) {
+      search(value, setSearchResults);
+    }
   }, [value, search]);
+
+  useEffect(() => {
+    requestNotes(searchResults?.map(({ item: note }) => note.id) || []);
+  }, [requestNotes, searchResults]);
```
kevinslin, line 100: why do we need to depend on `requestNotes` function here?

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -49,33 +53,84 @@ function DebouncedDendronSearchComponent(props: DendronPageWithNoteDataProps) {
...
   const dispatch = useCombinedDispatch();
   const { noteBodies, requestNotes } = useNoteBodies();
-  const { noteIndex, dendronRouter, search, error, loading, ensureIndexReady } =
-    props;
-  const [value, setValue] = React.useState("");
-  const [results, setResults] = React.useState<SearchResults>(undefined);
+  const lookup = useDendronLookup();
+  const { noteActive } = useNoteActive(dendronRouter.getActiveNoteId());
+  const initValue = noteActive?.fname || "";
+  const [value, setValue] = React.useState(initValue);
```
kevinslin, line 90: can we use a more descriptive name? eg. `searchQueryValue`

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -49,33 +53,84 @@ function DebouncedDendronSearchComponent(props: DendronPageWithNoteDataProps) {
...
 ) {
+  const { noteIndex, dendronRouter, search, error, loading, notes } = props;
+
+  const [searchResults, setSearchResults] =
+    React.useState<SearchResults>(undefined);
+  const [lookupResults, setLookupResults] = React.useState<NoteIndexProps[]>(
+    []
+  );
+  const [results, setResults] =
+    React.useState<"searchResults" | "lookupResults">();
```
kevinslin, line 84: When we have a small set of constants, prefer using enums

```ts
// good
enum ReturnStatus {
  ok = "ok",
  fail = "fail",
}

function foo(): ReturnStatus {
  ...
}

// bad
function foo(): "ok"|"fail"{
  ...
}
```

kevinslin: can we make the name more descriptive?

```tsx
enum SearchMode {
  LOOKUP_MODE = "lookupMode",
SEARCH_MODE = "searchMode"
}

[searchMode, setSearchMode] = React.useState<SearchMode>
```

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -49,33 +53,84 @@ function DebouncedDendronSearchComponent(props: DendronPageWithNoteDataProps) {
...
+
+  useEffect(() => {
+    requestNotes(searchResults?.map(({ item: note }) => note.id) || []);
+  }, [requestNotes, searchResults]);
+
   useEffect(() => {
-    requestNotes(results?.map(({ item: note }) => note.id) || []);
-  }, [results]);
+    if (value?.startsWith("?")) {
+      setResults("searchResults");
```
kevinslin, line 104: use enum value here

kevinslin (changes requested):
nice job! some comments about naming and small details. additional followups below:

1. can you comment to what was causing the title to mutate before?
1. in a future iteration, can we add realtime search back in without the debounce?

packages/nextjs-template/components/DendronSearch.tsx
5h1rU: Good question,

React doesn't know exactly when the requestNotes can change, in this case, is just a way to make sure nothing will change in the requestNotes function.

This is the kind of warnings thrown by the linter, that BTW is turned off.

This is a [good blogpost](https://reacttraining.com/blog/when-to-use-functions-in-hooks-dependency-array/) explaining exactly this behaviour.

5h1rU:
> nice job! some comments about naming and small details. additional followups below:
> 1. can you comment to what was causing the title to mutate before?
1. renderedTitle in the MatchTitle component was having an incorrect behavior with some keys duplicated. Isn't recommended to add raw arrays and push data in a React component since you can't guarantee immutability and this will trigger unexpected side effects. We have something similar in the MatchBody component but at the moment the outcome is not duplicating strings.

> 2. in a future iteration, can we add realtime search back in without the debounce?
2. Yes it's feasible, but first, we need to revamp some practices around rehydration in react and how the data is being updated in some places (see hooks file).

packages/nextjs-template/components/DendronSearch.tsx
kevinslin: got it. thanks for the explanation. feel free to re-enable the linting rules in that case in a separate PR!

just for my understanding, in this case, since `requestNotes` doesn't close over state, this technically isn't necessary but the linter doesn't know that and this could change in the future?

kevinslin (approved):

PR #1621, opened 2021-10-30
fix: Change Workspace command recognizes native workspaces

SeriousBug:


kevinslin (changes requested):
can we add a test for this?

PR #1631, opened 2021-11-01
enhance: adding error messaging and additional validation for when schemas are malformed

nickolay-kondratyev:
# enhance: adding error messaging and additional validation for when schemas are malformed

Adding error messaging and additional validation for when schemas are malformed.

# Pull Request Checklist

### Quality Assurance

## Follow up
Integrate the schemas malformed messaging with on save update (separate commit that is currently in review).


nickolay-kondratyev:
See that CLI tests are not passing will look into it.

packages/common-server/src/parser.ts
```diff
@@ -129,6 +145,31 @@ export class SchemaParserV2 extends ParserBaseV2 {
...
 
+  static createSchema(
+    opts: (SchemaOpts | SchemaRaw) & { vault: DVault }
+  ): SchemaProps {
+    this.validateSchemaObjOptsOnlyUseAllowedKeys(opts);
+
+    return SchemaUtils.create(opts);
+  }
+
+  private static validateSchemaObjOptsOnlyUseAllowedKeys(
```
kevinslin, line 156: we currently use [ajv](https://ajv.js.org/guide/typescript.html) library for pods to validate and error on schema. you might want to use this instead of creating our own (note: you'll need to add `ajv` to `common-server` as a dependency since its currently only used in `pods-core`)

kevinslin: see how we're using it [here](https://github.com/dendronhq/dendron-site/blob/dev/vault/pkg.pods-core.ref.validation.md#L17:L17)

kevinslin (changes requested):

packages/common-all/src/constants.ts
```diff
@@ -62,6 +62,10 @@ export enum ERROR_STATUS {
   UNKNOWN = "unknown",
 }
 
+export enum USER_MESSAGES {
```
jonathanyeung, line 65: nice, ty - let's start consolidating to this centralized location for user strings.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -146,9 +147,27 @@ export class FileStorage implements DStore {
...
+      if (resp.error && resp.error.payload) {
+        reason = JSON.parse(JSON.parse(resp.error.payload)[0].payload).message;
+      }
+    } catch (parseErr) {
+      reason = USER_MESSAGES.UNKNOWN;
+    }
+
     return new DendronError({
-      message: "schema malformed",
+      message: `Schema '${filePath}' is malformed. Reason: ${reason}. Use 'Go to File...' command to resolve malformed schema.`,
```
jonathanyeung, line 170: Speaking out loud - it'd be nice for us to include a button in the error notification such that they can click on it and be taken directly to the schema file for editing.  It'd be hard (and perhaps undesirable) to pack that info into a DendronError though. Something maybe for V2/V3

packages/common-all/src/constants.ts
nickolay-kondratyev: We might want to consider laying the groundwork in an extensible way  (to allow localization to happen in the future), but this is a step towards right direction.

packages/engine-server/src/drivers/file/storev2.ts
nickolay-kondratyev: Something like optional `errorData.fileAtFault` seems reasonable to have within DendronError. Dendron deals with files so fileAtFault in DendronError does not appear to break placement. Not sure if we can add buttons with `window.showErrorMessage` though.

jonathanyeung: It should be possible with the `items[]` param: https://code.visualstudio.com/api/references/vscode-api#window.showErrorMessage

packages/common-server/src/parser.ts
nickolay-kondratyev: Added AJV for the bulk the bulk of the schema validation.

NOTE: Added a brand new dependency ajv-errors (https://www.npmjs.com/package/ajv-errors) as part of this. It has MIT license and 10Milllion weekly downloads. This is used to format AJV errors in a user friendly way. 

Still kept some of the manual validation like validating that only the top schemas to require ids, since it appears overkill to create another separate AJV schema just for validation of id existence (if there are other cases specific to top level we can refactor into AJV validation).

kevinslin (changes requested):
looks like there are compiler issues preventing the build

packages/engine-server/src/drivers/file/storev2.ts
nickolay-kondratyev: Added a follow up ![[scratch.2021.11.04.134933.add-file-navigation-in-schema-malformed-message]]

nickolay-kondratyev:
Looks like splitting the createFromSchemaRaw/createFromSchemaOpts and possibly merge changes have caused the inline schema tests to break. Good thing tests are catching this, I dont see why its breaking quickly. Will have to investigate deeper possibly tomorrow.

jonathanyeung (approved):

Issue #1606, opened 2021-10-28
`Move Header` does not create a note if it does not already exists


mhijazi1:
## Please select the area your bug applies to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

When using the `Move Header` command, if a note that does not already exist is specified, the note is not created, but the header is still deleted from the current note.

## To Reproduce

1. Go to a note and place your cursor over a header you would like to move
2. Select `Dendron: Move Header` from the command pallet
3. Specify a note that does not already exist as the destination

Two things should happen:
1. The header you had selected will be deleted from the file
2. No new note is created.

## Expected behavior

A new note should be created and my selected header should be moved there

Additionally, if a new note cannot be created for any reason (say a permissions issue or some other failure), then my selected header should not be deleted and should fail as a no-op.

## Desktop (please complete the following information)

- OS: Verified on Windows and MacOS
- Version: 0.65.0

## Additional Context

This is a multi-vault workspace with two remote vaults. Not sure if that has an impact as I have not verified this with a fresh single vault workspace.

hikchoi:
@mhijazi1 

This has been shipped with v68 🎉 
Please try it out and let us know if there are still outstanding concerns regarding this issue. If there are no more comments in this thread after this point, this issue will automatically close in 4 days. Feel free to manually close this issue if you wish to do so.

PR #1646, opened 2021-11-03
fix: allow creation of new notes when move header destination doesn't exist yet

hikchoi:
# fix: allow creation of new notes when move header destination doesn't exist
This PR
- Allows new note creation in `MoveHeaderCommand` if destination note doesn't exist yet.

# Pull Request Checklist

### Quality Assurance

#### Special Cases
- [~] if your tests changes an existing snapshot, make sure that snapshots are [updated](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#updating-test-snapshots)
- [~] if you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in [test-workspace](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#test-workspace). We use this to manually inspect new changes and for auto regression testing

### Docs
- [~] If your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR

### DIFF
--- a/packages/plugin-core/src/commands/MoveHeader.ts
+++ b/packages/plugin-core/src/commands/MoveHeader.ts
@@ -22,7 +22,10 @@ import {
   Anchor,
   LinkUtils,
 } from "@dendronhq/engine-server";
-import { LookupControllerV3 } from "../components/lookup/LookupControllerV3";
+import {
+  LookupControllerV3,
+  LookupControllerV3CreateOpts,
+} from "../components/lookup/LookupControllerV3";
 import {
   NoteLookupProvider,
   NoteLookupProviderSuccessResp,
@@ -32,16 +35,25 @@ import { VSCodeUtils } from "../utils";
 import { BasicCommand } from "./base";
 import _ from "lodash";
 import path from "path";
-import { getEngine } from "../workspace";
+import { getDWorkspace, getEngine, getVaultFromUri } from "../workspace";
 import { EngineAPIService } from "../services/EngineAPIService";
 import { delayedUpdateDecorations } from "../features/windowDecorations";
 import { findReferences, FoundRefT } from "../utils/md";
 import { Location } from "vscode";
 import { file2Note, vault2Path } from "@dendronhq/common-server";
+import {
+  NewLocation,
+  PickerUtilsV2,
+  ProviderAcceptHooks,
+} from "../components/lookup/utils";
 
 type CommandInput =
   | {
       dest?: NoteProps;
+      initialValue?: string;
+      nonInteractive?: boolean;
+      useSameVault?: boolean;
+      vaultName?: string;
     }
   | undefined;
 type CommandOpts = {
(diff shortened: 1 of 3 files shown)


hikchoi:
DOCS: https://github.com/dendronhq/dendron-site/pull/263

hikchoi:
related: #1606

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -130,21 +147,34 @@ export class MoveHeaderCommand extends BasicCommand<
...
+      disableVaultSelection: opts?.useSameVault,
+      vaultSelectCanToggle: false,
+    };
+    const lookupController = LookupControllerV3.create(lcOpts);
     const lookupProvider = new NoteLookupProvider(this.key, {
-      allowNewNote: false,
+      allowNewNote: true,
       noHidePickerOnAccept: false,
     });
+    lookupProvider.registerOnAcceptHook(ProviderAcceptHooks.NewLocationHook);
```
kevinslin, line 160: lookup by default already gives us the quickpick value, this hooks seems to provide us with a vault if it is not set in the quickpick. instead of a hook, does it make sense to have this as a utility? eg. `PickerUtils.getVaultFromPickerOrEditor`?

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -32,16 +35,25 @@ import { VSCodeUtils } from "../utils";
...
+import {
+  NewLocation,
+  PickerUtilsV2,
+  ProviderAcceptHooks,
+} from "../components/lookup/utils";
 
 type CommandInput =
   | {
       dest?: NoteProps;
+      initialValue?: string;
```
kevinslin, line 53: why do we have both `dest` and `initialValue`? aren't they the same thing?

kevinslin: if we have to pick one, i would use `initialValue` to keep compatibility with `NoteLookup`

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -130,21 +147,34 @@ export class MoveHeaderCommand extends BasicCommand<
...
+    lookupProvider.registerOnAcceptHook(ProviderAcceptHooks.NewLocationHook);
+
+    let initialValue: string | undefined;
+    let nonInteractive: boolean = false;
+    if (opts?.dest) {
+      initialValue = opts.dest.fname;
+      nonInteractive = true;
+    } else if (opts?.initialValue) {
+      initialValue = opts.initialValue;
+      nonInteractive = true;
```
kevinslin, line 169: `initialValue` shouldn't apply `nonInteractive` that should be taken from the passed in value

kevinslin (changes requested):

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: doh. totally blanked on the fact that you can just grab the vault from the picker.

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: you are right. dest is a remnant of the old version of move header that I used for easier testing. will get rid of it

packages/plugin-core/src/commands/MoveHeader.ts
hikchoi: good point. this was intended for easier testing and not really to expose the args to the users. but we should keep it consistent with lookup.

PR #1664, opened 2021-11-05
feat:(next-export-pod): filter by navbar property

5h1rU:
### Status
This draft is ready for review. I don't want to publish as a PR yet since I want to ensure everything is working as expected. 

I found this improvement to be a step in the right direction but is at the limit of the 80/20 philosophy. The feature is working with a new file 10x lighter than notes.json (gzipped). Meaning we are loading all the initial data 10x faster. 

It's a good starting point for an eventual V2 on the next-js template and was an excellent opportunity for being more familiar with the codebase.

After making sure all tests are passing, I'll publish it as PR.


5h1rU:
I'm looking for early feedback on this feature.

The idea is to generate a new `sections.json` only with the sidebar sections that `nextjs-template` needs to use.

### Why?
Currently, `nextjs-template` is fetching the whole `notes.json,` which is significant and impacts the first render's performance.

This new JSON file will decrease the amount of data sent for navigation purposes; hence the performance and the render time will improve significantly. 

### Problem
The changes pushed on this draft aren't detected by the CLI,

#### Steps:
1. `yarn link` in the `pods-core` workspace
2.  `yarn link "@dendronhq/pods-core"` in the `dendron-cli` workspace
3. run `dendron exportPod --podId dendron.nextjs --config dest="/Users/MYUSER/Projects/dendron/packages/nextjs-template" --wsRoot /Users/MYUSER/Projects/org-workspace-incubator/seeds/dendron.dendron-site/`

The command did not create the new JSON file. Not sure if I'm pointing in the wrong direction.

packages/nextjs-template/utils/types.ts
```diff
@@ -38,8 +64,8 @@ export type DendronCommonProps = Partial<NoteData> & {
 export type DendronPageWithNoteDataProps = NoteData & DendronCommonProps;
 
 export function verifyNoteData(
-  noteData: Partial<NoteData>
-): noteData is NoteData {
+  noteData: NotePropsDict
```
kevinslin, line 67: why was this changed? this is meant to verify `NoteData`

kevinslin (changes requested):
some early comments, still working through the draft

packages/nextjs-template/pages/_app.tsx
```diff
@@ -45,24 +46,29 @@ function DendronApp({ Component, pageProps }: AppProps) {
...
       setLogLevel("INFO");
-      logger.info({ ctx: "fetchNotes:pre" });
-      const data = await fetchNotes();
-      logger.info({ ctx: "fetchNotes:got-data" });
+      const data = await fetchSections();
       setNoteData(data);
       batch(() => {
-        dispatch(browserEngineSlice.actions.setNotes(data.notes));
+        dispatch(
+          browserEngineSlice.actions.setNotes(data.notes as NotePropsDict)
```
kevinslin, line 53: this isn't actually true since `data` is `Section` instead of `NoteProps`. instead of casting it to something its not, we should add a `sections` property to `browserEngineSlice`

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -117,7 +126,7 @@ function DendronSearchComponent(
     }
     const out =
       qs === ""
-        ? NoteLookupUtils.fetchRootResults(notes)
+        ? NoteLookupUtils.fetchRootResults(notes as NotePropsDict)
```
kevinslin, line 129: don't cast notes to something its not, if `fetchRootResults` works with `sections` type, then add it to the type of `fetchRootResults`

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -42,10 +48,13 @@ type SearchFunction = (
...
 
-function DebouncedDendronSearchComponent(props: DendronPageWithNoteDataProps) {
+function DebouncedDendronSearchComponent(props: {
+  notes: SectionsData["notes"];
+  indexId: string;
+}) {
   // Splitting this part from DendronSearchComponent so that the debounced
   // function doesn't get reset every time value changes.
-  const results = useFetchFuse(props.notes);
+  const results = useFetchFuse(props.notes as NotePropsDict);
```
kevinslin, line 57: don't cast `props.notes` as something its not. change type signature instead or use a new method

kevinslin (changes requested):

PR #1670, opened 2021-11-08
feat: Native workspace enhancements

SeriousBug:
This PR does several things (sorry for creating a single big PR!):

### User-facing

- Dendron will now search for `dendron.yml` in native workspaces even if `dendron.yml` is not in the root of the workspace. For performance concerns, I limited this to 3 folders deep at most, and attempted to skip over some common folders we shouldn't search through, for example folders like `.git`.
- When in a non-Dendron workspace, Dendron will now activate a workspace watcher to watch for any `dendron.yml` files being added to the workspace. Once one is added, Dendron will automatically activate itself. Again this is limited by folder depth and ignored folders to avoid performance issues.
  - Workspaces initialized through this are tracked with analytics.
- We now track the type of workspace that has been initialized (e.g. code vs native) in analytics

### Internal

- Split the `DendronExtension.isActive` into `isActive` and `isValid`, with `isActive` checking whether the extension has been actually activated and `isValid` checking if the current workspace is an actual Dendron workspace.
- Edited test harnesses to avoid relying on `onWSInit`. This allows tests to fail instead of hanging indefinitely if there was an issue initializing the workspace.
- Adds tests to check for the initialization of native workspaces, both with `dendron.yml` at root and in nested folders.

Docs:  https://github.com/dendronhq/dendron-site/pull/266


SeriousBug:
I'm actually not sure if the "automatically initialize Dendron when a `dendron.yml` file is added" is a good feature. We'll be running Dendron to watch files being created in every workspace the user works with. This typically should not cause a problem, but there could be edge cases that cause problems for users. While this risk is very low, the impact would be high, which is why I'm hesitating. Any thoughts on keeping that feature or not?

packages/engine-server/src/workspace/utils.ts
```diff
@@ -68,14 +82,22 @@ export class WorkspaceUtils {
...
+    const dendronWorkspaceFolders = await Promise.all(
+      folders.map((folder) =>
+        findDownTo({
+          base: folder,
+          fname: CONSTANTS.DENDRON_CONFIG_FILE,
+          returnDirPath: true,
+        })
+      )
+    );
+    return dendronWorkspaceFolders.filter(isNotUndefined)[0];
```
kevinslin, line 100: instead of returning just the first entry, can we return all of them? in the short term, this lets us warn the user that we detected multiple and choose one. in the long term, we can eventually see if it makes sense to support this use case

packages/plugin-core/package.json
```diff
@@ -723,6 +723,11 @@
         "dendron.serverPort": {
           "type": "number",
           "description": "port for server. If not set, will be randomly generated at startup."
+        },
```
kevinslin, line 726: thoughts about adding this in `dendron.yml`  instead of the workspace? the reason is that we're working on removing configuration that relies on vscode

kevinslin (changes requested):
nice work. this will be a big enabler in using dendron in existing projects. a few minor comments

packages/plugin-core/package.json
SeriousBug: That wouldn't make sense because the watcher works **outside** Dendron workspaces.

kevinslin: ah, you're right. ignore that comment

packages/common-server/src/filesv2.ts
```diff
@@ -282,6 +283,73 @@ export function findUpTo(opts: {
...
 }
 
+export const WS_FILE_MAX_SEARCH_DEPTH = 3;
+
+/**
+ * Go to dirname that {fname} is contained in, going in (deeper into tree) from base.
+ * @param maxLvl Default 3, how deep to go down in the file tree. Keep in mind that the tree gets wider and this search becomes exponentially more expensive the deeper we go.
+ * @param returnDirPath - return path to directory, default: false
+ *
+ * One warning: this will not search into folders starting with `.` to avoid searching through things like the `.git` folder.
```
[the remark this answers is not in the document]
SeriousBug: Used `anymatch` (already a transitive dep from chokidar) and reused the list of glob patterns that we use for Chokidar.

packages/engine-server/src/workspace/utils.ts
SeriousBug: Revised this code to return all of them. Also added some code to Native Workspace initialization to prompt user if there are multiple roots. This prompt should be pretty rare because the only situation this would happen is if:
1. User has multiple workspace folders configured
2. There are 2 workspace folders, not subfolders of each other, that both contain `dendron.yml`
I tested this with a workspace I created with this exact setup:

PR #1673, opened 2021-11-08
feat: adding new command: create schema from hierarchy

nickolay-kondratyev:
# adding new command: create schema from hierarchy

Adding new command: `Dendron: Create Schema From Note Hierarchy` 

[PR Dendron Site Documentation](https://github.com/dendronhq/dendron-site/commit/9aa4870608b1ed189d991514363fc06314eb47e2)  
### Quality Assurance
```
   CreateSchemaFromHierarchyCommand tests
     HierarchyLevel tests:
       GIVEN 'languages.*.data' hierarchy level
         ✓ THEN have expected label
         ✓ THEN have expected top id
         ✓ THEN produce expected default schema name
         tokenize tests
           ✓ WHEN fname is 'languages.python.data' THEN tokenize to '["languages","*","data"]'.
           ✓ WHEN fname is 'languages.python.data.integer' THEN tokenize to '["languages","*","data","integer"]'.
           ✓ WHEN fname is 'languages.python.machine-learning' THEN tokenize to '["languages","*","machine-learning"]'.
         isCandidateNote tests
           ✓ WHEN testing 'languages' note THEN do NOT match.
           ✓ WHEN testing 'languages.python' note THEN do NOT match.
           ✓ WHEN testing 'languages.python.data' note THEN do match.
           ✓ WHEN testing 'languages.python.data.integer' note THEN do match.
           ✓ WHEN testing 'languages.python.machine-learning' note THEN do match.
     SchemaCreator tests:
       ✓ WHEN valid schema candidates THEN generate schema
     CreateSchemaFromHierarchyCommand
       sanityCheck tests:
         ✓ WHEN no active text editor THEN give error
       formatSchemaCandidates tests:
         WHEN formatting valid input idx=1 ['h1.h2.h3a', 'h1.h2.h3b']
           ✓ THEN include 'h1.*.h3a' label
      haveUserPickSchemaFileName tests:
        ✓ WHEN user picked non existent name THEN prompt use the name
        ✓ WHEN user picked pre existing name THEN prompt again
      haveUserSelectHierarchyLevel tests:
        ✓ WHEN happy input THEN return user picked hierarchy level
        ✓ WHEN hierarchy depth of current file is too small THEN undefined
        ✓ WHEN top id is already used by existing schema THEN undefined
      hasSelected tests:
        ✓ WHEN curr is greater than prev THEN true
        ✓ WHEN curr is equal to prev THEN false
        ✓ WHEN curr is less than prev THEN false
      hasUnselected tests:
        ✓ WHEN curr is greater thhan prev THEN false
        ✓ WHEN curr is equal to prev THEN false
        ✓ WHEN curr is less than prev THEN true
      findUncheckedItem tests:
        ✓ WHEN unchecked THEN get the candidate
      findCheckedItems tests:
        ✓ WHEN checked THEN get the candidate
      determineAfterSelect tests:
        WHEN a child of non selected hierarchy is checked
          ✓ THEN make sure length is as expected
          ✓ THEN retain the checked items
          ✓ THEN select parent of item
          ✓ THEN select grand-parent (ancestor) of item
      determineAfterUnselect tests:
        WHEN ancestor is unchecked
          ✓ THEN keep another hierarchy as is
          ✓ THEN unselect all the descendents
  33 passing (497ms)
```
Added `test-workspace/vault/dendron.cmd.create-schema-from-hierarchy.md`

### DIFF
--- a/packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
+++ b/packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
@@ -151,7 +151,7 @@ function getSchemaUri(vault: DVault, schemaName: string) {
  * asking user for input data.
  * */
 export class UserQueries {
-  static async haveUserPickSchemaFileName(
+  static async promptUserForSchemaFileName(
     hierarchyLevel: HierarchyLevel,
     vault: DVault
   ): Promise<string | undefined> {
@@ -172,7 +172,7 @@ export class UserQueries {
       alreadyExists = fs.existsSync(getSchemaUri(vault, schemaName).fsPath);
       if (alreadyExists) {
         vscode.window.showInformationMessage(
-          `Schema with name '${schemaName}' already exists. Pick a different name or cancel.`
+          `Schema with name '${schemaName}' already exists. Please choose a different name.`
         );
       }
     } while (alreadyExists);
@@ -180,14 +180,14 @@ export class UserQueries {
     return schemaName;
   }
 
-  static async haveUserSelectHierarchyLevel(currDocFsPath: string) {
+  static async promptUserToSelectHierarchyLevel(currDocFsPath: string) {
     const hierarchy = new Hierarchy(path.basename(currDocFsPath, ".md"));
 
     if (hierarchy.depth() <= 2) {
       // We require some depth to the hierarchy to be able to choose a variance
       // pattern within in it. More info within Hierarchy object.
       await vscode.window.showErrorMessage(
-        `Pick a note deeper within the note hierarchy.`
+        `Pick a note with note depth greater than 2.`
       );
 
       return undefined;
@@ -219,7 +219,7 @@ export class UserQueries {
     return hierarchyLevel;
   }
 
-  static haveUserPickPatternsFromCandidates(
+  static promptUserToPickPatternsFromCandidates(
     labeledCandidates: SchemaCandidate[]
   ): Promise<readonly SchemaCandidate[]> {
     return new Promise((resolve) => {
@@ -438,7 +438,7 @@ export class CreateSchemaFromHierarchyCommand extends BasicCommand<
     const vault = PluginVaultUtils.getVaultByNotePath({
       fsPath: currDocumentFSPath,
     });
-    const hierarchyLevel = await UserQueries.haveUserSelectHierarchyLevel(
+    const hierarchyLevel = await UserQueries.promptUserToSelectHierarchyLevel(
       currDocumentFSPath
     );
     if (hierarchyLevel === undefined) {
@@ -448,9 +448,9 @@ export class CreateSchemaFromHierarchyCommand extends BasicCommand<
 
     const candidates = this.getHierarchyCandidates(hierarchyLevel);
     const pickedCandidates =
-      await UserQueries.haveUserPickPatternsFromCandidates(candidates);
+      await UserQueries.promptUserToPickPatternsFromCandidates(candidates);
 
-    const schemaName = await UserQueries.haveUserPickSchemaFileName(
+    const schemaName = await UserQueries.promptUserForSchemaFileName(
       hierarchyLevel,
       vault
     );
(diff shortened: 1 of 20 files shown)


packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
```diff
@@ -0,0 +1,528 @@
...
+  );
+  return uri;
+}
+
+/**
+ * Encapsulates methods that are responsible for user interaction when
+ * asking user for input data.
+ * */
+export class UserQueries {
+  static async haveUserPickSchemaFileName(
```
jonathanyeung, line 154: super nit: replace `have` with `prompt`? `promptUserForSchemaFileName`?

packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
```diff
@@ -0,0 +1,528 @@
...
+
+      if (!schemaName) {
+        // Cancelled.
+        return schemaName;
+      }
+
+      alreadyExists = fs.existsSync(getSchemaUri(vault, schemaName).fsPath);
+      if (alreadyExists) {
+        vscode.window.showInformationMessage(
+          `Schema with name '${schemaName}' already exists. Pick a different name or cancel.`
```
jonathanyeung, line 175: super nit: drop 'or cancel'. 

"Schema with name '${schemaName}' already exists. Please choose a different name."

packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
```diff
@@ -0,0 +1,528 @@
...
+  }
+
+  static async haveUserSelectHierarchyLevel(currDocFsPath: string) {
+    const hierarchy = new Hierarchy(path.basename(currDocFsPath, ".md"));
+
+    if (hierarchy.depth() <= 2) {
+      // We require some depth to the hierarchy to be able to choose a variance
+      // pattern within in it. More info within Hierarchy object.
+      await vscode.window.showErrorMessage(
+        `Pick a note deeper within the note hierarchy.`
```
jonathanyeung, line 190: Should we be explicit in the error message that the depth must be greater than or equal to 2? Instead of saying 'deeper', which may be ambiguous.

packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
```diff
@@ -0,0 +1,528 @@
...
+        `Schema with top level id: '${hierarchy.topId()}' already exists.`,
+        msgGoToSchema
+      );
+      if (action === msgGoToSchema) {
+        const schema = PluginSchemaUtils.getSchema(hierarchy.topId());
+
+        await VSCodeUtils.openFileInEditor(getUriFromSchema(schema));
+      }
+
+      return undefined;
```
jonathanyeung, line 212: It'd be good if we can capture some of these exit points (`return undefined`) with some analytics in a finally block.  That way we can distinguish between a user who ran the command but gave up vs a user who went all the way through and ended up with a schema. This method should allow you to add props to the analytics payload that gets generated from the command invocation: https://github.com/dendronhq/dendron/blob/01d2dd8b69d6bff175acf78405bd903ac268b3b3/packages/plugin-core/src/commands/base.ts#L32

jonathanyeung (changes requested):
Looks great, just a few nitpicks.

packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
nickolay-kondratyev: Prompt sounds better agree.

packages/plugin-core/src/commands/CreateSchemaFromHierarchyCommand.ts
nickolay-kondratyev: Agreed, as discussed will circle back to this.

jonathanyeung (approved):

Issue #1533, opened 2021-10-16
"Vault Add" will add duplicate .dendron.cache.* line to .gitignore


djradon:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [x] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

"Vault Add" will add duplicate .dendron.cache.* line to .gitignore

## To Reproduce

1. create a new, empty remote Git repository
2. "Vault Add" to add it to a multi-vault workspace
3. notice that ".dendron.cache.*" gets added to .gitignore (possibly with an extraneous blank line at the start)
4. commit the "Vault Add"-related changes
5. "Vault Add" the repo to a different workspace (or the same one after removing it?)
6. notice an addition  ".dendron.cache.*"  line gets added

## Expected behavior

If .gitignore already contains  ".dendron.cache.*", no additional one gets added

djradon:
Also noticed that, when adding vaults to the workspace's .gitignore:
* there's an extraneous blank line added as well when new vaults are added
* if you remove and re-add a vault, a duplicate line gets added. (expected behaviour is to not add duplicates lines or blank lines)

djradon:
You could also argue that "Vault Remove" should cleanup the workspace's .gitignore vault reference, but since "Vault Remove" doesn't remove the folder from the filesystem, probably leaving the .gitignore line is safer.

kevinslin:
Good catch. Will get this fixed in the coming release!

SeriousBug:
Dendron will now avoid adding duplicate lines to gitignore with version 0.69. Let us know if you are having any trouble with this feature, otherwise this issue will close in a few days.

PR #1689, opened 2021-11-10
fix: vault add avoids adding duplicate lines & vault remove cleans up gitignore lines

SeriousBug:
#1533


SeriousBug:
I'm not sure if we should revert the change to clean up gitignore lines when removing the vault since the files are not deleted. I added it since it's a very simple change, should we keep it?

packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -169,15 +169,13 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
-            vaults,
-            async (resp: any, vault: DVault) => {
-              await resp;
-              await wsService.createVault({ vault });
-              return this.addVaultToWorkspace(vault);
-            },
-            Promise.resolve()
-          );
+          // Some things, like updating config, can't be parallelized so needs to be done one at a time
+          for (const vault of vaults) {
```
kevinslin, line 173: TIL - didn't realize that `for` loops can be used to run things in serial!

might be worth adding as a utility in common-all since we do this in a number of places

kevinslin:
> I'm not sure if we should revert the change to clean up gitignore lines when removing the vault since the files are not deleted. I added it since it's a very simple change, should we keep it?

its uncertain what the desired behavior is since removing a vault doesn't delete it. i would err on the side of doing the "dumb thing" - that is leave things as is. we can always add additional behavior in a subsequent pr

kevinslin (changes requested):
code looks great. just a comment about removing the action on removeVault

PR #1692, opened 2021-11-10
enhance: looks ups with dot at the end now look up hierarchies without having exact postfix

nickolay-kondratyev:
# Summary
Having a dot at the end will now look up child & descendant notes (bubbling up the child notes to the top) that match the query but without requiring the full match of the prefix note. 

# Usage details in test workspace
```
1. Look up `data.` 
Should see results with roughly the following order:
* `data.driven`     
    * first show matches with `data.` exact matched and `data.` highest in hierarchy
* `languages.python.data.string`
    * then show matches that have `data.` exact matched but lower in hierarchy.
* ...
* `languages.with-data.make-sense` 
    * later show matches that are not exact matched `with-data.`
* `languages.python.data.string.memory` 
    * lastly show matches that have not just children but grand children etc.
All matches should have `data.` and some value after `data.`
```

# Details
In the task note for this task it was proposed that one way to handle the `.` at the end would be to grab the closest hierarchy that matches. However, doing some testing it appeared we can improve the usability of dot ended queries more by doing special sort of the results so that we show results from multiple hierarchies that match the given dot ended query, while with special sorting helping to bubble up the most relevant results.
Task note: [[adding . to lookup should show all results in the level|scratch.2021.10.31.143958.adding--to-lookup-should-show-all-results-in-the-level]]

# Pull Request Checklist

### Quality Assurance

### DIFF
--- a/packages/engine-test-utils/src/presets/engine-server/utils.ts
+++ b/packages/engine-test-utils/src/presets/engine-server/utils.ts
@@ -1,5 +1,6 @@
 import { NoteProps, SchemaUtils } from "@dendronhq/common-all";
 import {
+  CreateNoteFactory,
   NOTE_PRESETS_V4,
   NoteTestUtilsV4,
   PreSetupHookFunction,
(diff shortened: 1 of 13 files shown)


packages/plugin-core/src/components/lookup/utils.ts
```diff
@@ -864,6 +934,11 @@ export const filterPickerResults = ({
     );
   }
 
+  // Ending the query with a dot adds special processing of showing matched descendents.
+  if (transformedQuery.originalQuery.endsWith(".")) {
+    return sortForQueryEndingWithDot(transformedQuery, itemsToFilter);
```
nickolay-kondratyev, line 939: Probably best not to return but rather assign to `itemsToFilter` to have a single return in the method as well as allowing any other filtering (although none apply right now but for future) to take place.

packages/engine-test-utils/src/presets/engine-server/utils.ts
```diff
@@ -45,6 +46,56 @@ export const setupBasic: PreSetupHookFunction = async ({
   await SCHEMA_PRESETS_V4.SCHEMA_SIMPLE.create({ vault, wsRoot });
 };
 
+/**
+ /vault1
```
kevinslin, line 50: awesome doc string :)

packages/engine-test-utils/src/presets/engine-server/utils.ts
```diff
@@ -45,6 +46,56 @@ export const setupBasic: PreSetupHookFunction = async ({
...
+
+/vault2
+ ├── root.md
+ └── root.schema.yml
+
+ /vault3
+ ├── root.md
+ └── root.schema.yml
+ * */
+export const setupHierarchy: PreSetupHookFunction = async ({
```
kevinslin, line 72: can we give this a more descriptive name? eg. `setupHierarchyForLookupTests`

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -412,6 +438,140 @@ suite("NoteLookupCommand", function () {
...
+   ├── bar.ch1.gch1.ggch1.md
+   ├── bar.ch1.gch1.md
+   ├── bar.ch1.md
+   ├── bar.md
+   ├── foo.ch1.gch1.ggch1.md
+   ├── foo.ch1.gch1.md
+   ├── foo.ch1.gch2.md
+   ├── foo.ch1.md
+   ├── foo.ch2.md
+   ├── goo.ends-with-ch1.no-ch1-by-itself.md
```
kevinslin, line 455: goo is new. guess you don't give a `baz` 😆

packages/plugin-core/src/components/lookup/utils.ts
kevinslin: good point on the assignment. besides allowing for future filters, it also keeps the filter function simple (current implementation would introduce conditional logic that short circuits other processors)

unless there are any downsides i'm missing, would go for assignment

kevinslin (changes requested):
nicely done! like that you took this task and implemented the more general solution :)

heads up that there are some conflicts from master because of some simplification work we did for subscribing to lookup inputs: https://github.com/dendronhq/dendron/pull/1693/files

packages/engine-test-utils/src/presets/engine-server/utils.ts
nickolay-kondratyev: As a nit for myself should surround this string with <pre></pre> so quic doc renders it nicely as well.

packages/plugin-core/src/components/lookup/utils.ts
nickolay-kondratyev: Updated.

kevinslin (approved):

PR #1696, opened 2021-11-11
fix:(layout): position sidebar, main content and footer

5h1rU:
- Replace runtime CSS rules on DendronLayout with layout components
- Move footer over the navbar [[scratch.2021.11.08.151528.move-footer-over-to-navbar]]

**Improvements**
- The layout takes the full width on screens < 1320px (See [docusaurus](https://docusaurus.io/docs))
- Remove the content layout shifting when the page is being resized.
- Remove runtime calculations

Attachment

https://user-images.githubusercontent.com/989826/141357288-34e36897-4028-41d2-867d-f795101989b1.mov

### DIFF
--- a/packages/nextjs-template/components/DendronLayout.tsx
+++ b/packages/nextjs-template/components/DendronLayout.tsx
@@ -1,4 +1,4 @@
-import { Layout, Row, Col, Divider } from "antd";
+import { Layout, Row, Col, Divider, Typography } from "antd";
 import { MenuOutlined } from "@ant-design/icons";
 import * as React from "react";
 import { DENDRON_STYLE_CONSTANTS } from "../styles/constants";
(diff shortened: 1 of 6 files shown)


SeriousBug (commented):
Looks good to me and feel pretty good to use. I think it's more responsive (as in it responds to input faster) but I have a one nitpick:

The line size looks pretty wide on large displays, I'm seeing around 145 characters per line on a 1440p display. Could we shorten them a bit? Suggestions online point to around 60 to 100 to be optimal, so I think something in that range would be better.

kevinslin (changes requested):
i don't know if its my local style but when I try to preview, the tree from our former footer text is causing a newline

https://www.loom.com/i/6980d6f090bc4eb78e8ab456fecdcc80

PR #1701, opened 2021-11-12
feat: Initialize Workspace command can create native workspaces

SeriousBug:
Docs: https://github.com/dendronhq/dendron-site/pull/276

If the user currently has a non-Dendron workspace open, we now offer to initialize that as a Dendron native workspace.

Fixed a bug with handling existing roots where cancelling the prompt would create a workspace anyway, and also improved the user-friendliness of it.

Finally, also fixed a bug where non-Dendron workspaces are recognized as Native Dendron workspaces.

### DIFF
--- a/packages/plugin-core/src/commands/SetupWorkspace.ts
+++ b/packages/plugin-core/src/commands/SetupWorkspace.ts
@@ -33,8 +33,8 @@ type CommandOutput = DVault[];
 
 export { CommandOpts as SetupWorkspaceOpts };
 
-const CODE_WS_LABEL = "Code";
-const CODE_WS_DETAIL = "New workspace";
+const CODE_WS_LABEL = "Code Workspace";
+const CODE_WS_DETAIL = undefined;
 
 enum EXISTING_ROOT_ACTIONS {
   CONTINUE = "continue",
@@ -66,14 +66,14 @@ export class SetupWorkspaceCommand extends BasicCommand<
             picked: true,
             label: CODE_WS_LABEL,
             description: CODE_WS_DETAIL,
-            detail: "A Dendron Code workspace that's for notes only.",
+            detail: "A dedicated IDE workspace for just your notes",
           },
           ...workspaceFolders.map((folder) => {
             const folderName = folder.name || folder.uri.fsPath;
             return {
               label: folderName,
               description: folder.uri.fsPath,
-              detail: `Take notes in ${folderName} with a Dendron Native workspace.`,
+              detail: `Take notes in ${folderName} alongside your existing project`,
             };
           }),
         ],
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/commands/SetupWorkspace.ts
```diff
@@ -14,39 +14,104 @@ import { BasicCommand } from "./base";
...
+          ignoreFocusOut: true,
+          title: "Workspace type to initialize",
+        }
+      );
+      if (initNative === undefined) return;
+      if (
+        initNative.label !== CODE_WS_LABEL ||
+        initNative.description !== CODE_WS_DETAIL
+      ) {
+        // Not sure if there's a better way to check for this, but this is if a native workspace option was selected
```
kevinslin, line 90: okay for now until/if we add more options

packages/plugin-core/src/commands/SetupWorkspace.ts
```diff
@@ -85,39 +162,35 @@ export class SetupWorkspaceCommand extends BasicCommand<
...
-        validateInput: async (value: string) => {
-          if (!_.includes(_.keys(options), value.toLowerCase())) {
-            return `not valid input. valid inputs: ${_.keys(options).join(
-              ", "
-            )}`;
-          }
-          return null;
-        },
-      });
+      const resp = await VSCodeUtils.showQuickPick(
```
kevinslin, line 165: kudos for including this. this has been on my backlog since the moment it was implemented 😅

packages/plugin-core/src/commands/SetupWorkspace.ts
```diff
@@ -14,39 +14,104 @@ import { BasicCommand } from "./base";
...
+    const workspaceFolders = vscode.workspace.workspaceFolders;
+    if (
+      workspaceFolders !== undefined &&
+      workspaceFolders.length > 0 &&
+      (await WorkspaceUtils.getWorkspaceType({
+        workspaceFolders,
+      })) === WorkspaceType.NONE
+    ) {
+      // If there's a non-Dendron workspace already open, offer to convert that to a Dendron workspace first
+      const initNative = await VSCodeUtils.showQuickPick(
```
kevinslin, line 63: for labels, can we make them more consistent and give more context? Something like the following:

- Code Workspace  - [no description]
    - A dedicated IDE workspace for just your notes

- Native Workspace - [no description]
    - A Dendron workspace that can be used alongside your existing projects

kevinslin (changes requested):

packages/plugin-core/src/commands/SetupWorkspace.ts
SeriousBug: We can't exactly do that because here may be more than 1 native workspace option if there are multiple workspace folders set up. I tried to change it as close as I could get it:

kevinslin: Hmm, can we swap out the label? instead of the name of the folder, can we use "Native Workspace"

For the detail, quote the folder name (or bold it if it supports markdown modifiers)

SeriousBug: No markdown support so I just quoted it:



Thanks for helping iterate on this!

kevinslin: any time. good job!

kevinslin (approved):

PR #1711, opened 2021-11-14
chore/3 refresh notes api

kevinslin:
chore(api): add `refreshNotes` api

API to update all note metadata. For example, links and tags. 
Useful when you've added a note externally from Dendron. 

[Docs](https://github.com/dendronhq/dendron-site/blob/dev/vault/pkg.dendron-engine.api.engine.md#L7:L7)


packages/common-all/src/types/typesv2.ts
```diff
@@ -494,6 +498,12 @@ export type DEngine = DCommonProps &
     queryNotesSync({ qs }: { qs: string; vault?: DVault }): NoteQueryResp;
     renameNote: (opts: RenameNoteOpts) => Promise<RespV2<RenameNotePayload>>;
     renderNote: (opts: RenderNoteOpts) => Promise<RespV2<RenderNotePayload>>;
+    /**
+     * Update note metadata.
+     * Use cases:
+     * - update notes if they've been changed outside of Dendron
+     */
+    refreshNotes: (opts: RefreshNotesOpts) => Promise<void | IDendronError>;
```
jonathanyeung, line 506: why not `RespV2`?

packages/engine-server/src/engineClient.ts
```diff
@@ -372,7 +368,7 @@ export class DendronEngineClient implements DEngineClient {
     if (_.isUndefined(noteClean)) {
       throw new DendronError({ message: "error updating note", payload: resp });
     }
-    await this.refreshNotes([noteClean]);
+    await this.refreshNotesV2([{ note: noteClean, status: "update" }]);
```
jonathanyeung, line 371: If we are getting rid of the existing `refreshNotes`, why rename to `V2`?

packages/engine-server/src/utils.ts
```diff
@@ -326,4 +328,39 @@ export class EngineUtils {
...
+    engine,
+    notesMap,
+  }: {
+    note: NoteProps;
+    engine: DEngineClient;
+    notesMap: Map<string, NoteProps>;
+  }) {
+    const maxNoteLength = ConfigUtils.getWorkspace(engine.config).maxNoteLength;
+    if (
+      note.body.length <
```
jonathanyeung, line 347: Is it possible to return something in case length > max length? At least let's add some note in the function description.

jonathanyeung (changes requested):

packages/common-all/src/types/typesv2.ts
kevinslin: no good reason, changed it

packages/engine-server/src/engineClient.ts
kevinslin: that method already existed, `refreshNotes` was just a wrapper around it. wanted to minimize code changes with introducing this API

packages/engine-server/src/utils.ts
kevinslin: added a note.

kevinslin:
pr comments addressed in this new PR -> https://github.com/dendronhq/dendron/pull/1716

PR #1725, opened 2021-11-17
fix: decorations working for long notes, improved task note decorations

SeriousBug:
- Thanks to an optimization, all decorations (like link highlighting) are now displayed even in very long notes.
- Fixed a bug with decorations not appearing in Native Workspaces.
- Task note decorations are colored with the same color as the link to make them easier to distinguish from regular text.
- Task notes say `priority:` instead of `prio:` for priorities.


packages/common-all/src/types/workspace.ts
```diff
@@ -356,13 +356,7 @@ export type DendronConfig = {
...
 
-  /** Notes that are too large can cause slowdowns when editing them. For notes
-   * longer than this many characters, some editor decorations like tag colors
-   * and broken link highlighting will be disabled. Other decorations will
-   * continue to function.
-   *
-   * Defaults to 4096 characters, which is about 4 KiB.
-   */
+  /** @deprecated No longer relevant, all decorations are now displayed for long notes. */
   maxNoteDecoratedLength?: number;
```
kevinslin, line 360: if we're no longer using, does it make sense to just remove altogether?

packages/plugin-core/src/features/windowDecorations.ts
```diff
@@ -126,114 +121,108 @@ export function delayedUpdateDecorations(
...
-    note = VSCodeUtils.getNoteFromDocument(activeEditor.document);
-    if (_.isUndefined(note)) return {};
-  } catch (error) {
+      if (decorateRange.start.line === 0) {
+        // Can't check frontmatter if frontmatter is not visible
+        if (_.isUndefined(frontmatter)) {
+          allWarnings.push(warnMissingFrontmatter(activeEditor.document));
+        } else {
+          allWarnings.push(
+            ...warnBadFrontmatterContents(activeEditor.document, frontmatter)
```
kevinslin, line 194: can we rename this method? `warnBadFrontmatterContents` makes me think that is bad frontmatter. `checkAndWarnIfBadFrontmatterContents` would be more descriptive

kevinslin (changes requested):
Found an edge case where if I start from a task note and then go to a note that links to it, the decorators show up 2x: https://www.loom.com/share/f04e23bb082642f98221aa912ffba1d2?from_recorder=1&focus_title=1

Repro by going to [[Prepare Report|task.2021.10.26.020813.prepare-report]] in test-workspace

SeriousBug:
> Found an edge case

Oops! Thought I had fixed this with the `mergeRanges` function but looks like I used it in the wrong order :)

Fixed the bug and applied other feedback now.

kevinslin (approved):

PR #1745, opened 2021-11-19
fix: add workspace root to workspace folders for Code Workspaces

SeriousBug:
Adds the workspace root to the Code Workspace template. The reasoning behind it is that people tend to get confused that they can't find `dendron.yml` in their workspace when they initialize it.

I remember talking about making this change a few months ago, but I don't remember if we decided not to add it.

Here's a screenshot of what the explorer looks like with a default workspace initialized:
Screenshot_20211119_041308

This also might be a little confusing since the vaults are shown both in their own sections and also under root. I couldn't figure out a way to hide them without potentially also hiding other folders in the workspace. What are your thoughts on doing it this way versus how it was before?

### DIFF
--- a/packages/engine-server/src/workspace/vscode.ts
+++ b/packages/engine-server/src/workspace/vscode.ts
@@ -109,12 +109,16 @@ export class WorkspaceConfig {
     });
     const jsonBody: WorkspaceSettings = _.merge(
       {
-        folders: cleanOpts.vaults
-          ? cleanOpts.vaults.map((ent) => ({
-              path: ent.fsPath,
-              name: ent.name,
-            }))
-          : [],
+        folders: [
+          {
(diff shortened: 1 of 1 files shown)


kevinslin (commented):
yeah, this is a tough one. its confusing not to be able to edit workspace files but arguably more confusing to see your workspace and folders show up twice. ideally, we shouldn't need to rely on the file explorer in vscode as the dendron tree view should show all vaults and notes (we'd need to extend it to see the dendron.yml)

I'd prefer not adding the root workspace since the `code-workspace` file would be the only thing that folks would need to edit and even that is now less important with native workspaces

SeriousBug:
Makes sense to me, in that case I'll drop the PR.

PR #1766, opened 2021-11-21
enhance(cli): cli uses separate port file

kevinslin:
The Dendron CLI will create its own instance of an API server for most commands. This engine, when spawned, will write a port file at `.dendron.port` so that other processes can discover the port that the engine server is running on.
The issue is that a running workspace has its own port file which will be overridden. This will make commands like `dendron {cmd} --attach` fail as the port is no longer accurate.

This change causes the dendron-cli to write the engine port at `dendron.port.cli` instead.

- enhance: cli write to own path
- enhance(cli): do not overwrite ws dendron port when activating CLI

### DIFF
--- a/bootstrap/.vscode/tasks.json
+++ b/bootstrap/.vscode/tasks.json
@@ -37,10 +37,11 @@
       "command": "./scripts/testAll.sh",
       "options": {
         "env": {
-          "LOG_LEVEL": "error"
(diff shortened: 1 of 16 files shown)


SeriousBug (approved):
Looks good to me. Can confirm it's creating the `.dendron.port.cli` file for the CLI commands.

PR #1768, opened 2021-11-22
chore(build): Use in memory verdaccio to speed up local publishing

kevinslin:
- NOTE: in memory verdaccio is much faster but doesn't persist contents after shutdown which makes it insuitable for publishing in pipeline (where building and pulling are in two different steps)

- chore(build): add sleep after publish to build step
- update build
- update timeout
- better lerna output
- update verdaccio config
- add local auth provider
- update patch script
- udpate verdaccio
- update verdaccio
- restore old buildPatch
- add npm proxy back
- Revert "chore(build): add sleep after publish to build step"
- add await
- update uplinks
- update buildPatch

### DIFF
--- a/bootstrap/scripts/buildPatch.sh
+++ b/bootstrap/scripts/buildPatch.sh
@@ -6,7 +6,7 @@ if [ $PUBLISH_ENDPOINT = "local" ] && [ -z $FAST ] ; then
 	sleep 3
 fi
 if [ $PUBLISH_ENDPOINT = "local" ] && [ $FAST ] ; then
-	echo "start verdaccio in fast mode"
+	echo "starting verdaccio with in-memory cache to speed up build time"
 	verdaccio -c ./bootstrap/data/verdaccio/config.yaml > verdaccio.log 2>&1 &
 	FOO_PID=$!
 	echo "$FOO_PID"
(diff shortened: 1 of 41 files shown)


bootstrap/scripts/buildPatch.sh
```diff
@@ -1,10 +1,18 @@
-if [ $PUBLISH_ENDPOINT = "local" ]; then
+if [ $PUBLISH_ENDPOINT = "local" ] && [ -z $FAST ] ; then
 	echo "start verdaccio"
 	verdaccio > verdaccio.log 2>&1 &
 	FOO_PID=$!
 	echo "$FOO_PID"
 	sleep 3
 fi
+if [ $PUBLISH_ENDPOINT = "local" ] && [ $FAST ] ; then
+	echo "start verdaccio in fast mode"
```
jonathanyeung, line 9: instead of calling it 'fast mode', can we describe it for what's actually happening? i.e. `starting verdaccio with in-memory cache to speed up build time`

jonathanyeung (approved):

bootstrap/scripts/buildPatch.sh
kevinslin: updated

PR #1770, opened 2021-11-22
feat: decorator improvements

SeriousBug:
Docs (internal): https://github.com/dendronhq/dendron-docs/pull/1

This re-adds the reverted decorator improvements, including the improved decorator speeds and task note decorator improvements. It also makes some changes to improve responsiveness of the decorators. Most importantly:

- We now give up early if calculating decorations takes too long instead of doing it based on the note length, which is a better experience when dealing with notes that are long but don't have a lot of links in them. We warn the user once in the session if this happens.
- Also added a debounce to the decorator updates to avoid the editor getting bogged down with a lot of async operations that haven't yet completed, which then blocks other async functions like note lookups.

This seems to have fixed the lagginess when editing wikilink dense files like the backlog, as the highlighting stops after some wikilinks at the top have been highlighted. I think this is generally better while we implement the `getNoteByFName` performance improvements.

For the review: you'll want to look at the commits after Nov 19th if you want to see the changes since the reversal.


kevinslin (changes requested):
Did some tests with the new changes - page is responsive but there seems to still be a leak. See the following:
- with decorator optimization: https://www.loom.com/share/54dc3d0344274b28a1b60fcecb1306b7
- with current master(f43ecbc56): https://www.loom.com/share/91d743732ac44c4cb4a9ca4f17d43d63

The file descriptor count was obtained by running 
```
    watch -n1 "lsof | awk '{print $1}' | uniq -c | grep Code | awk '{ sum += \$1 } END { print sum }'"
```

SeriousBug:
Figured out the issue with leaking file descriptors. It's actually deeper than just the decorators. The details are in `[[Decorator Leak|dendron://private/user.kevin.journal.2021.11.21.m.decorator-leak]]`, but I'll copy them here too:

## Experiments

By adding some logging code and analyzing it I found something very interesting. The logging code looks like this:

```ts
    const fds = readdirSync("/proc/self/fd");
    Logger.info({
      ctx,
      files: fds.map((fd) => {
        try {
          return readlinkSync(`/proc/self/fd/${fd}`);
        } catch {
          return undefined;
        }
      }),
    });
```
This basically find the corresponding files for all open file descriptors in the current process. The interesting part is when I plug the earliest and latest logs to node and process them a bit:
```ts
> first.files.length
78
> last.files.length // more files have been opened as we scroll
86
> _.uniq(last.files).length // the files being opened are not unique, they are duplicates!
51
> _.uniq(first.files).length
51
> last.files.slice(last.files.length - 12)
[
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  'pipe:[18315242]',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  '/home/kaan/.config/Code - Insiders/logs/20211123T003248/exthost8/dendron.@dendronhq/plugin-core/dendron.log',
  null,
  'pipe:[18315242]'
]
```

Upon further investigation, `getFM` method was creating a new Workspace service
on every call, which created a new logger every time, which then would open the
log file every time. The log file **must** be closed explicitly, otherwise the
file descriptors leak.

## Resolution

Deprecated `createLogger` and added `createDisposableLogger` to fix the issue
with minimal codebase changes. Made the Workspace service disposable, and
updated `getFM` to dispose the workspace service once done.

packages/common-server/src/logger.ts
```diff
@@ -50,6 +51,28 @@ function createLogger(
   }
 }
 
+/** Create a logger. The logger **must** be disposed after being used if the function returned a dispose callback, otherwise it will leak file descriptors and may lead to crashes. */
+function createDisposableLogger(
```
kevinslin, line 55: to reduce state, can we always return a `dispose()` function? for `stdout` case, it should just be a null-op

packages/engine-server/src/seed/service.ts
```diff
@@ -228,6 +231,7 @@ export class SeedService {
   }): Promise<SeedSvcResp> {
     const ws = new WorkspaceService({ wsRoot: this.wsRoot });
     const config = ws.config;
+    ws.dispose();
```
kevinslin, line 234: instead of disposing of the workspace everytime, can we add a singleton by using [getOrCreate](https://github.com/dendronhq/dendron/blob/c6cd54f99dc9c17e28566bbe80f3c23546fe935d/packages/lsp-server/src/engine-server.ts#L15)?

kevinslin (changes requested):
nice find and good job debugging! just a few comments on implementation

packages/engine-server/src/seed/service.ts
SeriousBug: Do you mean a global singleton for the `WorkspaceService`? I'm reluctant to implement it that way because some code could create one with a different `wsRoot`. We could go with a memoized `getOrCreate`, but we'll still be opening files and leaving them open as we keep workspace services in the memoize cache.

I simplified some of the code by using a static function instead.

kevinslin (approved):

PR #1771, opened 2021-11-22
fix: allow assets to open from preview view

nickolay-kondratyev:
# fix: allow assets to open from preview view

### Quality Assurance


packages/plugin-core/src/commands/ShowPreview.ts
```diff
@@ -94,31 +102,104 @@ export const extractNoteIdFromHref = (data: {
...
   MARKDOWN = "MARKDOWN",
   UNKNOWN = "UNKNOWN",
 }
 
 const classifyLink = ({ href }: NoteViewMessage["data"]): LinkType => {
   if (href && href.startsWith("vscode-webview")) {
-    return LinkType.WIKI;
-  }
-  if (href && (href.startsWith("http://") || href.startsWith("https://"))) {
+    if (href.includes("/assets/")) {
```
kevinslin, line 115: Assets are referenced via `/assets/`,  `assets/` and `./assets/`. The above would miss the second case. 

There's also an edge case where we could have someone with a vault named assets at which point the cross vault link would look like the following: `dendron://assets/{wiki-link}`

I forget if the vault name shows up as part of the `href`. I like to say no but good to double check

test-workspace/dendron.2021.11.22.113354516.migrate-configs.yml
```diff
@@ -0,0 +1,118 @@
+version: 3
+useFMTitle: true
```
kevinslin, line 2: this is an artifact of #1743 . shouldn't be in the commit

test-workspace/dendron.2021.11.22.172645373.migrate-configs.yml
```diff
@@ -0,0 +1,118 @@
+version: 3
+useFMTitle: true
```
kevinslin, line 2: see above comment

test-workspace/vault/dendron.ref.assets.md
```diff
@@ -0,0 +1,14 @@
+---
+id: Hf27I1UR3HvKyd6HRh8C0
+title: Assets
+desc: ''
+updated: 1637576810331
+created: 1637568178090
+---
+
+## PDF pasted with `Dendron: Paste File` command
+[dummy-pdf.pdf](assets/dummy-pdf.pdf)
```
kevinslin, line 10: should also add case of `./assets/` and `/assets/`.

also, can we add a vault or rename one of the local vaults to `assets` and create a cross vault link to it?

kevinslin (changes requested):

packages/plugin-core/src/commands/ShowPreview.ts
nickolay-kondratyev: Good call out didn't know `dendron://assets/{wiki-link}` was a thing. 

On the different asset formatting it appears by the time it gets to `href` they are all formatted the same added manual testing for that as well. Let me know if you meant some different way to reference the assets than the ones that are added in the manual test:

```
* [dummy-pdf.pdf](assets/dummy-pdf.pdf) 
* [dummy-pdf.pdf](./assets/dummy-pdf.pdf) 
* [dummy-pdf.pdf](/assets/dummy-pdf.pdf) 
```

test-workspace/vault/dendron.ref.assets.md
nickolay-kondratyev: Added 

New vault `assets` and asset link to it from `vault` file 
```
* [test-file-in-another-vault-which-is-named-assets](dendron://assets/assets/assets/test-file-in-asset-vault.md)
```

packages/plugin-core/src/commands/ShowPreview.ts
kevinslin: miscommunication here - when i say they could be a vault for assets, I meant someone can create a vault and name it `assets`. 

in that case, you would have `assets` appear in the url but actually not be an asset. so `[[ dendron://assets/foo]]` should open the note foo. if the vault name doesn't show up in the `href` then the point should be moot and we don't need to handle it

packages/plugin-core/src/commands/ShowPreview.ts
```diff
@@ -94,31 +102,138 @@ export const extractNoteIdFromHref = (data: {
...
   WEBSITE = "WEBSITE",
   MARKDOWN = "MARKDOWN",
   UNKNOWN = "UNKNOWN",
 }
 
 const classifyLink = ({ href }: NoteViewMessage["data"]): LinkType => {
-  if (href && href.startsWith("vscode-webview")) {
+  if (
+    href &&
+    (href.startsWith("dendron://") ||
```
kevinslin, line 116: we only support this format for cross vault links to dendron notes, not assets. 
its interesting to think how we would link to assets in other vaults but we'd want to think about how we want to support that. 

we're actually moving away from regular markdown links alltogether and handling assets according to this instead -> [[Note References over Note Types|dendron://private/task.markdown.2021.11.18.note-references-over-note-types]]

kevinslin (changes requested):

packages/plugin-core/src/commands/ShowPreview.ts
nickolay-kondratyev: Rolled back the change of 'dendron://'

kevinslin (approved):

PR #1774, opened 2021-11-22
enhance: lookup selection to items

hikchoi:
# enhance: lookup selection to items
This PR:
- Adds a new selection modifier `selection2Items`, which takes the selection from active note, extracts wikilinks from selection, and use the extracted wikilinks as possible lookup quickpick items.

This is a pre-cursor to [[Refactor from Selection|task.lookup.refactor-can-take-clipboard-as-multi-select-input]] that adds the needed base lookup functionality.

# Pull Request Checklist

### Quality Assurance

#### Special Cases

### Docs

### Git
        - `eg. feat/add-thisthing`

### Github Issue
- [~] If this pull request is addressing an existing issue, make sure to link this PR to the issue that it is resolving.
- [~] If the resolution comes with a document update, link the docs PR to the issue as well.

### DIFF
--- a/packages/engine-server/src/markdown/remark/index.ts
+++ b/packages/engine-server/src/markdown/remark/index.ts
@@ -14,6 +14,7 @@ export {
   ALIAS_NAME,
   LINK_CONTENTS,
   visit,
+  ParseLinkV2Resp,
 } from "./utils";
 export { wikiLinks, WikiLinksOpts, matchWikiLink } from "./wikiLinks";
 export {
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -694,6 +694,21 @@ export class LinkUtils {
       return true;
     } else return false;
   }
+
+  /**
+   * Given a source string, extract all wikilinks within the source.
+   *
+   * @param source string to extract wikilinks from
+   */
+  static extractWikiLinks(source: string) {
+    // chop up the source.
+    const regExp = new RegExp("\\[\\[(.+?)?\\]\\]", "g");
+    const matched = [...source.matchAll(regExp)].map((match) => {
+      return LinkUtils.parseLinkV2(match[1]);
+    });
+
+    return matched;
+  }
 }
 
 export class AnchorUtils {
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -115,6 +115,22 @@ export type LinkFilter = {
   loc?: Partial<DNoteLoc>;
 };
 
+export type ParseLinkV2Resp =
+  | {
+      alias?: string;
+      value: string;
+      anchorHeader?: string;
+      vaultName?: string;
+      sameFile: false;
+    }
+  | {
+      alias?: string;
+      value?: string;
+      anchorHeader: string;
+      vaultName?: string;
+      sameFile: true;
+    };
+
 export function hashTag2WikiLinkNoteV4(hashtag: HashTag): WikiLinkNoteV4 {
   return {
     ...hashtag,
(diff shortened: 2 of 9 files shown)


kevinslin (changes requested):
currently, using `selection2item` doesn't seem like it will do anything (eg. you can't open the selection or accept). for this feature actually result in an action, it should also trigger `multiSelect`

- if `selection2Item` is selected, also enable `multiselect` and check all items that are selected
    - when user enters okay, all items are `accepted`, resulting in everything being opened

hikchoi:
> currently, using `selection2item` doesn't seem like it will do anything (eg. you can't open the selection or accept). for this feature actually result in an action, it should also trigger `multiSelect`
> 
> * if `selection2Item` is selected, also enable `multiselect` and check all items that are selected
>   
>   * when user enters okay, all items are `accepted`, resulting in everything being opened

Do you think we should limit the use case for `selection2Items` with only multi-selecting everything? I can come up with a few scenarios where I would want to limit the scope of lookup but still don't want everything to actually be accepted.

> for this feature actually result in an action

I considered this feature being in action as "converting the selected wikilinks into quickpick items".
The intent of using this modifier would be limiting the scope of lookup, and accepting everything would be another intent. I don't think it we should couple those two intents into one. 

Unless we are creating a command that _specifically_ is made for limiting the scope of lookup and also selecting / doing something with the selections (i.e. the workflow of the proposed new workflow for refactor, which is separate to this PR)

kevinslin:
> Do you think we should limit the use case for selection2Items with only multi-selecting everything? I can come up with a few scenarios where I would want to limit the scope of lookup but still don't want everything to actually be accepted.
> I considered this feature being in action would be "converting the selected wikilinks into quickpick items".
> The intent of using this modifier would be limiting the scope of lookup, and accepting everything would be another intent. I don't think it we should couple those two intents into one.

yes, it would be more *pure* to have a modifier for "scope lookup to these items" and another for "select these items", but in practice, these two functions will be coupled most of the time. 

a general design philosophy for dendron is to make simple things simple. the simple thing when selecting a bunch of links is to perform an action over all of them (this triggers multi-select). until we come to a time where we need to decouple this or we get feedback from users that this is desired behavior, we should adopt a `just-enough-engineering` philosophy.

> edit: On second thought, I wouldn't even make the new refactor workflow default to trigger both multiselect and selection2Items at the same time. In the context of refactor, making a selection in the document and triggering selection2Items by its nature could sufficiently be considered as "scoping", where actually having those items selected in the quickpick would be "capturing" (which are two different stages of refactor)

going back to keeping simple things simple, if the user selects a bunch of links, i would enable both multiselect and `selection2Item`

hikchoi:
> but in practice, these two functions will be coupled most of the time.

I am not too sure about this point.

Do notes typically have all links neatly organized in one block so that the user can alway select exactly what they want selected/accepted?

I think the case your suggested behavior is preferred is if you have a note specifically written for the purpose of selecting and opening all notes. Most other times you would have links sparsely sprinkled around a note, and not all are neatly ordered in a way that you could simply select a block and open all of them.

Examples of general use cases I can think of where keeping it decoupled makes sense:
1. I have a note with a bunch of links. I want to lookup only the notes mentioned in this note. I do not have the intention to open them all at once, but want to look them up.
2. I have a note with a bunch of links. Say this is a list of all relevant links to a project. Maybe I want to open some of them, but I'm not sure. I would want to see them in the quickpick before deciding I want to select multiple items.

> we should adopt a just-enough-engineering philosophy

Wouldn't a JEE here be have the behavior generic and have be manually triggered? Adding an extra behavior so that triggering `selection2Items` would also trigger `multiSelect`, and also select everything in the quickpick is a reasonably simple task, so we don't really have the concern of "is this worth the time?", but we are still assuming that we know what the users want. I would argue that keeping things simple here is to not assume what the user want and not automate a small behavior that would otherwise could be done manually. In fact, coupling the two modifiers would make it less modularized which makes taking this button and using it in other commands harder.

I also want to step back and resurface my concern about the note lookup command becoming a command that does everything. Having a lot of moving parts means we are going to have this discussion about user intent every time a new feature is added to note lookup, and it's going to get more complex and we will be struggling to come up with a sensible default that can cover all of the different intents the user might have given this command.

kevinslin:
After further thought, I'm coming to your side of keeping `multiselect` and `selection2Item` decoupled. 

> I also want to step back and resurface my concern about the note lookup command becoming a command that does everything. Having a lot of moving parts means we are going to have this discussion about user intent every time a new feature is added to note lookup, and it's going to get more complex and we will be struggling to come up with a sensible default that can cover all of the different intents the user might have given this command.

We're adopting a plugin based approach to lookup. Lookup isn't doing everything - it just has the capability to accept modifiers (buttons in this case), that change its effect. Depending on what the command needs, it can take a different set of modifiers. I think the issue is that some modifiers require us to add additional properties into quickpick. But the properties we add are general for all lookups. 

I did a short write up of the current design [here](https://github.com/dendronhq/dendron-docs/blob/main/vault/pkg.plugin-core.t.lookup.md#L10:L10)

If you have a different idea for how we can architect it, feel free to schedule an async. Open to exploring other approaches.

PR #1781, opened 2021-11-24
enhance: add auto complete to note lookup

nickolay-kondratyev:
# enhance: add auto complete to note lookup

Note the tests currently hiccup around odd issue on initialization, haven't been able to pin point why this is happening yet. 
```
Cmd is not a constructor: TypeError: Cmd is not a constructor
	at /usr/local/workplace/dendron/packages/plugin-core/out/src/workspace.js:476:25
```

DOC PR: https://github.com/dendronhq/dendron-site/pull/292

Here are some video usages, quite satisfying to use this addition:
* https://www.loom.com/share/e2e6cd06c05d45fb9ed60542b96a069e
* https://www.loom.com/share/4e2be8f5ddb842cfa8136f2c9ac3b118

### Quality Assurance


hikchoi:
```
Cmd is not a constructor: TypeError: Cmd is not a constructor
at /usr/local/workplace/dendron/packages/plugin-core/out/src/workspace.js:476:25
```

Possibly because `NoteLookupAutoCompleteCommand` depends on `NoteLookupCommand`, and there is an indirect circular dependency somewhere. This one's a super cryptic error I've ran into a while ago and it drove me crazy 😄  

If you try / catch what's happening in `_setupCommands` in `workspace.ts`, you can see what `Cmd` it was trying to register that was undefined. It is undefined because the module loading order is mixed up and the `Cmd` doesn't exist yet.

A quick fix to this is pinpointing where the command is invoked (most likely in `NoteLookupAutoCompleteCommand`), and dynamically `require`ing the dependency instead of `import`

The real fix for this is to change our codebase so that we have full control over the module loading order, but that is a whole other big task so using the require hack for now should be good enough.

example of doing this: https://github.com/dendronhq/dendron/blob/a9ea0eaa4b44c7598831be9a625cd04df069725f/packages/plugin-core/src/utils.ts#L742
more info on this issue: https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de

packages/plugin-core/package.json
```diff
@@ -771,6 +777,12 @@
         "key": "ctrl+l",
         "when": "dendron:pluginActive"
       },
+      {
+        "command": "dendron.lookupNoteAutoComplete",
+        "mac": "Tab",
```
kevinslin, line 782: `key` alone should be good enough, `mac` is only if we have mac specific keybindings

packages/plugin-core/src/commands/NoteLookupAutoCompleteCommand.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class NoteLookupAutoCompleteCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput
+> {
+  key = DENDRON_COMMANDS.LOOKUP_NOTE_AUTO_COMPLETE.key;
+  async gatherInputs(): Promise<CommandInput | undefined> {
+    return {};
+  }
+  static num: number = 0;
```
kevinslin, line 19: did we mean to leave this in?

kevinslin (changes requested):
can we refactor the code so that:
- `UI_NOTE_LOOKUP_COMMAND` is not exported but accessed via a function
- avoid the circular dependency that @hikchoi is talking about
- make it possible to extend this paradigm to add autocomplete to other commands down the line?


should be able to accomplish this with something like the below:


- src/utils/autoCompleter.ts
```tsx
const _UI_AUTOCOMPLETE_COMMANDS = {
    "noteLookup": undefined,
}

// we will want to bring autocomplete in other lookup enabled commands (eg. rename, insert note, etc)
const getOrInitUIAutoComplete = (id: "noteLookup", cmd: BaseCommand) => {
    if ! _UI_AUTOCOMPLETE_COMMANDS[id] {
        _UI_AUTOCOMPLETE_COMMANDS[id] = cmd
    }
}

const runUIAutoComplete = (id: "noteLookup") => {
    return _UI_AUTOCOMPLETE_COMMANDS[id]
}
```

- src/commands/NoteLookupAutoCompleteCommand.ts
```ts
execute {
    getOrInitUIAutoComplete().onAutoComplete()
}
```

- src/commands/NoteLookupCommand.ts
```ts
constructor {
    
}
```

packages/plugin-core/src/constants.ts
```diff
@@ -358,6 +359,38 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
...
+  // But it is possible for user to tab into empty side panel which does not
+  // have a `view` context set, at that point if user still has look up open and
+  // presses tab, Tab will get intercepted by note auto complete.
+  //
+  // Ideally there would be a trigger event when quick pick goes in focus/focuses out
+  // but not able to find such hook.
+  LOOKUP_NOTE_AUTO_COMPLETE: {
+    key: "dendron.lookupNoteAutoComplete",
+
+    // TODO: Hide this command from command palette
```
jonathanyeung, line 380: Can we try this? https://stackoverflow.com/questions/55270915/vs-code-extension-hide-commands-from-command-palette

`package.json`:
```typescript
"contributes": {
"commands": [
    {
        "command": "extension.myHiddenCommand",
        "title": "Compile folder"
    }
],
"menus": {
    "commandPalette": [
        {
            "command": "extension.myHiddenCommand",
            "when": "false"
        }
    ]
}
```

packages/plugin-core/src/commands/NoteLookupAutoCompleteCommand.ts
nickolay-kondratyev: nope missed removing it

packages/plugin-core/src/constants.ts
nickolay-kondratyev: Will try

nickolay-kondratyev: Works

packages/plugin-core/src/commands/NoteLookupAutoCompleteCommand.ts
nickolay-kondratyev: taken out

packages/plugin-core/package.json
nickolay-kondratyev: removed

packages/plugin-core/src/commands/NoteLookupAutoCompleteCommand.ts
```diff
@@ -1,6 +1,10 @@
 import { DENDRON_COMMANDS } from "../constants";
 import { BasicCommand } from "./base";
-import { UI_NOTE_LOOKUP_COMMAND } from "./NoteLookupCommand";
+import {
+  AUTO_COMPLETABLE_COMMAND_ID,
+  UIAutoCompletableCmds,
+} from "../utils/autoCompleter";
+// import { AUTO_COMPLETABLE_COMMAND_IDS, getUIAutoCompletableUICmd } from "../utils/autoCompleter";
```
kevinslin, line 7: nitpick: remove comment

kevinslin:
@jonathanyeung , can you add https://stackoverflow.com/questions/55270915/vs-code-extension-hide-commands-from-command-palette to [[dev.cook.vscode]] inside our docs vault? (note doesn't exist yet)

packages/plugin-core/src/test/suite-integ/C_autoCompleter.test.ts
```diff
@@ -0,0 +1,85 @@
+/**
+ * File has C_ prefix to move it down in the list of tests.
+ * If C_ is removed it bubbles up to the top of the tests and for some reason
```
kevinslin, line 3: Will merge it so we can get feedback on the tab completion. Since circular dependencies are becoming more of an issue, can you spend a few more cycles here to see if you can diagnose it? (eg. 30min time box and writeup of what you found for next person to take over?)

packages/plugin-core/src/commands/NoteLookupAutoCompleteCommand.ts
kevinslin: will merge into mainline but could you remove the above in a separate commit? (you can bundle it with your next PR)

packages/plugin-core/src/test/suite-integ/C_autoCompleter.test.ts
nickolay-kondratyev: Fixed this particular instance https://github.com/dendronhq/dendron/pull/1810

Issue #1609, opened 2021-10-28
Refactor Hierarchy has complete regex (with replacement values) for robust rename many notes and change their wikilinks in vault


voothi:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are possible) ##

- [x] Onboard. Initial vault setup and import/export pods
- [ ] Create. Note creation, lookup, snippets and templates
- [x] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

After imported my notes with Dendron Import Pod Markdown I have flat structure of notes in my vault without any prefix. 
All notes located in one heap (vault). The structure of links between them is stored in certain root notes in this vault. 
For prefixes (typing) my leaf notes with robust change ref wikilinks I should have tool for it. 
It seems well fit Refactor Hierarchy for this, but his don't work for some cases.
For example
Compose regexp replace string for correct renaming notes in vault. Now i get not correct strings when entering arguments 
link to examples on regex101: https://regex101.com/r/3i2s4j/1/

regex: `journal.(\d{4})(\d{2})(\d{2})`
replacement: `$1.$2.$3`
test data in:
```
veles.daily.journal.20211013
veles.daily.journal.20211014
veles.daily.journal.20211015
veles.daily.journal.20211016
veles.daily.journal.20211017
veles.daily.journal.20211018
veles.daily.journal.20211019
```
test data out:
```
veles.daily.2021.10.13
veles.daily.2021.10.14
veles.daily.2021.10.15
veles.daily.2021.10.16
veles.daily.2021.10.17
veles.daily.2021.10.18
veles.daily.2021.10.19
```

In Dendron I now have
![image](https://user-images.githubusercontent.com/9587340/139229508-4131cf07-6118-466e-b97f-31a13a7d26a3.png)

## Describe the solution you'd like ##

In Dendron Refactor Hierarchy I have worked regex replacement such as `$1.$2.$3` in example above.

## Describe alternatives you've considered ##

As a more universal solution I have in Dendron Doctor command for 
run over all notes in around all connected vault in workspace 
and fix wikilinks to new wikilinks with added prexifex.
[...]

Issue #116, opened 2020-08-23
Support backreferences in `Refactor`


kevinslin:
Example

- files
```
lang.python.data
lang.ruby.data
lang.js.data
```

- refactor
```
match: lang.(.*).data
replace: lang.$1.d
```

- result
```
lang.python.d
lang.ruby.d
lang.js.dd
```

hikchoi:
looking at the `RefactorHierarchyCommandV2` now, looks like we are fixed to doing a `(.*)(${match})(.*)` to pick the candidates.

I think removing that and simply letting the users input a full regexp pattern would cause a bit of friction as it's a pretty significant breaking change.

I propose that we add a separate command (say `RefactorHierachyFullRegexCommand`), or add an config to the existing one so that users can have a choice. (and let them pick the default behaviour).

kevinslin:
I like that. Instead of doing a separate command, we can add an option
toggle to the existing command to go fully regex

PR #1796, opened 2021-11-26
enhance: enable scoping and full regex support for refactor hierarchy command

hikchoi:
# enhance: enable scoping and full regex support for refactor hierarchy command
This PR
- Adds scoping capability to refactor hierarchy command
  - Can choose to use all notes or use `selection2Items` and `multiSelect` to cherry pick notes to refactor through lookup
- Add full regular expression support for capturing stage
- Add full regular expression support for replacing stage

# Pending tasks
- [x] Add tests for new refactor capabilities
- [x] Update documents for refactor command
- [x] Write-up on future plans on introducing refactor based hierarchy manipulation capabilities.

# Pull Request Checklist

### Quality Assurance
- [~] If your tests involve running [[manual Testing|dendron://dendron.dendron-site/dendron.dev.qa.test#manual-testing]], make sure to update [[Test Workspace|dendron://dendron.dendron-site/dendron.dev.ref.test-workspace]] with any necessary notes/additions to perform manual test ^9jtWc6ov340p
- [x] After you submit your pull request, check the output of our [integration test](https://github.com/dendronhq/dendron/actions) and make sure all tests pass

#### Special Cases

### Docs

### Git
        - `eg. feat/add-thisthing`

### Analytics
- [~] If you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Github Issue
- [x] If this pull request is addressing an existing issue, make sure to link this PR to the issue that it is resolving.
- [x] If the resolution comes with a document update, link the docs PR to the issue as well.

### DIFF
--- a/packages/plugin-core/src/commands/RefactorHierarchyV2.ts
+++ b/packages/plugin-core/src/commands/RefactorHierarchyV2.ts
@@ -64,7 +64,7 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
   async promptScope(): Promise<NoteLookupProviderSuccessResp | undefined> {
     // see if we have a selection that contains wikilinks
     const { text } = VSCodeUtils.getSelection();
-    const wikiLinks = LinkUtils.extractWikiLinks(text as string);
+    const wikiLinks = text ? LinkUtils.extractWikiLinks(text) : [];
     const shouldUseSelection = wikiLinks.length > 0;
 
     // if we have a selection w/ wikilinks, selection2Items
@@ -108,8 +108,7 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
       });
       controller.show({
         title: "Decide the scope of refactor",
-        placeholder:
-          "Query for scope. Select 'Entire Workspace' to use all notes.",
+        placeholder: "Query for scope.",
         provider,
         selectAll: true,
       });
--- a/packages/plugin-core/src/commands/RefactorHierarchyV2.ts
+++ b/packages/plugin-core/src/commands/RefactorHierarchyV2.ts
@@ -1,20 +1,41 @@
-import { DVault, NoteUtils } from "@dendronhq/common-all";
+import {
+  DNodeProps,
+  DNodeUtils,
+  DVault,
+  NoteQuickInput,
+  NoteUtils,
+} from "@dendronhq/common-all";
 import { vault2Path } from "@dendronhq/common-server";
 import fs from "fs-extra";
 import _ from "lodash";
 import _md from "markdown-it";
+import { HistoryEvent } from "@dendronhq/engine-server";
 import path from "path";
 import { ProgressLocation, Uri, ViewColumn, window } from "vscode";
+import {
+  LookupControllerV3,
+  LookupControllerV3CreateOpts,
+} from "../components/lookup/LookupControllerV3";
+import {
+  NoteLookupProvider,
+  NoteLookupProviderSuccessResp,
+} from "../components/lookup/LookupProviderV3";
+import { NoteLookupProviderUtils } from "../components/lookup/utils";
 import { DENDRON_COMMANDS } from "../constants";
 import { FileWatcher } from "../fileWatcher";
 import { VSCodeUtils } from "../utils";
 import { getExtension, getDWorkspace } from "../workspace";
 import { BasicCommand } from "./base";
 import { RenameNoteOutputV2a, RenameNoteV2aCommand } from "./RenameNoteV2a";
+import {
+  MultiSelectBtn,
+  Selection2ItemsBtn,
+} from "../components/lookup/buttons";
 
 const md = _md();
 
 type CommandOpts = {
+  scope?: NoteLookupProviderSuccessResp;
   match: string;
   replace: string;
 };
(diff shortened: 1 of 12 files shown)


hikchoi:
related: #116 #1609

hikchoi:
NOTE:
As a pre-cursor to this PR, `selection2Items` modifier was added to Note lookup.
This is not what we actually want. As a follow up, I will be removing the modifier button from `NoteLookupCommand`, but this will not be the scope of this PR and will be in a separate one.

hikchoi:
A quick write-up for future plans. Use as a base for release announcement as necessary:

- Refactor hierarchy now has full support for regular expressions. This means you can now use complex expressions like positive / negative lookaround, backreferences, named/unnamed capture groups, and non-capture groups
- We are planning to add the ability to limit the scope based on vaults for multivault workspaces.
- We plan to use the full regular expression capabilities to create various refactor-derivative convenience commands that abstract away the regular expression and make the user experience of hierarchy manipulation much more easier.

packages/plugin-core/src/commands/RefactorHierarchyV2.ts
```diff
@@ -114,45 +229,61 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
...
+    matchRE: RegExp;
+    engine: DEngineClient;
+  }) {
+    const { scope, matchRE, engine } = opts;
+    const { notes, schemas, vaults, wsRoot } = engine;
+
+    const capturedEntries: NoteQuickInput[] = [];
+
+    const scopedItems = _.isUndefined(scope)
+      ? _.toArray(notes).map((note: DNodeProps) => {
```
kevinslin, line 243: this can be very expensive to do (eg. our workspace is +10k notes). instead of converting everything to `QuickInput`, can we do the reverse? Convert `QuickInput` items to notes?

packages/plugin-core/src/test/suite-integ/RefactorHierarchy.test.ts
```diff
@@ -1,212 +1,299 @@
...
+        const { wsRoot, vaults } = opts;
+        const vault = vaults[0];
+
+        refFooTest = await NoteTestUtilsV4.createNote({
+          vault,
           wsRoot,
-          fname: "refactor",
-          body: ["- [[refactor.one]]", "- [[refactor.two]]"].join("\n"),
+          fname: "dendron.ref.foo.test",
+          body: [""].join("\n"),
```
kevinslin, line 232: body is optional, if you're not using it, you can not define it

kevinslin (changes requested):
1. The scope selection at beginning of refactor is a little confusing:
    - by default, it pre-fills the current note name 
        - pressing enter limits the scope to the current note which is probably not what people want (that's what rename is for)
        - its also too similar to how refactor's current interface is (select match text) 
    - suggestions:
        - default to current behavior
            - title of quickpick should reflect stage in refactor (eg Pick a scope) 
            - by default, it should be "Entire Workspace"
                - instead of pressing escape, can we keep the same interface? (first option is always entire workspace)
1. For user experience, if the user selects text and activates refactor, we should default to `select2Items` being turned on

packages/plugin-core/src/commands/RefactorHierarchyV2.ts
```diff
@@ -32,27 +55,119 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
...
+        title: "Scope",
+        placeholder: "Query for scope. Press esc. to use all notes.",
+        provider,
+      });
+    });
+  }
+
+  async promptMatchText() {
+    const match = await VSCodeUtils.showInputBox({
+      prompt: "Enter match text. Leave blank to capture entire file name.",
```
jonathanyeung, line 104: I know this is parity with our current behavior, but should we add some more info here as to what 'match text' does? We can also split it up to use the 'title' prop and the 'prompt' prop of `InputBox`.  Maybe something like 

Title: "Enter match text"

Prompt: "The matched portion of the file name will be the part that gets modified. The rest will remain unchanged.  Leave blank to capture entire file name"

packages/plugin-core/src/commands/RefactorHierarchyV2.ts
```diff
@@ -32,27 +55,119 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
...
+    return match;
+  }
+
+  async promptReplaceText() {
+    let done = false;
     let replace: string | undefined;
+    do {
+      // eslint-disable-next-line no-await-in-loop
+      replace = await VSCodeUtils.showInputBox({
+        prompt: "Enter replace text.",
```
jonathanyeung, line 122: Going along the lines of the previous comment:

Title: "Enter replace text."

Prompt: "This will replace the matched portion of the file name.  Regex is supported, including capture groups."

Placeholder: contents here depend on the scope set in stage 1.  If the user is trying to do the standard, current refactor op, we can probably leave this blank. Or maybe something like `new.hierarchy`.  If the user is trying the prefix refactor on selective note selection, then we can have a placeholder with the regex capture group syntax.

hikchoi: I agree. I think it makes sense to do placeholders in the future for refactor-derivative commands that have very specific use cases, but for the vanilla refactor it's non-trivial to figure out a sensible placeholder for the replace stage

packages/plugin-core/src/commands/RefactorHierarchyV2.ts
hikchoi: fixed

hikchoi:
UPDATE:
1. On scoping stage, pressing escape will now cancel the command
2. On scoping stage, when no query input, a quickpick item "Entire Workspace" is sticked to the top.
  - We can now pass `extra items` that we can sticky to the top when there is no query.
3. On scoping stage, if the user has selection in the active editor that contains at least one wikilink, `selection2Items` and `multiSelect` is pressed by default
4. On scoping stage, no placeholder query is set
5. Now lookup derivative commands can optionally set custom titles for the quickpick (used to not be the case, this will fix every command that did pass in the title)
6. Match stage and Replace stage now provides more useful information.

image

image

image

kevinslin (changes requested):
1. By default, can we have everything checked in the quickpick and wikilinks are selected?
2. When links are selected, the option to select all workspaces goes away. the scope step is still a bit confusing because:
    - by default, you can "select entire workspace" **OR** individual notes (the latter isn't a frequent use case)
    - if you enable multi-select, then you can select "entire workspace" **AND** individual notes which also seems counterintuitive
    - I'm thinking the more intuitive interface should be:
        - the scope selector shows two options:
            - "Refactor over Entire Workspace"
            - "Refactor over specifc notes"
        - if the user chooses the latter, we go into `multi-select` and quickpick enabled
            - but this isn't a great experience until we get multi-select to pin selected items to the top 
    - given the current limitations, what do you think about the following?
        - if the user doesn't have wikilinks selected, don't even show the scope input but refactor over entire workspace (current refactor interface)
        - if the user has wikilinks selected, then show multi-select (without the option to select scope) 
    - if the user wants to refactor the entire workspace, they run command without wikilink selected
    - if the user wants to refactor selection, then they do the selection
    - for phase 2, we introduce scoping input when we have "sticky" selection feature shipped

hikchoi:
UPDATE:
1. scoping stage is skipped and defaults to entire workspace if there is no selection or the selection does not have any wikilinks
2. when refactoring from selection, `selection2Items` and `multiSelect` is now not togglable
3. when refactoring from selection, all items are selected on quickpick creation.
4. fixed a bug where triggering refactor when active note is not a markdown note would fail.

The _extra_ quickpick item `Entire Workspace` is not used, but preserved to be used in the next phase, where we introducing scoping through lookup by adding "sticky mode" and vault scoping.
For the same reason, the ability to pass in extra quickpick items to the provider is preserved even though it is not used anywhere yet.

kevinslin (changes requested):
Getting an error when refactoring in workspace from selection. See
https://www.loom.com/share/8b04abb0ff7745f89c1df64ccf91064f

and daily.journal.2021.10.26.md

hikchoi:
> Getting an error when refactoring in workspace from selection. See https://www.loom.com/share/8b04abb0ff7745f89c1df64ccf91064f
> 
> and daily.journal.2021.10.26.md

`selection2Items` was not deduping the wikilinks that point to the same note. Just pushed the fix as well as test for deduping.

hikchoi:
@Harshita-mindfire it could be worth hooking up a code action for refactor now that we can use selections to limit the scope of refactoring. any thought on that idea?

PR #1805, opened 2021-11-29
fix: fixing navigation to notes from preview that do not have vault specified

nickolay-kondratyev:
# fix: fixing navigation to notes from preview that do not have vault specified

Now when clicking on a link in preview that does not specify a vault (but lives in a different) vault we will navigate to the note. IF there are multiple notes in different vaults that match THEN we will choose the note with the oldest create time. 

IF there is a note with same file name in current vault we will choose the note in the same vault above all else. 

### Quality Assurance

### DIFF
--- a/packages/common-test-utils/src/noteUtils.ts
+++ b/packages/common-test-utils/src/noteUtils.ts
@@ -6,6 +6,7 @@ import {
   SchemaModuleProps,
   SchemaUtils,
   DNodePropsQuickInputV2,
+  NotePropsDict,
 } from "@dendronhq/common-all";
 import {
   file2Note,
@@ -88,6 +89,16 @@ export class TestNoteFactory {
       noteProps.push(await this.createForFName(fnames[i]));
     }
     return noteProps;
+  }
+
+  toNotePropsDict(notes: NoteProps[]): NotePropsDict {
+    const dict: NotePropsDict = {};
+
+    for (const note of notes) {
+      dict[note.id] = note;
+    }
+
+    return dict;
   }
 }
 
--- a/packages/plugin-core/src/commands/ShowPreview.ts
+++ b/packages/plugin-core/src/commands/ShowPreview.ts
@@ -8,6 +8,7 @@ import {
   ErrorFactory,
   getStage,
   NoteProps,
+  NotePropsDict,
   NoteUtils,
   NoteViewMessage,
   NoteViewMessageEnum,
@@ -29,6 +30,7 @@ import { VSCodeUtils, WSUtils } from "../utils";
 import { getEngine, getExtension } from "../workspace";
 import { BasicCommand } from "./base";
 import { GotoNoteCommand } from "./GotoNote";
+import { QuickPickUtil } from "../utils/quickPick";
 
 type CommandOpts = {};
 type CommandOutput = any;
@@ -84,7 +86,7 @@ export const extractNoteIdFromHref = (data: {
   href?: string;
 }): string | undefined => {
   if (data.href === undefined) {
-    return undefined;
+    throw ErrorFactory.createInvalidStateError({ message: `href is missing.` });
   }
   // For some cases such as markdown value='[head2](#head2)' the href isn't as nice
   // and looks like href='http://localhost:3005/vscode/note-preview.html?ws=WS-VALUE&port=3005#head2'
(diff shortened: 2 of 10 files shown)


packages/plugin-core/src/commands/ShowPreview.ts
```diff
@@ -216,6 +217,68 @@ const handleAssetLink = async ({
...
+    // If we could not find the note by the extracted id there is a good chance that the name
+    // of the note was in place of the id in the HREF (as in case of navigating to a note
+    // in a different vault without explicit vault specification). Hence we will attempt
+    // to find the note by file name.
+    const candidates = NoteUtils.getNotesByFname({ fname: noteId, notes });
+
+    // In case of multiple candidates matching the note we will grab the one with oldest (smallest)
+    // create date to make the navigation decently predictable (the predictability
+    // can be thrown off if a new vault is added to the workspace with matching name and older note).
+    if (candidates.length > 0) {
```
jonathanyeung, line 271: Do you think it's worth warning the user if we have more than 1 candidate? Like an info box saying something along the links of "Ambiguous Link - ${noteCount} notes exist with this name across ${vaultCount} vaults. Defaulting to oldest note."

packages/plugin-core/src/commands/ShowPreview.ts
```diff
@@ -216,6 +217,68 @@ const handleAssetLink = async ({
...
+      note = candidates.reduce((n1, n2) => (n1.created < n2.created ? n1 : n2));
```
jonathanyeung, line 272: Does it make sense to alter the logic to the following steps (or will this already be taken into account by the logic above `note === undefined`)?
1. Prefer same vault as current note, if one of the candidates is in the current vault.
2. Otherwise pick oldest note.

jonathanyeung (changes requested):

packages/plugin-core/src/commands/ShowPreview.ts
nickolay-kondratyev: The downside of the info box, that it doesn't go away on its own. But yea could be worthwhile to add. 

Was also considering showing a quick pick with the vaults however, that seemed a bit odd user experience to attempt to prop open a quick pick of vaults after a click.

packages/plugin-core/src/commands/ShowPreview.ts
nickolay-kondratyev: Yea that is already taken care off by `note === undefined` adding into a comment to clarify.

packages/plugin-core/src/commands/ShowPreview.ts
jonathanyeung: An alternative (and perhaps better) way we could handle this is to have a code warning whenever there's an ambiguous link; maybe even with a context UI for a quick right click > fix (ask user to specify the vault, and then we format the link for them).

nickolay-kondratyev: Yep the auto fix makes sense for subsequent versions.

For now added the info message when we are navigating to a link with duplicates in a different vault.

Did not add info message when navigating to a link in current vault (that has duplicates in other vault) since we cannot distinguish from the data that comes to us whether the vault was explicitly specified in the link or whether it is in the current vault link.

packages/common-test-utils/src/noteUtils.ts
```diff
@@ -89,6 +90,16 @@ export class TestNoteFactory {
     }
     return noteProps;
   }
+
+  toNotePropsDict(notes: NoteProps[]): NotePropsDict {
+    const dict: NotePropsDict = {};
+
+    for (const note of notes) {
```
kevinslin, line 97: nitpick -> for iteration, generally prefer `notes.forEach( note => {})` syntax as its more terse and common with es6

packages/plugin-core/src/commands/ShowPreview.ts
```diff
@@ -216,6 +217,75 @@ const handleAssetLink = async ({
...
+    // to find the note by file name.
+    const candidates = NoteUtils.getNotesByFname({ fname: noteId, notes });
+
+    if (candidates.length === 1) {
+      note = candidates[0];
+    } else if (candidates.length > 1) {
+      // In case of multiple candidates matching the note we will grab the one with oldest (smallest)
+      // create date to make the navigation decently predictable (the predictability
+      // can be thrown off if a new vault is added to the workspace with matching name and older note).
+      note = candidates.reduce((n1, n2) => (n1.created < n2.created ? n1 : n2));
```
kevinslin, line 275: its not intuitive to me that the older timestamp should be the valid note (eg. you could also add an existing vault with a note that was created with an earlier timestamp). as an alternative, what if we show the user a quickpick with the possible options? if you look at https://github.com/dendronhq/dendron/pull/1796, @hikchoi is adding the ability to create a lookup with only a few notes selected, seems like it would make a good use case for this

kevinslin (changes requested):

packages/plugin-core/src/commands/ShowPreview.ts
nickolay-kondratyev: Added quick pick choice if there are multiple notes in other vaults.

PR #1811, opened 2021-11-30
enhance: add fuzzy threshold configuration

nickolay-kondratyev:
# enhance: add fuzzy threshold configuration

Allows users to configure the fuzziness threshodl of lookups. 

PR doc: https://github.com/dendronhq/dendron-site/pull/303/files

### Quality Assurance
  * Tested with configuration
  * Tested without adding new configuration
  * Tested with note configuration entirely.

### DIFF
--- a/packages/common-all/src/constants/configs/commands.ts
+++ b/packages/common-all/src/constants/configs/commands.ts
@@ -61,6 +61,10 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     label: "Bubble Up Create New",
     desc: "Put `Create New` at the top of the lookup result.",
   },
+  fuzzThreshold: {
+    label: "Fuzziness threshold for lookup",
+    desc: "Sets the fuzziness level of lookups 0=exact matches only, 1=max fuzziness.",
+  },
 };
 
 /**
--- a/packages/common-all/src/fuse.ts
+++ b/packages/common-all/src/fuse.ts
@@ -28,43 +28,6 @@ export const FuseExtendedSearchConstants = {
   PrefixExactMatch: "^",
 };
 
-/**
- * Experimentally set.
- *
- * At the time of testing:
- *
- * At previous threshold of 0.5 string 'dendron' matched
- * 'scratch.2021.06.15.104331.make-sure-seeds-are-initialized-on-startup' with score 0.42.
- * Which is too fuzzy of a match.
- *
- * 'rename' fuzzy matches 'dendron.scratch.2020.11.07.publish-under-original-filenames' with 0.16.
- *
- * For reference
- * 'dendron rename' matches 'dendron.dev.design.commands.rename' with 0.001.
- *
- * Having this score too high gets too unrelated matches which pushes the
- * 'Create New' entry out of the view.
- * --------------------------------------------------------------------------------
- *
- * Note if you are going to be tweaking this value it is highly suggested to add a
- * temporary piece of code To be able to see the all the results that are matched by
- * fuse engine along with their scores, inside {@link FuseEngine.queryNote}
- * */
-//       const dir = `<YOUR-DIR>/${qs}`;
-//       try{
-//         require('fs').mkdirSync(dir)
-//       }catch (e){
-//       }
-//       const data = JSON.stringify(
-//         {
-//           qs: qs,
-//           fuseQueryString: fuseQueryString,
-//           results:results
-//         });
-//       const path = `${dir}/${THRESHOLD_VALUE}_${new Date().getTime()}.json`;
-//       require('fs').writeFile(path, data, ()=>{});
-export const DEFAULT_THRESHOLD_VALUE = 0.2;
-
 function createFuse<T>(
   initList: T[],
   opts: Fuse.IFuseOptions<T> & {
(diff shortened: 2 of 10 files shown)


packages/common-all/src/fuse.ts
```diff
@@ -63,7 +63,7 @@ export const FuseExtendedSearchConstants = {
 //         });
 //       const path = `${dir}/${THRESHOLD_VALUE}_${new Date().getTime()}.json`;
 //       require('fs').writeFile(path, data, ()=>{});
-const THRESHOLD_VALUE = 0.2;
+export const DEFAULT_THRESHOLD_VALUE = 0.2;
```
kevinslin, line 66: instead of having a hardcoded value here, can you add it as a config value via [[Add New Config|dendron://dendron.docs/pkg.common-all.dev.cook#add-new-config]]?

this should also fix the test failing

kevinslin (changes requested):

packages/common-all/src/fuse.ts
hikchoi: +1 on @kevinslin's suggestion. I think removing the constant and having `fuzzThreshold` config required and default to `0.2` would be better.

packages/common-all/src/types/configs/commands/lookup.ts
```diff
@@ -27,6 +27,7 @@ export type NoteLookupConfig = {
   confirmVaultOnCreate: boolean;
   leaveTrace: boolean;
   bubbleUpCreateNew: boolean;
+  fuzzThreshold?: number;
```
hikchoi, line 30: the fuse engine's fuzz threshold affects all lookup derivative commands. Should this config be in `commands.lookups.note`?

hikchoi: That makes me think... a lot of the configs for note lookup also affects other commands. Not sure what the most natural way to namespace this would be.

packages/engine-test-utils/src/__tests__/common-all/fuse.spec.ts
```diff
@@ -5,6 +5,10 @@ import {
 } from "@dendronhq/common-all";
 import { NoteTestUtilsV4 } from "@dendronhq/common-test-utils";
 import Fuse from "fuse.js";
+import {
+  DEFAULT_THRESHOLD_VALUE,
+  getThresholdValue,
+} from "@dendronhq/common-all/src";
```
hikchoi, line 11:
```suggestion
} from "@dendronhq/common-all";
```

hikchoi (changes requested):
on failing tests: 

from [[Cookbook|dendron://dendron.docs/pkg.common-all.dev.cook]]:

  - When adding a config key to the new namespace type(s), you also have to add a corresponding entry in the [DendronConfigEntryCollection](https://github.com/dendronhq/dendron/blob/6a7be61db3ec7e6fab61871b30ec215c47f1cb59/packages/common-all/src/constants/configs/dendronConfig.ts#L10)
      - This is an object that holds every possible config key's label and description that will later be used to automatically generate a configuration view.
      - If this step is omitted, Typescript will complain that `DendronConfigEntryCollection` is missing a key.

packages/common-all/src/types/configs/commands/lookup.ts
nickolay-kondratyev: Yes, it would affect the other commands (also right now as is currently this affects the schema look up as well).

packages/common-all/src/fuse.ts
nickolay-kondratyev: Added default config value

packages/common-all/src/fuse.ts
```diff
@@ -132,6 +95,18 @@ export type SerializedFuseIndex = ReturnType<
 
 type FuseEngineOpts = {
   mode?: DEngineMode;
+  /** If specified must be within 0-1 range. */
+  fuzzThreshold: number;
+};
+
+export const getThresholdValue = (configThreshold: number) => {
+  // Setting it to fallback threshold value in case configuration is incorrect.
+  let threshold = 0.2;
```
kevinslin, line 104: instead of hardcoding, can we use `ConfigUtils.getDefaults()` so we get the fallback value from the config?

packages/common-all/src/fuse.ts
```diff
@@ -132,6 +95,18 @@ export type SerializedFuseIndex = ReturnType<
...
+  if (configThreshold >= 0 && configThreshold <= 1) {
+    threshold = configThreshold;
```
kevinslin, line 106: nitpick: instead of assignment, can we return the default threshold if the threshold is bad? otherwise, return the regular threshold value

packages/common-all/src/fuse.ts
```diff
@@ -132,6 +95,18 @@ export type SerializedFuseIndex = ReturnType<
 
 type FuseEngineOpts = {
   mode?: DEngineMode;
+  /** If specified must be within 0-1 range. */
+  fuzzThreshold: number;
+};
+
+export const getThresholdValue = (configThreshold: number) => {
```
kevinslin, line 102: nitpick: for this, can we rename to `getCleanThresholdValue` to indicate that this value is sanitized/validated?

packages/common-all/src/constants/configs/commands.ts
```diff
@@ -61,6 +61,10 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     label: "Bubble Up Create New",
     desc: "Put `Create New` at the top of the lookup result.",
   },
+  fuzzThreshold: {
+    label: "Fuzziness threshold for lookup",
+    desc: "Sets the fuzziness level of lookups 0=exact matches only, 1=max fuzziness.",
```
kevinslin, line 66: @hikchoi for a future phase, we should expose json schema variables here so we can do validation (eg. fuzzThreshold range must be btw 0 and 1) when user saves

kevinslin (changes requested):

packages/common-all/src/fuse.ts
nickolay-kondratyev: Do we have a method like that already, doesnt come up for me when I do symbol search.

kevinslin: i was going off memory. the method is `genDefaultConfig`

packages/common-all/src/fuse.ts
nickolay-kondratyev: Sure will change to just return. Although, that can be up for debate on what's better given that its typically suggested go towards limiting return statements from the function to limit the escape points and hence typically making it easier to debug (this function is as simple as it comes so a bit of moot point on debugging).

Issue #1675, opened 2021-11-09
Schema note template overrides selection extract


aleksey-rowan:
## Please select the area your bug applies to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x] Misc (Choose this if your not sure)

## Describe the bug

When creating a new note using the `selection extract` lookup modifier, the schema note template overrides the extracted block.

## To Reproduce

1. Select a block of text in any note
2. `Ctrl + L` and tick the `selection extract` modifier
3. Create a new note in a hierarchy that has schema templates
4. Observe that selected extract was not inserted into the newly created note (the template was)

## Expected behavior

The selected extract should be prepended or appended to the note template. I think placing the extract at the beginning of the file is slightly better.

## Desktop (please complete the following information)

- Version: 0.66

Harshita-mindfire:
Thanks @aleksey-rowan for reporting the issue. I can repro this one and I agree placing the extract at the beginning of the file sounds better.

kevinslin:
this should be fixed in v72

aleksey-rowan:
Can confirm it's fixed. Thank you!

PR #1812, opened 2021-11-30
fix(schema): When applying a schema template, do not override the body but append to the end to it

tma66:
* Updated tests for dnode.ts
* Minor lint fixes in dnode.spec.ts
* Append template note's body if current note's body is not empty
* See https://github.com/dendronhq/dendron/issues/1675

Unit Test
*  ./node_modules/.bin/jest /Users/tulingma/workspace/dendron/packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts

Manual Test
* Run "Run:Extension Local (plugin-core) against test workspace
* Test creating note with template
* Test creating note with template with selection extract

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1262,8 +1262,13 @@ export class SchemaUtils {
       }
       const tempNoteProps = _.pick(tempNote, this.TEMPLATE_COPY_PROPS);
       _.forEach(tempNoteProps, (v, k) => {
-        // @ts-ignore
-        note[k] = v;
+        // If note body exists, append template's body instead of overriding
+        if (k === "body" && note[k] !== "") {
+          note[k] += `\n${v}`;
+        } else {
+          // @ts-ignore
+          note[k] = v;
+        }
       });
       return true;
     }
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1241,7 +1241,6 @@ type SchemaMatchResult = {
 export class SchemaUtils {
   /** The props of a template note that will get copied over when the template is applied. */
   static TEMPLATE_COPY_PROPS: readonly (keyof NoteProps)[] = [
-    "body",
     "desc",
     "custom",
     "color",
@@ -1262,14 +1261,15 @@ export class SchemaUtils {
       }
       const tempNoteProps = _.pick(tempNote, this.TEMPLATE_COPY_PROPS);
       _.forEach(tempNoteProps, (v, k) => {
-        // If note body exists, append template's body instead of overriding
-        if (k === "body" && note[k] !== "") {
-          note[k] += `\n${v}`;
-        } else {
-          // @ts-ignore
-          note[k] = v;
-        }
+        // @ts-ignore
+        note[k] = v;
       });
+      // If note body exists, append template's body instead of overriding
+      if (note.body) {
+        note.body += `\n${tempNote.body}`;
+      } else {
+        note.body = tempNote.body;
+      }
       return true;
     }
     return false;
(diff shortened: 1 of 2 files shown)


packages/common-all/src/dnode.ts
```diff
@@ -1262,8 +1262,13 @@ export class SchemaUtils {
       }
       const tempNoteProps = _.pick(tempNote, this.TEMPLATE_COPY_PROPS);
       _.forEach(tempNoteProps, (v, k) => {
-        // @ts-ignore
-        note[k] = v;
+        // If note body exists, append template's body instead of overriding
```
kevinslin, line 1265: instead of introducing an if statement here, can we remove `body` from `TEMPLATE_COPY_PROPS` and add the logic outside of the for loop?

packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts
```diff
@@ -4,14 +4,16 @@ import {
...
+import { runEngineTestV5 } from "../../engine";
+import { ENGINE_HOOKS } from "../../presets";
 
 describe(`NoteUtils tests:`, () => {
   describe(`genSchemaDesc tests`, () => {
     const vault = { fsPath: "/tmp/ws/vault1" };
     const SCHEMA_ID = "id-1";
 
-    async function test_genSchemaDesc(
+    async function testGenSchemaDesc(
```
kevinslin, line 16: thanks for fixing this!

packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts
```diff
@@ -95,3 +97,74 @@ describe(`NoteUtils tests:`, () => {
     });
   });
 });
+
+describe(`SchemaUtil tests:`, () => {
```
kevinslin, line 101: heads up for the tests: we follow a bdd style that is outlined here -> [[Style|dendron://dendron.docs/dev.process.qa.style]]

packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts
```diff
@@ -95,3 +97,74 @@ describe(`NoteUtils tests:`, () => {
...
+          const note = await noteFactory.createForFName("new note");
+
+          const resp = SchemaUtils.applyTemplate({
+            template: { id: "foo", type: "note" },
+            note,
+            engine,
+          });
+          expect(resp).toBeTruthy();
+          expect(note.body).toEqual(engine.notes["foo"].body);
+          return [];
```
kevinslin, line 116: you don't actually need to return anything here

packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts
```diff
@@ -95,3 +97,74 @@ describe(`NoteUtils tests:`, () => {
...
+          note.body = noteBody;
+
+          const resp = SchemaUtils.applyTemplate({
+            template: { id: "foo", type: "note" },
+            note,
+            engine,
+          });
+          expect(resp).toBeTruthy();
+          expect(note.body).toEqual(noteBody + "\n" + engine.notes["foo"].body);
+          return [];
```
kevinslin, line 140: see above

packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts
```diff
@@ -95,3 +97,74 @@ describe(`NoteUtils tests:`, () => {
...
+          const noteFactory = TestNoteFactory.defaultUnitTestFactory();
+          const note = await noteFactory.createForFName("new note");
+
+          const resp = SchemaUtils.applyTemplate({
+            template: { id: "foo", type: "snippet" },
+            note,
+            engine,
+          });
+          expect(resp).toBeFalsy();
+          return [];
```
kevinslin, line 161: see above

kevinslin (changes requested):

packages/common-all/src/dnode.ts
tma66: Sure that can be done. I'll still need to keep the if logic to check if body exists since that determines whether we add a \n or not. That work?

PR #1842, opened 2021-12-05
fix(views): update web uis on note creation

kevinslin:
Newly created notes were not synced with the view, leading to components
like tree view not updating when a new view was added


SeriousBug (approved):
Code looks good to me, and I can confirm the tree view updates when a note is created.

PR #1844, opened 2021-12-06
feat: Goto Note can open links to non-note files

SeriousBug:
This works by writing relative paths from the `wsRoot` in wikilinks. For example,
`[[/src/index.js]]` will open `${wsRoot}/src/index.js`.

- `[[/src/index.js]]` and `[[src/index.js]]` will function in the same way to maintain compatibility with how Dendron handles wikilinks.
- If the path starts with `/assets`, then it's handled in a special way where we'll instead search for it in `assets` folders within each vault first. This maintains compatibility with how assets work in publishing.
- This only works for existing files, creating non-existing files is not supported.
- If a note and a file with the same name exist (e.g. there's a `file.txt` under `wsRoot` and `file.txt.md` in some vault, and link is `[[file.txt]]`), then the note gets priority to avoid breaking existing links.

Docs: https://github.com/dendronhq/dendron-site/pull/313

### DIFF
--- a/packages/common-server/src/filesv2.ts
+++ b/packages/common-server/src/filesv2.ts
@@ -453,4 +453,60 @@ export function dot2Slash(fname: string) {
   return fname.replace(/\./g, "/");
 }
 
+/** Checks that the `path` contains a file. */
+export async function fileExists(path: string) {
+  try {
+    const stat = await fs.stat(path);
+    return stat.isFile();
+  } catch {
+    return false;
+  }
+}
+
+/** Finds if a file `fpath` is located in any vault.
+ *
+ * @param fpath A file name or relative path that we are searching inside vaults.
+ */
+async function findFileVault({
+  fpath,
+  wsRoot,
+  vaults,
+}: {
+  fpath: string;
+  wsRoot: string;
+  vaults: DVault[];
+}): Promise<{ vault: DVault; fullPath: string } | undefined> {
+  // Assets from later vaults will overwrite earlier ones.
+  vaults = [...vaults].reverse();
+  for (const vault of vaults) {
+    const fullPath = path.join(wsRoot, vault.fsPath, fpath);
+    // Doing this sequentially to simulate how publishing handles conflicting assets.
+    // eslint-disable-next-line no-await-in-loop
+    if (await fileExists(fullPath)) {
+      return { vault, fullPath };
+    }
+  }
+  return;
+}
+
+export async function findNonNoteFile(opts: {
+  fpath: string;
+  wsRoot: string;
+  vaults: DVault[];
+}): Promise<{ vault?: DVault; fullPath: string } | undefined> {
+  let { fpath } = opts;
+  // Especially for assets, `/assets` and `assets` refers to the same place.
+  fpath = _.trim(fpath, "/\\");
+  // Check if this is an asset first
+  if (fpath.startsWith("assets")) {
+    const out = await findFileVault(opts);
+    if (out !== undefined) return out;
+  }
+  // If not an asset, or if we couldn't find it in assets, then check from wsRoot for out-of-vault files
+  const fullPath = path.join(opts.wsRoot, fpath);
+  if (await fileExists(fullPath)) return { fullPath };
+  // Otherwise, it just doesn't exist
+  return undefined;
+}
+
 export { tmp, DirResult };
--- a/packages/common-server/src/filesv2.ts
+++ b/packages/common-server/src/filesv2.ts
@@ -467,7 +467,7 @@ export async function fileExists(path: string) {
  *
  * @param fpath A file name or relative path that we are searching inside vaults.
  */
-async function findFileVault({
+async function findFileInVault({
   fpath,
   wsRoot,
   vaults,
@@ -479,7 +479,7 @@ async function findFileVault({
   // Assets from later vaults will overwrite earlier ones.
   vaults = [...vaults].reverse();
   for (const vault of vaults) {
-    const fullPath = path.join(wsRoot, vault.fsPath, fpath);
+    const fullPath = path.join(wsRoot, VaultUtils.getRelPath(vault), fpath);
     // Doing this sequentially to simulate how publishing handles conflicting assets.
     // eslint-disable-next-line no-await-in-loop
     if (await fileExists(fullPath)) {
@@ -499,7 +499,7 @@ export async function findNonNoteFile(opts: {
   fpath = _.trim(fpath, "/\\");
   // Check if this is an asset first
   if (fpath.startsWith("assets")) {
-    const out = await findFileVault(opts);
+    const out = await findFileInVault(opts);
     if (out !== undefined) return out;
   }
   // If not an asset, or if we couldn't find it in assets, then check from wsRoot for out-of-vault files
--- a/packages/plugin-core/src/commands/GotoNote.ts
+++ b/packages/plugin-core/src/commands/GotoNote.ts
@@ -250,7 +250,9 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
 
     // Non-note files use `qs` for full path, and set vault to null
     if (vault === null) {
-      await VSCodeUtils.openFileInEditor(Uri.parse(qs));
+      await VSCodeUtils.openFileInEditor(
+        Uri.from({ scheme: "file", path: qs })
+      );
       return {
         kind: "nonNote",
         fullPath: qs,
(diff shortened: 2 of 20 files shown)


packages/common-server/src/filesv2.ts
```diff
@@ -453,4 +453,60 @@ export function dot2Slash(fname: string) {
...
+  vaults,
+}: {
+  fpath: string;
+  wsRoot: string;
+  vaults: DVault[];
+}): Promise<{ vault: DVault; fullPath: string } | undefined> {
+  // Assets from later vaults will overwrite earlier ones.
+  vaults = [...vaults].reverse();
+  for (const vault of vaults) {
+    const fullPath = path.join(wsRoot, vault.fsPath, fpath);
```
kevinslin, line 482: heads up that the `fullPath` changes depending if we have a workspace vault or seed vault. see https://github.com/dendronhq/dendron-docs/blob/main/vault/pkg.dendron-engine.dev.cook.md#getting-absolute-path-for-a-vault

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -180,13 +211,36 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
     const client = getExtension().getEngine();
+    const { wsRoot } = getDWorkspace();
 
     const { qs, vault } = (await this.processInputs(opts)) || opts;
     if (_.isUndefined(qs) || _.isUndefined(vault)) {
       // There was an error or the user cancelled a prompt
       return;
     }
 
+    const nonNoteFile = await findNonNoteFile({
```
kevinslin, line 222: don't we already do this in `processInputs`? why do this a second time?

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -135,27 +151,42 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
-            vault: currentVault,
-            fname: opts.qs,
-            vaultSelectionMode: selectionMode,
+          // Not an existing note
+          const nonNoteFile = await findNonNoteFile({
+            fpath: opts.qs,
+            wsRoot: engine.wsRoot,
+            vaults: engine.vaults,
           });
+          if (nonNoteFile?.vault !== undefined) {
```
kevinslin, line 160: can we be explicit about returning the kind of note here? so either a "note" or a "nonNote"?

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -33,7 +37,10 @@ type CommandOpts = {
 export { CommandOpts as GotoNoteCommandOpts };
 
 type CommandOutput =
-  | { note: NoteProps; pos?: Position; source?: string }
+  // When opening a note
+  | { type: "note"; note: NoteProps; pos?: Position; source?: string }
```
kevinslin, line 41: since note types will soon be a thing, can we change the key here to `kind` to avoid confusion?

kevinslin: also, can we use `nonNote` as the kind? prefer camelcase since we can use them in javascript without quoting

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -135,27 +151,42 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
-              ? VaultSelectionMode.smart
-              : VaultSelectionMode.alwaysPrompt;
-
-          const currentVault = PickerUtilsV2.getVaultForOpenEditor();
-          const selectedVault = await PickerUtilsV2.getOrPromptVaultForNewNote({
-            vault: currentVault,
-            fname: opts.qs,
-            vaultSelectionMode: selectionMode,
+          // Not an existing note
+          const nonNoteFile = await findNonNoteFile({
```
kevinslin, line 155: `processInputs` is getting quite large. can we extract processing `nonNote` into its own method?

kevinslin (changes requested):

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -78,9 +88,97 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+    if (notes.length === 1) {
+      // There's just one note, so that's the one we'll go with.
+      opts.vault = notes[0].vault;
+    } else if (notes.length > 1) {
+      // It's ambiguous which note the user wants to go to, so we have to
+      // guess or prompt.
+      const resp = await PickerUtilsV2.promptVault(
+        notes.map((ent) => ent.vault)
+      );
+      if (_.isUndefined(resp)) return null;
```
kevinslin, line 134: since the normal pattern for cancelled input is `undefined` for quickpick related events, why not just pass through `undefined` here?

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -78,9 +88,97 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+          wsRoot,
+          vaults,
+          fsPath: opts.qs,
+        });
+      }
+    }
+    return opts;
+  }
+
+  private async getExistingNote(opts: CommandOpts) {
```
kevinslin, line 119: can we rename this? eg. `maybeSetOptsFromExistingNote`

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -91,73 +189,33 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+      const existingNote = await this.getExistingNote(opts);
+      // User cancelled prompt
+      if (existingNote === null) return null;
+      opts = existingNote;
+    }
+    if (opts.vault === undefined) {
+      opts = await this.getExistingNonNoteFile(opts);
+    }
+    if (opts.vault === undefined) {
+      const newNote = await this.getNewNote(opts);
```
kevinslin, line 215: rename to `setOptsFromNewNote`

packages/plugin-core/src/test/suite-integ/GotoNote.test.ts
```diff
@@ -630,13 +633,44 @@ suite("GotoNote", function () {
...
+          expect(getActiveEditorBasename()).toEqual("test.txt");
+          expect(
+            VSCodeUtils.getActiveTextEditorOrThrow().document.getText().trim()
+          ).toEqual("Et voluptatem autem sunt.");
+          done();
+        },
+      });
+    });
+
+    test("non-note file, without slash in link", (done) => {
```
kevinslin, line 651: is there a reason we're not using the new style workspace test here?

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -85,9 +88,97 @@ export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+        opts.qs = note.fname;
+        opts.vault = note.vault;
+      } else {
+        const { wsRoot, vaults } = getDWorkspace().engine;
+        // Same file link within non-note file
+        opts.qs = path.relative(
+          wsRoot,
+          VSCodeUtils.getActiveTextEditorOrThrow().document.fileName
+        );
+        opts.vault = VaultUtils.getVaultByNotePath({
```
kevinslin, line 109: will this throw an error if `opts.qs` isn't a note?

kevinslin (changes requested):

packages/plugin-core/src/commands/GotoNote.ts
SeriousBug: No, it works correctly. It's just computing it based on the file name. I'll go ahead and rename the function since it's serving both purposes now.

packages/plugin-core/src/commands/GotoNote.ts
SeriousBug: If the user cancelled the prompt, then we want to stop and not try the other options. So we have to differentiate between "couldn't find a note" and "found multiple notes, but user cancelled the prompt". I thought returning null might be a good way to signal that, it's not just unset (undefined) but explicitly refused.

packages/plugin-core/src/test/suite-integ/GotoNote.test.ts
SeriousBug: This wouldn't have benefitted from the abilities of the new test style to share the same workspace across multiple tests, so I chose to go with the simpler option of copying and modifying existing tests. I'll go ahead and revise them though, you're right that we should still use the new style for new tests.

kevinslin (approved):

PR #1854, opened 2021-12-07
enhance(publish): Fallback to default SEO image if no image is set for published pages.

tma66:
* Updated image field of DendronSiteConfig to represent default SEO image
* See [[Fallback Seo|dendron://private/task.publish.2021.11.24.fallback-seo]]

Manual Test
* Update dendron.yml to include default image field
* Run NextJs against test workspace. [[Test|dendron://dendron.docs/pkg.nextjs-template.qa.test]]
* Open http://localhost:3000/notes/boy2/ and view page source
* Verified image in meta property="og:image"

### DIFF
--- a/packages/nextjs-template/components/DendronSEO.tsx
+++ b/packages/nextjs-template/components/DendronSEO.tsx
@@ -1,10 +1,11 @@
 import {
-  IntermediateDendronConfig,
+  ConfigUtils,
   DendronSiteConfig,
+  IntermediateDendronConfig,
   NoteProps,
+  PublishUtils,
   SEOProps,
   Time,
-  PublishUtils,
 } from "@dendronhq/common-all";
 import _ from "lodash";
 import { NextSeo, NextSeoProps } from "next-seo";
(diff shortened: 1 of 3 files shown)


packages/nextjs-template/components/DendronSEO.tsx
```diff
@@ -68,7 +70,8 @@ export default function DendronSEO({
 
   const title = cleanSeoProps.title;
   const description = cleanSeoProps.excerpt;
-  const images = cleanSeoProps?.image ? [cleanSeoProps.image] : [];
+  const defaultImages = siteConfig?.image ? [siteConfig.image] : [];
+  const images = cleanSeoProps?.image ? [cleanSeoProps.image] : defaultImages;
```
kevinslin, line 74: shouldn't line 69 already do this?
`_.defaults(noteSeoProps, siteSeoProps)`?

If not, you might want to use `_.defaultsDeep`

kevinslin (changes requested):

packages/nextjs-template/components/DendronSEO.tsx
tma66: It doesn't include `image` but I updated `getSEOPropsFromConfig` instead

kevinslin (approved):

Issue #963, opened 2021-07-13
Prompt user with more information and possible actions when using `Dendron: Delete Node`


hikchoi:
## Please Select if your Request is Either something new or an Enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [x] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##
Currently, A note deleted with `Dendron: Delete Node` will leave links to it untouched.

If `foo.md` is deleted, and `bar.md` had the link `[[foo]]`, `bar.md` will still point to `foo`.

Automatically modifying / deleting the existing links to a deleted note would not be preferable in many cases, but I think it is equally frustrating and confusing to unexpectedly stumble upon a link that leads to nothing.

There should be a safe option to bulk convert existing links to the note as well as a way to review what is going to be changed, and an option to not convert the links but still be reminded of what is going to be left after the deletion.

## Describe the solution you'd like ##
Prompt the user with a preview similar to what you would get in `Refactor Hierarchy`, with a list of links that would be left after deleting this note.

At this point the user can choose to do any of the following:
- cancel and use the list as a reference to which notes they should fix.
- convert aliased links into plain text of just the alias, and naked (no alias) links into the title of the note-to-be-deleted
  - display how it's going to be changed in the preview.
- delete anyways.

## Describe alternatives you've considered ##
Using the command as-is, but always checking the backlinks panel to see if the note has any outstanding links to be fixed before deleting.

hikchoi:
Issue resolved with #1885 and #1933

PR #1885, opened 2021-12-13
enhance: better delete node

hikchoi:
# enhance: better delete node
This PR:
- adds a preview of delete node when the command is run on a note that will create broken links if deleted

As a follow-up enhancement, various convenience commands to convert broken links will be added.

# Pull Request Checklist

### Quality Assurance
- [~] If your tests involve running [[manual Testing|dendron://dendron.dendron-site/dendron.dev.qa.test#manual-testing]], make sure to update [[Test Workspace|dendron://dendron.dendron-site/dendron.dev.ref.test-workspace]] with any necessary notes/additions to perform manual test ^9jtWc6ov340p

#### Special Cases

### Docs

### Git
        - `eg. feat/add-thisthing`

### Analytics
- [~] If you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Github Issue
- [ ] If this pull request is addressing an existing issue, make sure to link this PR to the issue that it is resolving.
- [ ] If the resolution comes with a document update, link the docs PR to the issue as well.

### DIFF
--- a/packages/plugin-core/src/commands/DeleteNodeCommand.ts
+++ b/packages/plugin-core/src/commands/DeleteNodeCommand.ts
@@ -3,6 +3,7 @@ import {
   DVault,
   EngineDeletePayload,
   NoteProps,
+  NotePropsDict,
   NoteUtils,
   SchemaUtils,
   VaultUtils,
@@ -18,6 +19,7 @@ import { VSCodeUtils } from "../vsCodeUtils";
 import { getDWorkspace, getEngine, getExtension } from "../workspace";
 import { BasicCommand } from "./base";
 import _md from "markdown-it";
+import fs from "fs-extra";
 
 type CommandOpts = {
   _fsPath?: string;
@@ -47,6 +49,33 @@ export class DeleteNodeCommand extends BasicCommand<
     return {};
   }
 
+  getBacklinkFrontmatterLineOffset(opts: {
+    link: DLink;
+    vaults: DVault[];
+    notes: NotePropsDict;
+    wsRoot: string;
+  }) {
+    const { link, vaults, notes, wsRoot } = opts;
+    const vault = VaultUtils.getVaultByName({
+      vaults,
+      vname: link.from.vaultName as string,
+    }) as DVault;
+    const noteWithLink = NoteUtils.getNoteByFnameV5({
+      fname: link.from.fname as string,
+      vault,
+      notes,
+      wsRoot,
+    }) as NoteProps;
+    const fsPath = NoteUtils.getFullPath({
+      note: noteWithLink,
+      wsRoot,
+    });
+    const fileContent = fs.readFileSync(fsPath).toString();
+    const lines = fileContent.split("\n");
+    lines.shift();
+    return lines.indexOf("---") + 2;
+  }
+
   async showNoteDeletePreview(note: NoteProps, backlinks: DLink[]) {
     let content = [
       "# Delete Node Preview",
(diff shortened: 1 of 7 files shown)


hikchoi:
related: #963

kevinslin (changes requested):
For the ux, there's two issues with this current implementation:
1. the line number's don't currently line up because it doesn't take into account the frontmatter
2. the common pattern after seeing this notice is to go to the files and fix the links. the problem is that we then lose access to the quickpick. we should make sure it doesn't dissapear on focus out

hikchoi:
> the line number's don't currently line up because it doesn't take into account the frontmatter

That's right. I should fix that.

> the common pattern after seeing this notice is to go to the files and fix the links. the problem is that we then lose access to the quickpick. we should make sure it doesn't dissapear on focus out

The quickpick already ignores focus out. But I think that doesn't make the experience different.
When the quickpick asks for confirmation, the user can either cancel or proceed with the deletion. Regardless of what they choose to do, the preview will be there and they can reference that to go find where the broken links are. 

Since we decided that we aren't going to introduce any bulk fix operations to delete node itself yet, delete node will only delete the current open note and show you a preview if there are broken links.

Besides, even if we make the quickpick ignore focus out, the user will eventually bring up the lookup quickpick to navigate to the notes to fix the broken links, which will hide the existing quickpick. As of now, I think we can keep it as is because whatever choices the user makes doesn't have any side effects related to broken links.

packages/plugin-core/src/commands/DeleteNodeCommand.ts
```diff
@@ -32,6 +38,8 @@ function formatDeletedMsg({
   return `${path.basename(fsPath)} (${VaultUtils.getName(vault)}) deleted`;
 }
 
+const md = _md();
```
kevinslin, line 41: can we load this inside of the command? in general, avoid having side effects on import (_md will also be initialized whenever delete node is imported)

kevinslin, line 41: can we load this inside of the command? in general, avoid having side effects on import (_md will also be initialized whenever delete node is imported)

packages/plugin-core/src/commands/DeleteNodeCommand.ts
```diff
@@ -41,6 +49,89 @@ export class DeleteNodeCommand extends BasicCommand<
...
+      fname: link.from.fname as string,
+      vault,
+      notes,
+      wsRoot,
+    }) as NoteProps;
+    const fsPath = NoteUtils.getFullPath({
+      note: noteWithLink,
+      wsRoot,
+    });
+    const fileContent = fs.readFileSync(fsPath).toString();
```
kevinslin, line 73: this is not the greatest heuristic since we could have quoted --- inside the frontmatter. 
we have frontmatter logic in a few places now that parse using unified. can we use that and also consolidate?
- [moveCursorPastFrontmatter](https://github.com/dendronhq/dendron/blob/8fd685d53e8dd31f19bcb6e8d3b5ded25d86a63d/packages/plugin-core/src/windowWatcher.ts#L164)
- [provideFoldingRanges](https://github.com/dendronhq/dendron/blob/1f0b25a112e3da630d38cfcf7966003cd8520984/packages/plugin-core/src/features/FrontmatterFoldingRangeProvider.ts#L17)

can we consolidate the frontmatter detection logic to [RemarkUtils](https://github.com/dendronhq/dendron/blob/90e083b5e10073acbc8967ad9649c0008aae381c/packages/engine-server/src/markdown/remark/utils.ts#L872) ?

kevinslin (changes requested):

kevinslin (approved):

PR #1890, opened 2021-12-13
DRAFT: scroll sync from editor to preview

nickolay-kondratyev:
## DRAFT: scroll sync from editor to preview

Main related note: ![[dendron://private/task.preview.scroll-in-sync-with-markdown]]
(At the bottom section: ![[dendron://private/task.preview.scroll-in-sync-with-markdown#user-triggered-sync,1:#*]] there is info on different steps of approach).

Good note to test this is ![[dendron.links.heading-anchors#heading-2,1]] in test workspace.

The approach is attempting to get cleaned version of the text from text editor focus and find the offset (how many elements in editor exist with such text) then find the offset element with the preview. However, the current challenge is that the links get rendered very differently (with diffrent text) which throws off this approach. 

Looking for suggestions if someone has a better approach to this problem or whether we should continue down this road and attempt to parse the links etc.. further for higher likelehood of correct match of offset of the element.

### DIFF
--- a/packages/plugin-core/src/commands/ShowPreview.ts
+++ b/packages/plugin-core/src/commands/ShowPreview.ts
@@ -291,6 +291,9 @@ export const handleLink = async ({
   }
 };
 
+// eslint-disable-next-line import/no-mutable-exports
+export let UI_SHOW_PREVIEW_CMD: ShowPreviewCommand | undefined;
+
 export class ShowPreviewCommand extends BasicCommand<
   CommandOpts,
   CommandOutput
@@ -301,6 +304,23 @@ export class ShowPreviewCommand extends BasicCommand<
   constructor(previewPanel: vscode.WebviewPanel) {
     super();
     this._panel = previewPanel;
+
+    if (UI_SHOW_PREVIEW_CMD === undefined) {
+      UI_SHOW_PREVIEW_CMD = this;
+    }
+  }
+
+  async syncEditorToPreviewPlacement(
+    cleanedText: string,
+    elementOffset: number
+  ) {
+    if (this._panel && this._panel.webview) {
+      this._panel.webview.postMessage({
+        type: "sync-editor-to-preview",
+        text: cleanedText,
+        elementOffset,
+      });
+    }
   }
 
   async sanityCheck() {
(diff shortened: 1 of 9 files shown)


kevinslin:
Would it help if links had a unique id in a data attribute?
Then you could figure out the size of a note ref based on the data attribute. 
One caveat is that if note references repeat, you'll need to keep track of repetitions and also encode that in the data note.

```
![[hello]]

<div data-note="hello">...</div>

![[hello#head]]
<div data-note="hello#head">...</div>

![[hello#head,1:#*]]
<div data-note="hello#head,1:#*">...</div>
```

packages/plugin-core/src/commands/ShowPreview.ts
```diff
@@ -301,6 +304,23 @@ export class ShowPreviewCommand extends BasicCommand<
   constructor(previewPanel: vscode.WebviewPanel) {
     super();
     this._panel = previewPanel;
+
+    if (UI_SHOW_PREVIEW_CMD === undefined) {
+      UI_SHOW_PREVIEW_CMD = this;
+    }
+  }
+
+  async syncEditorToPreviewPlacement(
```
jonathanyeung, line 313: nit: should this one actually be named 'sync-preview-to-editor'? As in - we're updating the position of the preview to the current position in the editor

packages/plugin-core/src/commands/SyncPreviewCommand.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    if (elementOffset === -1) {
+      Logger.warn({
+        msg: `'cleanText' must have not matched up with original element setting offset to '0'`,
+      });
+      elementOffset = 0;
+    }
+
+    console.log(
```
jonathanyeung, line 73: may be fine for tuning purposes as we're developing this, but this may get noisy.

jonathanyeung (commented):
minor stuff

jonathanyeung:
>The approach is attempting to get cleaned version of the text from text editor focus and find the offset (how many elements in editor exist with such text) then find the offset element with the preview. However, the current challenge is that the links get rendered very differently (with diffrent text) which throws off this approach.

Can you confirm my understanding of the challenge - say we're trying to scroll to the 10th instance of `foobar`, but in the middle of the note we have an inlined note ref, which also has the text `foobar` appear a few times - thus, we scroll to somewhere in the middle of the note ref, and not to the correct spot?

If so - in the logic below, can we somehow exclude all html elements from note reference renders, and don't add those to `allElements`?

```javascript
        const allElements = document.querySelectorAll('*');
        
        var myElements = [];
        for (var i = 0; i < allElements.length; i++) { 
            if (allElements[i].innerHTML.includes(targetText)) {
                myElements.push(allElements[i]);
            }
        }
```

PR #1895, opened 2021-12-14
feat: non-note file enhancements

SeriousBug:
This PR adds a several improvements for working with non-note files.

- Links to non-note files can now link to specific lines. The syntax for this is similar to what Github uses, and it's an extension of our existing syntax. For example: `[[src/code.js#L12]]`
- Copy Note Link now works inside non-note files, and will produce a correct link to that file. If the user has a portion of the file highlighted, then a link to that line will be created using the above syntax.
- Links to non-note files are now highlighted as existing if the linked file exists
- Hovering over links to non-note files will display a "click here to open this" message, and not the incorrect "this note is missing" message.


kevinslin (changes requested):
Looks like theres a bug if note has a `.` in front. Goto note will try to open the file instead
https://www.loom.com/share/9c41b41a749f44ef8d9b5014d93c41fb

packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -29,24 +37,33 @@ export class CopyNoteLinkCommand extends BasicCommand<
...
+    // Check if the file is in the assets of any vault. If it is, we can shorten the link.
+    for (const vault of vaults) {
+      const vaultPath = VaultUtils.getRelPath(vault);
+      if (isInsidePath(vaultPath, fsPath)) {
+        fsPath = path.relative(vaultPath, fsPath);
+        break;
+      }
     }
+    let anchor = "";
+    if (!editor.selection.isEmpty) {
```
kevinslin, line 54: from an UX point, when we introduce anchors, how will we diffentiate the user wanting to copy a line number vs copy an anchor?

SeriousBug: I think anchors would be the default option, with a configuration to allow line numbers instead. Since line numbers would get stale too quickly, I think we should avoid encouraging their use. It'll still be an option if a user doesn't want to introduce anchors into their code base.

SeriousBug:
Looks like it was a bug with Copy Note Link for files that are in vaults but not inside assets. Fixed and added a test for it, plus some additional tests for files/folders starting with dots.

packages/plugin-core/src/commands/CopyNoteLink.ts
kevinslin: sounds good. in a separate PR, can we add the following option?

```yml
commands:
  copyNoteLink:
    nonNote:
      defaultAnchorType: line|block|none
```

by default, this should be "none" until the user chooses to enable one. when we introduce the block anchor, we can switch the default to "block". I want to avoid introducing the "line" default only to switch it later.

kevinslin (approved):

PR #1897, opened 2021-12-15
enhance(views): Support new config to automatically show preview

tma66:
- Introduce new config automaticallyShowPreview
- When window changes (ie switching between notes), show preview if config is set to true
- When workspace is launched, show preview if config is set to true
- Default to false

Testing
- Manually verified preview behavior against test workspace
- Updated test cases for WindowWatcher.test.ts

### DIFF
--- a/packages/plugin-core/src/components/views/PreviewViewFactory.ts
+++ b/packages/plugin-core/src/components/views/PreviewViewFactory.ts
@@ -1,5 +1,6 @@
 import {
   assertUnreachable,
+  ConfigUtils,
   DendronEditorViewKey,
   DMessageEnum,
   getWebEditorViewEntry,
@@ -11,12 +12,16 @@ import {
 } from "@dendronhq/common-all";
 import _ from "lodash";
 import * as vscode from "vscode";
-import { handleLink, LinkType } from "../../commands/ShowPreview";
+import {
+  handleLink,
+  LinkType,
+  ShowPreviewCommand,
+} from "../../commands/ShowPreview";
 import { Logger } from "../../logger";
 import { sentryReportingCallback } from "../../utils/analytics";
 import { WebViewUtils } from "../../views/utils";
 import { VSCodeUtils } from "../../vsCodeUtils";
-import { DendronExtension } from "../../workspace";
+import { DendronExtension, getDWorkspace, getExtension } from "../../workspace";
 import { WSUtils } from "../../WSUtils";
 
 /**
@@ -28,6 +33,17 @@ export interface PreviewProxy {
    * @param note Note Props to update the preview contents with
    */
   updateForNote(note: NoteProps): void;
+
+  /**
+   * Show preview panel if applicable and update
+   * @param note Note to display in preview panel
+   */
+  showPreviewAndUpdate(note: NoteProps): void;
+
+  /**
+   * Return current panel. Can be undefined.
+   */
+  getPanel(): vscode.WebviewPanel | undefined;
 }
 
 export class PreviewPanelFactory {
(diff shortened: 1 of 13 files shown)


packages/plugin-core/src/components/views/PreviewViewFactory.ts
```diff
@@ -28,6 +33,17 @@ export interface PreviewProxy {
    * @param note Note Props to update the preview contents with
    */
   updateForNote(note: NoteProps): void;
+
+  /**
```
jonathanyeung, line 37: The Proxy was meant only as a temporary interface until we can refactor out other usages of 'updateForNote'.  Let's not further expose the panel by adding these methods.

Right now, we're still in the middle of refactoring the panel dependency issues, so it's hard to properly get an instance of the _panel properly. 

Rather than doing the `ConfigUtils.getPreview(config).automaticallyShowPreview` inside this method, can we do something like 

_in the code where we want to open the preview:_

```typescript
if (ConfigUtils.getPreview(config).automaticallyShowPreview) {
  const ws = getExtension();
  await new ShowPreviewCommand(PreviewPanelFactory.create(ws)).run();
}
```

I think this should have the same effect? 

Let me know if it's helpful to discuss further offline.

jonathanyeung:
On this: 

> "when change active editor, if preview not active, automatically show preview"

@kevinslin  - are you sure we actually want to do this? What if the user doesn't want the preview open? Then every time they switch notes, they have to close the preview again?

packages/plugin-core/src/windowWatcher.ts
```diff
@@ -122,6 +125,17 @@ export class WindowWatcher {
     return;
   }
 
+  /**
+   * Show note preview panel if applicable
+   */
+  async triggerNotePreviewUpdate({ document }: TextEditor) {
+    const maybeNote = WSUtils.tryGetNoteFromDocument(document);
+    if (maybeNote)
+      PreviewPanelFactory.getProxy().showPreviewAndUpdate(maybeNote);
```
jonathanyeung, line 134: Rather than calling the static `getProxy()` (which was meant only as a temporary method until we could finish refactoring), let's try to pass a instance of `PreviewProxy` into the constructor of `WindowWatcher`.  Then, in `workspace.ts`, we can pass in the static instance when constructing `WindowWatcher` (for now).

Then in testing, we can just use a dummy implementation of `PreviewProxy`.

However, if we want to do this show on window switching  functionality, we will still need to expose a `show()` method on `PreviewProxy`

jonathanyeung (changes requested):

kevinslin:
> On this:
> 
> > "when change active editor, if preview not active, automatically show preview"
> 
> @kevinslin - are you sure we actually want to do this? What if the user doesn't want the preview open? Then every time they switch notes, they have to close the preview again?

Then they wouldn't have this option enabled. We're bringing back the option from markdown preview enhanced. This is the same behavior there and we've had several people asking for this

kevinslin:
@tma66 can you rebase from master? should fix the tests

packages/plugin-core/src/components/views/PreviewViewFactory.ts
tma66: 1. Unfortunately calling ShowPreviewCommand is not enough as I should only call if the panel does not exist. Otherwise it leads to a strange visual side effect where the preview panel is always in effect (maybe easier to show than tell). I updated the existing method you exposed to do both show preview and update, moving the update to a private method of the factory. This way only one method gets exposed instead of two.

2. I need the _panel property for testing purposes. Is there another way to check if the panel exists?

packages/plugin-core/src/windowWatcher.ts
tma66: Good idea, I'll update the constructor

packages/plugin-core/src/components/views/PreviewViewFactory.ts
jonathanyeung: Ah - I see.  

Re point 2 - Maybe instead of exposing the panel directly, we can expose a property on PreviewProxy like `PanelVisible` or something like that.  My main worry is losing encapsulation on the preview panel again. Let's defer that change to next time.

jonathanyeung (approved):

kevinslin (approved):
nice job! excited to ship this

kevinslin:
one more thing. can you update the documentation? add it here -> [[Config|dendron.topic.preview#config]]

PR #1906, opened 2021-12-16
feat: open preview buttons for context menus

SeriousBug:
- Adds a "Dendron: Show Preview" button in the context menu when right clicking a markdown file in the explorer, as well as when right clicking a tab in the editor.
- Extends the Show Preview command to display the note passed in through the context menu for this feature to work.
- Makes Show Preview fall back to the default preview if a corresponding note is not found.
- Adds analytics to track when show preview is used from a context menu.

Combined with the following PR, this will replace the default "Show Preview" context menu buttons with Dendron's: https://github.com/microsoft/vscode/pull/139241


kevinslin (changes requested):
When switching between regular preview and dendron preview, this adds complexity to the users:
    - when the user navigates from a dendron markdown file to a regular markdown file outside of Dendron, the preview will stop working and it won't be clear that they have to click the preview again
    - the same problem happens vice versa except its a little more insidious because the builtin preview will try to render dendron markdown and will fail at dendron specific syntax

Instead of toggling between previews, can we use the Dendron preview for all markdown files?

This will require the following change: [[Dendron Should Show Regular Markdown|dendron://private/task.views.2021.12.16.dendron-should-show-regular-markdown]]

SeriousBug:
This is actually a bit more involved because the preview asks the engine to render the note, but only passes in the id of the note. I'll need to update the API to allow overriding the note body.

Issue #1312, opened 2021-09-08
'Extension host terminated unexpectedly' - Argdown


kzilla-git:
## Please select the area your Bug applys to. (Multiple selections are Possible)
- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x] Misc (Choose this if your not sure)

## Describe the bug
Getting 'Extension host terminated unexpectedly' error upon starting Dendron in VSCode no matter how many times I reloaded/restarted and rebooted my computer. 

## To Reproduce
1. Have the Argdown extension already installed.
2. Open Dendron in VSCode and either open the Calendar or the Dendron Preview V2.

## Expected behavior
'Extension host terminated unexpectedly' error appears.

## Desktop (please complete the following information)
- OS: MS Windows 10
- Browser: Chrome

## Additional context
After uninstalling Argdown extension, the problem was resolved.

kevinslin:
Thanks for the issue. We'll be investigating this separately when we have some more bandwidth 😅

lnicola:
I'm also seeing this. It appears to run out of memory:

![image](https://user-images.githubusercontent.com/308347/144711119-040c0ac3-21a0-4fa2-b16c-02cee0e0db7a.png)

Note that it's happening outside of Dendron workspaces (CC #1540).

lnicola:
This might have gotten worse recently, but disabling `dendron.watchForNativeWorkspace` seems to fix it for me.

SeriousBug:
Dendron 0.74 disables `dendron.watchForNativeWorkspace` by default now, and this crash should not occur anymore. If it continues to happen, please let us know. Otherwise this issue will close automatically in a few days.

lnicola:
Shouldn't this be kept open for tracking, since people might still enable that preference?

PR #1913, opened 2021-12-16
fix: extension crash in non-Dendron workspaces when there's many files

SeriousBug:
When there's a lot of files in non-Dendron workspaces, the native workspace creation watcher would cause Dendron to crash.
This PR disables the watcher by default, and instead simply reloads the workspace when a native workspace is created.

#1312


kevinslin (approved):
so we don't want non-dendron workspaces to crash but also not to add more friction for people to use dendron workspaces.

PR #1914, opened 2021-12-16
chore(lookup): Better schema lookup error message

tma66:
- Instead of a warning message, show a modal message
    
    Testing
    - Added test case to check for warning message
    - Verified manually by performing multilevel schema lookup

### DIFF
--- a/packages/plugin-core/src/components/lookup/LookupProviderV3.ts
+++ b/packages/plugin-core/src/components/lookup/LookupProviderV3.ts
@@ -621,7 +621,13 @@ export class SchemaLookupProvider implements ILookupProviderV3 {
               });
             }
           });
-        //window.showErrorMessage("schemas can only be one level deep");
+
+        HistoryService.instance().add({
+          source: "lookupProvider",
+          action: "done",
+          id: this.id,
+          data: { cancel: true },
+        });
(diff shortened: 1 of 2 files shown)


packages/plugin-core/src/components/lookup/LookupProviderV3.ts
```diff
@@ -608,7 +609,19 @@ export class SchemaLookupProvider implements ILookupProviderV3 {
...
+            ...["Note Lookup"]
+          )
+          .then(async (selection) => {
+            if (selection === "Note Lookup") {
+              await new NoteLookupCommand().run({
+                initialValue: picker.value,
+              });
+            }
+          });
+        //window.showErrorMessage("schemas can only be one level deep");
```
kevinslin, line 624: if you remove the `return` value, this will work. the reason its hanging is because of the following:

Note_Preview_—_dendron__Workspace_

Docs: [[Initialization|dendron://dendron.docs/pkg.plugin-core.t.lookup.arch#initialization]]

Essentially, that command only returns when it receives an event from the `HistoryService` which is never done in this case. This is actually a bug in our code.

kevinslin (approved):

PR #1920, opened 2021-12-17
Add links to dendron vault for a few dendron components

kevinslin:
Now that we have the ability to document both in the code and also in Dendron, the question is what to put where.

The pattern I've been using: 
- details should be put in code as regular comments
- high level overviews should be put in Dendron. these details currently are in the form of reference notes of existing packages. See docs [here](https://github.com/dendronhq/dendron-docs/commit/f49a192ad8f58824989bbba7feece38a57704e0f)

The reference notes for these packages can be found [here](https://github.com/dendronhq/dendron-docs/commit/dafc752b99b33697c8872eedf0bf0c195a6c17ed)

One outstanding issue:
- the `Summary` section overlaps with the docstring for a given module. see below for an example. potential solution: use block embeding from file to note to generate summary 

```ts
/**
 * Creates instances of `FooWidget`
 */
class FooWidgetFactory {
}
```

```md
## Summary
Creates instances of `FooWidget`
```

This is just the beginning - there's a lot more that can be had but in interest of iteration, starting with module references as a simple starting point that we can build on.

### DIFF
--- a/packages/plugin-core/src/WorkspaceWatcher.ts
+++ b/packages/plugin-core/src/WorkspaceWatcher.ts
@@ -63,6 +63,9 @@ interface DebouncedFunc<T extends (...args: any[]) => any> {
   flush(): ReturnType<T> | undefined;
 }
 
+/**
+ * See [[Workspace Watcher|dendron://dendron.docs/pkg.plugin-core.ref.workspace-watcher]] for more docs
+ */
 export class WorkspaceWatcher {
   /** The documents that have been opened during this session that have not been viewed yet in the editor. */
   private _openedDocuments: Map<string, TextDocument>;
--- a/packages/plugin-core/src/features/windowDecorations.ts
+++ b/packages/plugin-core/src/features/windowDecorations.ts
@@ -114,6 +114,7 @@ export const debouncedUpdateDecorations = debounceAsyncUntilComplete({
   trailing: true,
 });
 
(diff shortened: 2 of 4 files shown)


SeriousBug (approved):
This looks good, and when to document in vault makes sense. We can also update `[[dev.process.review]]` or `[[dev.process.code]]` to suggest that people writing code can document the changes and then link to them.

I'm not sure what you meant by "potential solution: use block embeding from file to note to generate summary" though, we can't embed files into notes yet. We could embed notes in files, but the hover support isn't there yet.

kevinslin:
> 



> This looks good, and when to document in vault makes sense. We can also update `[[dev.process.review]]` or `[[dev.process.code]]` to suggest that people writing code can document the changes and then link to them.
> 
> I'm not sure what you meant by "potential solution: use block embeding from file to note to generate summary" though, we can't embed files into notes yet. We could embed notes in files, but the hover support isn't there yet.

Talk moore about block embedding (aka block note reference) in our async discussion -> https://docs.google.com/document/d/1Pr0npNrs7xiiEMPsUi9_PrSDVl9NCoZ08PZjicGhHEY/edit

hikchoi:
looks good. in the future, we can potentially have the comments in code to conform with, say [TSDoc](https://tsdoc.org/), and then organize information that can potentially clutter the codebase with our note links. This way we can implement a tsdoc plugin that can correctly parse the wikilinks and point to their respective published urls.

hikchoi (approved):

PR #1939, opened 2021-12-21
chore(docs): document migratioon process

kevinslin:
Add documentation on how `dendorn.yml` configuration is changed during upgrades

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -427,6 +427,7 @@ export async function _activate(
       } else {
         // no migration changes.
         // see if we need to force a config migration.
+        // see [[Run Config Migration|dendron://dendron.docs/pkg.dendron-engine.t.upgrade.arch.lifecycle#run-config-migration]]
         const configMigrationChanges =
           await wsService.runConfigMigrationIfNecessary({
             currentVersion,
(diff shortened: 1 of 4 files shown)


kevinslin:
See https://github.com/dendronhq/dendron-docs/commit/72d2a698992a0187d358287e96d554ae6f0a89d1 for more docs

packages/plugin-core/src/_extension.ts
```diff
@@ -427,6 +427,7 @@ export async function _activate(
       } else {
         // no migration changes.
         // see if we need to force a config migration.
+        // see [[Run Config Migration|dendron://dendron.docs/pkg.dendron-engine.t.upgrade.arch.lifecycle#run-config-migration]]
```
SeriousBug, line 430: Should be `...#run-migration]]`

SeriousBug (approved):
Looks good, just one broken anchor.

Also, not related to this PR but there's a link for `[[2021 06 Deprecate Workspace Settings|proj.2021-06-deprecate-workspace-settings]]` in `_extension.ts`, which is actually in `private` vault which is not included in the workspace. We should move the note if we want it accessible.

packages/plugin-core/src/_extension.ts
kevinslin: fixed

PR #1942, opened 2021-12-21
chore: improve note highlighting and autocomplete performance

SeriousBug:
Implements new interfaces for getting notes by their name, `getNotesByFnameFromEngine` and `getNoteByFnameFromEngine`. These new interfaces use a newly added fname to note id map to speed up note lookups. The results is significantly improved performance in note highlighting and autocomplete.

The immediate end-user effect of this PR is that note highlighting and autocomplete will be faster.

I avoided migrating other parts of the codebase for now, and stuck with the most performance critical parts that I know of. We can slowly migrate over to this new API as we are able.


packages/common-all/src/dnode.ts
```diff
@@ -807,6 +807,21 @@ export class NoteUtils {
...
+    engine: DEngineClient;
+    vault?: DVault;
+  }): NoteProps[] {
+    let notes = engine.noteFnames.get(engine.notes, fname);
+    if (vault)
+      notes = notes.filter((note) => VaultUtils.isEqualV2(note.vault, vault));
+    return notes;
+  }
+
   static getNoteByFnameV5({
```
jonathanyeung, line 825: If we do these kinds of revision bumps, we should mark the old versions with an `@deprecated` tag. Otherwise, new code will likely continue using v5 as we scan the code base to see what other code is doing as an example(and seeing usages of v5).

packages/common-all/src/dnode.ts
```diff
@@ -1170,23 +1193,80 @@ export class NoteUtils {
...
+  }
+
   /** Generate a random color for `note`, but allow the user to override that color selection.
    *
    * @param note The fname of note that you want to get the color of.
    * @param notes: All notes in `engine.notes`, used to check the ancestors of `note`.
    * @returns The color, and whether this color was randomly generated or explicitly defined.
    */
-  static color(opts: { fname: string; vault?: DVault; notes: NotePropsDict }): {
+  static colorV2(opts: {
```
jonathanyeung, line 1214: It looks like there are only 2 references to the original `color` method.  Can we go ahead and update those to avoid creating a `colorv2`?

packages/common-all/src/dnode.ts
```diff
@@ -1203,24 +1283,24 @@ export class NoteUtils {
    *
    * @param opts.fname The fname of the note you are trying to get the ancestors of.
    * @param opts.vault The vault to look for. If provided, only notes from this vault will be included.
-   * @param opts.notes All notes in `engine.notes`.
+   * @param opts.engine The engine.
    * @param opts.includeSelf: If true, note with `fname` itself will be included in the ancestors.
    * @param opts.nonStubOnly: If true, only notes that are not stubs will be included.
    */
-  static *ancestors(opts: {
+  static *ancestorsV2(opts: {
```
jonathanyeung, line 1290: same comment as with colorV2

packages/common-all/src/types/typesv2.ts
```diff
@@ -364,6 +365,8 @@ export type GetDecorationsOpts = {
 export type DCommonProps = {
   /** Dictionary where key is the note id. */
   notes: NotePropsDict;
+  /** Dictionary where the key is lowercase note fname, and values are ids of notes with that fname. */
```
jonathanyeung, line 368: nit: would you mind expanding on this one a bit saying that the reason there can be multiple id's is because of multi-vault?

jonathanyeung (changes requested):
(oops: sorry I missed that this one is still in draft).  Left some comments anyway :)

packages/common-all/src/dnode.ts
SeriousBug: The remaining usages are in the UI, and the UI doesn't have access to `DEngineClient` but a subset of it. Switching it over is possible but it's more involved than just changing the name, so I'd rather leave it for later.

kevinslin (changes requested):

packages/common-all/src/utils.ts
```diff
@@ -210,6 +211,64 @@ export class DefaultMap<K, V> {
...
+    const values = this._internalMap.get(key);
+    if (values === undefined) return false;
+    return values.includes(value);
+  }
+}
+
+export class NoteFNamesDict {
+  private _internalMap = new ListMap<string, string>();
+
+  public constructor(initialNotes?: NoteProps[]) {
```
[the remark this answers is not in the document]
SeriousBug: NoteFNamesDict doesn't store notes, only ids. You may notice `get` actually requires you to pass in the notes dict exactly for that reason. The constructor and `add` functions accept the entire object only for convenience, it minimizes errors in my opinion.

packages/common-all/src/dnode.ts
kevinslin: the only usage here is in [[../packages/common-frontend/src/utils/tree-view.tsx]] correct? if that is the only usage, would be better to remove it and simplify the code and re-introduce the feature later

we'll want to expose an engine object in the web version as well so we can use a common API regardless where engine methods are being called from. can be added as a backlog task

packages/engine-server/src/engineClient.ts
```diff
@@ -318,8 +324,15 @@ export class DendronEngineClient implements DEngineClient {
           );
         }
         this.notes[id] = ent.note;
+        this.noteFnames.add(ent.note);
       }
     });
+    // There's a limitation with updates: we know what the new note is, but we don't know what the old note was.
+    // If the note was renamed, then we can't remove the old name from `noteFnames` because we don't know the old fname.
+    // To get around this, we'll just refresh the entire dict.
+    if (notes.some((note) => note.status === "update")) {
```
kevinslin, line 333: will this be expensive? its called every time the note changes (eg. [onDidChangeTextDocument](https://github.com/dendronhq/dendron/blob/9b58d335bcaa86692dfef895f15b5e48220e962b/packages/plugin-core/src/WorkspaceWatcher.ts#L160) gets called everytime the note changes)

can we update `NoteChangeEntry` signature to the following?
```ts
export type NoteChangeEntry = {
  notePrev: NoteProps;
  noteNext: NoteProps;
  status: "create" | "update" | "delete";
};
```

this is a pre-requisite to a bunch of later link optimizations (currently we calculate all links on every update but this lets us be more selective depending on what props have changed)

kevinslin (changes requested):

PR #1944, opened 2021-12-21
fix: circular dependencies part 1

nickolay-kondratyev:
# Pull Request Checklist

Part 1: of Circular Dependency Sequel.

In this part we start the split up/decoupling process. There isn't too many circular dependencies  knocked out (from 124 on master to 119) in this commit but it does the initial set up to start knocking out the circular dependencies. 

We will need to do this piece meal to try to minimize the merge conflicts as this sequel is going to be touching quite a lot of files. 

Using C# interface naming convention (to avoid attempting to rename current implementation). And file naming convention of suffixing `Interface.ts` to interface holding files to more align with our file naming convention. 

In this process, attempting to make no changes to the logic and move code around as much as possible in as is way. 

### Quality Assurance

### DIFF
--- a/packages/plugin-core/src/ExtensionProvider.ts
+++ b/packages/plugin-core/src/ExtensionProvider.ts
@@ -2,6 +2,14 @@ import { IDendronExtension } from "./dendronExtensionInterface";
 import { DendronError } from "@dendronhq/common-all";
 import _ from "lodash";
 
+/**
+ * Use this to statically get implementation of IDendronExtension without having to
+ * depend on concrete DendronExtension.
+ *
+ * Note: Prefer to get IDendronExtension injected into your classes upon their
+ * construction rather than statically getting it from here. But if that's not
+ * a fitting option then use this class.
+ * */
 export class ExtensionProvider {
   static extension: IDendronExtension;
 
@@ -13,6 +21,14 @@ export class ExtensionProvider {
     }
 
     return ExtensionProvider.extension;
+  }
+
+  static getDWorkspace() {
+    return ExtensionProvider.getExtension().getDWorkspace();
+  }
+
+  static getEngine() {
+    return ExtensionProvider.getExtension().getEngine();
   }
 
   static register(extension: IDendronExtension) {
--- a/packages/plugin-core/src/ExtensionProvider.ts
+++ b/packages/plugin-core/src/ExtensionProvider.ts
@@ -1,0 +1,21 @@
+import { IDendronExtension } from "./dendronExtensionInterface";
+import { DendronError } from "@dendronhq/common-all";
+import _ from "lodash";
+
+export class ExtensionProvider {
+  static extension: IDendronExtension;
+
+  static getExtension(): IDendronExtension {
+    if (_.isUndefined(ExtensionProvider.extension)) {
+      throw new DendronError({
+        message: `Extension is not yet registered. Maker sure initialization registers extension prior to usage.`,
+      });
+    }
+
+    return ExtensionProvider.extension;
+  }
+
+  static register(extension: IDendronExtension) {
+    ExtensionProvider.extension = extension;
+  }
+}
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -739,7 +739,7 @@ export async function _activate(
       // If automaticallyShowPreview = true, display preview panel on start up
       const note = WSUtils.getActiveNote();
       if (note) {
-        PreviewPanelFactory.getProxy().showPreviewAndUpdate(note);
+        PreviewPanelFactory.getProxy(getExtension()).showPreviewAndUpdate(note);
       }
 
       return true;
@@ -978,7 +978,8 @@ async function _setupCommands(
   const existingCommands = await vscode.commands.getCommands();
 
   ALL_COMMANDS.map((Cmd) => {
-    const cmd = new Cmd();
+    const cmd = new Cmd(ws);
+
     if (!existingCommands.includes(cmd.key))
       context.subscriptions.push(
         vscode.commands.registerCommand(
--- a/packages/plugin-core/src/commandFactory.ts
+++ b/packages/plugin-core/src/commandFactory.ts
@@ -1,0 +1,37 @@
+/**
+ * Class responsible for creation of commands, specifically useful when one
+ * command needs an instance of another command, this adds in a layer of
+ * decoupling.
+ */
+import { BaseCommand } from "./commands/base";
+import {
+  ShowPreviewCommandOpts,
+  ShowPreviewCommandOutput,
+} from "./commands/ShowPreviewInterface";
+import { ShowPreviewCommand } from "./commands/ShowPreview";
+import vscode from "vscode";
+import { ICommandFactory } from "./commandFactoryInterface";
+import {
+  GoToNoteCommandOpts,
+  GoToNoteCommandOutput,
+} from "./commands/GoToNoteInterface";
+import { GotoNoteCommand } from "./commands/GotoNote";
+import { IDendronExtension } from "./dendronExtensionInterface";
+
+export class CommandFactory implements ICommandFactory {
+  private readonly extension: IDendronExtension;
+
+  constructor(extension: IDendronExtension) {
+    this.extension = extension;
+  }
+
+  showPreviewCmd(
+    panel: vscode.WebviewPanel
+  ): BaseCommand<ShowPreviewCommandOpts, ShowPreviewCommandOutput> {
+    return new ShowPreviewCommand(panel);
+  }
+
+  goToNoteCmd(): BaseCommand<GoToNoteCommandOpts, GoToNoteCommandOutput> {
+    return new GotoNoteCommand(this.extension);
+  }
+}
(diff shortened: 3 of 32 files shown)


packages/plugin-core/src/ExtensionProvider.ts
```diff
@@ -0,0 +1,21 @@
+import { IDendronExtension } from "./dendronExtensionInterface";
+import { DendronError } from "@dendronhq/common-all";
+import _ from "lodash";
+
+export class ExtensionProvider {
```
nickolay-kondratyev, line 5: The idea with this class is that we can still query out statically to grab the extension (especially during transition) without having to depend on `workspace` file or concrete implementation of extension.

packages/plugin-core/src/_extension.ts
```diff
@@ -978,7 +978,8 @@ async function _setupCommands(
   const existingCommands = await vscode.commands.getCommands();
 
   ALL_COMMANDS.map((Cmd) => {
-    const cmd = new Cmd();
+    const cmd = new Cmd(ws);
```
nickolay-kondratyev, line 981: Commands are getting extension passed to their constructor to be able to use an instance of dependency injected extension interface rather than reaching out statically.

packages/plugin-core/src/commandFactory.ts
```diff
@@ -0,0 +1,37 @@
...
+import vscode from "vscode";
+import { ICommandFactory } from "./commandFactoryInterface";
+import {
+  GoToNoteCommandOpts,
+  GoToNoteCommandOutput,
+} from "./commands/GoToNoteInterface";
+import { GotoNoteCommand } from "./commands/GotoNote";
+import { IDendronExtension } from "./dendronExtensionInterface";
+
+export class CommandFactory implements ICommandFactory {
```
nickolay-kondratyev, line 21: This allows commands that depend on instances of other commands to get instances of those commands without creating tight coupling to those implementations.

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
```diff
@@ -141,7 +141,7 @@ export class CreateNoteWithTraitCommand extends BaseCommand<
       }
     }
 
-    await new GotoNoteCommand().execute({
+    await new GotoNoteCommand(getExtension()).execute({
```
nickolay-kondratyev, line 144: Since we already have getExtension() imported into these classes it doesn't make it worse to use it for now, but it also doesn't make it better as time goes on these should be refactored to use extension that gets passed into the instance of the command.

(or if something is used statically that we dont want to refactor right away then it should use ExtensionProvider to grab the interface of DendronExtension rather than coupling ourselves to concrete implementation of DendronExtension).

packages/plugin-core/src/dendronExtensionInterface.ts
```diff
@@ -0,0 +1,70 @@
...
+  "dendron.defaultLookupCreateBehavior": string;
+  "dendron.defaultTimestampDecorationFormat": string;
+  "dendron.rootDir": string;
+  "dendron.dendronDir": string;
+  "dendron.logLevel": string;
+  "dendron.trace.server": string;
+  "dendron.serverPort": string;
+}>;
+
+export interface IDendronExtension {
```
nickolay-kondratyev, line 31: Going forward we should use IDendronExtension in places of concrete DendronExtension

kevinslin:
Can you write out a high level description of what is being refactored here? will make the logic here a little easier to follow.

If there are best practices, we should add them to the `dev` section of [[Dev|dendron://dendron.docs/pkg.plugin-core.dev]]

packages/plugin-core/src/_extension.ts
jonathanyeung: A command with a no-arg constructor still works fine here right?

packages/plugin-core/src/commands/base.ts
```diff
@@ -64,7 +65,7 @@ export abstract class BaseCommand<
 
   abstract enrichInputs(inputs: TGatherOutput): Promise<TOpts | undefined>;
 
-  abstract execute(opts: TOpts): Promise<TOut>;
+  abstract execute(opts?: TOpts): Promise<TOut>;
```
jonathanyeung, line 68: hmm, was this a stray change?

packages/plugin-core/src/ExtensionProvider.ts
jonathanyeung: Sounds good - can you add a comment to this class w/ this explanation? Also - make a note there that any _new_ classes should **not** use these statics and instead get an instance of `IDendronExtension` passed into their constructors

packages/plugin-core/src/dendronExtensionInterface.ts
jonathanyeung: Can you go ahead and add some comments info for `IDendronExtension`? Doesn't need to be complete, we can build upon the docstrings over time.

packages/plugin-core/src/services/EngineAPIService.ts
```diff
@@ -83,62 +84,71 @@ export class EngineAPIService implements DEngineClient {
   get trustedWorkspace(): boolean {
     return this._trustedWorkspace;
   }
+
   set trustedWorkspace(value: boolean) {
     this._trustedWorkspace = value;
   }
 
-  public get notes(): NotePropsDict {
```
jonathanyeung, line 90: nit: Why the removal of `public`? Was this done by an auto-code formatter? 

I understand that `public` is the default access level in typescript - but I personally like it there to explicitly indicate that it's meant for external consumption.

packages/plugin-core/src/workspace.ts
```diff
@@ -571,7 +552,7 @@ export class DendronExtension {
     }
   }
 
-  private setupBacklinkTreeView() {
```
jonathanyeung, line 574: was it necessary to expose this one?

jonathanyeung (changes requested):
Looking good

packages/plugin-core/src/services/EngineAPIService.ts
nickolay-kondratyev: It was an automatic refactor. 

Will add it back.

jonathanyeung:
When we do check this in, would you mind running the circ dependency checker again, and adding the results to the bottom of [[Cyclical Dependency Fixing|dendron://private/scratch.2021.12.02.143546.cyclical-dependency-fixing]] (at least for plugin core circular deps)? So that we can track our progress and what's remaining.

nickolay-kondratyev:
> Can you write out a high level description of what is being refactored here? will make the logic here a little easier to follow.

Added approach at the bottom of the task note [[dendron://private/task.workspace.2021.12.21.fix-blocking-circular-dependencies]]


> If there are best practices, we should add them to the `dev` section of [[Dev|dendron://dendron.docs/pkg.plugin-core.dev]]

Sounds good will do after we merge this first commit.

packages/plugin-core/src/workspace.ts
nickolay-kondratyev: No, put it back.

packages/plugin-core/src/commands/base.ts
nickolay-kondratyev: This one was purposeful. 

We have commands that do not require options when they execute. Example `GoToNote`, So now with CommandFactory we are calling execute through the interface of BaseCommand. 

(As example we want to allow ShowPreview to call GoToNote command but we dont want ShowPreview to know the exact implementation of GoToNote).

packages/plugin-core/src/ExtensionProvider.ts
nickolay-kondratyev: Added explanation, and @deprecated getExtension() in workspace.ts (along with getDWorkspace(),  getEngine()) with a comment to go their mirroring ExtensionProvider methods.

packages/plugin-core/src/_extension.ts
nickolay-kondratyev: Yep tested note lookup it has no-arg constructor and appears to work fine. 

```
NoteLookupCommand
  constructor() {
  ```

File: packages/plugin-core/src/dendronExtensionInterface.ts (hunk 5 above)
Comment (nickolay-kondratyev) [reply packages/plugin-core/src/dendronExtensionInterface.ts:31]: Good point, added.

**Comment by nickolay-kondratyev** [comment]
Note commit that was blocked due to GoToNote circular dependency now passes : https://github.com/dendronhq/dendron/pull/1891

**Review by jonathanyeung — APPROVED** [review]

**Comment by jonathanyeung** [comment]
Looks good to me. @kevinslin - I'll let you review/approve prior to merging.

File: packages/plugin-core/src/commands/GotoNote.ts [hunk 9]
```diff
@@ -80,8 +57,19 @@ type FoundLinkSelection = NonNullable<
 /**
  * Open or create a note. See {@link GotoNoteCommand.execute} for details
  */
-export class GotoNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
+export class GotoNoteCommand extends BasicCommand<
+  GoToNoteCommandOpts,
+  GoToNoteCommandOutput
+> {
   key = DENDRON_COMMANDS.GOTO_NOTE.key;
+  private extension: IDendronExtension;
```
kevinslin, line 65: not required for this PR but does it make sense to make extension and wsUtils part of the command interface?

kevinslin (approved):
looks good, nice job!

packages/plugin-core/src/commands/GotoNote.ts
nickolay-kondratyev: By command interface do you mean adding IDendronExtension to BaseCommand?

PR #1960, opened 2021-12-22
enhance: add configuration for vault selection behavior, change prompt vault selection default to true

nickolay-kondratyev:
# enhance: add configuration for vault selection behavior, change prompt vault selection default to true

enhance: add configuration for vault selection behavior, change prompt vault selection default to true

## General
Doc PR: https://github.com/dendronhq/dendron-site/pull/329

### Quality Assurance

### DIFF
--- a/packages/common-all/src/constants/configs/commands.ts
+++ b/packages/common-all/src/constants/configs/commands.ts
@@ -53,7 +53,7 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     label: "Confirm Vault on Create.",
     desc: "Prompt for vault selection when creating new note.",
   },
-  vaultModeOnCreate: {
+  vaultSelectionModeOnCreate: {
     label: "Vault Selection Mode On Note Creation.",
     desc: "Whether to smartly guess the vault for the note when creating new note.",
   },
--- a/packages/common-all/src/constants/configs/commands.ts
+++ b/packages/common-all/src/constants/configs/commands.ts
@@ -53,6 +53,10 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     label: "Confirm Vault on Create.",
     desc: "Prompt for vault selection when creating new note.",
   },
+  selectVaultModeOnCreate: {
+    label: "Vault Selection Mode On Note Creation.",
+    desc: "Whether to smartly guess the vault for the note when creating new note.",
+  },
   leaveTrace: {
     label: "Leave Trace",
     desc: "Leaves a reference of the destination note in the source note while using selectionExtract",
@@ -62,8 +66,8 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     desc: "Put `Create New` at the top of the lookup result.",
   },
   fuzzThreshold: {
-    label: "Fuzziness threshold for lookup",
-    desc: "Sets the fuzziness level of lookups 0=exact matches only, 1=max fuzziness.",
+    label: "Fuzziness Threshold For Lookup",
+    desc: "Sets the fuzziness level of lookup",
   },
 };
 
--- a/packages/common-all/src/constants/configs/commands.ts
+++ b/packages/common-all/src/constants/configs/commands.ts
@@ -53,7 +53,7 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     label: "Confirm Vault on Create.",
     desc: "Prompt for vault selection when creating new note.",
   },
-  selectVaultModeOnCreate: {
+  vaultModeOnCreate: {
     label: "Vault Selection Mode On Note Creation.",
     desc: "Whether to smartly guess the vault for the note when creating new note.",
   },
--- a/packages/common-all/src/types/configs/commands/lookup.ts
+++ b/packages/common-all/src/types/configs/commands/lookup.ts
@@ -31,7 +31,7 @@ export type LookupConfig = {
  */
 export type NoteLookupConfig = {
   selectionMode: LookupSelectionMode;
-  selectVaultModeOnCreate: LookupSelectVaultModeOnCreate;
+  vaultModeOnCreate: LookupSelectVaultModeOnCreate;
   confirmVaultOnCreate: boolean;
   leaveTrace: boolean;
   bubbleUpCreateNew: boolean;
@@ -47,7 +47,7 @@ export function genDefaultLookupConfig(): LookupConfig {
     note: {
       selectionMode: LookupSelectionModeEnum.extract,
       confirmVaultOnCreate: true,
-      selectVaultModeOnCreate: LookupSelectVaultModeOnCreateEnum.smart,
+      vaultModeOnCreate: LookupSelectVaultModeOnCreateEnum.smart,
       leaveTrace: false,
       bubbleUpCreateNew: true,
       /**
--- a/packages/common-all/src/types/configs/commands/lookup.ts
+++ b/packages/common-all/src/types/configs/commands/lookup.ts
@@ -7,11 +7,18 @@ export enum LookupSelectionModeEnum {
   none = "none",
 }
 
+export enum LookupSelectVaultModeOnCreateEnum {
+  smart = "smart",
+  alwaysPrompt = "alwaysPrompt",
+}
+
 /**
  * String literal type generated from {@link NoteLookupSelectionBehaviorEnum}
  */
 export type LookupSelectionMode = keyof typeof LookupSelectionModeEnum;
 
+export type LookupSelectVaultModeOnCreate =
+  keyof typeof LookupSelectVaultModeOnCreateEnum;
 /**
  * Namespace for configuring lookup commands
  */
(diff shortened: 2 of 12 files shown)


packages/common-all/src/constants/configs/commands.ts
```diff
@@ -53,6 +53,10 @@ const NOTE: DendronConfigEntryCollection<NoteLookupConfig> = {
     label: "Confirm Vault on Create.",
     desc: "Prompt for vault selection when creating new note.",
   },
+  selectVaultModeOnCreate: {
+    label: "Vault selection mode on note creation.",
+    desc: "When new note is created this configures whether Dendron should smartly guess the vault for the note or always prompt to select a vault. Values:[smart, alwaysPrompt].",
+  },
```
hikchoi, line 59: super nit: these descriptions are not used anywhere yet, but will be used in a webview for config in the future. can we make the tone consistent with the others?

1. Consistent capitalization of label
1. short form description with consistent tone with others.
1. remove description of possible values in the `desc` field as these will be provided by the to-be-implemented webview as a dropdown form.

```suggestion
  selectVaultModeOnCreate: {
    label: "Vault Selection Mode on Create.",
    desc: "Sets what vault selection mode should be used when creating new note.",
  },
```

hikchoi: while we are at it, can we also fix the label and description for `fuzzThreshold`?

```
fuzzThreshold: {
  label: "Fuzziness Threshold for Lookup",
  desc: "Sets the fuzziness level of lookup".
}
```

packages/plugin-core/src/components/lookup/vaultSelectionModeKeeper.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * We want to start with the selection from configuration value
+ * However, if the user has toggled the selection during the session
+ * we want to at least respect their wishes during the session and use
+ * the value that they have toggled to during the session.
+ *
+ * (V2: in the future we should highly consider modifying the configuration
+ * value if the user has changed the toggle to persist user's wishes across
+ * the sessions without users having to modify the config yaml).
+ * */
```
hikchoi, line 18: Some thoughts on the keeper class in general:

Right now, the user experience we provide with configuration is less than optimal (having to open a yaml file to modify it), but this I think is something we should address separately.

I think keeping the user's expectations simple is beneficial for both the users and the developers: 
1. users are happy that there are no surprises (when they toggle the vault selection mode that deviates from the config, are they notified that this mode will persist for this session? and that it will reset when a new session starts?), and 
2. developers have less moving parts to be concerned (especially for a common component that is shared across many features of Dendron)

Right now, the immediate expectation mismatch I'm seeing is that the user will not know that these modes are going to be remembered for the session.

I definitely think this could provide a smoother experience if we introduce this too keeper behavior to other configs as well, but currently there is no guardrails against the users thinking the other togglable configs will be remembered as well. (Not suggesting we should implement this for everything right now)

If we decided that we are going with this solution, I think we should properly notify the users that whatever they toggle right now is going to be remembered for this session and will be reset to the config value when they start a new one.

That being said, maybe it is worth introducing this to the users and gather feedback on how it feels. If the users like this, we can generalize the keeper component so that we can reuse it in other places that makes sense.

Any thoughts on this @kevinslin ?

hikchoi (changes requested):
left some comments.

packages/plugin-core/src/components/lookup/vaultSelectionModeKeeper.ts
kevinslin: I agree with @hikchoi - lets add the ability to make this selection sticky in future work

hikchoi:
This will hopefully resolve https://github.com/dendronhq/dendron/issues/2042 as well

packages/plugin-core/src/components/lookup/vaultSelectionModeKeeper.ts
nickolay-kondratyev: >  lets add the ability to make this selection sticky in future work

Could we clarify whether we want to:
1. Add a notification to the user that they have toggled the value and that the value will persist for the duration of this session. 
2. Not persist the value for duration of the session.

packages/common-all/src/types/configs/commands/lookup.ts
```diff
@@ -38,7 +46,8 @@ export function genDefaultLookupConfig(): LookupConfig {
   return {
     note: {
       selectionMode: LookupSelectionModeEnum.extract,
-      confirmVaultOnCreate: false,
+      confirmVaultOnCreate: true,
+      selectVaultModeOnCreate: LookupSelectVaultModeOnCreateEnum.smart,
```
hikchoi, line 50: I got confused with the wording because `selectVaultModeOnCreate` read like a boolean property but it's actually describing what the vault selection mode is.

Can we follow the convention in [[Config|dendron://dendron.docs/dev.process.code.config]]:

```
1. Use bare nouns only for configuration keys with values that are direct objects / literals.
    - A list of vaults, for example, should be in a key simply called `vaults`.
```

Maybe something like `vaultModeOnCreate` similar to how `selectionMode` is named.

hikchoi (changes requested):
Nitpick on config key naming but otherwise tried it out and works as described

packages/common-all/src/types/configs/commands/lookup.ts
nickolay-kondratyev: Good call, renaming.

kevinslin: on the naming part, one more comment. since we aren't changing the *type of the vault* but how we are selecting vaults, can we use `vaultSelectionModeOnCreate` to be super explicit?

packages/plugin-core/src/components/lookup/vaultSelectionModeConfig.ts
```diff
@@ -0,0 +1,44 @@
+import { VaultSelectionMode } from "./types";
+import { ConfigUtils } from "@dendronhq/common-all";
+import { ExtensionProvider } from "../../ExtensionProvider";
+
+/**
+ * Class responsible for proxying interaction with vault
+ * selection mode configuration.
+ * */
+export class VaultSelectionModeConfig {
```
kevinslin, line 9: write now, we have an existing convention for proxying with configuration. can we stick to that? see [ConfigUtils](https://github.com/dendronhq/dendron/blob/ec063f6a0c2e026056d34128f9eab3b0238b5f09/packages/common-all/src/utils.ts#L539)

if we want to make something that is workspace specific, create a new class there

kevinslin (changes requested):
added a comment in the naming thread in addition to this pr review

packages/common-all/src/types/configs/commands/lookup.ts
hikchoi: That makes more sense. It's always good to be as explicit as possible.

packages/plugin-core/src/components/lookup/vaultSelectionModeConfig.ts
hikchoi: Hmm. My intuition is that `ConfigUtils` is more of an abstraction that pertains to retrieval of config values and what we have here is more closely coupled to _how vault selection works_. 

Of course, it's up to debate where we draw the distinction for each utility class' responsibility / abstraction level. My initial impression is that this class is closer to the vault selection implementation detail than a mechinism for config value retrieval.

kevinslin: multiple things going on. when adding utility methods, two things to consider:
1. what package does it go under?
2. where in the package does it go under?

for 1, we always want a general utility to go in the most accessible location. so in order of preference:
- common-all
- common-server
- engine-server
- end client (eg. dendron-cli, plugin-core, etc)

so to answer 1, is this a utility that can be used across multiple clients and is not client specific? if so, it should be put into a `common-*` package

for 2, this depends on context. number one convention - when in doubt, go by what's already done. right now, configuration utilities (anything that is stricly getting a config or a derived config value) is handled in `ConfigUtils`.
furthermore, we already have a [PublishUtils](https://github.com/dendronhq/dendron/blob/763c797c4c2f7821ef747376c980e4a4b0eace8e/packages/common-all/src/utils.ts#L486) in the same file that calculates derived config values for publishing. so we should go with this convention here and add `VaultUtils` here.
some optimizations we can make:
- split off config utilities in a separate file
- make name more specific (eg. `PublishConfigUtils`)

nickolay-kondratyev: From levels of abstraction it would seem that `ConfigUtils` is a lower level abastraction than `VaultSelectionModeConfig` but if you guys want me to move the logic into `ConfigUtils` let me know.

nickolay-kondratyev: Renamed to VaultSelectionModeConfigUtils

kevinslin (approved):

hikchoi (approved):

PR #1979, opened 2021-12-25
enhance(publish): add showChildLinks override to notes

kevinslin:
Be able to control whether a page shows child links using note frontmatter

- cleanup various DNode properties
- separate implicit and explicit DNodeProps

### DIFF
--- a/packages/common-all/src/utils.ts
+++ b/packages/common-all/src/utils.ts
@@ -1,33 +1,32 @@
 import GithubSlugger from "github-slugger";
 import _ from "lodash";
 import minimatch from "minimatch";
+import path from "path";
 import querystring from "querystring";
 import semver from "semver";
 import { COLORS_LIST } from "./colors";
 import {
+  DendronSiteConfig,
+  DHookDict,
+  DVault,
   NoteProps,
   SEOProps,
-  DVault,
-  DHookDict,
-  DendronSiteConfig,
 } from "./types";
 import { TaskConfig } from "./types/configs/workspace/task";
 import {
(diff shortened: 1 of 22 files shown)


packages/common-all/src/utils.ts
```diff
@@ -594,6 +580,26 @@ export class ConfigUtils {
...
+      opts &&
+      opts.note &&
+      opts.note.config &&
+      opts.note.config.global &&
+      !_.isUndefined(opts.note.config.global.enableChildLinks)
+    ) {
+      return opts.note.config.global.enableChildLinks;
+    }
+    return true;
+  }
```
hikchoi, line 601: I think it's a bit confusing to introduce the global config namespace for just this key when we haven't introduced the namespace yet. 

I personally would have added them into the `publishing` and `preview` namespace separately (and explicitly) and then have the config util assume those namespaces' defaults instead.

For the users: there will be cases where they want separate control, but having a global namespace that is implied here like so would have clashing mental models and user intent.

For migration when we eventually introduce the global namespace: we want things in global to be separately available in each respective config namespaces as well. So it makes more sense to introduce them now as such rather than with an assumed global namespace. This way when the global namespace is introduced: 1. we don't have to posthumously add `enableChildLinks` to the two other namespaces (it's already there), and 2. we simply introduce the global namespace and what we have here will be naturally supported.

hikchoi (commented):
left a comment on config namespace concerns

kevinslin:
got it - haven't forgotten about this, will have a pr for this later this week

PR #1982, opened 2021-12-26
chore(internal): Pass Extension into Plugin Command

kevinslin:
Since we deprecated `getExtension`, we need a way for commands to get an instance of the extension.
This is necessary to access engine specific variables in the airtable pod.

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -1005,7 +1005,7 @@ async function _setupCommands(
       vscode.commands.registerCommand(
         DENDRON_COMMANDS.RELOAD_INDEX.key,
         sentryReportingCallback(async (silent?: boolean) => {
-          const out = await new ReloadIndexCommand().run({ silent });
+          const out = await new ReloadIndexCommand(ws).run({ silent });
           if (!silent) {
             vscode.window.showInformationMessage(`finish reload`);
           }
@@ -1021,7 +1021,7 @@ async function _setupCommands(
       vscode.commands.registerCommand(
         DENDRON_COMMANDS.GO_NEXT_HIERARCHY.key,
         sentryReportingCallback(async () => {
-          await new GoToSiblingCommand().execute({ direction: "next" });
+          await new GoToSiblingCommand(ws).execute({ direction: "next" });
         })
       )
     );
@@ -1031,7 +1031,7 @@ async function _setupCommands(
       vscode.commands.registerCommand(
         DENDRON_COMMANDS.GO_PREV_HIERARCHY.key,
         sentryReportingCallback(async () => {
-          await new GoToSiblingCommand().execute({ direction: "prev" });
+          await new GoToSiblingCommand(ws).execute({ direction: "prev" });
         })
       )
     );
@@ -1043,7 +1043,7 @@ async function _setupCommands(
       vscode.commands.registerCommand(
         DENDRON_COMMANDS.RENAME_NOTE.key,
         sentryReportingCallback(async (args: any) => {
-          await new MoveNoteCommand().run({
+          await new MoveNoteCommand(ws).run({
             allowMultiselect: false,
             useSameVault: true,
             ...args,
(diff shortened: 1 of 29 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -1005,7 +1005,7 @@ async function _setupCommands(
       vscode.commands.registerCommand(
         DENDRON_COMMANDS.RELOAD_INDEX.key,
         sentryReportingCallback(async (silent?: boolean) => {
-          const out = await new ReloadIndexCommand().run({ silent });
+          const out = await new ReloadIndexCommand(ws).run({ silent });
```
nickolay-kondratyev, line 1008: nit: we are passing the extension instance explicitly to ReloadIndexCommand while `ReloadIndexCommand` command does not have a constructor that accepts extension. Even though this will compile in TS a bit weird to pass something explicitly to this instance if it doesn't accept it yet.

packages/pods-core/src/v2/pods/export/AirtableExportPodV2.ts
```diff
@@ -42,10 +43,21 @@ export class AirtableExportPodV2
 {
   private _config: RunnableAirtableV2PodConfig;
   private _airtableBase: Base;
-
-  constructor(airtable: Airtable, config: RunnableAirtableV2PodConfig) {
+  // @ts-ignore
```
nickolay-kondratyev, line 46: Why do we need an ignore here? seems like we are pasing in the same type as an argument to the constructor.

packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
```diff
@@ -49,9 +50,16 @@ export abstract class BaseExportPodCommand<
...
    */
-  constructor(hierarchySelector: HierarchySelector) {
+  constructor({
+    hierarchySelector,
+    extension,
+  }: {
+    hierarchySelector: HierarchySelector;
+    extension: IDendronExtension;
+  }) {
     super();
```
nickolay-kondratyev, line 60: Should we pass the extension up to the super()/BaseCommand since it accepts it?

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -62,7 +62,7 @@ export class GotoNoteCommand extends BasicCommand<
   GoToNoteCommandOutput
 > {
   key = DENDRON_COMMANDS.GOTO_NOTE.key;
-  private extension: IDendronExtension;
+  protected extension: IDendronExtension;
```
nickolay-kondratyev, line 65: Probably makes most sense to remove this field of GoToCommand level now and use the BaseCommand.extension, otherwise I believe we are hiding the BaseCommands field and that becomes unintuitive maintainance wise.

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -366,7 +367,7 @@ suite("NoteLookupCommand", function () {
           });
         },
         onInit: async ({ vaults }) => {
-          const cmd = new NoteLookupCommand();
+          const cmd = new NoteLookupCommand(getFakeExtensionForTest());
```
nickolay-kondratyev, line 370: nit: could extract a method along the lines of `createNoteLookupCmdForTest/getNoteLookupCmdForTest` that way if there are more changes to the way we do initialization of this command for this test the diff will be just a single line instead of ~30

packages/plugin-core/src/commands/PublishDevCommand.ts
```diff
@@ -9,13 +9,20 @@ import fs from "fs-extra";
...
 import { DENDRON_EMOJIS } from "@dendronhq/common-all";
+import { IDendronExtension } from "../dendronExtensionInterface";
 
 type CommandOutput = {
   pid: number;
 };
 
 export class PublishDevCommand extends BasicCommand<CommandOutput> {
   key = DENDRON_COMMANDS.PUBLISH_DEV.key;
+  protected extension: IDendronExtension;
```
nickolay-kondratyev, line 20: Attracts caution to re-declare the same field that exists in the base class, any reason why we can't use the `extension` field that is coming from the base class?

nickolay-kondratyev:
One more step away from circles! :D

kevinslin:
going to hold off on this PR so we can spend some time on design of commands going forward

PR #1983, opened 2021-12-27
enhance(pod): support linked records for airtable export

kevinslin:
- support for linked records
- add metadata utilities in [[../packages/engine-server/src/util/noteMetadataUtils.ts]]
- add test for AirpodV2 spec
- introduce new utility method, `isFalsy`
- clean `AirtableExportPodV2` internals
- throw error if note does not have airtable id


kevinslin:
heads up @jonathanyeung that this is already pushed, feel free to close or comment after you've had a chance to look

PR #1990, opened 2021-12-27
fix(lookup): full length word matches should be case insensitive

nickolay-kondratyev:
# fix(lookup): full length word matches should be case insensitive

### Quality Assurance


kevinslin (commented):
should not all matches be case insensitive? when do we want a match to be case sensitive?

nickolay-kondratyev:
Yes, all matches should be case insensitive which they are, added unit test for that in the follow up commit for case insensitive match. 

This fix is to address when we had full length match (length of result is exactly the same as query) (that didn't have special query characters) we werent being case insensitive.

kevinslin (approved):

PR #1997, opened 2021-12-28
fix(commands): seed commands broken by refactor

kevinslin:
Seed command expects a panel as first argument. Take it out of command
registration. Also did some refactoring of registering commands

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -93,6 +93,9 @@ import { WorkspaceInitFactory } from "./workspace/workspaceInitializer";
 import { WSUtils } from "./WSUtils";
 import { AutoCompletableRegistrar } from "./utils/registers/AutoCompletableRegistrar";
 import { isAutoCompletable } from "./utils/AutoCompletable";
+import { IBaseCommand } from "./types";
+import { SeedAddCommand } from "./commands/SeedAddCommand";
+import { SeedRemoveCommand } from "./commands/SeedRemoveCommand";
 
 const MARKDOWN_WORD_PATTERN = new RegExp("([\\w\\.\\#]+)");
 // === Main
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -1053,42 +1057,69 @@ async function _setupCommands(
...
-        DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key,
-        sentryReportingCallback(async () => {
-          await new ShowNoteGraphCommand(
-            NoteGraphPanelFactory.create(ws)
-          ).run();
-        })
-      )
-    );
-  }
+  // // NOTE: graph commands with create a panel when instantiated, hence passing them in via a closure
```
nickolay-kondratyev, line 1094: `with` -> `will`

packages/plugin-core/src/_extension.ts
```diff
@@ -1053,42 +1057,69 @@ async function _setupCommands(
...
+    context,
+    key: DENDRON_COMMANDS.SHOW_SCHEMA_GRAPH.key,
+    cmd: () => new ShowSchemaGraphCommand(SchemaGraphViewFactory.create(ws)),
+    existingCommands,
+  });
+
+  addCommand({
+    context,
+    key: DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key,
+    cmd: () => new ShowNoteGraphCommand(NoteGraphPanelFactory.create(ws)),
```
nickolay-kondratyev, line 1105: If we are subscribing this closure instantiator function to a subscription event we are ok with new instances of the command being instantiated for each subscription event correct?

nickolay-kondratyev (approved):

packages/plugin-core/src/_extension.ts
kevinslin: good catch. reverted this part

Issue #1830, opened 2021-12-02
`nav_order` should define order of nodes in tree view index


hrmnjt:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [x] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup - Dendron's Lookup Command
- [x] Views - Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [ ] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

As of now, `nav_order` provides a way to sort nodes while publishing but not on the tree view. Not having an order on the tree view creates a need to run the NextJS server to check if `nav_order` would work as intended.

## Describe the solution you'd like ##

It would be amazing if `nav_order` property which when defined could order the nodes in tree view to give a consistent experience.

## Describe alternatives you've considered ##

None possible alternatives to order nodes as of now.

## Additional context ##

Link to possibly related issue #440 and discussion on Discord - https://discord.com/channels/717965437182410783/735365126227493004/916094980878123011

kevinslin:
This should come out in today's v72 release :)

hrmnjt:
Hi @kevinslin,

I see this is not working well in v0.72 release. Sharing details here.

Dendron Version
![image](https://user-images.githubusercontent.com/10371494/145705099-7707de8d-5c1a-4749-9dbd-e207153c32f3.png)

`nav_order: 1` shows up first in tree view index
![image](https://user-images.githubusercontent.com/10371494/145705123-dc1d1e55-05c5-4c54-9c1a-1572ac2bb1bf.png)

`nav_order: 2` shows up last (4th) in tree view index
![image](https://user-images.githubusercontent.com/10371494/145705150-70b41dfb-8fd6-4650-bfc5-1604b0675977.png)

`nav_order: 99` shows up 3rd in tree view index
![image](https://user-images.githubusercontent.com/10371494/145705179-4691bdd8-9257-4524-83b9-ff5fe2137d8e.png)

This problem doesn't happen on the publishing but only on VSCode while editing. Let me know if there is a configuration that needs to be edited for this.

[...]

PR #2014, opened 2021-12-30
fix: update tree order when a note changes order

SeriousBug:
The tree view wouldn't update when a note's `nav_order`, title, or `updated' changes (these all affect order in the tree). This PR fixes this issue.

#1830

### DIFF
--- a/packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
@@ -15,12 +15,15 @@ type OnExpandFunc = TreeProps["onExpand"];
 type OnSelectFunc = TreeProps["onSelect"];
 const DendronTreeExplorerPanel: DendronComponent = (props) => {
   const logger = createLogger("DendronTreeExplorerPanel");
-  const noteActive = props.ide.noteActive;
   const engine = props.engine;
   const { config, notes } = engine;
+  const numNotes = _.size(notes);
+  const noteActive = props.ide.noteActive;
   const [activeNoteIds, setActiveNoteIds] = useState<string[]>([]);
   const [roots, setRoots] = useState<DataNode[]>([]);
-  const numNotes = _.size(notes);
+  // Used to avoid recomputing tree data unnecessarily
+  const [noteActiveId, setNoteActiveId] = useState<string>();
+  const [numNotesLast, setNumNotesLast] = useState<number>(numNotes);
 
   logger.info({
     msg: "enter",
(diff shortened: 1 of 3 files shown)


packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
```diff
@@ -15,12 +15,15 @@ type OnExpandFunc = TreeProps["onExpand"];
 type OnSelectFunc = TreeProps["onSelect"];
 const DendronTreeExplorerPanel: DendronComponent = (props) => {
   const logger = createLogger("DendronTreeExplorerPanel");
-  const noteActive = props.ide.noteActive;
   const engine = props.engine;
   const { config, notes } = engine;
+  const numNotes = _.size(notes);
```
kevinslin, line 20: since users can pass files as virtual notes, will this not also change based on that? maybe its not worth handling. if we decide not to handle it, can you add a `// OPTIMIZE:` comment as well as a backlog task for this?

speaking of which, can we add a new cookbook item in [[Views|dendron://dendron.docs/pkg.plugin-core.dev.cook#views]] for how to pass virtual notes into the view?

packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
```diff
@@ -15,12 +15,15 @@ type OnExpandFunc = TreeProps["onExpand"];
...
+  const noteActive = props.ide.noteActive;
   const [activeNoteIds, setActiveNoteIds] = useState<string[]>([]);
   const [roots, setRoots] = useState<DataNode[]>([]);
-  const numNotes = _.size(notes);
+  // Used to avoid recomputing tree data unnecessarily
+  const [noteActiveId, setNoteActiveId] = useState<string>();
+  const [numNotesLast, setNumNotesLast] = useState<number>(numNotes);
```
kevinslin, line 26: this would actually be useful for all views when calculating state. can you add this to [[../packages/common-frontend/src/features/ide/slice.ts]]?

introduce a `noteActivePrev` field and in [[setNoteActive|../packages/common-frontend/src/features/ide/slice.ts#L40]], set it to the active note. 
it will then get set automatically in [[../packages/dendron-plugin-views/src/components/DendronApp.tsx#L89]]

packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
```diff
@@ -44,10 +47,24 @@ const DendronTreeExplorerPanel: DendronComponent = (props) => {
...
+    if (
+      roots.length !== 0 &&
+      noteActiveId !== noteActive?.id &&
+      numNotesLast === numNotes
+    ) {
+      logger.info({
+        msg: "calcRoots:noteChange",
+        noteActiveId: noteActive?.id,
+      });
+      setNoteActiveId(noteActive?.id);
```
kevinslin, line 63: this should be unecessary when we keep track of `noteActivePrev` in `ide.slice`

kevinslin (changes requested):

packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
SeriousBug: > since users can pass files as virtual notes, will this not also change based on that?

I don't think so actually, the virtual notes aren't sent to the tree view, only note preview.

> can we add a new cookbook item

Got it!

kevinslin (approved):

Issue #656, opened 2021-04-21
Refactoring changes title


funnym0nk3y:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

When refactoring the title of a note gets changed. 

## Expected behavior

I'd expect that refatoring preserves the title. For example some titles have spaces or start with a capital letter one added when creating the note. That all gets shmushed. 

## Desktop (please complete the following information)

- vscode: 1.55.2
- dendron: 0.38.0

kevinslin:
Hmm, do you have an example? Refactoring follows the same convention as rename - if the title is the same as what Dendron would autogenerate (kebab-case capitalized), then refactor will change the title as well. Otherwise, it (should) preserve the existing title

funnym0nk3y:
Unfortunately not, but I tried it with another file: 

`Frobeniusmatrix -> frobeniusmatrix` when changing the file hierarchy `mathematik.frobeniusmatrix -> mathematik.lineareAlgebra.frobeniusmatrix`

kpathakota:
I'm not sure that's unexpected. I thought we wouldn't preserve capitalization especially if it's in camel case. @funnym0nk3y - are you still seeing this affect you often?

funnym0nk3y:
I avoided refactoring recently because I didn't want to mess with my files and  transitioned to the recommended kebab-case.

kpathakota:
Try as I might I cannot repro this behavior. Might've been fixed in a recent update, but not sure. @funnym0nk3y, I'd say the behavior is a bug if it continues to repro for you (I literally tried that frobenius -> linearAlgebra example and all the casing stayed just fine). Closing out.

PR #2015, opened 2021-12-30
fix: rename operations modify unnecessary files

hikchoi:
# fix: rename operations modify unnecessary files

This PR:
- Fixes an issue with `FileStorage.renameNote` updating notes with link to a renamed note that doesn't actually have changes.
  - This bug would change notes that:
    - don't have note body changes at all, but is referencing a note that has a header rename.
      - e.g. note `foo` has a header that was changed by `Rename Header`, the note `foo`'s name hasn't actually changed, but internally `Rename Header` uses `renameNote` with the same note location to trigger reference updates. The bug here is that all notes that are not referencing the particular header but the note itself will be counted as modified, which add undesirable changes to the file (adding a `\n` at the end, changing `"` to `'`, etc.)
    - don't have note body (or frontmatter tag) changes, but used double quotes (`"`) in frontmatter, or
      - e.g. lots of linter-like changes as found in https://github.com/dendronhq/dendron-site/pull/338/files
    - don't have note body (or frontmatter tag) changes, but had "mistyped" frontmatter (type mismatched by frontmatter's standard)
      - e.g. `changelog.md` in `dendron-site` had a backlink to the note`weekly.journal.2021.11.09`, which had the frontmatter `title: 09`. A write note operation treated this as a number (probably because of the leading zero) and overwrote it with some garbage value. (if you change this to a very large number like 1000000000000000000000009, the write operation will overwrite this to a scientific notation)
- This bug affected 
  - commands `Rename Header`, `Move Note`, `Refactor Hierarchy`
  - `RenameProvider`
  - `WindowWatcher`

# Cause
- `FileStorage.renameNote` finds note with links to a renamed note in order to find and update all references to the old note if applicable.
- When it grabbed all notes with links to the note being renamed, it passed notes that don't really have meaningful changes at all to `updateOldNoteReferences`, which does a `writeNote` that updates existing notes in the file system, which updated the frontmatter.
  - This is fixed by only passing notes to `updateOldNoteReferences` that have note body change or frontmatter tag changes.

# Pull Request Checklist

### Quality Assurance
- [x] After you submit your pull request, check the output of our [integration test](https://github.com/dendronhq/dendron/actions) and make sure all tests pass

#### Special Cases
- [ ] If your tests changes an existing snapshot, make sure that snapshots are [updated](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#updating-test-snapshots)
- [ ] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in [test-workspace](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#test-workspace). We use this to manually inspect new changes and for auto regression testing 

### Docs

### Git
        - `eg. feat/add-thisthing`

### Analytics
- [ ] If you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Github Issue
- [ ] If this pull request is addressing an existing issue, make sure to link this PR to the issue that it is resolving.
- [ ] If the resolution comes with a document update, link the docs PR to the issue as well.

### DIFF
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -50,6 +50,7 @@ import {
 import {
   DLogger,
   file2Note,
+  genHash,
   getAllFiles,
   getDurationMilliseconds,
   note2File,
@@ -958,12 +959,14 @@ export class FileStorage implements DStore {
         },
         _n
       );
-      n.body = noteMod.body;
-      n.tags = noteMod.tags;
       const shouldChange = !(
         n.body === noteMod.body && n.tags === noteMod.tags
       );
-      if (shouldChange) notesToChange.push(n);
+      if (shouldChange) {
+        n.body = noteMod.body;
+        n.tags = noteMod.tags;
+        notesToChange.push(n);
+      }
     });
 
     /**
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -633,154 +633,154 @@ export class FileStorage implements DStore {
   }
 
   /** Update the links inside this note that need to be updated for the rename from `oldLoc` to `newLoc` */
-  private async processNoteChangedByRename({
-    note,
-    oldLoc,
-    newLoc,
-  }: {
-    note: NoteProps;
-    oldLoc: DNoteLoc;
-    newLoc: DNoteLoc;
-  }): Promise<NoteChangeUpdateEntry> {
-    const prevNote = { ...note };
-    const vault = note.vault;
-    const wsRoot = this.wsRoot;
-    const vaultPath = vault2Path({ vault, wsRoot });
-    // read note in case its changed
-    const _n = file2Note(path.join(vaultPath, note.fname + ".md"), vault);
-    const foundLinks = LinkUtils.findLinks({
-      note: _n,
-      engine: this.engine,
-      filter: { loc: oldLoc },
-    });
-    let allLinks = _.orderBy(
-      foundLinks,
-      (link) => {
-        return link.position?.start.offset;
-      },
-      "desc"
-    );
-    if (
-      oldLoc.fname.toLowerCase() === newLoc.fname.toLowerCase() &&
-      oldLoc.vaultName === newLoc.vaultName &&
-      oldLoc.anchorHeader &&
-      newLoc.anchorHeader
-    ) {
-      // Renaming the header, only update links that link to the old header
-      allLinks = _.filter(allLinks, (link): boolean => {
-        // This is a wikilink to this header
-        if (link.to?.anchorHeader === oldLoc.anchorHeader) return true;
-        // Or this is a range reference, and one part of the range includes this header
-        return (
-          link.type === "ref" &&
-          isNotUndefined(oldLoc.anchorHeader) &&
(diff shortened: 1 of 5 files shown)


hikchoi:
possibly related: https://github.com/dendronhq/dendron/issues/656

hikchoi:
As a follow-up we should identify frontmatter generated by the extension that can possibly be casted during write operations to avoid unwanted frontmatter corruptions.

hikchoi:
Looks like I've broken some tests. I'll change this to draft in the meantime and figure out what's going on.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -649,161 +649,149 @@ export class FileStorage implements DStore {
...
+          return _n;
+        },
+        _n
+      );
+      n.body = noteMod.body;
+      n.tags = noteMod.tags;
+      const shouldChange = !(
+        n.body === noteMod.body && n.tags === noteMod.tags
+      );
+      if (shouldChange) notesToChange.push(n);
```
SeriousBug, line 794: I think (but I'm not sure) that `n` here is a reference to a note in the engine, so changes made like `n.body = noteMod.body` will still cause the note to change in the engine. Updating the code to clone `n` so the note in the engine doesn't change causes the change to not apply at all, so the code seems to rely on this behavior. Hope this helps figure out why the tests are broken.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -649,161 +649,149 @@ export class FileStorage implements DStore {
...
+          });
+          _n.body = newBody;
+          return _n;
+        },
+        _n
+      );
+      n.body = noteMod.body;
+      n.tags = noteMod.tags;
+      const shouldChange = !(
+        n.body === noteMod.body && n.tags === noteMod.tags
```
SeriousBug, line 792: small nitpick: could compare content hash instead

SeriousBug (commented):
Left some thoughts that might help fix the broken tests.

Also, should we add a test for the bug this is fixing (i.e. a test that would fail without this fix) if it isn't too difficult to reproduce?

packages/engine-server/src/drivers/file/storev2.ts
hikchoi: I was being a dummy and checking if there's a change _after_ I set `n.body` to `noteMod.body`. 😅

packages/engine-server/src/drivers/file/storev2.ts
hikchoi: we could, but `noteMod` that we are comparing `n` with doesn't have a contentHash to start with (file2Note doesn't calculate hash). We can pass in `calculateHash` to `file2Note` and get it when we are reading `_n`, but then we also have to re-calculate the content hash of `noteMode` right before we check if there's a change. I think it would be better to just use what we already know might have changed to check here instead of adding hash calculation to an operation that can possibly span out to many notes.

hikchoi:
Having a bit of trouble getting some tests to pass after the rebase. Looks like there were some changes to rename in `storev2` from @SeriousBug that I'm not quite getting right. Will get the fix in soon.

kevinslin (approved):

SeriousBug (approved):

PR #2023, opened 2022-01-01
enhance(pods): various airtable pod enhancements

kevinslin:
- chore(internal): clean up unused bootstrap scripts
- error message when single tag field is wrong
- introduce new note props type
- - make onExportComplete be async - await on async exports inside BaseExportPod
- update export pod logic

# Pull Request Checklist

## Testing
### Docs
### Analytics

### DIFF
--- a/packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
+++ b/packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-await-in-loop */
 import {
   assertUnreachable,
   DNodeProps,
@@ -18,8 +19,8 @@ import path from "path";
 import * as vscode from "vscode";
 import { HierarchySelector } from "../../components/lookup/HierarchySelector";
 import { PodUIControls } from "../../components/pods/PodControls";
+import { ExtensionProvider } from "../../ExtensionProvider";
 import { VSCodeUtils } from "../../vsCodeUtils";
-import { getDWorkspace, getExtension } from "../../workspace";
 import { BaseCommand } from "../base";
 
 /**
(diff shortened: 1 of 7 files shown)


packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
```diff
@@ -179,21 +180,16 @@ export abstract class BaseExportPodCommand<
             break;
           }
           case PodExportScope.Note: {
-            const promises = [];
```
jonathanyeung, line 182: Removing this pattern leads to an ESLint warning about having await inside a for loop.

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -178,6 +179,11 @@ export class AirtableUtils {
     return _.flatten(out);
   }
 
+  /**
+   * Maps a {@linkk SrcFieldMappingV2["type"]} to a field on {@link NoteProps}
```
jonathanyeung, line 183: nit: linkk

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -390,7 +398,7 @@ export class AirtableUtils {
       records.map(async (ent) => {
         const airtableId = ent.id;
         const dendronId = ent.fields["DendronId"] as string;
-        const note = engine.notes[dendronId];
+        const note = engine.notes[dendronId] as NotePropsWithOptionalCustom;
```
jonathanyeung, line 401: Is this actually necessary? what if we do something akin to `isStringNullOrEmpty()` on the NoteAirtableId below.  Can we avoid a cast and a new type definition?

jonathanyeung (changes requested):

packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
kevinslin: that's why I added a es-lint ignore  at the top of the file
`/* eslint-disable no-await-in-loop */`

packages/pods-core/src/builtin/AirtablePod.ts
kevinslin: fixed

packages/pods-core/src/builtin/AirtablePod.ts
kevinslin: our actual `NoteProps` definition is wrong. `custom` is optional and can be undefined. we actually need to refactor the definition (or the code) but that was a much bigger change, hence this type as a stopgap.

we'll fix this as part of january core feature maturity milestones

packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
jonathanyeung: But why change it if the current pattern doesn't warn?

jonathanyeung (approved):

PR #2026, opened 2022-01-01
enhance(markdown): various updates to backlinks and child links

kevinslin:
BREAKING: change backlinks and child links from `<h2> to <strong>`
BREAKING: change order of sections (child links come before backlinks)
BREAKING: remove backlinks and child links from toc


SeriousBug (approved):
Looks good, tested both preview and publishing.

PR #2027, opened 2022-01-02
chore(internal): remove analytics code during tests

kevinslin:
NOTE: I removed the test from `engine-test-utils/src/__tests__/dendron-cli/commands/base.spec.ts` for the following reasons:

- telemetry should be disabled for testing and this changes introduces a test guard to automatically detect and disable telemetry when run in CI or local testing
- while we could implement this by stubbing out the test guard, we would also need to stub out every actual call that reports telemetry
  - this is brittle and will need to be updated whenever we make telemetry related changes

To actually test the telemetry logic, next step is to introduce a single test suite that only runs for telemetry related changes.
Proposal is to create a new package (eg. `dendron-integ-tests`), that tests for expensive end to end csaes (eg. telemetry) that are not handled in the current test suite.


packages/common-server/src/etc.ts
```diff
@@ -14,6 +16,31 @@ export class NodeJSUtils {
   }
 }
 
+/**
+ * Various utilities that are not categorized
+ */
+export class EtcUtils {
```
jonathanyeung, line 22: How about something like `RuntimeUtils` or maybe bundled with the `Environent` stuf? (`getStage`) etc.?

kevinslin: done

hikchoi (approved):

PR #2034, opened 2022-01-03
fix: warn for frontmatter issues even if the frontmatter is not visible

SeriousBug:
The optimizations for note decorations caused frontmatter warnings to be skipped if the frontmatter is not visible within the editor. This PR fixes this by always checking for frontmatter warnings, even if it's not visible. This is actually preferred because warnings should always be displayed even if not visible, because the user could use something like "go to previous issue" to navigate to them.


hikchoi (approved):
looks good 👍🏼

PR #2035, opened 2022-01-03
fix: Publishing dev server keeps running after exiting on Windows

SeriousBug:
Fixes an issue where on Windows, the next server would continue running even after the Dendron CLI is interrupted with Ctrl+C.

I believe this change will cause a terminal window to appear when using the publish dev plugin command, but I was unfortunately unable to test that. This is necessary because it appears that Windows otherwise doesn't let applications handle an interrupt, so node/execa can't stop the server when Dendron exits. See: https://github.com/nodejs/node/issues/29837


packages/pods-core/src/builtin/NextjsExportPod.ts
```diff
@@ -150,7 +148,7 @@ export class NextjsExportPodUtils {
   static async startNextDev(opts: { nextPath: string; quiet?: boolean }) {
     const { nextPath, quiet } = opts;
     const cmdDev = quiet ? "npm run --silent dev" : "npm run dev";
-    const out = $$(cmdDev, { cwd: nextPath });
+    const out = $$(cmdDev, { cwd: nextPath, windowsHide: false });
```
kevinslin, line 151: can we explicitly pass this in as an option from the plugin?

kevinslin (changes requested):

kevinslin:
looks like there's a compiler error

SeriousBug:
Works now!

SeriousBug:
The failing windows plugin tests work fine locally.

kevinslin (approved):

PR #2004, opened 2021-12-29
chore: adding lookup controller and lookup provider interaces and usage

nickolay-kondratyev:
# Pull Request Checklist

-6 circular dependencies compared to master.

### Quality Assurance
- [ ] Add a [test](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#writing-tests) for the new feature
- [x] Make sure all the existing [tests](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#running-all-tests) pass
- [x] Do a spot check by running your feature with our [test workspace](https://wiki.dendron.so/notes/cb22bd36-d45a-4ffd-a31e-96c4b39cb7fb.html#test-workspace)
```
	modified:   src/commands/ConvertLink.ts
	        - Added an ran through manual tests in [[dendron.cmd.convert-link]]
	modified:   src/commands/InsertNoteCommand.ts
	        - Tested being about insert note contents
	modified:   src/commands/InsertNoteLink.ts
	        - Tested being able to insert note link
	modified:   src/commands/MoveHeader.ts
	        - Tested header moved
	modified:   src/commands/MoveNoteCommand.ts
	        - Tested note moved to different vault
	modified:   src/commands/NoteLookupCommand.ts
	        - Tested note look up and auto complete
	        - Tested note creation
	modified:   src/commands/RefactorHierarchyV2.ts
	       X- Looks like there is pre-existing issue (not able to refactor on nightly build either)
	       Nightly build shows message: "Command 'Dendron: Refactor Hierarchy' resulted in an error (Cannot read property 'message' of undefined)"
	modified:   src/commands/RenameNoteV2a.ts
	        - Tested renaming a note
	modified:   src/commands/SchemaLookupCommand.ts
	        - Tested schema looku
	 ```
- [ ] If your tests involve running [[manual Testing|dendron://dendron.dendron-site/dendron.dev.qa.test#manual-testing]], make sure to update [[Test Workspace|dendron://dendron.dendron-site/dendron.dev.ref.test-workspace]] with any necessary notes/additions to perform manual test ^9jtWc6ov340p
- [ ] After you submit your pull request, check the output of our [integration test](https://github.com/dendronhq/dendron/actions) and make sure all tests pass
  - NOTE: If you running MacOS or Linux, pay special attention to the Windows output and vice versa if you are developing on Windows

#### Special Cases
[...]

PR #2037, opened 2022-01-03
fix: regression on onTriggerButton not scoping properly

hikchoi:
# fix: regression on onTriggerButton not scoping properly
This PR:
- Fixes a regression introduced by #2004 where the `onTriggerButton` was not properly getting access of `this`.

- This is fixed by changing the `onTriggerButton` back to an arrow function instead of it being a function expression.
- We can instead keep `onTriggerButton` as a function expression but instead bind them to `this` in the class. Arguably the _better_ way to do it as it can have potential performance / inheritance issues, but since lookup controllers are always created and disposed in the lifecycle of a lookup (derivative) command, I don't think that's a necessary change that we should introduce right now. We should maybe consider it in the future if this no longer is the case.

# Pull Request Checklist

## Testing
- [~] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in the [test workspace](https://docs.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)
### Docs
### Analytics

### DIFF
--- a/packages/plugin-core/src/components/lookup/LookupControllerV3.ts
+++ b/packages/plugin-core/src/components/lookup/LookupControllerV3.ts
@@ -225,7 +225,7 @@ export class LookupControllerV3 implements ILookupControllerV3 {
     Logger.info({ ctx, msg: "exit" });
   }
 
-  async onTriggerButton(btn: QuickInputButton) {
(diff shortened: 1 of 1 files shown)


hikchoi:
fyi @nickolay-kondratyev

PR #2047, opened 2022-01-05
fix: hover preview containing local images on Windows 

SeriousBug:
Fixes an issue with hover previews containing local images where the image would fail to be previewed.

#876


jonathanyeung:
@SeriousBug  - is this one still in Draft, or ready for review?

SeriousBug:
> @SeriousBug - is this one still in Draft, or ready for review?

There's still some issues with tests failing. I'm working on getting the tests working while also verifying it works on Windows.

SeriousBug:
Okay, I think I need some help with this PR because I've been spending too much time on it.

I've managed to solve the issue, when processing the reference in a hover preview you just need to switch to V5 proc, and use `runSync` so that the plugin runs. The problem is, doing so seems to break many tests in weird ways. What I came up with is not clean, but breaks the tests most minimally, but some tests are still failing. I'd appreciate if anyone could take a look at this to see if I'm missing something.

kevinslin:
just a heads up that this is on my radar. will try to get to this today

Issue #2021, opened 2021-12-31
\begin elements not working


themurtazanazir:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [x] Markdown.  Markdown features, such as syntax support and features

## Describe the bug
Any Latex block with `\begin` and `\end` isn't rendering.

## To Reproduce

![Screenshot from 2022-01-01 00-54-09](https://user-images.githubusercontent.com/34167018/147836976-300d7480-56cb-4f23-b720-f5edb0395bfe.png)

## Desktop (please complete the following information)

- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here

SeriousBug:
Hi! Looks like Dendron is using an older version of Katex that doesn't support `\begin{equation}...\end{equation}`. I'll see if we can get an up-to-date version for next week's release.

SeriousBug:
I've looked into this, and unfortunately we are not able to update our dependencies until VSCode allows ESM module support: https://github.com/microsoft/vscode/issues/130367

SeriousBug:
Actually nevermind that, we can update it slighltly which is recent enough to support begin...end blocks. This should be in the next version!

themurtazanazir:
Is this included n the pre release?

SeriousBug:
I believe it is, please let me know if you are having problems with it and I can investigate.

themurtazanazir:
It is working but not the equation numbers. 

```latex
$$
\begin{equation}
f_x = \frac{\partial f}{\partial x}_{(x_0, y_0)} = \lim_{\Delta x \rightarrow 0}\frac{f(x_0+ \Delta x, y_0)-f(x_0, y_0)}{\Delta x}
\end{equation}
$$
```

![Screenshot from 2022-01-09 10-30-27](https://user-images.githubusercontent.com/34167018/148670121-ad314da4-862f-4cf8-92ae-f74f3ad9890a.png)

[...]

PR #2048, opened 2022-01-05
chore: update katex to allow more math syntax

SeriousBug:
Updates our `katex` version as up-to-date as we can get before an ESM-only version, which enables more math syntax to be used.

#2021


kevinslin (approved):

SeriousBug:
> can you fix the compile issue before merging?

FYI the error is in master too, not a change in this PR

Issue #2045, opened 2022-01-05
Template variables creating single-digit numbers for day and month


tsnieman:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [x] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When using [the example provided on the template variables docs](https://wiki.dendron.so/notes/861cbdf8-102e-4633-9933-1f3d74df53d2/#template-variables), a date such as `2021.1.4` is produced rather than the documented `2021.01.04`

## To Reproduce

1. Create a template
2. Fill it with something like: `{{ CURRENT_YEAR }}.{{ CURRENT_MONTH }}.{{ CURRENT_DAY }}` (per the [aforementioned example](https://wiki.dendron.so/notes/861cbdf8-102e-4633-9933-1f3d74df53d2/#template-variables))
3. Use the template
4. Date is `2021.1.4`

## Expected behavior

I expect a two-digit day & month, (per the [aforementioned example](https://wiki.dendron.so/notes/861cbdf8-102e-4633-9933-1f3d74df53d2/#template-variables)):
```md
CURRENT_MONTH: The month as two digits (example '02')
CURRENT_DAY: The day of the month as two digits (example '08')
```

## Screenshots

![image](https://user-images.githubusercontent.com/595711/148147726-3273d575-d29a-4aa7-8f9a-ecb201072d75.png)

## Desktop (please complete the following information)

- OS: [e.g. iOS] ubuntu
- Browser: n/a
- Version: Dendron v0.76.0

## Additional context

n/a

## Dendron Log file

n/a

kevinslin:
Thanks for the catch, we'll get on this right away!

kevinslin:
this has been fixed. heads up that the template variable syntax has changed -> https://wiki.dendron.so/notes/GelEQPZrSgr3CK9y10Nrg.html#template-variable-replacement

PR #2064, opened 2022-01-05
fix(schema): Ensure month/day/time has two digits when doing data variable substitution

tma66:
Background: https://github.com/dendronhq/dendron/issues/2045

Testing:
Updated unit tests and tested manually using {{ CURRENT_DAY }}
Screen Shot 2022-01-05 at 12 57 37

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1380,12 +1380,12 @@ export class SchemaUtils {
       const currentDate = Time.now();
       const compiledSchemaTemplate = _.template(note.body);
       note.body = compiledSchemaTemplate({
-        CURRENT_YEAR: currentDate.year,
-        CURRENT_MONTH: this.addPadding(currentDate.month),
-        CURRENT_DAY: this.addPadding(currentDate.day),
-        CURRENT_HOUR: this.addPadding(currentDate.hour),
-        CURRENT_MINUTE: this.addPadding(currentDate.minute),
-        CURRENT_SECOND: this.addPadding(currentDate.second),
+        CURRENT_YEAR: currentDate.toFormat("yyyy"),
+        CURRENT_MONTH: currentDate.toFormat("LL"),
+        CURRENT_DAY: currentDate.toFormat("dd"),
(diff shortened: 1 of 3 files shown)


packages/common-all/src/dnode.ts
```diff
@@ -1945,6 +1945,16 @@ export class SchemaUtils {
...
   }
 
+  /**
+   * Adds leading zero if number is a single digit.
+   *
+   * Example: addPadding(3) -> "03"
+   *          addPadding(23) -> "23"
+   */
+  static addPadding(number: number) {
+    return number >= 0 && number < 10 ? "0" + number : "" + number;
```
kevinslin, line 1955: generally not cosure to add a "string" to a number without explicit casting since it can lead to unexpected bugs

also, since you are using luxon, there is native format strings which you can (and should) use

https://github.com/moment/luxon/blob/master/docs/formatting.md

kevinslin (changes requested):

packages/common-all/src/dnode.ts
tma66: Gotcha, makes sense

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -794,14 +794,15 @@ suite("NoteLookupCommand", function () {
           });
           const document = VSCodeUtils.getActiveTextEditor()?.document;
           const newNote = WSUtils.getNoteFromDocument(document!);
+          const currentDate = Time.now();
```
kevinslin, line 797: this test is problematic because it runs the same logic to test as the code. better to stub the date and then test if the format is correct. see https://stackoverflow.com/questions/31591098/how-do-i-stub-new-date-using-sinon

kevinslin (changes requested):

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
tma66: fixed

kevinslin (approved):

PR #2049, opened 2022-01-05
Chore: refactor NoteSyncService

jonathanyeung:
## Chore: refactor NoteSyncService

Restructuring how NoteSyncService works to remove circular dependencies:

- NoteSyncService no longer exposes methods to signify the changing of a note (Except for one instance of `syncNoteMetadata` in FileWatcher, which will be refactored later).  NoteSyncService now encapsulates logic of when the engine needs to be updated with NoteProps.
- NoteSyncService now exposes a vscode.Event that fires when a set of NoteProps has been updated on the Engine. This is useful for places like Preview and BacklinksPanel.  The eventing model also further untangles the PreviewPanel lifecycle: if the Panel isn't visible, then it doesn't need to have a callback actively subscribed to the NoteSyncService event. This should also help solve timing issues we have with `selection2link` and the backlinks panel.

Corresponding dev doc change: https://github.com/dendronhq/dendron-docs/pull/7 

---

# Pull Request Checklist

If this is your first time submitting a pull request to Dendron, copy and paste the full [Dendron Review Checklist](https://docs.dendron.so/notes/1EoNIXzgmhgagqcAo9nDn.html) into this request and check off each item as necessary. 

This template contains the short checklist which is used by the Dendron core team. 

## Testing
- [ ] [Write Tests](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#writing-tests) 
- [ ] [Confirm existing tests pass](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#executing-tests)
- [ ] [Confirm manual testing](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#manual-testing) 
- [ ] If your tests changes an existing snapshot, make sure that snapshots are [updated](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#updating-test-snapshots)
- [ ] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in the [test workspace](https://docs.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)
### Docs
- [ ] if your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR
### Analytics
- [ ] if you are adding analytics related changes, make sure the [Telemetry](https://docs.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

PR #2082, opened 2022-01-07
fix(views): tree view refresh and circ dependency removal

jonathanyeung:
## fix(views): tree view refresh and circ dependency removal

- https://github.com/dendronhq/dendron/commit/b66032fef1b8cb5f7a6fa522a5e0ad14ac4d8388 - this change introduced a fix to have tree view update when the frontmatter of a note got edited.  However, it also introduced a bug where tree view would re-appear when editing a note, even if it were hidden
- https://github.com/dendronhq/dendron/pull/2049 - this accidentally removed both the fix and the bug of the above change during merge resolution

This change re-adds the fix but also addresses the panel pop up bug.  Furthermore, I did a few small changes to remove circular dependency issues associated with `DendronTreeViewV2.ts`

NOTE: some refresh scenarios still don't work in TreeViewV2 (and weren't working, even after https://github.com/dendronhq/dendron/commit/b66032fef1b8cb5f7a6fa522a5e0ad14ac4d8388 ).  For example, running `RenameNote` causes staleness in TreeView. Those are out of scope of this change and will be addressed later.

---

# Pull Request Checklist

## Testing
### Docs
### Analytics

### DIFF
--- a/packages/plugin-core/src/views/utils.ts
+++ b/packages/plugin-core/src/views/utils.ts
@@ -8,9 +8,10 @@ import {
 import { findUpTo, WebViewCommonUtils } from "@dendronhq/common-server";
 import path from "path";
 import * as vscode from "vscode";
+import { IDendronExtension } from "../dendronExtensionInterface";
 import { Logger } from "../logger";
 import { VSCodeUtils } from "../vsCodeUtils";
-import { DendronExtension, getDWorkspace, getExtension } from "../workspace";
+import { getDWorkspace, getExtension } from "../workspace";
 
 export class WebViewUtils {
   /**
@@ -100,7 +101,7 @@ export class WebViewUtils {
     key,
     webviewView,
   }: {
-    ext: DendronExtension;
+    ext: IDendronExtension;
     key: DendronTreeViewKey;
     webviewView: vscode.WebviewView;
   }) {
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/views/utils.ts
```diff
@@ -100,7 +101,7 @@ export class WebViewUtils {
     key,
     webviewView,
   }: {
-    ext: DendronExtension;
+    ext: IDendronExtension;
```
hikchoi, line 104: The new lookup view also calls this method. It currently grabs the instance of the extension with the deprecated bare `getExtension` method. Would changing the signature here cause any issue there?

hikchoi (commented):
left a comment on a slightly tangential concern but otherwise the change looks good 👍🏼

hikchoi:
I just pulled this branch and tested out. I don't think it's related to the changes in this PR but was the scroll behavior in the webview always a bit weird?
It does update the scroll position on refresh (when a note changes), but sometimes it scrolls all the way down or not at all.

packages/plugin-core/src/views/utils.ts
jonathanyeung: If it's compiling, it should be fine - changing this siganture shouldn't change anything functionally.

jonathanyeung:
> I just pulled this branch and tested out. I don't think it's related to the changes in this PR but was the scroll behavior in the webview always a bit weird? It does update the scroll position on refresh (when a note changes), but sometimes it scrolls all the way down or not at all.

Yeah - I can repro to.  There's still quite a bit of weirdness in the tree view.

packages/plugin-core/src/views/utils.ts
hikchoi: can confirm it works fine

kevinslin (approved):

Issue #1608, opened 2021-10-28
Import Pod Markdown is modified my own notes body content


voothi:
## Please select the area your bug applies to. (Multiple selections are Possible)

- [x] Onboard. Initial vault setup and import/export pods
- [ ] Create. Note creation, lookup, snippets and templates
- [x] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [x] Misc (Choose this if your not sure)

## Describe the bug

Is it possible to make import pod markdown without changing my note's body?
Found that after imports changed
  - `__Tag__` на `**Tag**` if this is not [[wikilink]]
  - add `\` before `_` 
  - `\\`
  - `<>` wrap links
  - `&lt;` in begin some lines
This all сreates difficulties in using the note with the code. The code is not everywhere goes to ```

## To Reproduce

I. Prepare
1. Create $dir1 (Source files)
`./Dendron/Import/Src/20211028102351-import-changing-body `
2. Copy file 
from
[20211028102351_Test-Import_Pod_Markdown.md](https://github.com/dendronhq/dendron/files/7432350/20211028102351_Test-Import_Pod_Markdown.md)
to 
$dir1
3. Create $dir2 (Destination files)
`./Dendron/Import/Dst`
4. Open VSCode
5. Copy $dir2 path to clipboard
6. Run Dendron Initialize Workspace
7. Paste $dir2 path
8. Copy 20211028102351-import-changing-body to clipboard
9. Run Dendron Add Vault
10. Paste $name

II. Import Pod Markdown
1. Edit config.import.yml
src: $dir1 
vaultName: $name
2. Run Dendron Import Pod Markdown

III. Compare files
1. src file from $dir1/20211028102351_Test-Import_Pod_Markdown.md
2. dst file from $dir2/20211028102351_test-import_pod_markdown.md

This steps shown in video 
Look links to ## Screenshots below

## Expected behavior

After complete Dendron Import Pod Markdown in each my imported markdown notes will have
- filename in only lower case (as well as now) - OK
- Dendron added frontmatter unique section  - OK
- my own content is not modefied in no way! - Fail

## Screenshots

https://www.youtube.com/watch?v=g5e993KDLx4&ab_channel=Voothi
or
https://youtu.be/g5e993KDLx4

## Desktop (please complete the following information)

- OS: 
```
OS Name:                   Microsoft Windows Server 2012 R2 Standard
[...]

PR #2084, opened 2022-01-07
enhance: Markdown import pod avoids reformatting files

SeriousBug:
Markdown import pod now only modifies lines containing links directly, and avoids reformatting imported notes. This leads to cleaner imports.


Harshita-mindfire (approved):
Tested with the file provided in #1608, looks good.

jonathanyeung (approved):

PR #2095, opened 2022-01-08
enahnce(pod): introduce filters field for airtable config that allows filtering by fname

kevinslin:
Docs: https://github.com/dendronhq/dendron-site/commit/e4387010bebb390f344fedbe573cc07638e2fbb2

Some future work:
- make this generic for all export pods
- support filtering by vault


jonathanyeung:
What's the use case you had in mind for this? 

I feel like this kind of thing should be defined or configured in the export scope, not in airtable pod config.

Harshita-mindfire (commented):
If the use case is to `filter out` notes when we perform bulk export, it looks good and works fine.

PR #2100, opened 2022-01-08
enhance(pod): Support Null and Required Properties for Scalar Values

kevinslin:
Docs:
https://github.com/dendronhq/dendron-site/commit/9f9e1ae26b5cc57d775accd16e0556125075b004

### DIFF
--- a/packages/engine-server/src/util/noteMetadataUtils.ts
+++ b/packages/engine-server/src/util/noteMetadataUtils.ts
@@ -51,7 +51,7 @@ export class NoteMetadataUtils {
     });
   }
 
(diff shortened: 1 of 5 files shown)


packages/engine-server/src/util/noteMetadataUtils.ts
```diff
@@ -31,60 +51,83 @@ export class NoteMetadataUtils {
     });
   }
 
+  static checkIfAllowNullOrUndefind(
```
jonathanyeung, line 54: nit: sp on `undefined`

jonathanyeung (approved):

PR #2109, opened 2022-01-10
fix(cli): using `--noBuild` with export from CLI will cause command to hang

kevinslin:


hikchoi (approved):
good catch! sorry about that.

PR #2113, opened 2022-01-10
enhance: detect config / client compatibility mismatch in cli

hikchoi:
# enhance: detect config / client compatibility mismatch in cli
This PR:
- adds compatibility validation for config and client.
- adds validation logic to CLI commands
  - all cli commands excluding `dendron dev` will have have this validation logic running prior to execution.
  - when there is a client / config compatibility mismatch, the process will exit and display a message informing how to fix it.
  - image
  - image
  - later this will be enhanced to have an option to pass `--yes` to automatically fix (if possible) the mismatch (useful in CI/CD)
- adds tracking of client / config mismatch events
- fixes client version fetching logic in CLI (will now fetch from package.json rather than `process.env.npm_package_version` which will alway be `undefined` because we are not running an npm script.)

NOTE
- splitting some changes that will be more immediately useful from this task: [[Config Migration Implementation|dendron://private/task.workspace.2022.01.03.config-migration-implementation]] in the spirit of iteration / timely delivery.

# Pull Request Checklist

## Testing
### Docs
### Analytics

### DIFF
--- a/packages/dendron-cli/src/commands/base.ts
+++ b/packages/dendron-cli/src/commands/base.ts
@@ -14,7 +14,11 @@ import {
   SegmentClient,
   TelemetryStatus,
 } from "@dendronhq/common-server";
-import { WorkspaceService, WorkspaceUtils } from "@dendronhq/engine-server";
+import {
+  ALL_MIGRATIONS,
+  DConfig,
+  WorkspaceUtils,
+} from "@dendronhq/engine-server";
 import _ from "lodash";
 import yargs from "yargs";
 import { CLIAnalyticsUtils } from "../utils/analytics";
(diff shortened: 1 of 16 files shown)


hikchoi:
just rebased in case the recent bunch of refactoring in the codebase affects the features added here.

jonathanyeung (approved):

jonathanyeung:
minor nit: I find one aspect of the console message a little bit confusing. 

- Maybe when config is out of date, let's not show the line "minimum compatible client version" (and just the other 3 lines)
- When Client is out of date, let's not show "minimum compatible config version"?

Just a minor nit, otherwise the console prompt is very helpful.

packages/dendron-cli/src/commands/base.ts
```diff
@@ -102,6 +106,51 @@ export abstract class CLICommand<
...
+        `${reason} is out of date.`,
+        "------------------------------------------------",
+        `current client version:            ${clientVersionOkay} ${clientVersion}`,
+        `minimum compatible client version:    ${minCompatClientVersion}`,
+        `current config version:            ${configVersionOkay} ${configVersion}`,
+        `minimum compatible config version:    ${minCompatConfigVersion}`,
+        "------------------------------------------------",
+        instruction,
+      ].join("\n");
+      console.log(message);
```
kevinslin, line 142: can you add eslint disable next line?

packages/dendron-cli/src/commands/base.ts
```diff
@@ -102,6 +106,51 @@ export abstract class CLICommand<
     this.L.info({ msg: `Telemetry is disabled? ${segment.hasOptedOut}` });
   }
 
+  async validateConfig(opts: { wsRoot: string }) {
+    const { wsRoot } = opts;
+    const ws = new WorkspaceService({ wsRoot });
+    const dendronConfig = ws.config;
+    const configVersion = ConfigUtils.getProp(dendronConfig, "version");
```
kevinslin, line 113: if the `version` doesn't exist, we currently assume that it equals 4. can this get us into trouble?

packages/dendron-cli/src/commands/base.ts
```diff
@@ -102,6 +106,51 @@ export abstract class CLICommand<
...
+    const clientVersion = CLIUtils.getClientVersion();
+    const validationResp = ConfigUtils.configIsValid({
```
kevinslin, line 115: its okay to say this is out of scope for this but we should also add a followup to check if the config is **actually** valid for the given version.
a user can keep the verison but manually change the settings. ideally, we shhould use json schema to validate this (lets not use a mess of if statements).
each config version can come with its own json schema and we can use `ajv` to validate

kevinslin (changes requested):

packages/dendron-cli/src/commands/base.ts
hikchoi: yes this is the plan down the line.

packages/dendron-cli/src/commands/base.ts
hikchoi: That's right. would need to look at the raw value and not assume a default with `getProp`.
For now I think it makes sense to stop all execution if the `version` field doesn't exist.
Later down the line once we introduce json schema validation, we will have to take snapshots of every version of the JSON schema and use that to determine the real `version`

hikchoi:
UPDATE:

1. made the message less confusing:

image

image

2. will now display a message to make sure `version` is present and correct in `dendron.yml`
3. now use `this.print` instead of bare `console.log`

kevinslin (approved):

PR #2114, opened 2022-01-10
chore(pods): migrate notion export pod to v2

Harshita-mindfire:
migrates notion export to pods v2

### DIFF
--- a/packages/plugin-core/src/commands/pods/NotionExportPodCommand.ts
+++ b/packages/plugin-core/src/commands/pods/NotionExportPodCommand.ts
@@ -1,0 +1,257 @@
+import { ErrorFactory, NoteProps, ResponseUtil } from "@dendronhq/common-all";
+import {
+  Client,
+  ConfigFileUtils,
+  createRunnableNotionV2PodConfigSchema,
+  ExportPodV2,
+  ExternalConnectionManager,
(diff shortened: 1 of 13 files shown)


packages/plugin-core/src/commands/pods/NotionExportPodCommand.ts
```diff
@@ -0,0 +1,259 @@
...
+      const mngr = new ExternalConnectionManager(getExtension().podsDir);
+      // If the apiKey doesn't exist, see if we can first extract it from the connectedServiceId:
+      if (opts?.connectionId) {
+        const config = mngr.getConfigById<NotionConnection>({
+          id: opts.connectionId,
+        });
+
+        if (!config) {
+          window.showErrorMessage(
+            `Couldn't find service config with the passed in connection ID ${opts.connectionId}.`
```
jonathanyeung, line 76: Maybe let's enhance this msg a bit to make it more user friendly:

"Couldn't find service config with the specified connection ID ${id}. Ensure that the connection ID is correct. You can check existing connections with the `Dendron: Configure Service Connection` command.

jonathanyeung (changes requested):

jonathanyeung (approved):

PR #2116, opened 2022-01-10
fix(pods): display all configs on export pod v2

Harshita-mindfire:
```markdown
This PR allows the `export pod V2` command to display a pod config when `exportScope` is omitted from the custom config.
```

## Testing
### Docs
### Analytics

### DIFF
--- a/packages/pods-core/src/v2/PodConfigManager.ts
+++ b/packages/pods-core/src/v2/PodConfigManager.ts
@@ -60,10 +60,11 @@ export class PodV2ConfigManager {
       const config = ConfigFileUtils.getConfigByFPath({
         fPath,
       });
-
-      // if (isExportPodConfigurationV2(config)) {
-      //   configs.push(config);
-      // }
+      /**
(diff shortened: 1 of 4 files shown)


packages/pods-core/src/v2/PodConfigManager.ts
```diff
@@ -62,9 +61,10 @@ export class PodV2ConfigManager {
         fPath,
       });
 
-      if (isExportPodConfigurationV2(config)) {
-        configs.push(config);
-      }
+      // if (isExportPodConfigurationV2(config)) {
+      //   configs.push(config);
+      // }
+      configs.push(config);
```
jonathanyeung, line 67: Don't leave the commented code in - let's just delete it. 

Can we also add a TODO - we should give some UI feedback to a user if their configuration has issues that makes it not conform to what a V2 config requires.  Either
- perhaps we can even have some visual cue within the quick pick itself (like a detail that says "Config missing properties: podId"
- When the config is open, have some sort of code warning.

jonathanyeung (changes requested):

PR #2124, opened 2022-01-11
enhance: allow notes to override the pretty refs setting

SeriousBug:
Notes are now allowed to override the pretty refs setting for themselves. All references inside that note will be rendered following the override. The override doesn't apply to nested references, the user would need to manually override the nested notes as well if needed.

Docs: https://github.com/dendronhq/dendron-site/pull/354

This PR doesn't change the circular dependency count.

### DIFF
--- a/packages/engine-server/src/markdown/remark/noteRefsV2.ts
+++ b/packages/engine-server/src/markdown/remark/noteRefsV2.ts
@@ -277,6 +277,12 @@ function convertNoteRef(opts: ConvertNoteRefOpts): {
     vault,
     engine,
   });
+  if (prettyRefs === undefined) {
+    prettyRefs = ConfigUtils.getEnablePrettyRefs(config, {
+      note: containingNote,
+      shouldApplyPublishRules,
+    });
+  }
   if (
     containingNote?.custom.usePrettyRefs !== undefined &&
     _.isBoolean(containingNote.custom.usePrettyRefs)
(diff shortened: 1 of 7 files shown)


jonathanyeung:
> This PR doesn't change the circular dependency count.

Thanks for checking!

packages/engine-server/src/markdown/remark/noteRefsV2.ts
```diff
@@ -260,14 +262,28 @@ function convertNoteRef(opts: ConvertNoteRefOpts): {
...
   }
 
+  // The note that contains this reference might override the pretty refs option for references inside it.
+  const containingNote = NoteUtils.getNoteByFnameFromEngine({
+    fname,
+    vault,
+    engine,
+  });
+  if (
+    containingNote?.custom.usePrettyRefs !== undefined &&
```
kevinslin, line 281: a few points:
- we now have a lot of reserved frontmatter keys, we want to scope them to some top level modules
- we already have a option for pretty ref in new config, lets use that `enablePrettyRef`
- with config migration, we are moving all config keys into modules, so lets also do that

for now, since we haven't migrated the `site` namespace to `publish`, i would recommend using the `global` namepsace. so the end result would look like this:

```yml
config:
  global:
    enablePrettyRefs
```

you can see how we currently do it for childLinks here -> [getEnableChildLinks](https://github.com/dendronhq/dendron/blob/93f256103fc8d99ca2777b38fe5c26d074eb7de6/packages/common-all/src/utils.ts#L666)

kevinslin (changes requested):

PR #2152, opened 2022-01-12
enhance/fix clipboard

kevinslin:
- chore: publish patch
- fix(views): enable copy plaintext from preview

### DIFF
--- a/packages/common-server/src/etc.ts
+++ b/packages/common-server/src/etc.ts
@@ -112,6 +112,21 @@ export class WebViewCommonUtils {
           copyToClipboard(htmlSelection);            
         }  
       });
+
+      /**
+       * Based off https://stackoverflow.com/questions/822452/strip-html-from-text-javascript
+       */
+      function clean(text) {
+        const out = text
+          .replace(/<style[^>]*>.*<\\/style>/gm, "")
+          // Remove script tags and content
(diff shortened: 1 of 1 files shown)


nickolay-kondratyev:
The change in packages/common-server/src/etc.ts  looks reasonable. 2 things though:

1. There are 33 files modified in this commit, could we decouple those seemengly unrelated changes from the change that is within packages/common-server/src/etc.ts 

2. Just manual validation in the PR comment: 
  2a. Does it work when copy pasting into Gmail(or gmail like client)?
  2b. Does it work when copy pasting into VSCode editor?

kevinslin:
Ah, 

> The change in packages/common-server/src/etc.ts looks reasonable. 2 things though:
> 
>     1. There are 33 files modified in this commit, could we decouple those seemengly unrelated changes from the change that is within packages/common-server/src/etc.ts
> 
>     2. Just manual validation in the PR comment:
>        2a. Does it work when copy pasting into Gmail(or gmail like client)?
>        2b. Does it work when copy pasting into VSCode editor?

Had non-rebased changes. Rebased and cleaned up. Only one file fix.

2a. Does it work when copy pasting into Gmail(or gmail like client)? yes
2b. Does it work when copy pasting into VSCode editor? yes

nickolay-kondratyev (approved):

PR #2158, opened 2022-01-13
fix: fix schema template match when there is same grandchild from two different schema parents

nickolay-kondratyev:
# Pull Request Checklist

## Testing
- [x] If your tests changes an existing snapshot, make sure that snapshots are [updated](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#updating-test-snapshots)tml)

## Manual testing
Created testSchema:
```
version: 1
schemas:
  - id: testSchema
    children:
      - child1
      - child2
    title: testSchema
    parent: root
  - id: child1
    children:
      - sharedGrandchild
  - id: child2
    children:
      - sharedGrandchild
  - id: sharedGrandchild
    template: template.test
```

Created new notes 
* testSchema.child1.sharedGrandchild
* testSchema.child2.sharedGrandchild

Prior to this change only child2 used to have the template applied after this change both of the newly created files got the template applied to them.

### DIFF
--- a/packages/common-server/src/parser.ts
+++ b/packages/common-server/src/parser.ts
@@ -327,10 +327,7 @@ export class SchemaParserV2 extends ParserBaseV2 {
               childClone.data === undefined ||
               childClone.data.pattern === undefined
             ) {
-              if (childClone.data === undefined) {
-                childClone.data = {};
-              }
-              childClone.data.pattern = childId;
+              _.set(childClone, "data.pattern", childId);
             }
 
             // Parent likely already has a reference to the original child identifier
--- a/packages/common-server/src/parser.ts
+++ b/packages/common-server/src/parser.ts
@@ -24,6 +24,7 @@ import AjvErrors from "ajv-errors";
 
 class AJVProvider {
   static ajv: Ajv;
+
   static getAjv() {
     if (this.ajv === undefined) {
       this.ajv = new Ajv({ allErrors: true, allowUnionTypes: true });
@@ -50,6 +51,7 @@ export class ParserBaseV2 {
     return this.opts.logger;
   }
 }
+
 const DEFAULT_LOG_CTX = "parsingSchemas";
 
 async function getSchemasFromImport(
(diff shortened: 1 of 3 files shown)


nickolay-kondratyev:
Tests are passing now so should be good to review.

hikchoi (approved):
LGTM. manually tested and confirmed working 👍🏼

hikchoi:
madge is reporting 5 more circ dependencies on this branch fyi.

kevinslin (changes requested):
I think this PR needs a change and also a followup discussion. 

Today, one of the limitations we've set (by design) is to have a note only have one parent.
This fundamentally changes that approach by allowing a note to have multiple parents. 

I think it makes sense to have in schemas but by adding it to `DNodeProps`, we're now saying its applicable for **notes** and **schemas**.

For this PR, `parents` should only be added to [SchemaProps](https://github.com/dendronhq/dendron/blob/fdd1178ed6d91a4a5017f706e9f2c6da90f3dfb2/packages/common-all/src/types/foundation.ts#L227). 

Some bigger design followups:
1. if we introduce `parents` to `SchemaProps`, then all schemas should have multiple parents. the `if/then` case of checking for multiple parent type will be bug prone and causes extra overhead
2. if we introduce `parents` to `SchemaProps`, does it also make sense to re-visit having multiple parents in `NoteProps`? if so, how does that impact `lookup`?

For next steps, we should schedule an async to discuss the implications

nickolay-kondratyev:
Updated the PR with cloning the diamond grandchild instead of creating multi parents.

packages/common-server/src/parser.ts
```diff
@@ -301,16 +303,51 @@ export class SchemaParserV2 extends ParserBaseV2 {
...
+        const child = schemasDict[childId];
 
-          addConnections(child);
+        if (child) {
+          if (child.parent === null) {
+            // Child does not have a parent pointers hence we can linkup the parent child as is.
+            DNodeUtils.addChild(parent, child);
+            addConnections(child);
+          } else {
+            // Child already contains a parent pointer hence we need to create a schema clone
```
jonathanyeung, line 315: Hmm, what if child.parent === parent? Do we need to deep clone in that case? 

I'm looking through the surrounding code, not sure I _fully_ understand it w/o utilizing a debugger haha.  But if I have a schema like 

```yml
- id: foo
  children:
    - fooschild
    - wildchild

- id: fooschild
  parent: foo
  
 - id: wildchild
```

Then, at this point in the code
- the node with id `fooschild` has a parent pointer to `foo`?
- 'wildchild' has no parent pointer?
- 'foo' has 2 children?

jonathanyeung (changes requested):
Mostly just questions

packages/common-server/src/parser.ts
```diff
@@ -301,16 +303,51 @@ export class SchemaParserV2 extends ParserBaseV2 {
...
+            // otherwise we would end up writing over the already existing parent pointer.
+            const childClone = _.cloneDeep(child);
+            // We use a dictionary hence we need to generate a brand new identifier for our clone.
+            childClone.id = `${childId}_${genUUID()}`;
```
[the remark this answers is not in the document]
hikchoi: The schema graph view does show two separate grand children

image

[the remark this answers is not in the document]
jonathanyeung: o - I think that's fine - I meant like the GUID (i.e. something like `63cee820-76e8-4ce8-b452-00a12a6aab05` showing up)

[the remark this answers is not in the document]
nickolay-kondratyev: Haven't seen the GUID showing up in lookup, or anywhere else, as far as I have seen thus far should be internal.

packages/common-server/src/parser.ts
nickolay-kondratyev: Interesting case. We would end up cloning fooschild in this case (even though we don't really have to). However, it should not hurt us since we will take out the original fooschild reference from the parent. 

Putting some print statements for this case we end up with:
```
Starting to process parent 'foo' with children: 'fooschild,wildchild'
Processing 'fooschild' which has parent 'foo'
Cloning child: 'fooschild'
Starting to process parent 'fooschild_9q31V7vxlhhzzF0odUraW' with children: ''
Parent 'foo' children: 'wildchild,fooschild_9q31V7vxlhhzzF0odUraW'
Processing 'wildchild' which has parent 'null'
Starting to process parent 'wildchild' with children: ''
Parent 'foo' children: 'wildchild,fooschild_9q31V7vxlhhzzF0odUraW'
Starting to process parent 'root' with children: ''
```

packages/common-server/src/parser.ts
```diff
@@ -301,16 +303,51 @@ export class SchemaParserV2 extends ParserBaseV2 {
...
+
+            // When it comes to schemas we often use the id as a matching pattern
+            // unless there is a pattern explicitly specified. Since we created a randomized
+            // identifier the id of our clone no longer can be used for pattern matching.
+            // Hence if there isn't a pattern we will set the pattern to match original id.
+            if (
+              childClone.data === undefined ||
+              childClone.data.pattern === undefined
+            ) {
+              if (childClone.data === undefined) {
```
kevinslin, line 330: heads up that you can use `_.set(childClone, "data.pattern", childId)` to simplify

packages/common-server/src/parser.ts
```diff
@@ -301,16 +303,51 @@ export class SchemaParserV2 extends ParserBaseV2 {
...
+                childClone.data = {};
+              }
+              childClone.data.pattern = childId;
```
kevinslin, line 333: does this take into account if the parent was a `namespace`? i think it works but can we add a test case for it?

kevinslin (changes requested):
can you add a test for the following cases?
- parent is a namespace
- the diamond shape is used as an import to another schema

packages/common-server/src/parser.ts
nickolay-kondratyev: Added _.set

packages/common-server/src/parser.ts
nickolay-kondratyev: Added tests for namespace and inclusion cases.

jonathanyeung (approved):

PR #2164, opened 2022-01-14
fix(refactor): bad refactor logic

kevinslin:
details of fix process in [[Refactor Issue|dendron://private/user.kevin.journal.2022.01.13.s.refactor-issue]]

### DIFF
--- a/packages/engine-server/src/utils.ts
+++ b/packages/engine-server/src/utils.ts
@@ -409,6 +409,8 @@ export class EngineUtils {
         note,
         wsRoot: engine.wsRoot,
       });
+      // update links for note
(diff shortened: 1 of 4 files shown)


packages/engine-server/src/utils.ts
```diff
@@ -409,6 +409,8 @@ export class EngineUtils {
         note,
         wsRoot: engine.wsRoot,
       });
+      // update links for note
+      note.links = links.concat(links);
```
hikchoi, line 413: oh wow I feel dumb for missing this 😓 thank you for looking into it

hikchoi (approved):

PR #2176, opened 2022-01-14
enhance(workspace): conflicting extension warning

hikchoi:
# enhance(workspace): conflicting extension warning
This PR:
- Adds warning for conflicting extensions
  - Will only warn when in Dendron Workspace
  - Will show a warning message with a button that opens a webview
  - Webview will show instructions for how to disable certain extensions per workspace
  - Webview will list conflicting extensions along with a `commandUri` that will open the `Extensions` panel in vscode with the corresponding extension id pre-filled in the query.

# Screenshots

Warning message displayed on activation if in Dendron workspace
image

Webview with instructions
image

Extensions panel with query pre-filled open when clicking `View Extension`
image

# Manual testing

Since it isn't straightforward to write integ tests for this, I've manually tested with all extensions listed in `CONFLICTING_EXTENSIONS` and confirmed that they trigger the warning and correctly show up in the webview.

To try out yourself:
1. Pull this branch
2. Build
3. Install any extension in `CONFLICTING_EXTENSIONS`
3. `Run Extension:Remote (plugin-core)` with Dendron disabled.

# Pull Request Checklist

## Testing
### Docs
### Analytics

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -62,7 +62,7 @@ import { PreviewPanelFactory } from "./components/views/PreviewViewFactory";
 import { SchemaGraphViewFactory } from "./components/views/SchemaGraphViewFactory";
 import {
   CONFIG,
-  CONFLICTING_EXTENSIONS,
+  INCOMPATIBLE_EXTENSIONS,
   DendronContext,
   DENDRON_COMMANDS,
   GLOBAL_STATE,
@@ -100,6 +100,8 @@ import { DendronCodeWorkspace } from "./workspace/codeWorkspace";
 import { DendronNativeWorkspace } from "./workspace/nativeWorkspace";
 import { WorkspaceInitFactory } from "./workspace/WorkspaceInitFactory";
 import { WSUtils } from "./WSUtils";
+import { DoctorCommand, PluginDoctorActionsEnum } from "./commands/Doctor";
+import { IDendronExtension } from "./dendronExtensionInterface";
 
 const MARKDOWN_WORD_PATTERN = new RegExp("([\\w\\.\\#]+)");
 // === Main
@@ -665,7 +667,9 @@ export async function _activate(
         codeWorkspacePresent,
       });
 
-      warnConflictingExtensions();
+      if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
+        await warnIncompatibleExtensions({ ext: ws });
+      }
 
       if (stage !== "test") {
         await ws.activateWatchers();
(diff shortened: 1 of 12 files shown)


hikchoi:
Pending tasks:
- [ ] update analytics docs.
- [ ] add docs for conflict details. (possibly link it within the preview as well)
- [ ] confirm warning display cadence.

hikchoi:
Note on warning display cadence:

Currently this will show up as long as a user has these conflicting extensions enabled, but I imagine some will want to go through troubleshooting and resolving conflicts (if it is possible) and use it with Dendron concurrently.

The first guard against bombarding the user with excessive warnings is only displaying them when activated (in Dendron workspace). This is already done.

Adding a `Don't show this again` button will block us from properly warning the users later down the line if they install one of the extensions.

I think a sane candence for this warning would be on every upgrade, but I would love some feedback on this before I go change the behavior.

Another possibility is to introduce a config in the `workspace` namespace that users can set to disable the warning, and keep the current cadence of warning on every activation.

hikchoi:
madge report count on master: 67
madge report count on current branch: 67

hikchoi:
UPDATE: added link to `dendron.guides.troubleshooting.conflicting-extensions` in the preview

image

jonathanyeung (approved):
lgtm

jonathanyeung:
> warning cadence is still per startup and need to be discussed.

Good point on this one - maybe we should add additional buttons for 'dismiss' and 'remind me later'? 

Also, in case a user has accepted the risks/experience of having conflicting extensions installed, we shouldn't bother them anymore

hikchoi:
> Good point on this one - maybe we should add additional buttons for 'dismiss' and 'remind me later'?

I think we should at least wait until they accept the toast message, and then inside the webview give them an option to dismiss for the future. It's a bit unfortunate that the information message can't be more rich, but as per vscode's best practices, we shouldn't be bombarding users on activation with a lot of information upfront.

A nice thing I found out about webviews is that we can embed command uris. (see [this line](https://github.com/dendronhq/dendron/pull/2176/files#diff-8135b32904386c7b4b949c00cf19215884bf9a2a65a295b9bdf7072a5a64dd57R1297) that adds a link that will trigger a workbench.extensions.search when clicked) (we should use this more for simple webviews that we use that are not nextjs based to augment a button)

> Also, in case a user has accepted the risks/experience of having conflicting extensions installed, we shouldn't bother them anymore

We can add an internal command that sets a state saying this user has dismissed and we should never show this again, and then add the command uri of it within the webview as a link.

Same goes for remind me later. Or maybe remind me later should be a button on the toast message instead of inside the webview?

hikchoi:
Just a thought I had right now:

The vim keybinding resolution I introduced a while ago is pretty unreliable and has been causing a lot of issues for our users. Now that we are warning people about conflicting extensions, should we also just merge the vim keybinding behavior into this as well?

The plus side is that we aren't automatically running a potentially destructive operation that would be sorta fine since we always back up the original keybinding.json, but is still an additional point of potential failure that is pretty catastrophic.

The downside is that when people dismiss the conflicting extension warning, we are essentially blocking them from using lookup, and in turn will have some rippling effects (in first week conversion rates, stickiness, initial user funnel, all sorts of metrics on this).

jonathanyeung:
### On re-prompting

I would err on the side of being less-intrusive to begin with; we can make the notice more "intrusive" over time if we find it's useful for users (especially given our starting point right now, which is no notification at all).

To that end, I think it'd be better to have the default be 'show only once', regardless of whether the user clicked 'Fix Conflicts' or Dismissed. I think making the user aware of potential conflict is 90% of the battle.

We could maybe make it such that the user can re-open the conflicting-extensions webview via a different command (perhaps under doctor). If we go this route, we should notify users of this by pointing it out in the webview text, and maybe if they hit dismiss, we pop another toast saying they can revisit later via the doctor cmd.

### Re: VIM Keybinding Resolution. 

Whatever we decide there, let's make that a separate change. 

One thing we should understand is how many new installs have VIM keybindings ext installed as well.

On the topic though - I think it does make sense to merge these workflows and maybe not touch keybindings.json automatically.  I think as you mentioned, there are a lot of avenues to prevent it, from any of the solutions you mentioned, or by somehow warning users who are going through the tutorial and have VIM ext installed.

hikchoi:
Makes sense. I'll start a separate async on the discussion on vim keybinding so we can focus on discussing how (+ how often) we want to prompt a warning.

packages/plugin-core/src/_extension.ts
```diff
@@ -1243,3 +1248,67 @@ function updateEngineAPI(port: number | string): void {
...
+      .then(async (resp) => {
+        if (resp === "Fix conflicts...") {
+          await showConflictingExtensionPreview(installedExtensions);
+        }
+      });
+  }
+}
+
+async function showConflictingExtensionPreview(installedExtensions: string[]) {
+  const md = _md();
```
kevinslin, line 1277: instead of using `markdown-it`, the dendron preview now supports rendering arbitrary markdown. wrote a guide for it here -> [[Using the Dendron Preview on arbitrary markdown|dendron://dendron.docs/pkg.plugin-core.dev.cook#using-the-dendron-preview-on-arbitrary-markdown]]

one thing to check is if this has some activation requirements (eg. workspace needs to be activated) before this works. if there are no conflicts, we should prefer using this going forward

packages/plugin-core/src/_extension.ts
```diff
@@ -652,6 +654,9 @@ export async function _activate(
         workspaceType: ws.type,
         codeWorkspacePresent,
       });
+
+      warnConflictingExtensions();
```
kevinslin, line 658: future optimization (create a next task, put in backlog and add a comment here, only run this check on first install and if the user has installed new extensions

kevinslin (changes requested):
what about turning this into a doctor command (`eg. detectIncompatibleExtensions`)?

for now, we just prompt this once (when dendron first installs) to check for conflicts. 
if a user has conflicts, we can ask them to run the doctor command

for vim key bindings, we've been getting too many issues of bad keybindings.

lets merge vim keybindings into this in a separate commit. offer a doctor command to fix it (but have the user actually run it instead of running automatically)

packages/plugin-core/src/_extension.ts
hikchoi: `NoteUtils.createForFile` looks like it's used to open files that are not in a location that Dendron recognizes. (e.g. a markdown file that's outside a vault)

It takes a file path so I assume this is meant for an actual file that exists, just not in a vault.

Are you suggesting we should create a file with the content we want to show in the preview, and then show it?

kevinslin: ah, `NoteUtils.createForFile` creates a dummy note for a file. for this change, we can add `NoteUtils.createForText` to pass in a note

hikchoi: I can git `NoteUtils.createForFile` a random path, and it will indeed create a `NoteProps` for me, but I don't know if we want to do that.

kevinslin (approved):

hikchoi:
Updates:

- Change all mentions of `conflicting` extensions to `incompatible` extensions
- Moved warning into new doctor command `findIncompatibleExtensions`
- Changed warning to only display once on initial install, which will display a warning message with a button that will trigger the doctor command.
- Refactored doctor command related types
  - made reports 1 less circular dependencies compared to master
- Refactored DoctorCommand to have `IDendronExtension` injected through the constructor
- Exposed `fileWatcher` in the `IDendronExtension` interface.
- Changed preview to show all install statuses regardless of it's existence (The preview will list extensions not installed as `Not Installed`) to provide a better picture of the detected results.

image

Note:
- Had partial success in displaying the markdown with Dendron Preview, but since this is an ephemeral note and doesn't have a concrete file, when the preview is rendered, the window watcher grabs the active note and immediately re-renders the current active note. This will be a little bit more involved than what this PR is concerned, so will be followed up as an enhancement

packages/plugin-core/src/_extension.ts
```diff
@@ -665,7 +667,9 @@ export async function _activate(
         codeWorkspacePresent,
       });
 
-      warnConflictingExtensions();
+      if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
+        await warnIncompatibleExtensions({ ext: ws });
```
jonathanyeung, line 671: nit: do we actually need to await here?  The rest of initialization doesn't need to be blocked on this warning right?

hikchoi:
updated wording in docs: https://github.com/dendronhq/dendron-site/pull/361

jonathanyeung (changes requested):
one small nit. 

Also added a few comments to the docs pr: https://github.com/dendronhq/dendron-site/pull/361

packages/plugin-core/src/_extension.ts
hikchoi: right. it will be immediately resolved after the warning message is displayed, but worth to be correct.

hikchoi: fixed

jonathanyeung (approved):

PR #2190, opened 2022-01-17
chore: keep export pods v2 commands behind enableExportPodV2 dev config option

Harshita-mindfire:
This PR adds a dev config option `enableExportPodV2` for the enablement of `Export Pod V2` commands. 
Manually tested this with:
- WHEN enableExportPodV2 is undefined or false
     - [x]  THEN `Dendron: Export Pod V2` command should not be visible in command pallet.
     - [x]  THEN `Dendron: Configure Export Pod V2` command should not be visible in command pallet.
     - [x]  THEN `Dendron: Configure Service Connection` command should not be visible in command pallet.
     - [x]  THEN export pod v2 keybinding shortcuts should not run the export v2 commands
- WHEN enableExportPodV2 is true
     - [x]  THEN `Dendron: Export Pod V2` command should be visible in command pallet.
     - [x]  THEN `Dendron: Configure Export Pod V2` command should be visible in command pallet.
     - [x]  THEN `Dendron: Configure Service Connection` command should be visible in command pallet.
     - [x]  THEN export pod v2 keybinding shortcuts should execute the export v2 commands.

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -4,6 +4,7 @@ import {
   SubProcessExitType,
 } from "@dendronhq/api-server";
 import {
+  ConfigEvents,
   ConfigUtils,
   CONSTANTS,
   DendronError,
@@ -637,6 +638,10 @@ export async function _activate(
         DendronContext.ENABLE_EXPORT_PODV2,
         dendronConfig.dev?.enableExportPodV2 ?? false
       );
+
+      if (dendronConfig.dev?.enableExportPodV2) {
+        AnalyticsUtils.track(ConfigEvents.EnabledExportPodV2);
+      }
       // round to nearest 10th
       let numNotes = _.size(getEngine().notes);
       if (numNotes > 10) {
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -633,13 +633,10 @@ export async function _activate(
       );
 
       //used for enablement of export pod v2 command
-      if (!_.isUndefined(dendronConfig.dev?.enableExportPodV2)) {
-        VSCodeUtils.setContext(
-          DendronContext.ENABLEEXPORTPODV2,
-          dendronConfig.dev?.enableExportPodV2 ?? false
-        );
-      }
-
+      VSCodeUtils.setContext(
+        DendronContext.ENABLE_EXPORT_PODV2,
+        dendronConfig.dev?.enableExportPodV2 ?? false
+      );
       // round to nearest 10th
       let numNotes = _.size(getEngine().notes);
       if (numNotes > 10) {
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -632,6 +632,14 @@ export async function _activate(
         MarkdownUtils.hasLegacyPreview()
       );
 
+      //used for enablement of export pod v2 command
+      if (!_.isUndefined(dendronConfig.dev?.enableExportPodV2)) {
+        VSCodeUtils.setContext(
+          DendronContext.ENABLEEXPORTPODV2,
+          dendronConfig.dev?.enableExportPodV2 ?? false
+        );
+      }
+
       // round to nearest 10th
       let numNotes = _.size(getEngine().notes);
       if (numNotes > 10) {
--- a/packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
+++ b/packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
@@ -2,6 +2,7 @@ import { getAllExportPods, PodClassEntryV4 } from "@dendronhq/pods-core";
 import { PodCommandFactory } from "../../components/pods/PodCommandFactory";
 import { PodUIControls } from "../../components/pods/PodControls";
 import { DENDRON_COMMANDS } from "../../constants";
+import { ExtensionProvider } from "../../ExtensionProvider";
 import { BaseCommand, CodeCommandInstance } from "../base";
 
 type CommandOutput = void;
@@ -32,6 +33,12 @@ export class ExportPodV2Command extends BaseCommand<
    * undefined if the user didn't select anything.
    */
   async gatherInputs(podId: string): Promise<CommandInput | undefined> {
+    // added check to return if export pod v2 is not enabled in dev config and is run using pod keyboard shortcuts
+    const { config } = ExtensionProvider.getDWorkspace();
+    if (!config.dev?.enableExportPodV2) {
+      return;
+    }
+
     // If a podId is passed in, use this instead of prompting the user
     if (podId) {
       return PodCommandFactory.createPodCommandForStoredConfig({ podId });
--- a/packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
+++ b/packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
@@ -70,4 +70,11 @@ export class ExportPodV2Command extends BaseCommand<
   async execute(opts: CommandOpts) {
     opts.run();
   }
+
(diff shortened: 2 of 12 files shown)


jonathanyeung (approved):
lgtm

test-workspace/dendron.yml
```diff
@@ -38,6 +38,7 @@ site:
 dev:
     enableWebUI: true
     enablePreviewV2: true
+    enableExportPodV2: false
```
jonathanyeung, line 41: nit: better to leave as true by default for test-workspace?

packages/plugin-core/src/_extension.ts
```diff
@@ -632,6 +632,14 @@ export async function _activate(
         MarkdownUtils.hasLegacyPreview()
       );
 
+      //used for enablement of export pod v2 command
+      if (!_.isUndefined(dendronConfig.dev?.enableExportPodV2)) {
+        VSCodeUtils.setContext(
+          DendronContext.ENABLEEXPORTPODV2,
+          dendronConfig.dev?.enableExportPodV2 ?? false
```
hikchoi, line 639: isn't this equivalent to just `dendronConfig.dev?.enableExportPodV2`?
or are you guarding against the possibility that the value there would not be a boolean?

packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
```diff
@@ -32,6 +33,12 @@ export class ExportPodV2Command extends BaseCommand<
    * undefined if the user didn't select anything.
    */
   async gatherInputs(podId: string): Promise<CommandInput | undefined> {
+    // added check to return if export pod v2 is not enabled in dev config and is run using pod keyboard shortcuts
```
hikchoi, line 36: fyi: keyboard shortcuts can also take when clauses.

packages/plugin-core/src/constants.ts
```diff
@@ -16,6 +16,7 @@ export enum DendronContext {
   HAS_CUSTOM_MARKDOWN_VIEW = "hasCustomMarkdownPreview",
   NOTE_LOOK_UP_ACTIVE = "dendron:noteLookupActive",
   BACKLINKS_SORT_ORDER = "dendron:backlinksSortOrder",
+  ENABLEEXPORTPODV2 = "dendron:enableExportPodV2",
```
hikchoi, line 19: super nitpick but this should be UPPER_SNAKE_CASE

hikchoi (changes requested):
some nitpicks and fyi on enablement

packages/plugin-core/src/constants.ts
```diff
@@ -541,12 +542,12 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
   CONFIGURE_SERVICE_CONNECTION: {
     key: "dendron.configureServiceConnection",
     title: `${CMD_PREFIX} Configure Service Connection`,
-    when: DendronContext.PLUGIN_ACTIVE,
+    when: DendronContext.PLUGIN_ACTIVE && DendronContext.ENABLEEXPORTPODV2,
```
hikchoi, line 545: I know this is super confusing so let me go over this again (I had to look this up AGAIN as well 😅)

from api docs (still confusing)
https://code.visualstudio.com/api/extension-guides/command#enablement-of-commands

from source code (much more clear)
https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/menusExtensionPoint.ts#L546:L549

So if we want to really block people from using this by any means unless they explicitly set the flag to true, we should add the "enablement" property to the "commands" contribution point rather than adding the "when" property to individual user-facing UI (command palette, menu, keybinding, etc)

adding
```
...
  "commands": [
    ...
    {
      "command": "dendron.configureServiceConnection",
      "title": ...,
      "enablement": "dendron.pluginActive && "dendron.enableExportPodV2"
    }
```

will disable this from all user facing UIs (command palette, menus, keybindings). It won't stop people from opening up the debug console within VSCode and manually triggering `vscode.commands.executeCommand("dendron.configureServiceConnection", ...)`, but at that point we should let them know it could be enabled by a flag.

packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
hikchoi: see https://github.com/dendronhq/dendron/pull/2190#discussion_r786635567 instead

packages/plugin-core/src/constants.ts
kevinslin: when we ship this, @Harshita-mindfire , can you also add documentation to [[Commands|dendron://dendron.docs/pkg.plugin-core.dev.cook#commands]] about how to add a command in preview following the process here? link to this PR as a reference

kevinslin: when you add that doc, can you add me and @hikchoi as reviewewrs?

packages/plugin-core/src/constants.ts
kevinslin: not nitpick. we should be consistent about this

kevinslin (changes requested):
we should have a metric to see how many people enable this flag. a dashboard that would be useful for next week:

- num users that enable export pod v2
- num users that use export pod v2 (with pod ids)

packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
Harshita-mindfire: I did try the "enablement" property to the commands contribution point, it does block the command to appear up in the command palette but not the keybinding shortcuts to pod v2.

hikchoi: That's weird. I guess it's not blocking any keybindings if the user adds it themselves (only blocking the ones we explicitly put in the contribution points)? We could probably check for the flag before registering the command but I don't think we have to go that far.

Let's just leave it as is and add a comment here to remind ourselves that this check should be removed when we announce it publicly 😄

packages/plugin-core/src/_extension.ts
Harshita-mindfire: 5f2a11b

packages/plugin-core/src/constants.ts
Harshita-mindfire: https://github.com/dendronhq/dendron-docs/pull/11

jonathanyeung:
Telemetry commit lgtm.  @hikchoi - could you re-review the latest command package.json changes and approve if it looks good? Thanks!

packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
```diff
@@ -70,4 +70,11 @@ export class ExportPodV2Command extends BaseCommand<
   async execute(opts: CommandOpts) {
     opts.run();
   }
+
+  addAnalyticsPayload(opts: CommandOpts) {
+    return {
+      configured: true,
+      pod: opts.key,
+    };
+  }
```
hikchoi, line 79: using the export pod will result in two tracked events (`dendron.exportpodV2` and `dendron.markdownexportv2`)

Should these two payloads be combined? the property `pod` is basically implied by the second tracked event

hikchoi (commented):
package.json looks good. Manually tested as well.

added a comment on the analytics payload. It might be worth combining the two to make querying them a bit simpler.

packages/plugin-core/src/commands/pods/ExportPodV2Command.ts
Harshita-mindfire: Yeah, you are right, each export pod results in tracking both the export pod and the selected pod-specific command. This just adds an option to filter metrics based on export pod command and not go through individual ones. We can totally skip this if it does not make sense.

kevinslin (approved):

Issue #2144, opened 2022-01-12
Search result link jumps to top of page


HopperTech:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

I have several longer notes. I have noticed that when that note is not open and I click on the search results, or use the open definitions feature, the cursor will jump to the top of the page.  If the note is already open then the clicking on the results link will take me to the found line in the note. 

## To Reproduce

1. Create a longer (>30 lines) note with some search term towards the bottom of the note
2. Search for the term 
3. Click on the resulting link. - This will work as expected, staying on the found line
4. Reload the window 
5. Search for the term again
6. Click on the resulting link - The cursor will be at the top of the note just under the frontmatter

Note: Reloading does not always immediately cause the issue, but once the issue begins on a note it is consistent from that point on. I was also able to reproduce this issue by copying and pasting the same text into a new note. At first the search link works correctly, but eventually the issue occurs. I have not been able to identify a trigger for when the issue begins.

## Expected behavior

When clicking on a search result, or using the open definition function, the new window should stay on the found line.

## Desktop (please complete the following information)

- OS: [e.g. iOS] Windows
- Browser: [e.g. chrome, safari] VSCode 1.63.2
- Dendron Version: [e.g. 22]   0.77.1

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here
[dendron.log](https://github.com/dendronhq/dendron/files/7855972/dendron.log)

SeriousBug:
Thanks for reporting this issue, I can reproduce it.

HopperTech:
Tested and works as expected.  Thank you

PR #2193, opened 2022-01-18
fix: cursor is moved when opening file through search interface

SeriousBug:
Fixes an issue where the cursor would get moved to the start of the document if the user opened the document through something like the search window, which is supposed to move the cursor to the matching position.

#2144

### DIFF
--- a/packages/common-all/src/utils.ts
+++ b/packages/common-all/src/utils.ts
@@ -865,3 +865,19 @@ export function cleanName(name: string): string {
 export function normalizeUnixPath(fsPath: string): string {
   return path.posix.format(path.parse(fsPath));
 }
+
+/** Wrapper(s) for easier testing, to wrap functions where we don't want to mock the global function. */
+export class Wrap {
+  /** A useless wrapper around `setTimeout`. Useful for testing.
+   *
+   * If you are testing code that uses `setTimeout`, you can switch that code over to this wrapper instead,
+   * and then mock the wrapper. We can't entirely mock `setTimeout` because that seems to break VSCode.
+   */
+  static setTimeout(
+    callback: (...args: any[]) => void,
+    ms: number,
+    ...args: any[]
+  ) {
+    return setTimeout(callback, ms, ...args);
+  }
+}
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -423,15 +425,44 @@ export class WorkspaceWatcher {
...
+        // Since we delayed, a new document could have opened. Make sure we're still in the document we expect
+        if (
+          VSCodeUtils.getActiveTextEditor()?.document.uri.fsPath === startFsPath
+        ) {
+          const { line, character } = editor.selection.active;
+          // Move the cursor, but only if it hasn't already been moved by VSCode, another extension, or a very quick user
+          if (line === 0 && character === 0) {
+            editor.selection = new Selection(position, position);
+          } else {
+            Logger.info({
```
jonathanyeung, line 454: do we actually need these in the logs? maybe debug level or something lower than `info`?

jonathanyeung:
apart from comments above, lgtm

SeriousBug:
> can we add a test for this by manually overriding the position in a test and using https://sinonjs.org/releases/v1.17.7/fake-timers/ ?

That doesn't work unfortunately. If faketimer is set up, VSCode completely freezes. I implemented the tests by wrapping around `setTimeout` and simulating how the watcher works instead.

packages/common-all/src/utils.ts
```diff
@@ -865,3 +865,19 @@ export function cleanName(name: string): string {
 export function normalizeUnixPath(fsPath: string): string {
   return path.posix.format(path.parse(fsPath));
 }
+
+/** Wrapper(s) for easier testing, to wrap functions where we don't want to mock the global function. */
+export class Wrap {
```
kevinslin, line 870: can you add documentation for this in [[Cook|dendron://dendron.docs/dev.process.qa.test#cook]]?

feel free to merge afterward (and add to weekly)

kevinslin (approved):

Issue #2178, opened 2022-01-15
numbering with no text are squashed together in the preview


imalightbulb:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [X] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

numbering with no text are squashed together in the preview

## To Reproduce

1. create a number list with no text

## Screenshots

For example, the following markdown text:
```md
### Amazing things that happened today

1.
2.
3.
```

becomes this:
![image](https://user-images.githubusercontent.com/66677431/149612981-b48c11f4-d8b1-4276-84b2-441ac416bf3c.png)

## Desktop (please complete the following information)

- OS: Arch Linux 
- Browser: Brave
- Version: 0.77.1

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here

SeriousBug:
Thanks for the report, we'll try to fix this soon.

SeriousBug:
Hi, this has been fixed with Dendron 0.80. Please let us know if it continues to happen, otherwise this issue will close in a few days.

PR #2219, opened 2022-01-20
fix: numbered lists without content stack on top of each other

SeriousBug:
This PR fixes numbered lists stacking on top of each other in publishing/preview when they have no content.

CSS: 
-  dimensions 
    - [x] sm
    - [x] lg
    - [x] xxl
 - browsers
    - [x] firefox
    - [x] chrome

#2178

### DIFF
--- a/packages/dendron-11ty/raw-assets/sass/base.scss
+++ b/packages/dendron-11ty/raw-assets/sass/base.scss
@@ -92,6 +92,8 @@ pre {
 
 li {
   margin: 0.25em 0;
+  // Otherwise empty list items stack on top of each other
+  height: 1rem;
 }
 
 img {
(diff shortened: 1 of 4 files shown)


packages/dendron-11ty/raw-assets/sass/base.scss
```diff
@@ -92,6 +92,8 @@ pre {
 
 li {
   margin: 0.25em 0;
+  // Otherwise empty list items stack on top of each other
+  height: 1rem;
```
jonathanyeung, line 96: (not a css expert)

Does this affect multiline contents? Should this be `min-height` instead?
```md
- something super ... long
- something with additional contents below
    > quote
- foo

SeriousBug: Good catch! That was a problem and using `min-height` fixes it.

kevinslin (approved):

jonathanyeung (approved):

Issue #1856, opened 2021-12-08
Preview does not display images


SR--:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

No images are visible in Dendron Preview.

## To Reproduce

1. Use `Show Preview` command. No images are visible
![image](https://user-images.githubusercontent.com/1008059/145155270-dfc8745e-ff4b-46fe-a60f-d0139878caa9.png)

![image](https://user-images.githubusercontent.com/1008059/145155293-41cd71a0-aa0c-493e-bc7c-6d57ca7e82ee.png)

The image links are inserted automatically and are correct. The behaviour was OK a few versions back. The images are visible in other previews and are openable on link click. Another machine displays the images in the same vaults OK.

It might be because this computer has two internal drives. Both machines are Windows.

## Expected behavior

Preview displays images.

@kevinslin has the logs

SeriousBug:
I've been trying to reproduce this issue, but haven't had any luck. I'm using Windows 10, 2 drives, the workspace is in the second drive. I followed the logs to reproduce the exact folder and file names too, but no luck.

Looking at the logs, it looks like:
1. The previews asks the Dendron engine for the image
2. Dendron engine finds the image and responds with it successfully
3. The preview doesn't receive the image, or can't display it for some reason

The only possibility I can think of is that there might be something blocking the connection between the preview and the engine, but only for images. Is there a firewall or antivirus software on the computer having issues? Any details you could share that could help me reproduce the problem?

Another thing to do would be to launch developer tools inside VSCode:
- Open the preview, then use the "Developer: Open Webview Developer Tools" command
- Switch to the "Console" tab in the developer tools
- Type anything inside the note
- Copy and send us (me or Kevin) the output from the console

SeriousBug:
Another possible point to diagnose:
- With developer tools open, click the "cursor inside a square" symbol on the top left of the developer tools
- Click on the broken image, this should highlight something that says "<img src=..." in the developer tools
[...]

PR #2237, opened 2022-01-20
chore: add option for url encoded images

SeriousBug:
Adds a `dev` config option which enables URL encoding of images for the preview. This builds the images into the generated HTML for the note, which avoids having to proxy images through the engine.

This may help with #1856 , where it looks like the engine can send the rendered note to the preview but the preview doesn't seem to be able to read these images.

### DIFF
--- a/packages/common-all/src/constants/configs/dev.ts
+++ b/packages/common-all/src/constants/configs/dev.ts
@@ -26,4 +26,8 @@ export const DEV: DendronConfigEntryCollection<DendronDevConfig> = {
     label: "Enable Preview V2",
     desc: "Use preview V2 as the default preview.",
   },
+  enablePreviewDirectImage: {
+    label: "Enable Preview Direct Image",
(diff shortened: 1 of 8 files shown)


kevinslin (changes requested):
Is there a less drastic solution we can try instead of encoding the image?

As you pointed out, this is expensive and doesn't solve this problem in general (and image handling is common enough that we want to solve in general)

 MPE extension adds a webview encoded URL for images -> [useExternalAddFileProtocolFunction](https://github.com/shd101wyy/vscode-markdown-preview-enhanced/blob/50ebcb2d139fbae70896f90e9fd88f9d33e77142/src/preview-content-provider.ts#L55). Can we do something like this instead?

SeriousBug:
I wasn't aware of that, should have thought to look at MPE! Thanks for the suggestion. Since the PR will be completely different I'll go ahead and close this PR, and submit one using the suggested change.

PR #2239, opened 2022-01-20
enhance/schemas support cross vault links

tma66:
This change will support allowing schemas to reference template notes in different vaults through 2 ways.

1. Cross vault notation (e.g dendron://vault1/foo). When this notation is used, we use the template specified in that vault and throw an error otherwise (even if it is found in a different vault)
2. Lookup + prompt. We will search the engine for all notes that match the template. If one is found, use that. If multiple is found, prompt user to select vault. If none, throw an error

Other changes:
* Modified SchemaUtils to take in template note instead of SchemaTemplate. This means that `applyTemplate` works only on valid template notes

Testing:
* Using test workspace, create schema to reference `egg` which lies in vault, vault2, and vaultThree. Verified prompt
* Other edge cases also manually created and through integ test cases

See [[Support Templates in Multivault|dendron://private/task.schema.2022.01.05.support-templates-in-multivault]] for more info


packages/common-all/src/vault.ts
```diff
@@ -4,7 +4,11 @@ import { normalizeUnixPath } from ".";
 import { CONSTANTS } from "./constants";
 import { DendronError } from "./error";
 import { DVault, WorkspaceFolderRaw } from "./types";
+import { QuickPickItem } from "vscode";
```
kevinslin, line 7: you can't import `QuickPickItem` in `common-all` because it doesn't have access to vscode. it only works in development because of the monorepo

packages/common-all/src/dnode.ts
```diff
@@ -6,6 +6,7 @@ import YAML, { JSON_SCHEMA } from "js-yaml";
 import _ from "lodash";
 import minimatch from "minimatch";
 import path from "path";
+import * as vscode from "vscode";
```
kevinslin, line 9: you can't import vscode in common-all

PR #2266, opened 2022-01-24
chore(pods): improve error feedback to user

Harshita-mindfire:
This PR improves the error feedback to a user.

### DIFF
--- a/packages/pods-core/src/builtin/AirtablePod.ts
+++ b/packages/pods-core/src/builtin/AirtablePod.ts
@@ -1,6 +1,5 @@
 import Airtable, { FieldSet, Records } from "@dendronhq/airtable";
 import {
-  assertUnreachable,
   DendronCompositeError,
   DendronError,
   DEngineClient,
@@ -303,7 +302,12 @@ export class AirtableUtils {
         return { data: recordIds.map((ent) => ent!.id) };
       }
       default:
-        assertUnreachable(fieldMapping);
+        return {
+          error: new DendronError({
+            message: `The type ${type} provided in srcFieldMapping is invalid. Please update the pod config by running Dendron: Configure Export Pod V2 command.`,
+            severity: ERROR_SEVERITY.MINOR,
+          }),
+        };
     }
   }
 
--- a/packages/pods-core/src/builtin/AirtablePod.ts
+++ b/packages/pods-core/src/builtin/AirtablePod.ts
@@ -304,7 +304,7 @@ export class AirtableUtils {
       default:
         return {
           error: new DendronError({
-            message: `The type ${type} provided in srcFieldMapping is invalid. Please update the pod config by running Dendron: Configure Export Pod V2 command.`,
+            message: `The type ${type} provided in srcFieldMapping is invalid. Please check the valid types here: https://wiki.dendron.so/notes/oTW7BFzKIlOd6iQnnNulg.html#sourcefieldmapping and update the pod config by running Dendron: Configure Export Pod V2 command.`,
             severity: ERROR_SEVERITY.MINOR,
           }),
         };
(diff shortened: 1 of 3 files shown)


packages/pods-core/src/v2/pods/export/AirtableExportPodV2.ts
```diff
@@ -59,11 +65,28 @@ class AirtableUtilsV2 {
         try {
           const _records = await func(record);
           return _records;
-        } catch (error) {
+        } catch (error: any) {
+          let message;
+          if (error?.statusCode === 422) {
```
jonathanyeung, line 70: Instead of magic number 422, can we use Status Codes lib? 

`import { StatusCodes } from "http-status-codes";`

packages/pods-core/src/v2/pods/export/AirtableExportPodV2.ts
```diff
@@ -59,11 +65,28 @@ class AirtableUtilsV2 {
...
+            const airtableError = error as AirtableError;
+            if (airtableError.error === "INVALID_MULTIPLE_CHOICE_OPTIONS") {
+              // example airtable error message: 'Insufficient permissions to create new select option ""scope.xyz""'
+              const value = airtableError.message.split('""')[1];
```
jonathanyeung, line 74: we need to check that [1] index in array actually exists; otherwise let's just print the default error message if it's formatted in an unexpected way.

packages/pods-core/src/v2/pods/export/AirtableExportPodV2.ts
```diff
@@ -59,11 +65,28 @@ class AirtableUtilsV2 {
...
+                : "";
+            } else if (airtableError.error === "INVALID_VALUE_FOR_COLUMN") {
+              message = airtableError.message;
+            }
+          }
           const _error = new DendronError({
             innerError: error as Error,
             payload: record,
-            message: "error with airtable",
+            message: `Error with Airtable. ${message}`,
```
jonathanyeung, line 89: let's change to `Error during Airtable Export. ${message}`

jonathanyeung (changes requested):

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -303,7 +302,12 @@ export class AirtableUtils {
         return { data: recordIds.map((ent) => ent!.id) };
       }
       default:
-        assertUnreachable(fieldMapping);
+        return {
+          error: new DendronError({
+            message: `The type ${type} provided in srcFieldMapping is invalid. Please update the pod config by running Dendron: Configure Export Pod V2 command.`,
```
kevinslin, line 307: can we link to the docs with a list of valid mappings?

kevinslin (changes requested):

packages/pods-core/src/v2/pods/export/AirtableExportPodV2.ts
Harshita-mindfire: We do have a check for the `field` while creating the message. I updated the message to be the default one in case the value is undefined.
We could just print the default error message but I remember there was a suggestion to include the `field` in the error. For eg. in our case, if the field `status` has value `in-progress`, this error message will let us know that problem lies in the `status` field.

jonathanyeung: > We could just print the default error message but I remember there was a suggestion to include the field in the error.

Right - for the happy case - yes, let's include field. 

But what if airtable decides to reformat their message to say `'Insufficient permissions to create new select option scope.xyz'` without the extra quotation marks?  Then we're going to hit an array out of bounds exception.

jonathanyeung: As discussed offline - should be good.

PR #2267, opened 2022-01-24
chore: clean up incorrect `assertUnreachable` usages

SeriousBug:
At several points in the codebase, `assertUnreachable` was used incorrectly with no parameter. When used this way, the function does no type checking and the code path may be actually reachable.

This PR fixes this by:
- Requiring a parameter for `assertUnreachable`
- Adding an additional example for the function, and a warning to not use it without parameters
- Where possible, updating the usage to pass a parameter
- Where not possible, adding an actual exception that will give information about the failure, rather than the "this code path should be unreachable!" error message you'd get when `assertUnreachable` is used incorrectly.

### DIFF
--- a/packages/common-all/src/error.ts
+++ b/packages/common-all/src/error.ts
@@ -189,20 +189,34 @@ export class ErrorMessages {
  *
  * An example of how this function may be used is below:
  *
- *     type Names = "bar" | "baz";
+ * ```ts
+ * type Names = "bar" | "baz";
  *
- *     function foo(name: Names) {
- *       if (name === "bar") { ... }
- *       else if (name === "baz") { ... }
- *       else assertUnreachable(name);
- *     }
+ * function foo(name: Names) {
+ *   if (name === "bar") { ... }
+ *   else if (name === "baz") { ... }
+ *   else assertUnreachable(name);
+ * }
+ * ```
  *
  * Let's say someone changes the type Names to `type Names = "bar" | "baz" | "ham";`. Thanks to this
  * assertion, the compiler will warn them that this branch is now reachable, and something is wrong.
  *
- * @param x
+ * Here's another example:
+ *
+ * ```
+ * switch (msg.type) {
+ *   case GraphViewMessageType.onSelect:
+ *   // ...
+ *   // ... all the cases
+ *   default:
+ *     assertUnreachable(msg.type);
+ * }
+ * ```
+ *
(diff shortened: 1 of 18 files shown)


packages/common-all/src/error.ts
```diff
@@ -189,20 +189,34 @@ export class ErrorMessages {
...
+ * switch (msg.type) {
+ *   case GraphViewMessageType.onSelect:
+ *   // ...
+ *   // ... all the cases
+ *   default:
+ *     assertUnreachable(msg.type);
+ * }
+ * ```
+ *
+ * Warning! Never use this function without a parameter. It won't actually do any type checks then.
```
jonathanyeung, line 217: I see - changing `_never?` -> `_never` should probably clear up most of the confusion, but it may also be worth clarifying in the comments that this function isn't meant as a parameter-type-not-recognized kind of Runtime Exception, which I presume was the intention of the existing parameter-less usages.

To further make it clear, maybe we should remove the return value of `never` from this method signature, as well as the `throw new Error()`, as that will never happen during run-time.

EDIT: I suppose it could happen if we change our TS compiler strictness? In which case, maybe worth leaving in the throw _just in case_ - but then maybe let's add something like the following to the error? `ErrorMessages.formatShouldNeverOccurMsg("TS never variable contains valid value. Check the TS compiler strictness settings");`

jonathanyeung (changes requested):

packages/common-all/src/error.ts
SeriousBug: The runtime error can also happen if someone made an incorrect cast. I think the alternative (the code just keeps running along despite the wrong cast) is probably worse, the exception will at least get caught in testing.

The return type of `never` is actually needed, otherwise you hit errors like this:

```ts
static anchor2string(anchor: DNoteAnchor | DNoteAnchorBasic): string {
  if (anchor.type === "block") return `^${anchor.value}`;
  if (anchor.type === "header") return anchor.value;
  if (anchor.type === "line") return `L${anchor.line}`;
  assertUnreachable(anchor); // not all code paths return a value
}
```

The `never` type lets the compiler know that that path won't actually occur, so not returning anything is safe.

jonathanyeung: Ah - good points.  In that case, let's leave signature as is.

jonathanyeung (approved):

kevinslin (approved):

PR #2268, opened 2022-01-25
chore: deprecate MDUtilsV4

SeriousBug:
# Pull Request Checklist

This PR marks `MDUtilsV4` as deprecated, moves several unrelated methods out of `MDUtilsV4`, and refactors a few low-hanging fruit to switch to MDUtilsV5.

Switching more of the code will take substantial work because changes quickly break tests, and some code relies on `MDUtilsV4` features that were not brought over to `MDUtilsV5`, like being able to pass overrides for certain options. I think it makes sense to take the time to think how those should be implemented going forward, which I'll continue working on.

This PR does not affect the number of cyclic dependencies.

Docs PR: https://github.com/dendronhq/dendron-docs/pull/13

## Testing
### Docs
### Analytics

### DIFF
--- a/packages/engine-server/src/markdown/utils.ts
+++ b/packages/engine-server/src/markdown/utils.ts
@@ -514,7 +514,7 @@ export class MDUtilsV4 {
 }
 
 /** Contains functions that help dealing with mdast trees. */
-export class NodeUtils {
+export class MdastUtils {
   /** Find the index of the list element for which the predicate `fn` returns true.
    *
    * @returns The index where the element was found, -1 otherwise.
--- a/packages/engine-server/src/markdown/utils.ts
+++ b/packages/engine-server/src/markdown/utils.ts
@@ -135,99 +135,8 @@ type VisitorParentsIndices = ({
   ancestors: ParentWithIndex[];
 }) => boolean | undefined | "skip";
 
+/** @deprecated Please use {@link MDUtilsV5} instead. */
 export class MDUtilsV4 {
-  /** Find the index of the list element for which the predicate `fn` returns true.
-   *
-   * @returns The index where the element was found, -1 otherwise.
-   */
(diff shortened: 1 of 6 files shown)


packages/engine-server/src/markdown/utils.ts
```diff
@@ -631,6 +513,128 @@ export class MDUtilsV4 {
   }
 }
 
+/** Contains functions that help dealing with mdast trees. */
+export class NodeUtils {
```
kevinslin, line 517: node is a little confusing since it could mean many things. can we rename to `MdastUtils`?

kevinslin (changes requested):
can we add a note about the deprecation and next steps in [[Changelog|dendron://dendron.docs/dev.changelog]]?

jonathanyeung (changes requested):

packages/engine-server/src/markdown/utils.ts
```diff
@@ -631,6 +513,128 @@ export class MDUtilsV4 {
   }
 }
 
+/** Contains functions that help dealing with mdast trees. */
```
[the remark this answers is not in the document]
SeriousBug: Done!

Agreed, maybe after `'MDUtilsV4` is fully deprecated we can rename `MDUtilsV5` to something like `MDProcessor`, since it only houses the processor now.

kevinslin (approved):
looks good but build is failing (think its what @hikchoi mentioned). rebase and should be good

packages/engine-server/src/markdown/utils.ts
hikchoi: `mdast` is now called `remark`, so maybe something else. We already have a class called `RemarkUtils` though...

jonathanyeung (approved):

PR #2269, opened 2022-01-25
fix(views): engine events; update DendronTreeView reliably

jonathanyeung:
## fix(views): engine events; update DendronTreeView reliably

This change introduces a new model for updating plugin-core UI components based on an eventing system using `vscode.Event`-like `event` and `EventEmitter`.  The change also refactors the native DendronTreeView to utilize the new events.  This change provides a UI model that helps us improve UI reliability, performance, and also allows us to further unwind circular dependencies.

### Engine Events

- EngineAPIService implements a new interface called `EngineEvents`, which exposes a new event that fires when notes have been created, updated, and deleted.
- Some fixes were made in `engineClient` and `storev2` to have the correct data payloads for `NoteChangeEvent`s. 

### Tree View

- Renamed `DendronTreeView` -> `NativeTreeView`
- Separated out TreeView controller, Engine Data Provider, and TreeNote into separate files.
- Changed the generic type of EngineNoteProvider from `string` to `NoteProps`
- The Tree View no longer exposes `refresh()` public methods, but instead has an instance of `EngineEvents` injected into its constructor. All UI updates are handle internally.

---

# Pull Request Checklist

## Testing
### Docs
### Analytics


packages/plugin-core/src/services/EngineAPIService.ts
```diff
@@ -270,4 +339,20 @@ export class EngineAPIService implements DEngineClient, IEngineAPIService {
...
+        this._onNoteCreatedEmitter.fire(entry.note);
+        break;
+      case "update":
+        this._onNoteChangeEmitter.fire(entry.note);
+        break;
+      case "delete":
+        this._onNoteDeletedEmitter.fire(entry.note);
+        break;
+      default:
+        assertUnreachable();
```
jonathanyeung, line 355: TODO: Amend to `assertUnreachable(entry)`

packages/plugin-core/src/services/NoteSyncService.ts
```diff
@@ -1,14 +1,16 @@
 import {
```
jonathanyeung, line 1: Changes in this file were not meant to be included in this PR

kevinslin (changes requested):
Some thoughts:
- I like the event emitter interface and eventing
- I don't like that this is in `plugin-core` vs `engine-server` since this is something that other clients should also be able to re-use
- I don't like that this duplicates the function of `HistoryInstance` which is in `engine-server`

Some proposals:
- if we want to start off with a clean slate for eventing, we should at least move this out of `plugin-core`. Attach these functions to `DendronEngineClient` in `engine-server` instead
- event emitters should emit `NoteChangeEntry[]` as their payload, not just the `Note`
- can we have a rough migration plan on how we would merge `HistoryInstance` with the new eventing interface? (eg. interface required to fully replicate functionality )

Other comments:
- we currently have a way of "pausing" the watcher for bulk operations (eg. refactor) to avoid updating the tree view until everything is done. do we have that capability in this new interface?
- similarly, how do we stop listening to events?

jonathanyeung:
### Proposals
- Moved to engine-service; added a version of `Event` and `EventEmitter`
- consolidated create/update/delete to a single event with type <NoteChangeEntry[]>
- As Discussed offline, we can gradually move functions from HistoryService into the common `EngineEvents` interface (or a renamed interface)

### Other Comments
- re: pausing - I intentionally left this out.  I don't think any code outside of the engine/engine eventing owner should be allowed to pause events.  What if a client wanted to hear those events?  If we allowed pausing, then the contract between publisher and subscriber of getting notified for EngineChanges would be broken.  The reason we pause today is because we're afraid UI's can't update in time.  However, some clients may be able to handle the load, we can't make judgments from a controller on which subscribers can or cannot handle the event load.  For those subscribers that cannot, it's the responsibility of the subscribing client to be able to handle the event volume, which they can do via their own debouncer.  I added a 250 ms debouncer to `NativeTreeView`. 
- stop listening to events: this is part of `Event` / `vscode.Event`.  It returns a disposable.  Calling `disposable.dispose()` will unsubscribe.

SeriousBug (approved):
Looks good to me, just one nitpick for docs.

Will this interface be replacing the `HistoryInstance` then? Or will we be keeping both since `HistoryInstance` has some features like lookback that this doesn't?

packages/engine-server/src/engineClient.ts
```diff
@@ -141,6 +146,13 @@ export class DendronEngineClient implements DEngineClient {
     };
   }
 
+  get onNoteChanged(): Event<NoteChangeEntry[]> {
```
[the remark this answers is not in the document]
jonathanyeung: It's documented on the `EngineEvents` interface, but it'd be good to call out here as well for completeness. 

But maybe we should rename the property to `onNoteStateChanged()`? or maybe `onEngineNoteStateChanged()`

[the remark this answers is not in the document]
jonathanyeung: Updated to `onEngineNoteStateChanged()` and added docstring.

jonathanyeung:
> Will this interface be replacing the `HistoryInstance` then? Or will we be keeping both since `HistoryInstance` has some features like lookback that this doesn't?

They will eventually be converged - `History` has some features that this doesn't currently have, but in its current state it's a bit hard to abstract it to an interface.  So I'm keeping them separate for now to reduce circular dependencies, but yeah eventually it should be converged.

jonathanyeung:
@kevinslin - as I was working on some tests for the events, I noticed that some scenarios don't seem to have the correct payloads for what's changed in the engine, as reported by `storev2`.  I made some fixes (but there seem to still be other pre-existing issues), but then I have questions about how the engine API's are supposed to work:

- When exactly should `updateNote()` be used versus `writeNote()`? It seems that `updateNote` can still be used if the note specified doesn't exist yet. Similarly, `writeNote()` looks like it can be called on a note that already exists.
- `EngineUpdateNodesOptsV2.newNode`  - what is this parameter for?
- `EngineWriteOptsV2.updateExisting` - what constitutes updating an existing note vs. overwriting?  Is it if the ID changes? 
- `refreshNotes` / `refreshNotesV2` - does this actually need to be public? Seems like it should be an operation internal to only the engine code.  Same question with `sync()` 
- `renameNote` currently will return a change payload with a `create` and a `delete`.  Is this by-design (does file watcher need it to be this way?).  Because semantically, it seems this should instead report a single `updated` event.

We should add docstrings to these methods detailing their exact intended behavior, especially if they are widely used API's.

kevinslin:
> @kevinslin - as I was working on some tests for the events, I noticed that some scenarios don't seem to have the correct payloads for what's changed in the engine, as reported by `storev2`. I made some fixes (but there seem to still be other pre-existing issues), but then I have questions about how the engine API's are supposed to work:
> 
>     * When exactly should `updateNote()` be used versus `writeNote()`? It seems that `updateNote` can still be used if the note specified doesn't exist yet. Similarly, `writeNote()` looks like it can be called on a note that already exists.
> 
>     * `EngineUpdateNodesOptsV2.newNode`  - what is this parameter for?
> 
>     * `EngineWriteOptsV2.updateExisting` - what constitutes updating an existing note vs. overwriting?  Is it if the ID changes?
> 
>     * `refreshNotes` / `refreshNotesV2` - does this actually need to be public? Seems like it should be an operation internal to only the engine code.  Same question with `sync()`
> 
>     * `renameNote` currently will return a change payload with a `create` and a `delete`.  Is this by-design (does file watcher need it to be this way?).  Because semantically, it seems this should instead report a single `updated` event.
> 
> 
> We should add docstrings to these methods detailing their exact intended behavior, especially if they are widely used API's.

> We should add docstrings to these methods detailing their exact intended behavior, especially if they are widely used API's.

Yep, agreed. This has been an internal API and can definitely be improved. 

> When exactly should updateNote() be used versus writeNote()? It seems that updateNote can still be used if the note specified doesn't exist yet. Similarly, writeNote() looks like it can be called on a note that already exists.

The difference is if they write notes to disk. `updateNote` is meant to update the internal `this.notes` dictionary. The reason new notes are valid is because the user added the note outside of dendron, its picked up by the filewatcher, and dendron needs to know that there is a new note

`writeNote` is called when we need to actually write the note to the `store`. 

> EngineUpdateNodesOptsV2.newNode - what is this parameter for?

This parameter is for the case when `updateNote` is adding a note that was written to disk outside of dendron. we need to add parents to the note and do some other actions for this case. see [[../packages/engine-server/src/drivers/file/storev2.ts#L966]]

>  EngineWriteOptsV2.updateExisting - what constitutes updating an existing note vs. overwriting? Is it if the ID changes?

`updateExisting` is used in refactoring commands. we are writing the note to disk but want to merge it with the note that the engine already has

>     refreshNotes / refreshNotesV2 - does this actually need to be public? Seems like it should be an operation internal to only the engine code. Same question with sync()

`refreshNotes` does not. `sync` is called when the engine is already initialized and another client connects to it. eg. [[../packages/engine-server/src/topics/connector.ts#L139]]

>     renameNote currently will return a change payload with a create and a delete. Is this by-design (does file watcher need it to be this way?). Because semantically, it seems this should instead report a single updated event.

Yep. we notify that the old note was deleted and new note is created. this is used by `refreshNotes` and some other apis to properly update the engine.

kevinslin:
looks like there's a build error

jonathanyeung:
> looks like there's a build error

should be patched now

packages/common-all/src/types/events.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * An event emitter can be used to create and manage an {@link Event} for others
+ * to subscribe to. One emitter always owns one event.
+ *
+ * This mimics the interface signature of vscode.EventEmitter but can be used
+ * outside of plugin-core.
+ */
+export class EventEmitter<T> {
+  _emitter: NodeEventEmitter = new NodeEventEmitter();
```
kevinslin, line 29: is this available in the browser? if not, should it be moved to `common-server`?

packages/engine-server/src/engineClient.ts
```diff
@@ -59,7 +62,9 @@ type DendronEngineClientOpts = {
   ws: string;
 };
 
-export class DendronEngineClient implements DEngineClient {
+export class DendronEngineClient implements DEngineClient, EngineEvents {
```
kevinslin, line 65: instead of having the client implement `EngineEvents`, can we add a method so that it returns a class that implements this interface?

eg. `getEngineEventEmitter(): EngineEvents`

packages/engine-server/src/EngineEvents.ts
```diff
@@ -0,0 +1,11 @@
+import { Disposable, Event, NoteChangeEntry } from "@dendronhq/common-all";
+
+/**
+ * Interface for events signaling changes that have been made to engine state
+ */
+export interface EngineEvents extends Disposable {
```
kevinslin, line 6: since this isn't events but an emitter for it, can we rename to `EngineEventsEmitter`?

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+/**
+ * Tests that the EngineEvents interface of the client-side DendronEngineClient signals properly
+ */
+describe("GIVEN a DendronEngineClient running on client-side", () => {
+  describe("WHEN writing a new note", () => {
+    test("THEN create event fired with the correct props AND update note event fired for its parent", async (done) => {
+      await runEngineTestV5(
+        async ({ engine, vaults, wsRoot }) => {
+          // TODO: Get rid of this casting once the proper way to expose EngineEvents of DendronEngineClient is implemented
+          const engineClient = engine as DendronEngineClient;
```
kevinslin, line 15: see earlier comments. if engine can return an event emitter, then we would not need to typecast

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+          const newNote = await NoteTestUtilsV4.createNote({
+            fname: "alpha",
+            body: `[[beta]]`,
+            vault: vaults[0],
+            wsRoot,
+          });
+
+          engineClient.onEngineNoteStateChanged(
+            (noteChangeEntries: NoteChangeEntry[]) => {
+              try {
```
kevinslin, line 26: why do we have try/catch here?

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+                expect(noteChangeEntries.length).toEqual(2);
+
+                noteChangeEntries.forEach((entry) => {
```
kevinslin, line 29: this is hard to read. can we add a helper function, especially since `NoteChangeUpdateEntry` will be used quite a bit?

```
extractEntries(status) => NoteChangeUpdateEntry[]
```

we can add it to `engine-test-utils/src/engine.ts`

kevinslin: apply feedback to subsequent tests

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+          expect,
+          preSetupHook: ENGINE_HOOKS.setupBasic,
+          createEngine: createEngineFromServer,
+        }
+      );
+    });
+  });
+
+  // TODO: This scenario doesn't seem to work - parent is updated, but children appear to get orphaned.
+  describe.skip("WHEN writing over an existing note by changing its ID", () => {
```
kevinslin, line 118: you need to pass in `updateExisting: true` as the second option

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+  /**
+   * If test fails due to timeout, then it means at least one of the expected
+   * onNoteCreated/onNoteDeleted callbacks was not received.
+   *
+   * TODO: Current storeV2 behavior is to delete and then create a note during a
+   * rename (because the file system name has been changed). Instead of
+   * reporting a create + delete, should we actually just report a single
+   * update, as the ID doesn't change on the note? Furthermore, an update is
+   * reported on the Root node, even though prev/new states are identical. This
+   * bug should be fixsed.
```
kevinslin, line 357: nitpick: typo

kevinslin: for these comments, can we also create a task note and add it in the comment?

packages/plugin-core/src/services/EngineAPIService.ts
```diff
@@ -36,13 +37,19 @@ import {
...
+  EngineEvents,
+  HistoryService,
+} from "@dendronhq/engine-server";
 import _ from "lodash";
 import { IEngineAPIService } from "./EngineAPIServiceInterface";
 
-export class EngineAPIService implements DEngineClient, IEngineAPIService {
-  private internalEngine: DEngineClient;
-
+export class EngineAPIService
```
kevinslin, line 48: see comment in engineClient.ts about returning an emitter

kevinslin (changes requested):

packages/engine-server/src/EngineEvents.ts
jonathanyeung: good call

packages/common-all/src/types/events.ts
jonathanyeung: Is there a definitive way to tell during compile time (or do we just get a runtime error if running in browser with a non-browser supported function)?  I was thinking that our `tsconfig.json` build setup should catch this if it were a problem, but I wasn't sure.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: This won't help though unless we also add `getEngineEventEmitter()` into `DEngineClient`, which I didn't want to do.  A DEngineClient does not necessarily need to also be the `EventsEngineEmitter`; the emitter data may come from elsewhere.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: Because we're in a callback, I found that if an `expect()` fails, the test will just hang (as it results in a failed promise instead of explicit thrown exception).  It's necessary to call `done(err)` for the test to properly fail when there's an expect error.

packages/engine-server/src/engineClient.ts
jonathanyeung: Replied on the other comment on `[EngineEvents.spec.ts]`.  Given that, I'm not sure this would help?

packages/plugin-core/src/services/EngineAPIService.ts
jonathanyeung: let's discuss on other comment thread in `[EngineEvents.spec.ts]`

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: I see - should `updateExisting: true` be the default behavior?  Which use cases have `updateExisting: false`?

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: I'll remove the part about create+delete vs update, as it looks like we need that for now for fileWatcher to properly work.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
kevinslin: got it. but for this specific test, are we expecting a failure? and if this is just a safety catch all, can we extract this in a common utility to reduce boilerplate code?

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: Created [[writeNote Orphans Child Nodes|dendron://private/task.2022.02.08.writenote-orphans-child-nodes]]

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: We're not expecting a failure in the sense that the test case is expected to fail today, but we should expect a failure to potentially occur in the future if we have a product code bug.

imo - there isn't much to extract here (it's basically just a try catch block), but I pulled it out into a method with a header to describe why it's necessary.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
jonathanyeung: But if we do that static method, then we'll just move the casting call to inside that `EngineEventsEmitter.create()` method, which will be worse - I don't think it's that bad to have a cast here, since we're in test code.  If the cast is invalid, the test will fail, which is what we want.  If we move the cast to product code, then if that cast fails we're in a much worse spot.

The proper fix here is to expose EngineEventEmitter to be one of the arg options inside the `runEngineTestV*`, i.e.

```typescript
await runEngineTestV5(
    async ({ engine, vaults, wsRoot, eventEmitter }) => {
    ...
```

But that's a bit more involved, which is why I left it as a `TODO` comment.

packages/common-all/src/types/events.ts
jonathanyeung: tried webpack locally, it seems to have worked - so I think we should be ok?  Let me know if you think there are more solid ways to test this out.

kevinslin: lgtm

PR #2273, opened 2022-01-25
enhance: Workspace Sync warns about bad states, can pull with uncommitted changes, and has better commit messages

SeriousBug:
# Pull Request Checklist

This PR brings in many improvements to the Workspace Sync:
- it can now stash and unstash local changes to pull even when there are uncommitted changes
- it can detect merge conflicts and ongoing rebases to avoid committing bad notes (may help #2280)
- it can detect bad or misconfigured remotes and warn about them
- it can detect that a push would fail, and avoid pushing to remote
- it can detect that a push is not needed, and skip pushing
- The message displayed to the user will switch between an information, warning, and error message type depending on what happened during the sync.
- Commit messages now include the version of Dendron, the names of vaults in that repository, and the hostname of the committing machine #2281

Docs: https://github.com/dendronhq/dendron-site/pull/378

The error messages displayed to the user could use more work, but I think it's minor enough that we can push that off until later.

## Testing
### Docs
### Analytics

Screenshot_20220128_181112
Screenshot_20220128_181440
Screenshot_20220128_182939

### DIFF
--- a/packages/dendron-cli/src/commands/workspaceCLICommand.ts
+++ b/packages/dendron-cli/src/commands/workspaceCLICommand.ts
@@ -84,7 +84,7 @@ export class WorkspaceCLICommand extends CLICommand<
         }
         case WorkspaceCommands.ADD_AND_COMMIT: {
           const ws = new WorkspaceService({ wsRoot });
-          await ws.commitAndAddAll();
+          await ws.commitAndAddAll({ engine: engine! });
           break;
         }
         case WorkspaceCommands.PUSH: {
@@ -101,7 +101,7 @@ export class WorkspaceCLICommand extends CLICommand<
         case WorkspaceCommands.SYNC: {
           const ws = new WorkspaceService({ wsRoot });
           this.print("commit and add...");
-          await ws.commitAndAddAll();
+          await ws.commitAndAddAll({ engine: engine! });
           this.print("pull...");
           await ws.pullVaults();
           this.print("push...");
--- a/packages/dendron-cli/src/commands/workspaceCLICommand.ts
+++ b/packages/dendron-cli/src/commands/workspaceCLICommand.ts
@@ -84,6 +84,10 @@ export class WorkspaceCLICommand extends CLICommand<
         }
         case WorkspaceCommands.ADD_AND_COMMIT: {
           const ws = new WorkspaceService({ wsRoot });
+          if (!engine) {
+            this.printError("Can't find the engine");
+            process.exit(1);
+          }
           await ws.commitAndAddAll({ engine: engine! });
           break;
         }
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -3,6 +3,7 @@ import {
   CONSTANTS,
   CURRENT_CONFIG_VERSION,
   DendronError,
+  DEngineClient,
   Disposable,
   DuplicateNoteAction,
   DUser,
@@ -36,6 +37,7 @@ import {
 import fs from "fs-extra";
 import _ from "lodash";
 import path from "path";
+import os from "os";
 import { DConfig } from "../config";
 import { MetadataService } from "../metadata";
 import {
@@ -53,30 +55,15 @@ import {
 } from "../utils";
 import { WorkspaceUtils } from "./utils";
 import { WorkspaceConfig } from "./vscode";
-import { IWorkspaceService } from "./workspaceServiceInterface";
+import {
+  IWorkspaceService,
+  SyncActionResult,
+  SyncActionStatus,
+} from "./workspaceServiceInterface";
 
 const DENDRON_WS_NAME = CONSTANTS.DENDRON_WS_NAME;
 
 export type PathExistBehavior = "delete" | "abort" | "continue";
-
-export enum SyncActionStatus {
-  DONE = "",
-  NO_CHANGES = "it has no changes",
-  NO_REMOTE = "it has no remote",
-  NO_UPSTREAM = "the current branch has no upstream",
-  SKIP_CONFIG = "it is configured so",
-  NOT_PERMITTED = "user is not permitted to push to one or more vaults",
-  NEW = "newly clond repository",
-  CANT_STASH = "failed to stash changes in working directory",
-  CANT_RESTORE = "failed to restore stashed changes",
-  ERROR = "error while syncing",
-}
-
-export type SyncActionResult = {
-  repo: string;
-  vaults: DVault[];
-  status: SyncActionStatus;
-};
 
 export type WorkspaceServiceCreateOpts = {
   wsRoot: string;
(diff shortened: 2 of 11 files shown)


packages/engine-server/src/workspace/service.ts
```diff
@@ -864,12 +866,33 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+              this.logger.info({ ctx, vaults, repo, stashed });
+              // this shouldn't fail, but for safety's sake
+              if (_.isEmpty(stashed) || !git.isValidStashCommit(stashed)) {
+                throw new DendronError({
+                  message: "unable to stash changes",
+                  payload: { stashed },
+                });
+              }
+              // stash create doesn't change the working directory, so we need to get rid of the tracked changes
+              await git.reset("hard");
```
jonathanyeung, line 884: just double checking - but a user won't lose files that are covered under `.gitignore` in this case right? 

To me looks good, but just trying to think of any edge cases where someone may lose a file.

jonathanyeung (commented):
lgtm - @kevinslin want to take a pass?

packages/engine-server/src/workspace/service.ts
SeriousBug: `git reset --hard` doesn't touch untracked files (new files that haven't been `git add`ed) or ignored files, only tracked files.

The only risk of losing files I think would be if Dendron crashes after reset and before restore. If that does happen though, I immediately log the stash commit before resetting, so we could direct the user to read the log and `git stash apply ...` themselves.

I guess it would be a good idea to document that gotcha somewhere just in case someone hits it. I'll add a docs PR and link it in the main body of this PR.

packages/engine-server/src/workspace/service.ts
```diff
@@ -864,12 +866,33 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
-            };
+
+          let stashed: string | undefined;
+          if (await git.hasChanges({ untrackedFiles: "no" })) {
+            // stash away changes, otherwise pull will fail
+            try {
+              stashed = await git.stashCreate();
+              this.logger.info({ ctx, vaults, repo, stashed });
+              // this shouldn't fail, but for safety's sake
+              if (_.isEmpty(stashed) || !git.isValidStashCommit(stashed)) {
```
kevinslin, line 877: something that is quite hard to do right now is debug, in a multi-vault, which vault failed. can we mention the vault name as part of the error?

packages/engine-server/src/workspace/service.ts
```diff
@@ -879,6 +902,11 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
               message: `error pulling vault${stderr}`,
               payload: { err, repoPath: repo },
             });
+          } finally {
+            // Try to restore changes if we stashed them, even if there were errors. We don't want to lose the users changes.
+            if (stashed) {
+              git.stashApplyCommit(stashed);
```
kevinslin, line 908: can this fail?

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
```diff
@@ -390,6 +396,71 @@ suite("workspace sync command", function () {
       });
     });
   });
+
+  describe("WHEN there are tracked, uncommitted changes", () => {
+    test("THEN Dendron stashes and restores the changes", (done) => {
+      runLegacyMultiWorkspaceTest({
```
kevinslin, line 402: why are we using legacy vs the new style tests?

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
```diff
@@ -390,6 +396,71 @@ suite("workspace sync command", function () {
       });
     });
   });
+
+  describe("WHEN there are tracked, uncommitted changes", () => {
```
kevinslin, line 400: one failure mode: user stashes changes, git pull causes rebase. what happens in this case? can we add a test?

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
```diff
@@ -390,6 +396,71 @@ suite("workspace sync command", function () {
...
+    test("THEN Dendron stashes and restores the changes", (done) => {
```
kevinslin, line 401: also add tests for when pull fails. we want to be more diligent with tests - they don't need to be exhaustive but we should test at least the happy case and 2-3 edge cases

kevinslin (changes requested):

packages/engine-server/src/workspace/service.ts
SeriousBug: Tested (and added automated tests) if there are conflicting changes. It doesn't fail, it just creates a merge conflict (`<<<< updated upstream...`). I'll add some code to detect that, and warn the user that a merge conflict happened.

I'm not sure if there are other cases where it could fail, let me know if you have something in mind and I'll add a test for that too.

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
SeriousBug: What do you mean by "pull fails"? Something where we can't pull at all (like a permissions issue), or the rebase hitting a conflict? I'll add tests for these, let me know if you had anything else in mind.

kevinslin: rebase hitting a conflict is the main case i'm thinking of

packages/engine-server/src/workspace/service.ts
kevinslin: for merge conflict, can we actually fail and warn the user? would be better to be explicit then to commit a conflict

packages/plugin-core/src/test/suite-integ/SyncCommand.test.ts
jonathanyeung: on this note, we should do a deprecation-tagging pass on test methods (separate to this review). I get tripped up on this as well.

packages/dendron-cli/src/commands/workspaceCLICommand.ts
```diff
@@ -84,7 +84,7 @@ export class WorkspaceCLICommand extends CLICommand<
         }
         case WorkspaceCommands.ADD_AND_COMMIT: {
           const ws = new WorkspaceService({ wsRoot });
-          await ws.commitAndAddAll();
+          await ws.commitAndAddAll({ engine: engine! });
```
kevinslin, line 87: instead of force casting engine, can we add a check if engine exists and show error if thats the case?

packages/engine-server/src/workspace/service.ts
```diff
@@ -849,36 +876,116 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+                message: `Failed to pull ${vaultNames}: ${stderr}`,
+                payload: {
+                  err,
+                  vaults,
+                  repo,
+                  stashed,
+                },
+              });
+            }
+          } finally {
```
kevinslin, line 984: if the stash apply is successful, do we drop the stash? otherwise, would we end up with lots of stashed commits?

packages/plugin-core/src/commands/Sync.ts
```diff
@@ -73,38 +86,123 @@ export class SyncCommand extends BasicCommand<CommandOpts, CommandReturns> {
...
-      ...SyncCommand.filteredRepoNames(pushed, SyncActionStatus.NO_UPSTREAM),
-      ...SyncCommand.filteredRepoNames(pulled, SyncActionStatus.NO_UPSTREAM),
-    ]).join(", ");
-    if (noUpstream.length > 0) {
-      message.push(
-        `Skipped pulling or pushing repos ${noUpstream} because they don't have upstream branches configured.`
-      );
-    }
+    let maxMessageSeverity: MessageSeverity = MessageSeverity.INFO;
+    const makeMessage = (
```
kevinslin, line 90: nitpick, probably worhwile to extract this as a separate function that encapsulates all `makeMessage` calls

kevinslin (changes requested):
nice job with this! a few small comments

packages/engine-server/src/workspace/service.ts
SeriousBug: `git stash create` doesn't actually create a stash entry like `git stash save` does. It only creates a dangling merge commit. If my understanding of `git gc` is correct, git will eventually automatically collect this commit since it's not referenced anywhere.

PR #2274, opened 2022-01-25
enhance/schemas support cross vault links

tma66:
**Background**
This change will support allowing schemas to reference template notes in different vaults through 2 ways.

1. Cross vault notation (e.g dendron://vault1/foo). When this notation is used, we use the template specified in that vault and throw an error otherwise (even if it is found in a different vault)
2. Lookup + prompt. We will search the engine for all notes that match the template. If one is found, use that. If multiple is found, prompt user to select vault. If none, throw an error

See [[Support Templates in Multivault|dendron://private/task.schema.2022.01.05.support-templates-in-multivault]] for more info

**Other changes:**
* Modified SchemaUtils to take in template note instead of SchemaTemplate. This means that applyTemplate works only on valid template notes

**Testing:**
* Using test workspace, create schema to reference egg which lies in vault, vault2, and vaultThree. Verified prompt
* Other edge cases also manually created and through integ test cases

Updated docs for [[template|dendron://dendron.dendron-site/dendron.topic.schema#template]]

**Madge:**
Before: 130
After: 130

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -38,7 +38,6 @@ import {
   SchemaProps,
   SchemaPropsDict,
   SchemaRaw,
-  SchemaTemplate,
 } from "./types";
 import {
   DefaultMap,
@@ -430,7 +429,20 @@ export class NoteUtils {
       NoteUtils.addSchema({ note, schemaModule, schema });
       const maybeTemplate = schema.data.template;
       if (maybeTemplate) {
-        SchemaUtils.applyTemplate({ template: maybeTemplate, note, engine });
+        // TODO: Support xvault with user prompting for this flow
+        const tempNote = NoteUtils.getNoteByFnameFromEngine({
+          fname: maybeTemplate.id,
+          vault: note.vault,
+          engine,
+        });
+
+        if (tempNote) {
+          SchemaUtils.applyTemplate({
+            templateNote: tempNote,
+            note,
+            engine,
+          });
+        }
       }
     }
     return note;
@@ -870,7 +882,10 @@ export class NoteUtils {
       vault: toVault,
     });
 
-    if (maybeNotes.length > 1) {
+    if (maybeNotes.length === 1) {
+      // Only one match so use that as note
+      existingNote = maybeNotes[0];
+    } else if (maybeNotes.length > 1) {
       // If there are multiple notes with this fname, default to one that's in the same vault first.
       const sameVaultNote = _.filter(maybeNotes, (n) =>
         VaultUtils.isEqual(n.vault, fromVault, wsRoot)
@@ -882,10 +897,7 @@ export class NoteUtils {
         // Otherwise, just pick one, doesn't matter which.
         existingNote = maybeNotes[0];
       }
-    } else {
-      // Just 1 note
-      existingNote = maybeNotes[0];
-    }
+    } // Else no match in which case we return undefined
     return existingNote;
   }
 
@@ -1349,27 +1361,28 @@ export class SchemaUtils {
     "image",
   ];
 
+  /**
+   * Apply template note to provided {@param note}.
+   *
+   * Changes include appending template note's body to end of provided note.
+   */
   static applyTemplate(opts: {
-    template: SchemaTemplate;
+    templateNote: NoteProps;
     note: NoteProps;
     engine: DEngineClient;
   }) {
-    const { template, note, engine } = opts;
-    if (template.type === "note") {
-      const tempNote = _.find(_.values(engine.notes), { fname: template.id });
-      if (_.isUndefined(tempNote)) {
-        throw Error(`no template found for ${template.id}`);
-      }
-      const tempNoteProps = _.pick(tempNote, this.TEMPLATE_COPY_PROPS);
+    const { templateNote, note } = opts;
+    if (templateNote.type === "note") {
+      const tempNoteProps = _.pick(templateNote, this.TEMPLATE_COPY_PROPS);
       _.forEach(tempNoteProps, (v, k) => {
         // @ts-ignore
         note[k] = v;
       });
       // If note body exists, append template's body instead of overriding
       if (note.body) {
-        note.body += `\n${tempNote.body}`;
+        note.body += `\n${templateNote.body}`;
       } else {
-        note.body = tempNote.body;
+        note.body = templateNote.body;
       }
 
       // Apply date variable substitution to the body based on lodash interpolate delimiter if applicable
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -869,16 +869,16 @@ export class NoteUtils {
 
   static getNoteFromMultiVault(opts: {
     fname: string;
-    notes: NotePropsDict | NoteProps[];
+    engine: DEngineClient;
     fromVault: DVault;
     wsRoot: string;
     toVault?: DVault;
   }) {
-    const { fname, notes, fromVault, toVault, wsRoot } = opts;
+    const { fname, engine, fromVault, toVault, wsRoot } = opts;
     let existingNote: NoteProps | undefined;
-    const maybeNotes = NoteUtils.getNotesByFname({
+    const maybeNotes = NoteUtils.getNotesByFnameFromEngine({
       fname,
-      notes,
+      engine,
       vault: toVault,
     });
 
--- a/packages/plugin-core/src/WSUtilsV2.ts
+++ b/packages/plugin-core/src/WSUtilsV2.ts
@@ -99,8 +99,12 @@ export class WSUtilsV2 implements IWSUtilsV2 {
   /**
    * See {@link IWSUtilsV2.findNoteFromMultiVaultAsync}.
    */
-  async findNoteFromMultiVaultAsync(opts: { fname: string; vault?: DVault }) {
-    const { fname, vault } = opts;
+  async findNoteFromMultiVaultAsync(opts: {
+    fname: string;
+    quickpickTitle: string;
+    vault?: DVault;
+  }) {
+    const { fname, quickpickTitle, vault } = opts;
     let existingNote: NoteProps | undefined;
     const engine = ExtensionProvider.getEngine();
     const maybeNotes = NoteUtils.getNotesByFnameFromEngine({
@@ -117,7 +121,7 @@ export class WSUtilsV2 implements IWSUtilsV2 {
       const vaults = maybeNotes.map((noteProps) => {
         return {
           vault: noteProps.vault,
-          label: VaultUtils.getName(noteProps.vault),
+          label: `${fname} from ${VaultUtils.getName(noteProps.vault)}`,
         };
       });
 
@@ -128,7 +132,7 @@ export class WSUtilsV2 implements IWSUtilsV2 {
           : vaultPickerItem.vault.fsPath,
       }));
       const resp = await vscode.window.showQuickPick(items, {
-        title: "Select Vault",
+        title: quickpickTitle,
       });
 
       if (!_.isUndefined(resp)) {
--- a/packages/plugin-core/src/WSUtilsV2.ts
+++ b/packages/plugin-core/src/WSUtilsV2.ts
@@ -2,9 +2,11 @@ import { IDendronExtension } from "./dendronExtensionInterface";
 import vscode from "vscode";
 import path from "path";
 import {
+  DendronError,
   DVault,
   NoteProps,
   NoteUtils,
+  RespV3,
   VaultUtils,
 } from "@dendronhq/common-all";
 import _ from "lodash";
@@ -103,7 +105,7 @@ export class WSUtilsV2 implements IWSUtilsV2 {
     fname: string;
     quickpickTitle: string;
     vault?: DVault;
-  }) {
+  }): Promise<RespV3<NoteProps | undefined>> {
     const { fname, quickpickTitle, vault } = opts;
     let existingNote: NoteProps | undefined;
     const engine = ExtensionProvider.getEngine();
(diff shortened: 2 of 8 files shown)


jonathanyeung (changes requested):
Can we also run a circular dependency check? https://docs.dendron.so/notes/773e0b5a-510f-4c21-acf4-2d1ab3ed741e/#ensure-no-new-circular-dependencies-are-being-added

packages/common-all/src/dnode.ts
```diff
@@ -430,7 +429,18 @@ export class NoteUtils {
       NoteUtils.addSchema({ note, schemaModule, schema });
       const maybeTemplate = schema.data.template;
       if (maybeTemplate) {
-        SchemaUtils.applyTemplate({ template: maybeTemplate, note, engine });
+        // TODO: Support xvault with user prompting for this flow
+        const tempNote = _.find(_.values(engine.notes), {
```
SeriousBug, line 433: This should use `NoteUtils.getNoteByFnameFromEngine`.

packages/common-all/src/dnode.ts
```diff
@@ -882,10 +895,7 @@ export class NoteUtils {
         // Otherwise, just pick one, doesn't matter which.
         existingNote = maybeNotes[0];
       }
-    } else {
-      // Just 1 note
-      existingNote = maybeNotes[0];
-    }
+    } // Else no match in which case we return undefined
```
SeriousBug, line 898: nitpick: This change is unnecceary, if `maybeNotes.length === 0` then `maybeNotes[0] === undefined`.

packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -506,6 +510,51 @@ export class NoteLookupCommand
...
+    });
+    if (_.isUndefined(maybeNote)) {
+      throw new DendronError({
+        message: `No template found for ${schemaTemplate.id}`,
+      });
+    }
+    return maybeNote;
+  }
+
+  private parseDendronURI(linkString: string) {
```
SeriousBug, line 543: We already have `parseRef`, could we use that here instead? https://github.com/dendronhq/dendron/blob/68185d0895df3e0514d2ec88c7b778c5f24de18d/packages/plugin-core/src/utils/md.ts#L389

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -749,6 +752,433 @@ suite("NoteLookupCommand", function () {
...
+                  id: "food",
+                  parent: "root",
+                  fname: "food",
+                  children: ["ch2"],
+                  vault,
+                }),
+                SchemaUtils.createFromSchemaRaw({
+                  id: "ch2",
+                  template: {
+                    id: `dendron://${vaults[1].fsPath}/template.ch2`,
```
SeriousBug, line 855: This should use `VaultUtils.getName` instead

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -749,6 +752,433 @@ suite("NoteLookupCommand", function () {
...
+    );
+
+    describeMultiWS(
+      "WHEN schema template references to a template note that lies in a different vault using xvault notation",
+      {
+        ctx,
+        preSetupHook: async ({ wsRoot, vaults }) => {
+          await ENGINE_HOOKS_MULTI.setupBasicMulti({ wsRoot, vaults });
+          // Template is in vault2
+          await NoteTestUtilsV4.createNote({
```
SeriousBug, line 829: Can we add another note with the same fname and a different body to a different vault, so we can test that it's actually finding the right template note?

packages/plugin-core/src/WSUtilsV2.ts
```diff
@@ -95,6 +96,48 @@ export class WSUtilsV2 implements IWSUtilsV2 {
     });
   }
 
+  /**
+   * See {@link IWSUtilsV2.getNoteFromMultiVault}
+   */
+  async getNoteFromMultiVault(opts: { fname: string; vault?: DVault }) {
```
SeriousBug, line 102: Why is this duplicated here when we have `NoteUtils.getNoteFromMultiVault` already? Can we refactor this to share the same code? Code duplication will make maintenance significanly harder in the future.

SeriousBug (changes requested):
A few low-level changes that I think needs to be addressed.

packages/plugin-core/src/WSUtilsV2.ts
tma66: This one is different from the `NoteUtils` one as this one will prompt the user to pick the vault for multiple matches. Long term I'd like to deprecate the `NoteUtils` version to use this non-static version but I'll like to address that in a later refactoring

packages/plugin-core/src/commands/NoteLookupCommand.ts
tma66: Oh nice, thanks for letting me know

packages/common-all/src/dnode.ts
tma66: Functionally you are correct, but I'll prefer to explicitly check for certain lengths for readability purposes

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
tma66: fixed

packages/common-all/src/dnode.ts
tma66: Sounds good

packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -506,6 +510,51 @@ export class NoteLookupCommand
...
+        vaults: engine.vaults,
+      });
+    }
+
+    const maybeNote = await WSUtilsV2.instance().getNoteFromMultiVault({
+      fname: link,
+      vault: maybeVault,
+    });
+    if (_.isUndefined(maybeNote)) {
+      throw new DendronError({
```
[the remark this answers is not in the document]
tma66: That makes sense to me. I'll let the note creation go through then. Button suggestion sounds good, but I'll probably look into it in a follow up pr.

FYI regarding madge, I have the before/after already in the description above

packages/plugin-core/src/WSUtilsV2.ts
SeriousBug: Please mark `NoteUtils.getNoteFromMultiVault` as deprecated then (`@deprecated, see {@link WSUtils.getNoteFromMultiVault}`.

tma66: Sure, I'll deprecate after I start looking into refactoring which will be in another pr. I'll leave a TODO note here to remind myself

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
tma66: This should be covered in the subsequent test case "WHEN schema template references to a template note that lies in multiple vaults". In this case, user is prompted to pick the right vault

SeriousBug: That's a different test. I mean if the template note is in the form of `dendron://vault1/note` and there are multiple notes with the same name in different vaults, then the code should automatically pick the one in `vault1`.

This test is already testing that, it's just that there are no other notes with the same name so the test would pass even if the template was set to `note' without the vault specified.

packages/plugin-core/src/WSUtilsV2.ts
SeriousBug: Please just mark it as deprecated now. I'm not asking you to investigate all the usages of it or anything like that, just mark it as deprecated with a comment so nobody else uses it in the future. That's the same thing as putting a TODO there, but this way VSCode will show a warning in the editor so people now not to use it.

```ts
/** @deprecated see {@link WSUtils.getNoteFromMultiVault} */
async getNoteFromMultiVault() {
```

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
tma66: Ah I see what you mean. Yea that makes sense, added note

packages/plugin-core/src/WSUtilsV2.ts
tma66: So on second look, these two methods are actually quite different. The `NoteUtils` one gets a list of notes by fname and picks one that matches the current vault whereas the `WSUtils` one gets a list of notes by fname and prompts user to pick one. Very different use cases. They also don't have much overlapping code that is worth refactoring out.

I can rename mine so it doesn't get confusing

SeriousBug: I think you could refactor it into a common function that accepts a callback to resolve the "more than 1 maybe note" case, but it's fine if you'd rather deal with that later.

tma66: That sounds like a good idea. I'll take a look into it when I try to refactor the code later

packages/plugin-core/src/commands/NoteLookupCommand.ts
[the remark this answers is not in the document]
jonathanyeung: Sorry, missed it!  Also - for now, we can just run madge under packages/plugin-core (scoping it to plugin-core instead of the entire code base).

packages/plugin-core/src/WSUtilsV2.ts
```diff
@@ -95,6 +96,48 @@ export class WSUtilsV2 implements IWSUtilsV2 {
...
+      const vaults = maybeNotes.map((noteProps) => {
+        return {
+          vault: noteProps.vault,
+          label: VaultUtils.getName(noteProps.vault),
+        };
+      });
+
+      const items = vaults.map((vaultPickerItem) => ({
+        ...vaultPickerItem,
+        label: vaultPickerItem.label
```
SeriousBug, line 126: There might be a small bug here:

Screenshot_20220128_150011

Steps to reproduce:
- Open the test workspace. You may need to close and restart after opening it for the first time.
- Use the lookup bar, type "templates.book", select the "prompt for vault" button (box shape, on the very left of the top bar), then select "Create New" and hit enter.
- Write something in the new template note.
- Create a note "book.something", the prompt will have the wrong vault names.

The problem goes away after restarting though. I would almost suspect `getNotesByFnameFromEngine` returning an incorrect result, but it does return 2 notes hence the 2 vaults.

SeriousBug: Also a nitpick, but can we make it clearer what's being selected here? Maybe this function should accept a parameter to override the item labels and title or something, but this is especially confusing if you enable the prompt for which vault the note should be created in.

Something like "Select which template to apply" as the title, and the labels being something like "templates.book from vault1" would be easier to understand I think. Feel free to leave that to a future task though, I don't want to block you from merging this feature.

SeriousBug (commented):
Look good, but there is one bug I encountered which should be fixed. Also a nitpick, but that can be left for future iterations.

packages/common-all/src/dnode.ts
```diff
@@ -870,7 +882,10 @@ export class NoteUtils {
       vault: toVault,
     });
 
-    if (maybeNotes.length > 1) {
+    if (maybeNotes.length === 1) {
```
kevinslin, line 885: can we refactor the signature here? this function calls ` NoteUtils.getNotesByFname`. we should use `getNotesByFnameFromEngine` instead

kevinslin (changes requested):
just a minor comment. could you also rebase from master to remove the merge conflicts?

packages/common-all/src/dnode.ts
tma66: sure

packages/plugin-core/src/WSUtilsV2.ts
tma66: I like that suggestion. I'll go ahead and update the title/labels

tma66: Regarding the incorrect vault name, it looks like an indexing issue with the engine. For example, after I create a new note in vaultThree, when I inspect `engine.notes`, it thinks the vault is vault1. Restarting fixes the vault name

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -749,6 +753,430 @@ suite("NoteLookupCommand", function () {
...
+              schemas.map((s) => {
+                schema.schemas[s.id] = s;
+              });
+              return schema;
+            },
+          });
+        },
+      },
+      () => {
+        test("THEN template body gets applied to new note", async () => {
```
kevinslin, line 805: can you make this more descriptive? "THEN template body gets applied to new note FROM other vault"

packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -749,6 +753,430 @@ suite("NoteLookupCommand", function () {
...
+              schemas.map((s) => {
+                schema.schemas[s.id] = s;
+              });
+              return schema;
+            },
+          });
+        },
+      },
+      () => {
+        test("THEN correct template body gets applied to new note", async () => {
```
kevinslin, line 882: see comment above. "FROM vault referred to be xvault link"

SeriousBug (approved):
Looks good to me.

kevinslin (changes requested):
when i have multiple schemas and i escape out, I get the following

```
Warning: Problem with journal schema. No template found for templates.daily
```

- we should have: title in lookup widget (press escape to not apply a schema)
- in this case, not show the warning message

kevinslin (approved):

Issue #2228, opened 2022-01-20
File links with spaces


aglucky:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ x] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

File links will not work if there is a space or %20 in the name

## To Reproduce

1. Add a file with spaces in the name to assets
2. Link the file to a note and view the markdown preview
3. Click link

## Expected behavior

Linked file should appear.

## Desktop (please complete the following information)

- OS: [e.g.  windows]
- Browser: [e.g. chrome, safari]
- Version: [e.g. 22]

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here

hikchoi:
Thank you for the report.

We will be investigating this and pushing a fix as soon as possible.

hikchoi:
This has been resolved with v80.

Feel free to let us know in this issue if you still have outstanding concerns. Otherwise this issue will be automatically closed in 3 days.

cwtowns:
This is not fixed in v88 or v89.  I have an image in `assets/images/foo bar.png`.  If I use the text `![foo](assets/images/foo bar.png)` the text is rendered literally.  If I use the text `![foo](assets/images/foo%20bar.png)` it renders a broken image icon.

bewillcott:
This is still a problem in v0.119.0.
I am trying to use a link to a pdf file that has spaces in the name:
e.g.: "My local file.pdf"

\[My local file](file:///home/myname/test/My local file.pdf)

If, however, I rename the file to: "My_local_file.pdf"

\[My local file](file:///home/myname/test/My_local_file.pdf)

Then it works as expected and opens the file in the system's default application, which is Firefox.
Also, using "assets/..." gives the same results in both cases.

PR #2279, opened 2022-01-26
fix: decode urlencoded spaces in asset path before opening

hikchoi:
# fix: decode urlencoded spaces in asset path before opening

This PR:
- fixes an issue with asset paths not opening correctly with the system's default app when there is a urlencoded space (%20) in the filename.

# Manual testing
- Since this is not straightforward to integ-test, I'm doing the following:

1. open `dendron.ref.assets` in `test-workspace`
2. open preview
3. click on link to asset with space
4. confirm opening in default app

# Pull Request Checklist

## Testing
- [x] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in the [test workspace](https://docs.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)

### Docs
### Analytics


hikchoi:
confirmed manual testing in 
- [ ] mac
- [ ] windows
- [ ] linux

hikchoi:
related: #2228

hikchoi:
madge reports no additional circ dependencies

jonathanyeung (approved):

kevinslin (changes requested):
can we have a test for this?

hikchoi:
> can we have a test for this?

Do you have any suggestions on how to write an integ test for this?

The behavior itself is triggered from the webview, which the plugin-core does listens to. the end result we want to verify is that a system default app has been used to open an asset file.

I can test if the path that we get at the end would be properly decoded, but if there's anything else that I'm blanking on here let me know.

packages/plugin-core/src/test/suite-integ/ShowPreview.test.ts
```diff
@@ -19,7 +19,71 @@ suite("ShowPreview utility methods", () => {
...
+            done();
+          },
+        });
+      });
+    });
+  });
+
+  // this test block passes because it doesn't wait for the actual test to run.
+  // the actual test does not pass.
+  describe.skip(`handleLink`, () => {
```
hikchoi, line 86: I've skipped this test because this test was pre-emptively passing before the actual code was executed.
Even with proper test resolution, this test code will fail.

hikchoi: I'll try to block some time to fix this, but I don't think this would be within the scope of this PR.

kevinslin (approved):
this looks good. can you rename `ShowPreview.test.ts` to `PreviewLinkHandler.test.ts` to make it more obvious for the next person?

PR #2293, opened 2022-01-26
chore: update github label policy

kevinslin:
issue labeler should only update labels on first open instead of
everytime someone edits

### DIFF
--- a/.github/workflows/issue-labler.yml
+++ b/.github/workflows/issue-labler.yml
@@ -1,7 +1,7 @@
 name: "Issue Labeler"
 on:
   issues:
-    types: [opened, edited]
+    types: [opened]
 
 jobs:
   triage:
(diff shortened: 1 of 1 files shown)


hikchoi:
Ticking the checkboxes unfortunately is an `edit`, but do we also want to not have the labeler add back `triage-needed` when there is a legitimate edit? (and other label changes).

I guess the person on-call could make note of the discussion and follow up without relying on the labeler bot. I don't really have a strong opinion on this.

hikchoi:
Since we are fixing the bots here, can we also change [this line](https://github.com/dendronhq/dendron/blob/master/.github/no-response.yml#L4) to match what [this document](https://github.com/dendronhq/dendron-docs/blob/main/vault/dev.process.issues.md#L54:L54) is describing? (doc says no-response closes after 7 days, but the config is set to 3 right now)

jonathanyeung:
aside from comment from @hikchoi , lgtm

kevinslin:
@hikchoi when users edit, we should get notified when sorting issues by last updated. will add a line in the oncall sop to check for that

kevinslin:
i've updated the docs to switch to 3 days. until we get our backlog under control, lets err to getting more tickets closed out sooner

PR #2295, opened 2022-01-26
enhance(schemas): Support date subsitution via string replacement

tma66:
As followup from https://github.com/dendronhq/dendron/discussions/2207, we will use string replacement instead of lodash to support date variable substitution. Since we don't need javascript evalaution, we find this will be simpler. We will still use lodash default interpolate delimiter

**Testing**
Create template as follows:
```
Year is <%= CURRENT_YEAR %>
Month is <%=CURRENT_MONTH %>
Hour is <%=CURRENT_HOUR%>

This links goes to [[daily.journal.<%=CURRENT_YEAR%>.<%= CURRENT_MONTH%>.<%= CURRENT_DAY%>]]
```
Creating note with template leads to:
```
Year is 2022
Month is 01
Hour is 13

This links goes to [[daily.journal.2022.01.26]]
```

Updated docs for [[Template Variables|dendron://dendron.dendron-site/dendron.topic.templates#template-variables]]

**Madge**
Before: 130
After: 130

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1372,22 +1372,34 @@ export class SchemaUtils {
         note.body = tempNote.body;
       }
 
-      // Apply date variable substitution to the body based on mustache delimiter if applicable
-      // E.g. if template has {{ CURRENT_YEAR }}, new note will contain 2021
-      // Use mustache delimiter
-      // TODO: Renable date variable substitution when new format is thought through
-      /*_.templateSettings.interpolate = /\{\{([\s\S]+?)\}\}/g;
-
+      // Apply date variable substitution to the body based on lodash interpolate delimiter if applicable
+      // E.g. if template has <%= CURRENT_YEAR %>, new note will contain 2021
       const currentDate = Time.now();
-      const compiledSchemaTemplate = _.template(note.body);
-      note.body = compiledSchemaTemplate({
-        CURRENT_YEAR: currentDate.toFormat("yyyy"),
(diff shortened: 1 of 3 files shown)


jonathanyeung:
Code looks good to me - but we should have a corresponding docs change to detail the new format?

packages/common-all/src/dnode.ts
```diff
@@ -1372,22 +1372,34 @@ export class SchemaUtils {
...
-        CURRENT_YEAR: currentDate.toFormat("yyyy"),
-        CURRENT_MONTH: currentDate.toFormat("LL"),
-        CURRENT_DAY: currentDate.toFormat("dd"),
-        CURRENT_HOUR: currentDate.toFormat("HH"),
-        CURRENT_MINUTE: currentDate.toFormat("mm"),
-        CURRENT_SECOND: currentDate.toFormat("ss"),
-      });*/
+
+      note.body = note.body.replace(
+        new RegExp("<%=\\s*CURRENT_YEAR\\s*%>", "g"),
```
SeriousBug, line 1380: Nitpick: Could we use regular expression literals? Like `/<%=\s*CURRENT_YEAR\s*%>/g`

SeriousBug (approved):
Just one minor nitpick, but we can merge without fixing that. It looks good to me, tested manually and it works fine.

packages/common-all/src/dnode.ts
tma66: Sure, no problem

kevinslin (approved):
this looks good.nice job!

PR #2346, opened 2022-02-03
chore(workspace): LookupJournal/ScratchNoteCommand cleanup

hikchoi:
# chore(workspace): LookupJournal/ScratchNoteCommand cleanup
This PR:
- Renames `Dendron: Lookup Journal Note` and `Dendron: Lookup Scratch Note` commands to `Dendron: Create Journal Note` and `Dendron: Create Scratch Note`
- Modifies `NoteLookupCommand` so that it can accept custom controllers prior to `gatherInput`
- Supply custom controllers in `CreateJournal/SratchNoteCommand` with own set of buttons and quickpick title

# Pull Request Checklist

## Testing
- [~] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in the [test workspace](https://docs.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)

### Docs

### Analytics

### DIFF
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -186,12 +186,12 @@
         "title": "Dendron: hidden"
       },
       {
-        "command": "dendron.lookupJournalNote",
-        "title": "Dendron: Lookup Journal Note"
+        "command": "dendron.createJournalNote",
+        "title": "Dendron: Create Journal Note"
       },
       {
-        "command": "dendron.lookupScratchNote",
-        "title": "Dendron: Lookup Scratch Note"
+        "command": "dendron.createScratchNote",
+        "title": "Dendron: Create Scratch Note"
       },
       {
         "command": "dendron.lookupSchema",
@@ -513,11 +513,11 @@
           "when": "dendron:pluginActive && dendron:noteLookupActive && !editorFocus && !view"
         },
         {
-          "command": "dendron.lookupJournalNote",
+          "command": "dendron.createJournalNote",
           "when": "dendron:pluginActive"
         },
         {
-          "command": "dendron.lookupScratchNote",
+          "command": "dendron.createScratchNote",
           "when": "dendron:pluginActive"
         },
         {
@@ -962,7 +962,7 @@
         "when": "dendron:pluginActive && dendron:noteLookupActive && !editorFocus && !view"
       },
       {
-        "command": "dendron.lookupJournalNote",
+        "command": "dendron.createJournalNote",
         "key": "ctrl+shift+j",
         "mac": "cmd+shift+j",
         "args": {
@@ -971,7 +971,7 @@
         "when": "dendron:pluginActive"
       },
       {
-        "command": "dendron.lookupScratchNote",
+        "command": "dendron.createScratchNote",
         "key": "ctrl+shift+s",
         "mac": "cmd+shift+s",
         "when": "dendron:pluginActive"
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -77,7 +77,7 @@
           "name": "Lookup View",
           "contextualTitle": "Lookup View",
           "type": "webview",
-          "when": "dendron:pluginActive && dendron:noteLookupActive"
+          "when": "dendron:pluginActive && dendron:noteLookupActive && dendron:shouldShowLookupView"
         },
         {
           "id": "dendron.backlinks",
--- a/packages/plugin-core/src/commands/CreateJournalNoteCommand.ts
+++ b/packages/plugin-core/src/commands/CreateJournalNoteCommand.ts
@@ -1,15 +1,26 @@
-import { Logger } from "../logger";
+// import { Logger } from "../logger";
 import { DENDRON_COMMANDS } from "../constants";
 import { BasicCommand } from "./base";
+import { CommandOpts as NoteLookupCommandOpts } from "./NoteLookupCommand";
+import { LookupControllerV3CreateOpts } from "../components/lookup/LookupControllerV3";
+import { ILookupControllerV3 } from "../components/lookup/LookupControllerV3Interface";
 import {
-  CommandOutput as NoteLookupOutput,
-  CommandRunOpts as NoteLookupRunOpts,
-} from "./NoteLookupCommand";
+  ILookupProviderV3,
+  ILookupProviderOptsV3,
+  NoteLookupProviderSuccessResp,
+} from "../components/lookup/LookupProviderV3Interface";
+import {
+  CopyNoteLinkBtn,
+  HorizontalSplitBtn,
+  JournalBtn,
+} from "../components/lookup/buttons";
+import { IDendronExtension } from "../dendronExtensionInterface";
+import { NoteLookupProviderUtils } from "../components/lookup/NoteLookupProviderUtils";
+import { HistoryEvent } from "@dendronhq/engine-server";
 import { AutoCompletableRegistrar } from "../utils/registers/AutoCompletableRegistrar";
-import { LookupNoteTypeEnum } from "../components/lookup/types";
 
-type CommandOpts = NoteLookupRunOpts;
-type CommandOutput = NoteLookupOutput | undefined;
+type CommandOpts = void;
+type CommandOutput = void;
 
 export { CommandOpts as CreateJournalNoteOpts };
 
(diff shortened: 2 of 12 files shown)


hikchoi:
docs: https://github.com/dendronhq/dendron-site/pull/388

kevinslin (changes requested):
to contain complexity for the command, can we not show the toggle for scratch note when in "create scratch note" and same for "journal note"? 

so support the concept of a button that is toggled but not visible

jonathanyeung:
> to contain complexity for the command, can we not show the toggle for scratch note when in "create scratch note" and same for "journal note"?
> 
> so support the concept of a button that is toggled but not visible

Hmm - I'm not sure if toggling the availability of one button would reduce complexity of the command.  In a way, I think having the button still visible can be helpful - it helps the user understand that `Create Journal Note` is equivalent to Lookup with the Journal Modifier button enabled.

hikchoi:
UPDATE:
I took the liberty of removing irrelevant buttons from each command. I've also made journal/scratch modifiers not toggable, but let me know how to progress with this.

1. Each command now only has the following set of modifiers: vault select, journal (or scratch), copy note link, horizontal split.
2. Journal / Scratch is alway _on_ and not togglable
3. There is little to no use in the lookup view popping up when using these two commands, so I disabled them.
4. Since this is still a note lookup command and we just want to make it simpler, I modified the note lookup command so that it can accept a custom controller (that only defines buttons that we want to show + a custom title for the quickpick). This way we don't have to make these two commands a lookup-derivative command and separately implement the accept logic (just reuse the note lookup logic)
  - If a custom controller is given, it will use the provided controller rather than setting up a new one during `gatherInput`.
  - Now note lookup command will explicitly dispose of the controller when the quickpick is hidden. A new controller will be created every time a new note lookup is triggered. (this was not the case until now and it only disposed the quickpick.)
    - This is necessary to use the bulk of the note lookup command's logic with a custom controller in the two new commands added

jonathanyeung:
> There is little to no use in the lookup view popping up when using these two commands, so I disabled them.

I'm not exactly sure what this means - does it mean running Create Scratch Note command will just create the new scratch note without further user input?

kevinslin:
> > to contain complexity for the command, can we not show the toggle for scratch note when in "create scratch note" and same for "journal note"?
> > so support the concept of a button that is toggled but not visible
> 
> Hmm - I'm not sure if toggling the availability of one button would reduce complexity of the command. In a way, I think having the button still visible can be helpful - it helps the user understand that `Create Journal Note` is equivalent to Lookup with the Journal Modifier button enabled.

I'm going off an observation thhat you made during metrics review which is when people use lookup, they don't use lookup and then decide they want to use a "journal note". rather, they use journal note directly. 

i think journal note/scratch note should be note traits that don't actually have buttons in the future (we can create lookup commands that apply traits to notes)

kevinslin:
> UPDATE: I took the liberty of removing irrelevant buttons from each command. I've also made journal/scratch modifiers not toggable, but let me know how to progress with this.
> 
>     1. Each command now only has the following set of modifiers: vault select, journal (or scratch), copy note link, horizontal split.
> 
>     2. Journal / Scratch is alway _on_ and not togglable
> 
>     3. There is little to no use in the lookup view popping up when using these two commands, so I disabled them.
> 
>     4. Since this is still a note lookup command and we just want to make it simpler, I modified the note lookup command so that it can accept a custom controller (that only defines buttons that we want to show + a custom title for the quickpick). This way we don't have to make these two commands a lookup-derivative command and separately implement the accept logic (just reuse the note lookup logic)
> 
> 
>     * If a custom controller is given, it will use the provided controller rather than setting up a new one during `gatherInput`.
> 
>     * Now note lookup command will explicitly dispose of the controller when the quickpick is hidden. A new controller will be created every time a new note lookup is triggered. (this was not the case until now and it only disposed the quickpick.)
>       
>       * This is necessary to use the bulk of the note lookup command's logic with a custom controller in the two new commands added

>  I've also made journal/scratch modifiers not toggable, but let me know how to progress with this.

Can we actually hide the button all together? I don't see the value of having the button here and it needlessly adds complexity to the UI

> Each command now only has the following set of modifiers: vault select, journal (or scratch), copy note link, horizontal split.

Why only these? the other modifiers are still applicable here (eg. selection2link)

> There is little to no use in the lookup view popping up when using these two commands, so I disabled them.

Why? if there are modifiers, then the lookup view should display. there's a few tweaks we should make for lookup view UX (eg. if the sidebar is open, it shouldn't need to load, better belnding with background, fix in linux) so its okay to hide it for now but this is a relevant use case

> I modified the note lookup command so that it can accept a custom controller (that only defines buttons that we want to show + a custom title for the quickpick).

Sounds good. Can you update the dev docs? probably [[Cook|dendron://dendron.docs/pkg.plugin-core.dev.cook]]

jonathanyeung:
> Can we actually hide the button all together? I don't see the value of having the button here and it needlessly adds complexity to the UI

+1 - yea that's the best approach imo.  Except instead of just hiding, we can remove the associated journal/scratch button code altogether.

hikchoi:
@jonathanyeung 

> I'm not exactly sure what this means - does it mean running Create Scratch Note command will just create the new scratch note without further user input?

I am talking about the lookup panel view, not the quickpick

hikchoi:
@kevinslin 

> Can we actually hide the button all together? I don't see the value of having the button here and it needlessly adds complexity to the UI

A lot of the logic of lookup and its modifiers are tied to the buttons and their existence. This PR was already a simple name fix that became something much larger. Do we want to keep this going or should we stop here and add a new task for the next sprint?

> Why only these? the other modifiers are still applicable here (eg. selection2link)

I'm sorry. `selection2link` and `selectionExtract` does make sense to be here. I originally had them here but they got lost in the process. However, there are other modifiers that do not make sense in the context of _creating_ a journal / scratch note.

> Why? if there are modifiers, then the lookup view should display. there's a few tweaks we should make for lookup view UX (eg. if the sidebar is open, it shouldn't need to load, better belnding with background, fix in linux) so its okay to hide it for now but this is a relevant use case

I could have been a bit more clear here. Let me rephrase.

The lookup view was made with the assumption that it will be used with the full lookup, where there are multiple types of modifiers, and each type of modifiers do similar things. The view helps visually organize them. (as opposed to a linear presentation of buttons in the quickpick). With a few buttons removed for the two new commands, the situation is flipped. The Lookup view will have a dropdown with one item / multiselect form with one item, a checkbox that does nothing, etc, which is much more confusing than having the quickpick buttons. On top of the tweaks you mentioned, lookup view was also made with the assumption that all buttons will be present. So hiding the buttons in the two commands would cause a mismatch in the view, and eventually an error when the user selects a modifier that doesn't exist. The changes made in the PR is already overextending what the PR was originally meant to be, and fixing this would add more to this. The lookup view _as-is_ provides little benefit.

hikchoi:
There are a few conflicting requests / opinions here so let's summarize this before moving forward:

1. remove quickpick buttons completely
  - this requires a bit of rework of how the modifiers work as the buttons themselves are tied to the modifier logic
2. but keep modifiers that make sense
  - vault select, copy note link, selection2link, selectionExtract, Journal/Scratch button, horizontal split button
3. present these modifiers with the lookup view
  - make the lookup view stable first.
  - lookup view also should be modified to make sense with lookup-derivative commands

The remaining question is: Do we want to cut it here and start a new task?

jonathanyeung:
> This PR was already a simple name fix that became something much larger. Do we want to keep this going or should we stop here and add a new task for the next sprint?

Haha - yea this became more like a functional spec discussion as opposed to a code review.  Apologies for that.  IMO - let's cut it here. 

> 1. remove quickpick buttons completely - this requires a bit of rework of how the modifiers work as the buttons themselves are tied to the modifier logic

Yeah - this one is going to be non-trivial, let's certainly not do that as part of this PR.  I've been tracking this item too for circular dep work as the button code is responsible for a lot of our remaining circular dependencies.

> 3. present these modifiers with the lookup view

I also think we should defer this one and take a closer look at the way the lookup view is structured ( + lookupController / provider interfaces).  This is the biggest piece remaining in the circ dependency work.

kevinslin:
agree with above comments. let keep pr as is. can you rebase to solve merge conflicts?

hikchoi:
It seems like part of my changes are introducing new circ dependencies related to the buttons.

madge is reporting 7 more in plugin-core compared to master

kevinslin:
> It seems like part of my changes are introducing new circ dependencies related to the buttons.
> 
> madge is reporting 7 more in plugin-core compared to master

i know we have a separate task to make buttons stateless. I think its okay, thoughts @jonathanyeung ?

jonathanyeung:
> > It seems like part of my changes are introducing new circ dependencies related to the buttons.
> > madge is reporting 7 more in plugin-core compared to master
> 
> i know we have a separate task to make buttons stateless. I think its okay, thoughts @jonathanyeung ?

@hikchoi  - would you mind posting the diff on the circ dependencies?

jonathanyeung:
I think it's from this one: 

packages/plugin-core/src/components/lookup/LookupControllerV3Interface.ts:
```typescript
import { LookupView } from "../../views/LookupView";
```

with the primary circle being

`src/components/lookup/LookupControllerV3Interface.ts` -> `src/views/LookupView.ts` -> `src/components/lookup/LookupControllerV3.ts`

I did already have a task open for circ deps on `LookupView.ts` ([[Circ Dep Lookup View|dendron://private/task.dev.2022.01.11.circ-dep-lookup-view]]).  Refactoring out circular deps around LookupView/LookupController is a big item. Given how this change has already been dragged out, let's defer that to be part of the above task.

jonathanyeung (approved):

PR #2350, opened 2022-02-03
fix(pod): markdown import to update asset references

Harshita-mindfire:
Related to  https://github.com/dendronhq/dendron/issues/1237

### DIFF
--- a/packages/engine-test-utils/src/__tests__/pods-core/MarkdownPod.spec.ts
+++ b/packages/engine-test-utils/src/__tests__/pods-core/MarkdownPod.spec.ts
@@ -299,8 +299,8 @@ function setupImport(src: string) {
     { path: "A1/B1.md" },
     { path: "A1/B2.md" },
     { path: "A1/B1 B2.md" },
-    { path: "Folder/something.md" },
-    { path: "Folder/assets/image.JPEG" },
+    { path: "Folder/Media/something.md" },
+    { path: "Folder/Media/image.JPEG" },
   ]);
 }
 
@@ -488,7 +488,6 @@ describe("markdown import pod", () => {
             noAddUUID: true,
           },
         });
-        console.log("wsRoot", wsRoot);
         const assetsDir = fs.readdirSync(path.join(vpath, "assets"));
         const fileBody = fs.readFileSync(path.join(vpath, "project.p4.md"), {
           encoding: "utf8",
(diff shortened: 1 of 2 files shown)


packages/engine-test-utils/src/__tests__/pods-core/MarkdownPod.spec.ts
```diff
@@ -487,6 +489,7 @@ describe("markdown import pod", () => {
             noAddUUID: true,
           },
         });
+        console.log("wsRoot", wsRoot);
```
kevinslin, line 492: remove console.log

kevinslin (changes requested):
can you make sure the github pr follows our commit style message? we should be able to copy the title of the message and add it to the changelog

[[Commit Style|dendron://dendron.docs/dev.process.review#commit-style]]

packages/engine-test-utils/src/__tests__/pods-core/MarkdownPod.spec.ts
```diff
@@ -503,11 +505,21 @@ describe("markdown import pod", () => {
...
+          path.join(vpath, "folder.something.md"),
+          {
+            encoding: "utf8",
+          }
+        );
         expect(fileBody.match("test.txt")).toBeTruthy();
         const assetPath = path.join("assets", "test.txt").replace(/[\\]/g, "/");
+        const imagePath = path
+          .join("assets", "image.JPEG")
+          .replace(/[\\]/g, "/");
```
SeriousBug, line 518: nitpick: `path.posix.join` will always use unix style joins instead if that's the intention. Or just type `assets/image.JPEG` directly :)

packages/pods-core/src/builtin/MarkdownPod.ts
```diff
@@ -353,7 +353,7 @@ export class MarkdownImportPod extends ImportPod<MarkdownImportPodConfig> {
   }
 
   static cleanAssetPath(path: string): string {
-    return path.replace(/[\\|/.]/g, "");
+    return path.toLowerCase().replace(/[\\|/.]/g, "");
```
SeriousBug, line 356: Won't this cause problems on case sensitive OSs like Linux/Mac? I think we should preserve the case of the file paths

SeriousBug (changes requested):
I have one concern about case-sensitive OSs. Could you explain a bit why we need to lowercase the file path? What was the problem in the referenced issue? How does lowercasing it fix that?

packages/engine-test-utils/src/__tests__/pods-core/MarkdownPod.spec.ts
```diff
@@ -503,11 +505,21 @@ describe("markdown import pod", () => {
...
         expect(fileBodyContent).toContain(`[test-pdf](/${assetPath})`);
         expect(fileBody2Content).toContain(`[test-pdf](/${assetPath})`);
-        expect(assetsDir.length).toEqual(3);
+        expect(fileBody3Content).toContain(`[image](/${imagePath})`);
```
SeriousBug, line 521: This test doesn't seem to test the linked issue. In the linked issue, the markdown file contains a link where the asset is in the same directory as the note, and the link looks like ``. The link is relative from the note file itself and not the root folder we're importing from.

SeriousBug (approved):
After talking to Harshita I see how this PR fixes the issue. Looks good to me.

kevinslin (approved):

kevinslin:
Looks good. can you update the github issue link so it doesn't autoclose? see updated SOP on that in https://github.com/dendronhq/dendron-docs/commit/683f8a4c7f5fc16dda6429e40c0fb4efc4881e8e

afterwards, feel free to add to release notes according to [[Ready to Ship|dendron://private/area.product.sop.ready-to-ship]]. note the changes in https://github.com/dendronhq/org-private/commit/1f69fa963d1c3aa29132078e3330166e38cbf410

Issue #2345, opened 2022-02-03
publishPod from CLI: Text isn't output to terminal/console for dendron.markdown or dendron.html, and --vault arg is ignored


ScriptAutomate:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [x] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

Using the CLI, I'm not able to get `publishPod` to work as expected. This seems to be the case for both Markdown and HTML, the two I have tested.

In the future, we should look at the following to work via CLI:
```sh
dendron publishPod --wsRoot . --podId dendron.html --query root
```
At the moment, it doesn't work, even if the pod was first generated via CLI:
```sh
dendron publishPod --wsRoot . --podId dendron.html --genConfig
```
It also seems to require `vaultName` to be present in the config yml, otherwise it will fail. `--vault` looks to be ignored at the CLI, with or without `vaultName` present in the yml config.

## To Reproduce

1. Run `dendron publishPod --wsRoot . --podId dendron.html --genConfig` to generate the HTML publish pod config
2. Run `dendron publishPod --wsRoot . --vault vault --podId dendron.html --query root` (where `vault` is named `vault`)

This errors out. Looks like `vaultName` is required to be configured in the `dendron.html/config.publish.yml`. After that is configured, I can run the command, but with zero output.

3. Run `dendron publishPod --wsRoot . --podId dendron.html --query root`
4. See zero output. No error and no text output to terminal.

## Expected behavior

Two things were expected (which means this may require two issues?):
- `dendron publishPod --wsRoot . --vault vault --podId dendron.html --query root` should work, even if `vaultName` is not configured in the `.yml`, since I am providing `vault` via `--vault` (which says in the `--help` is a valid arg).
- `dendron publishPod --wsRoot . --vault vault --podId dendron.html --query root` should output the result in the console/terminal output

For another example, I imagine that the following should be the same commands:

```sh
# Output `root.md` as GitHub Flavored Markdown
# This command currently works without issue
[...]

PR #2352, opened 2022-02-03
fix(pod): acknowledge cli args for publish pod

Harshita-mindfire:
Related to : https://github.com/dendronhq/dendron/issues/2345

### DIFF
--- a/packages/dendron-cli/.vscode/tasks.json
+++ b/packages/dendron-cli/.vscode/tasks.json
@@ -44,37 +44,25 @@
     {
       "label": "export-pod",
       "type": "shell",
-      "command": "node --inspect /usr/local/bin/dendron-cli exportPod --wsRoot ~/Dendron --podId dendron.airtable --podPkg @dendronhq/airtable-pod --podSource remote --showConfig",
-      "options": {
-        "cwd": "/Users/kevinlin/Dendron"
-      },
+      "command": "node --inspect ${workspaceFolder}/lib/bin/dendron-cli exportPod --wsRoot ${workspaceFolder}/../../test-workspace --podId dendron.airtable --podPkg @dendronhq/airtable-pod --showConfig",
       "problemMatcher": []
     },
     {
       "label": "gdoc-import-pod",
       "type": "shell",
-      "command": "node --inspect dendron_setup/dendron/packages/dendron-cli/lib/bin/dendron-cli importPod --wsRoot ~/Dendron --podId dendron.gdoc --podSource builtin",
-      "options": {
-        "cwd": "D:/"
-      },
+      "command": "node --inspect ${workspaceFolder}/lib/bin/dendron-cli importPod --wsRoot ${workspaceFolder}/../../test-workspace --podId dendron.gdoc --podSource builtin",
       "problemMatcher": []
     },
     {
       "label": "orbit-import-pod",
       "type": "shell",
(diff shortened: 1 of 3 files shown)


packages/dendron-cli/.vscode/tasks.json
```diff
@@ -67,6 +67,15 @@
         "cwd": "D:/dendron_setup/dendron/"
       },
       "problemMatcher": []
+    },
+    {
+      "label": "publish-pod",
+      "type": "shell",
+      "command": "node --inspect dendron/packages/dendron-cli/lib/bin/dendron-cli publishPod --wsRoot ${workspaceFolder}/../../test-workspace --podId dendron.markdown --vault vault --query root --podSource builtin",
+      "options": {
+        "cwd": "E:/dendron-code/"
```
kevinslin, line 76: since this is personal to your workspace, can we not check this in to the global task.json? can we get to this path using builtin variables like `${workspaceFolder}` instead? and if so, can you add this to `dev.process` so future folks can also do it?

kevinslin (changes requested):

packages/dendron-cli/.vscode/tasks.json
Harshita-mindfire: https://github.com/dendronhq/dendron-docs/pull/17

packages/dendron-cli/src/commands/pod.ts
```diff
@@ -169,16 +177,25 @@ export function enrichPodArgs(opts: {
...
-      } else {
-        // eslint-disable-next-line no-console
-        console.log(
-          `WARN: --query parameter not implemented for podType ${podType}`
-        );
       }
+    } else {
+      // eslint-disable-next-line no-console
+      console.log(
+        `WARN: --query and --vault parameter not implemented for podType ${podType}`
```
hikchoi, line 197: the base command has a `print` method which you can use to avoid the eslint disable every time.

packages/dendron-cli/.vscode/tasks.json
hikchoi: not related to this PR, but I see a few `/Users/kevinlin/ ...` directories in `tasks.json` 😄 maybe we should clean that up in a separate PR as well.

packages/dendron-cli/src/commands/pod.ts
hikchoi: are there pod types that don't actually require / implement these options?

hikchoi (changes requested):
Some small nitpick and a Q, but otherwise looks good

packages/dendron-cli/src/commands/pod.ts
Harshita-mindfire: yeah, --query is not required by any pod except `publish` and export pods have `vaults` config option which is of type object.

packages/dendron-cli/.vscode/tasks.json
Harshita-mindfire: addressed in the same PR : [f5d96d6](https://github.com/dendronhq/dendron/pull/2352/commits/f5d96d6fba56f55ea0b78fe468db14226961893d).

packages/dendron-cli/src/commands/pod.ts
Harshita-mindfire: updated console.logs with print [f5d96d6](https://github.com/dendronhq/dendron/pull/2352/commits/f5d96d6fba56f55ea0b78fe468db14226961893d)

kevinslin (approved):

hikchoi (approved):

PR #2358, opened 2022-02-04
chore: remove lint-staged

SeriousBug:
`lint-staged` causes the unstaged hunks inside partially-staged files to be staged as well. This may cause people (especially myself) to accidentally commit changes they did not intend to commit. This PR removes `lint-staged`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,9 @@
     "hooks:pre-commit": "node ./hooks/pre-commit.js",
     "hooks:pre-push": "node ./hooks/pre-push.js",
     "gen:meta": "node ./bootstrap/scripts/genMeta.js",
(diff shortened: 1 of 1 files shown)


kevinslin (changes requested):
`lint-staged` is actually really useful to make sure we have a consistent code base and if anything, we should be more aggressive with formatting. instead of removing it, can we:

- implement the workaround here: https://github.com/okonet/lint-staged/issues/62#issuecomment-286830310
- add a section in [[Dendron Review Process|dendron://dendron.docs/dev.process.review]] about staging partial commits using `git stash -k`

SeriousBug:
Thanks for the feedback, added the workaround and tested it. It works correctly.

PR #2359, opened 2022-02-04
chore: avoid workspace watcher crashing if folder is deleted

SeriousBug:
`WorkspaceWatcher` was using the deprecated `WorkspaceService.isPathInWorkspace` to check if something is in workspace. This function does `this.config`, which runs `WorkspaceService.getOrCreateConfig`, which will attempt to read the config file, or write it if the config file is missing.

If the folder that should contain the config file is missing, this will cause a `ENOENT` error.
I think this can happen if the user loads up a Dendron workspace, but then deletes the folder that contains the workspace while VSCode is still running.


kevinslin:
looks like a build error

SeriousBug:
My bad, fixed now.

hikchoi (commented):
Is there a reason we are fixing the deprecated method instead of the new method?

jonathanyeung:
> Is there a reason we are fixing the deprecated method instead of the new method?

Looks like there are 12 references to the deprecated method.  Could we try to migrate those instances instead? Then, we can also clean up `IWorkspaceService`

SeriousBug:
My bad, I was trying to fix the issue causing the overload of errors in our logs and tried to make the fix as quickly as possible.

The deprecated function is fully removed now.

SeriousBug:
This PR drops the total circular dependency count from 136 to 123.

kevinslin (approved):

PR #2361, opened 2022-02-04
chore: disable inactive user survey prompt

kevinslin:
There's currently a bug where users that use Dendron every day are still
getting prompted. Disabling until we root cause and fix


SeriousBug (approved):
Looks good to me.

kevinslin:
that should be correct (we haven't had any reports of folks running into it)

PR #2365, opened 2022-02-04
chore: add tests for schema matches

SeriousBug:
~~The way we build patterns, a pattern that starts with `!(thing)` means "anything that's not `thing`".
However, if the same pattern is used for the top level (schema part with `root` as the parent),
then the final pattern becomes
`!(thing)/...`. Because the pattern starts with `!`, this means the entire pattern is negated.~~

~~By adding the option `nonegate` we disable this behavior, so `!(thing)` works the same way no matter which part of the schema it is used in.~~

Nevermind, while trying to write tests I found that the top level of the schema doesn't allow/respect patterns so you can't have this case anyway. I'm adding the tests anyway though because I don't think there are any schema tests that check if expected fnames match the schema.

# Pull Request Checklist

## Testing
### Docs
### Analytics


SeriousBug:
@kevinslin Sure, here's an example (it's right under the example schema block): https://wiki.dendron.so/notes/c5e5adde-5459-409b-b34d-a0d75cbb1052/#pattern

The example for negation is at the bottom of that section, and looks like this:
`"!(temp|scratch)" matches notes, recipes etc. but not temp or scratch, if you are trying to exclude just those hierarchies`

So this can be used to exclude specific hierarchies. And to clarify, this PR is not enabling this functionality. The functionality is already enabled, there's just this edge case where if the topmost hierarchy's pattern begins with `!` then the entire pattern gets negated and not just that level of the hierarchy.

kevinslin:
> @kevinslin Sure, here's an example (it's right under the example schema block): https://wiki.dendron.so/notes/c5e5adde-5459-409b-b34d-a0d75cbb1052/#pattern
> 
> The example for negation is at the bottom of that section, and looks like this: `"!(temp|scratch)" matches notes, recipes etc. but not temp or scratch, if you are trying to exclude just those hierarchies`
> 
> To clarify, this PR is not enabling this functionality. The functionality is already enabled, there's just this edge case where if the topmost hierarchy's pattern begins with `!` then the entire pattern gets negated and not just that level of the hierarchy. This PR changes the options for minimatch so that doesn't happen.

Got it. Can we add a test for this?

SeriousBug:
Nevermind, while trying to write tests I found that the top level of the schema doesn't allow/respect patterns so you can't have this case anyway. I'm adding the tests anyway though because I don't think there are any schema tests that check if expected fnames match the schema.

packages/engine-test-utils/src/__tests__/common-all/dnode.spec.ts
```diff
@@ -241,4 +245,202 @@ describe(`SchemaUtil tests:`, () => {
...
+      checks: {
+        "cli.fd": "cli",
+        "cli.fd.bar.usage": false,
+        "cli.fd.cmd.usage": "cmd",
+        "cli.fd.env": "env",
+        "foo.bar": false,
+      },
+    });
+
+    testSchemaMatches({
```
kevinslin, line 281: I'm guilty of introducing the `testWithEngine` pattern but have reconsidered it because:
- it's implementation specific (assumes jest, doesn't work with mocha test)
- not easily compsable (doesn't let us use modifiers like `.only` easily)

Can we use a utility that doesn't actually run a test but just sets up the test case?

kevinslin (changes requested):
ah, nice discovery. in terms of schema testing, a few things:

- since schemas tests might be used across a number of different scenarios, can you add it as a top level test utilities? see https://github.com/dendronhq/dendron/pull/2371/files#diff-dd647564a4b4b5bddaa97ffdb7af6e3f47d09b214dd29ea1fd97bbc22fc1f185 for example
    - after adding it, can you also update the docs here? [[Test Utilities|dendron://dendron.docs/dev.process.qa.test#test-utilities]]

SeriousBug:
Added the docs to `[[Testing Schema Matches|dendron://dendron.docs/pkg.common-test-utils.ref.test-utils#testing-schema-matches]]` and revised how it works. The common function just generates the tests now, and there's a local function in `dnode.spec.ts` that runs them. The run function also supports `.skip` and '.only` for convenience!

PR #2366, opened 2022-02-06
chore(dev): update pr templates for community and team contributions

kevinslin:
Team contributions have a few more steps compared to community
contributions - make this explicit by splitting out PR checklist into
two separate templates

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,17 +1,6 @@
 
 # Pull Request Checklist
 
-If this is your first time submitting a pull request to Dendron, copy and paste the full [Dendron Review Checklist](https://docs.dendron.so/notes/1EoNIXzgmhgagqcAo9nDn.html) into this request and check off each item as necessary. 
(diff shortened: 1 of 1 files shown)


hikchoi (commented):
Is there a reason we are linking to a gist instead of the notes in `docs.dendron.so`?

kevinslin:
> Is there a reason we are linking to a gist instead of the notes in `docs.dendron.so`?

mainly because the output in `docs.dendorn.so` is HTML encoded and we want a pure plaintext version to copy and paste

jonathanyeung:
Does the gist URL encode a specific commit hash? If so, can it instead point generically to the file so that we can update the note without changing `pull_request_template.md`?

kevinslin:
> Does the gist URL encode a specific commit hash? If so, can it instead point generically to the file so that we can update the note without changing `pull_request_template.md`? Otherwise lgtm

updated

jonathanyeung (approved):

hikchoi (approved):

Issue #1972, opened 2021-12-24
File with subcategory that starts with 'md' does not show up in the Tree View properly


rickbsgu:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

As an example, for the category 'medical', the following subcategory files might be named thusly:

- medical.doctor.md
- medical.mdanderson
- medical.md-anderson.md

'medical.doctor' is fine and lists in the Tree View properly.  However, the other two will list as 'medical', treating the subcategory as a '.md' file suffix.

## To Reproduce

Reproduce the hierarchy above.

## Expected behavior

Tree View should list 'mdanderson' in the first case and 'md-anderson' in the next case.

Instead, it shows 'medical'.

## Suggested Fix

Fix should be easy - just check the string length past the '.' to the end of the string for a fixed length of 2 when building the Tree View.

rickbsgu:
Well, I cloned this to see if I could fix it quickly.

No dice.  That's a hell of a big codebase.  I need a dev guide, minimally.

Otherwise, I'll just have to wait until you all can fix this.

jonathanyeung:
@rickbsgu - thanks for filing the issue and giving it a go with the fix. We are in the process of cleaning up our dev docs, but you can find what we have now here: https://docs.dendron.so/

We'll add this to our backlog.

Harshita-mindfire:
Hi, this has been fixed with Dendron v0.82. Please let us know if it continues to happen, otherwise, this issue will close in a few days.

Issue #856, opened 2021-06-23
Having '.md' anywhere inside new note name will break title generation


LoPawlie:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

When creating a new note, any hierarchy element that starts with 'md' (and is not a domain) will break the auto title generation by using the previous hierarchy element to title the note/stub that starts with 'md' and any subsequent children instead of using the proper ones.

## To Reproduce

1. Go to Lookup.
2. Write `learn.mdn.css.text`
3. Frontmatter shows `title: Learn` for the new note.
3. Expand tree view to see the whole new hierarchy titled 'learn'
    ```
    .
    └── learn
        └── learn
            └── learn
                └── learn
    ```

## Expected behavior

The note would be titled 'Text' and the tree view would look like this:
```
.
└── learn
    └── mdn
        └── css
            └── text
```

## Desktop

- OS: macOS 10.15
- VSCode: 1.57.1
- Dendron: 0.47.0

## Dendron Log file

Please attach the output of `>Dev:Dendron: Open Logs` here

[dendron.log](https://github.com/dendronhq/dendron/files/6698655/dendron.log)

kevinslin:
That's a bug. Thanks for catching this. We'll look into this for the upcoming release

chmac:
I think #1972 is probably related to this. I'd guess that there's a regex somewhere which is `\.md` instead of `\.md$`.

@kevinslin Any pointers as to where this issue might be in the codebase? I could have a look and see if I can find the faulty regex if you're able to offer a general pointer, like which package this code would be in, etc.

Harshita-mindfire:
Hi, this has been fixed with Dendron v0.82. Please let us know if it continues to happen, otherwise, this issue will close in a few days.

chmac:
Looks fixed to me.

PR #2369, opened 2022-02-07
fix(workspace): correct title generartion for notes with sub-hierarchy starting with .md

Harshita-mindfire:
### Basics
- [x] code should follow [[Code Conventions|dendron://dendron.docs/dev.process.code]]
- [x] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [[Avoiding Circular Dependencies|dendron://dendron.docs/dev.process.code.best-practices#avoiding-circular-dependencies]].

### Extended
- General
- Operations
- Architecture

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -94,11 +94,9 @@ export class DNodeUtils {
   }
 
   static basename(nodePath: string, rmExtension?: boolean) {
-    if (rmExtension) {
-      const idx = nodePath.lastIndexOf(".md");
-      if (idx > 0) {
-        nodePath = nodePath.slice(0, idx);
-      }
+    //skip for nodePath that has a sub-hierarchy starting with .md eg: learn.mdone.test, learn.md-one.new
+    if (rmExtension && nodePath.endsWith(".md")) {
+      nodePath = nodePath.slice(undefined, -3);
     }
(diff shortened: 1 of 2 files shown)


Harshita-mindfire:
Related issue: https://github.com/dendronhq/dendron/issues/1972 and https://github.com/dendronhq/dendron/issues/856

madge report: now new circular dependencies(64)

packages/common-all/src/dnode.ts
```diff
@@ -94,7 +94,8 @@ export class DNodeUtils {
   }
 
   static basename(nodePath: string, rmExtension?: boolean) {
-    if (rmExtension) {
+    //skip for nodePath that has a sub-hierarchy starting with .md eg: learn.mdone.test, learn.md-one.new
+    if (rmExtension && nodePath.endsWith(".md")) {
```
SeriousBug, line 98: minor nitpick: could simplify this the rest of this part, if path ends with ".md" then `nodePath = nodePath.slice(undefined, -3)`

SeriousBug (approved):
Looks good, left one nitpick on simplifying the code.

packages/common-all/src/dnode.ts
Harshita-mindfire: Updated. Thanks!

kevinslin (approved):

PR #2371, opened 2022-02-08
enhance(cli): enable fix frontmatter action for doctor in cli

kevinslin:
This PR also fixes up some tests:
- create doctor test utils for doctor test
- use doctor test utils in doctor tests for both plugin and cli
- refactor doctor plugin test
- create `utils/index.ts` from `utils.ts` in `common-test-utils`
- split `AssertUtils` into own file in `utils/assert`

Individual Commits
- add doctor command to cli
- add tests
- refactor common test utils
- update tests


.github/workflows/ci.yml
```diff
@@ -79,9 +79,9 @@ jobs:
         with:
           path: |
             packages/*/lib/*
-          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}
+          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}-1
           restore-keys: |
-            ${{ runner.os }}-yarn-
+            ${{ runner.os }}-yarn-1
```
SeriousBug, line 84: Is this intentional, or was it accidentally left in?

packages/engine-server/src/doctor/service.ts
```diff
@@ -158,9 +159,15 @@ export class DoctorService {
     let doctorAction: (note: NoteProps) => Promise<any>;
     switch (action) {
       case DoctorActionsEnum.FIX_FRONTMATTER: {
-        console.log(
-          "the CLI currently doesn't support this action. please run this using the plugin"
-        );
+        const changed = await new BackfillService().updateNotes({
```
SeriousBug, line 162: This is not the right way to write doctor actions, you need to set `doctorAction` with an async function and place the code inside, like the other actions below. This also doesn't respect the scope selected by the doctor because of that.

packages/engine-test-utils/src/__tests__/dendron-cli/commands/doctor.spec.ts
```diff
@@ -122,6 +123,60 @@ const runDoctor = (opts: Omit<DoctorCLICommandOpts, "server">) => {
...
+        {
+          expect,
+          preSetupHook: async (opts) => {
+            TestDoctorUtils.createNotesWithBadIds(opts);
+          },
+        }
+      );
+    });
+  });
+});
```
SeriousBug, line 179: Could you also add a test where you check it with different scopes? E.g. if you run the doctor action with a scope of 1 note, it should only replace the id in that note and not other notes.

SeriousBug (changes requested):
I had a concern about the doctor actions implementation, and a concern with testing. Otherwise looks good.

packages/engine-server/src/doctor/service.ts
kevinslin: this is still a draft, just converted it to such

.github/workflows/ci.yml
kevinslin: nope intentional. just posted an async explaining this: https://github.com/dendronhq/dendron-docs/blob/main/vault/dev.process.code.gotchas.changing-file-to-folder.md#L16:L16

PR #2377, opened 2022-02-09
enhance: Block anchor support for non-note files

SeriousBug:
Dendron now supports block anchors in non-note files. This makes it possible to reliably link to specific lines in non-note files, like linking to a function inside a code file.

In specific, this PR adds:
- Goto Note can navigate to block anchors in non-note files
- Editor wikilink highlighting will highlight links to non-note files with block anchors as existing if the file exists (whether the block anchor exists or not is not checked since that would be expensive)
- Copy note link can now insert block anchors in non-note files. The default is inserting block anchors, but users can change this to use line numbers or to prompt every time.
- Copy Note Link now tracks (telemetry) the type of file it was used in (note vs non-note) and the type of anchor it created (none, header, block, line).

This PR also refactors `DConfig.writeConfig` to be async.

Docs: https://github.com/dendronhq/dendron-site/pull/398

# Pull Request Checklist

## Testing
### Docs
### Analytics


SeriousBug:
This PR does not affect the circular dependency count.

kevinslin:
see some build errors

SeriousBug:
My bad, I must have missed the action status. It should be good now.

jonathanyeung (commented):
Code overall looks good. 

I gave it a test - but I'm not seeing the prompt to pick a block anchor type?  The steps I took: 

1. Create a new file `test.ts`. 
2. Write some ts code in there
3. Put my cursor on a line, run `Copy Note Link`

I only got a link like [[vault/test.ts]], but didn't get prompted for a block anchor type.

I also left a comment on the docs PR

SeriousBug:
@jonathanyeung You need to have some text selected, similar to how block anchors already work with notes.

packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -37,8 +45,52 @@ export class CopyNoteLinkCommand extends BasicCommand<
     window.showInformationMessage(`${link} copied`);
   }
 
+  private async getUserLinkAnchorPreference(): Promise<
+    "line" | "block" | null
+  > {
+    const { config, wsRoot } = ExtensionProvider.getDWorkspace();
+    let anchorType = ConfigUtils.getNonNoteLinkAnchorType(config);
+    if (anchorType === undefined) {
+      // The preferred anchor type is not set, so ask the user if they want line numbers or block anchors
```
kevinslin, line 54: makes me think if this should be a third option (eg. prompt) when anchor type is not set.

packages/common-all/src/types/configs/commands/copyNoteLink.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Namespace for configuring {@link CopyNoteLinkCommand}
+ */
+export type CopyNoteLinkConfig = {
+  nonNoteFile?: {
+    anchorType?: NonNoteFileLinkAnchorType;
+  };
+};
+
+export function genDefaultCopyNoteLinkConfig(): CopyNoteLinkConfig {
+  // don't set a default for `nonNoteFiles`, we want to prompt the user whether they want lines or block anchors
```
kevinslin, line 14: rationale for prompting users vs sticking with the behavior we have for note files? (which is adding an anchor block)

I think there's an argument to be made whether that is the right default but unless there's good reason (and there might be - I can see users not wanting to accidentally add an anchor to some external file), we should have the same default for note files as well as non note files

In which case, we might want to revisit or default for copy note link and set the default there to be a prompt as well (followup task)

test-workspace/other-files/config.ts
```diff
@@ -0,0 +1,63 @@
...
+  IntermediateDendronConfig,
+  DendronSiteConfig,
+  Time,
+} from "@dendronhq/common-all";
+import { readYAML, writeYAML } from "@dendronhq/common-server";
+import fs from "fs-extra";
+import _ from "lodash";
+import path from "path";
+
+export class DConfig {
```
kevinslin, line 13: can we add this to [[Test Workspace|dendron://dendron.docs/dev.ref.test-workspace]] as something that people can use to test non-note files?

packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -37,8 +45,52 @@ export class CopyNoteLinkCommand extends BasicCommand<
...
+            label: "line",
+            description: "Use line numbers `L123`",
+            detail: "Links may point to the wrong place once code is changed.",
+          },
+        ],
+        {
+          canPickMany: false,
+          ignoreFocusOut: true,
+          title:
+            "What type of anchors should Dendron create for links in non-note files?",
```
kevinslin, line 73: its not clear from this prompt that this will actually write the preference in `dendron.yml`. In general, we haven't had a precedent for modifying the configuration and I can see this as a valid case as well (`prompt` mode). a proposal:

- set the default mode to `prompt` which will give this prompt on selection
- if the user escapes from prompt, should just copy without any anchors
- do not update the config on user selection

kevinslin (changes requested):

jonathanyeung:
nit: the description in the quick pick gets cut off (although it can still be read with hover).  Can we shorten it just a little bit for better readability? Maybe:

"Always links to the right place, but a short comment must be inserted into the file"

image

packages/common-all/src/types/configs/commands/copyNoteLink.ts
SeriousBug: I added the option because that's what we previously discussed to do :sweat_smile: 

I think that makes sense. I think with notes it might be better to use the read-only status introduced in RFC 42 later so we generate line numbers for read-only vaults and block anchors for writeable vaults. But we can iterate on the UX of that later.

packages/plugin-core/src/commands/CopyNoteLink.ts
SeriousBug: > do not update the config on user selection

To confirm, are you saying it should prompt every time until the user goes and modifies the configuration file?

kevinslin: yeah. we shouldn't update the user configuration, this should be the behavior if dendron is in prompt mode

packages/common-all/src/types/configs/commands/copyNoteLink.ts
kevinslin: yep, i remember making that proposal 😅
now seeing it in action, I realize that I was wrong

test-workspace/other-files/config.ts
SeriousBug: Good idea, added.

kevinslin (approved):

PR #2188, opened 2022-01-17
chore: hook for soft warning on circular dependency

jonathanyeung:
## chore(internal): hook for soft warning on circular dependency

BREAKING: adds madge as a dev-dependency; requires madge installation in your local dev environment

Simple hook for now to add warning text to the husky hook with a madge report on the count of circular dependencies. Adding a delta is a bit more involved (need to modify local git state), so I'll defer on that for now.

---

# Pull Request Checklist

If this is your first time submitting a pull request to Dendron, copy and paste the full [Dendron Review Checklist](https://docs.dendron.so/notes/1EoNIXzgmhgagqcAo9nDn.html) into this request and check off each item as necessary. 

This template contains the short checklist which is used by the Dendron core team. 

## Testing
- [ ] [Write Tests](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#writing-tests) 
- [ ] [Confirm existing tests pass](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#executing-tests)
- [ ] [Confirm manual testing](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#manual-testing) 
- [ ] If your tests changes an existing snapshot, make sure that snapshots are [updated](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#updating-test-snapshots)
- [ ] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), make sure that it is included in the [test workspace](https://docs.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)
### Docs
- [ ] if your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR
### Analytics
- [ ] if you are adding analytics related changes, make sure the [Telemetry](https://docs.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

PR #2379, opened 2022-02-09
enhance(publish): static tree menu

eleweek:
enhance(publish): static tree menu

1. Heads up: this adds ~186kb of `tree.json` for `org-workspace-incubator/dendron`. I think this could be reduced a bit by minifying json and reducing length of property names (I don't know if there is any kind of zipping).
2. I moved some of the code to `TreeUtils` as discussed with Kevin. Workspace version still relies on `TreeViewUtils` and we agreed that simplifying/refactoring `TreeViewUtils` is outside of the scope of this task.
3. `getTree()` is pretty similar to `getNotes()` but this is intentional. There is no simple way to merge them aside from creating a factory function for get-or-return-cached function, and a factory function seemed like an overkill in this case. 

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended


packages/nextjs-template/components/DendronTreeMenu.tsx
```diff
@@ -46,31 +48,28 @@ export default function DendronTreeMenu(
 
   // --- Verify
   // If no data, show spinner component
-  if (!verifyNoteData(props)) {
+  if (!props.tree || !noteActiveId || !props.noteIndex) {
```
kevinslin, line 51: is it possible not to have `props.tree`? since we are passing it in at build time, shouldn't this prop be non-optional?
same with `noteIndex` and `noteActiveId`?

packages/pods-core/src/builtin/NextjsExportPod.ts
```diff
@@ -455,6 +460,10 @@ export class NextjsExportPod extends ExportPod<NextjsExportConfig> {
         encoding: "utf8",
         spaces: 2,
       }),
+      fs.writeJSON(treeDstPath, tree, {
```
kevinslin, line 463: can we add a test for this in src/__tests__/pods-core/NextjsExportPod.spec.ts?

kevinslin (changes requested):
nice work. a few comments

kevinslin (changes requested):

kevinslin:
also heads up that the ci/cd is failing. originally it was because the bundled zip did not have a `tree.json` file. i uploaded a new zip file but now its failing  because unzip needs to overwrite an existing file (due to caching). we'll want to replace the line in https://github.com/dendronhq/dendron/blob/enhance/static-tree-menu-publishing/.github/workflows/ci.yml#L111:L111 with a `-o` flag to force overwrite.

https://superuser.com/questions/100656/linux-unzip-command-option-to-force-overwrite

PR #2388, opened 2022-02-09
chore(workspace) First pass at resolving cyclic dependencies wrt windowwatcher/workspacewatcher

tma66:
Madge Before: 53
Madge After: 48

Some notables changes in this commit:
- Update extension interface and extensionprovider to expose `isActive`, `getWorkspaceConfig`, and `getTreeView`
- Remove dependency on `workspace.ts` from `WorkspaceWatcher.ts`
- Update constructor of `WorkspaceWatcher.ts` to pass in windowwatcher instance
- Move `refreshLookupView` from `PickerUtilsV2` to class method of `LookupView`
- Update constructor of `LookupView` to take in extension instance

TODO: Resolve cyclic dependencies around `fileWatcher.ts` which should resolve rest of watcher specific issues

### DIFF
--- a/packages/plugin-core/src/ExtensionProvider.ts
+++ b/packages/plugin-core/src/ExtensionProvider.ts
@@ -44,6 +44,9 @@ export class ExtensionProvider {
     return ExtensionProvider.getExtension().getWorkspaceConfig();
   }
 
+  /**
+   * @deprecated. See {@link IDendronExtension.getTreeView}
+   */
   static getTreeView(key: DendronTreeViewKey) {
     return ExtensionProvider.getExtension().getTreeView(key);
   }
--- a/packages/plugin-core/src/ExtensionProvider.ts
+++ b/packages/plugin-core/src/ExtensionProvider.ts
@@ -1,5 +1,5 @@
 import { IDendronExtension } from "./dendronExtensionInterface";
-import { DendronError } from "@dendronhq/common-all";
+import { DendronError, DendronTreeViewKey } from "@dendronhq/common-all";
 import _ from "lodash";
 import { IWSUtilsV2 } from "./WSUtilsV2Interface";
 
@@ -34,6 +34,18 @@ export class ExtensionProvider {
 
   static getWSUtils(): IWSUtilsV2 {
     return ExtensionProvider.getExtension().wsUtils;
+  }
+
+  static isActive() {
+    return ExtensionProvider.getExtension().isActive();
+  }
+
+  static getWorkspaceConfig() {
+    return ExtensionProvider.getExtension().getWorkspaceConfig();
+  }
+
+  static getTreeView(key: DendronTreeViewKey) {
+    return ExtensionProvider.getExtension().getTreeView(key);
   }
 
   static register(extension: IDendronExtension) {
--- a/packages/plugin-core/src/components/lookup/utils.ts
+++ b/packages/plugin-core/src/components/lookup/utils.ts
@@ -8,7 +8,6 @@ import {
   DNodeUtils,
   DNoteLoc,
   DVault,
-  LookupModifierStatePayload,
   NoteProps,
   NoteUtils,
   OrderedMatcher,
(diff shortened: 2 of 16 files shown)


packages/plugin-core/src/views/LookupView.ts
```diff
@@ -127,6 +129,23 @@ export class LookupView implements vscode.WebviewViewProvider {
...
+          pressed: button.pressed,
+        };
+      }
+    );
+
+    const lookupView = this._extension.getTreeView(
+      DendronTreeViewKey.LOOKUP_VIEW
+    ) as LookupView;
+    lookupView.refresh(payload);
+  }
```
hikchoi, line 147: This probably can be merged with the `refresh` method below and simplified: 

```suggestion
  public refresh(opts: { buttons: DendronBtn[] }) {
    const { buttons } = opts;
    const payload: LookupModifierStatePayload = buttons.map(
      (button: DendronBtn) => {
        return {
          type: button.type,
          pressed: button.pressed,
        };
      }
    );

    if (this._view) {
      this._view.webview.postMessage({
        type: LookupViewMessageEnum.onUpdate,
        data: { payload },
        source: "vscode",
      });
    }
  }
```

packages/plugin-core/src/components/lookup/utils.ts
```diff
@@ -678,26 +676,10 @@ export class PickerUtilsV2 {
...
-    );
-
-    // TODO: swpa this out for `ExtensionProvider.getExtension()`
-    // once IDendronExtension interface has the tree view properties.
-    const lookupView = getExtension().getTreeView(
+    const lookupView = ExtensionProvider.getTreeView(
       DendronTreeViewKey.LOOKUP_VIEW
     ) as LookupView;
-    lookupView.refresh(payload);
+    lookupView.refreshLookupView({ buttons: opts.quickpick.buttons });
```
hikchoi, line 682:
```suggestion
    lookupView.refresh({ buttons: opts.quickpick.buttons });
```

See suggestion in `LookupView.ts`

hikchoi (changes requested):
Added some suggestions on lookup view method

packages/plugin-core/src/ExtensionProvider.ts
```diff
@@ -36,6 +36,18 @@ export class ExtensionProvider {
...
 
+  static isActive() {
+    return ExtensionProvider.getExtension().isActive();
+  }
+
+  static getWorkspaceConfig() {
+    return ExtensionProvider.getExtension().getWorkspaceConfig();
+  }
+
+  static getTreeView(key: DendronTreeViewKey) {
```
jonathanyeung, line 47: Can we go ahead and slap a `@deprecated` tag on this one? :D 

No issue with adding this in now to get around some circ deps, but going forward with the eventing pattern, we shouldn't need to expose any tree views anymore.

packages/plugin-core/src/dendronExtensionInterface.ts
```diff
@@ -98,4 +99,18 @@ export interface IDendronExtension {
...
+  isActive(): boolean;
+
+  /**
+   * Get Global Workspace configuration
+   */
+  getWorkspaceConfig(
+    section?: string | undefined
+  ): vscode.WorkspaceConfiguration;
+
+  getTreeView(key: DendronTreeViewKey): vscode.WebviewViewProvider;
```
jonathanyeung, line 115: Same comment as above w/ `@deprecated` tag

packages/plugin-core/src/test/MockDendronExtension.ts
```diff
@@ -163,4 +170,16 @@ export class MockDendronExtension implements IDendronExtension {
     }
     return this._engine as IEngineAPIService;
   }
+
+  isActive(): boolean {
+    throw new Error("Method not implemented in MockDendronExtension.");
```
jonathanyeung, line 175: we can probably safely just `return true` in this impl for tests?

jonathanyeung (changes requested):

packages/plugin-core/src/views/LookupView.ts
tma66: makes sense, thanks for pointing out

packages/plugin-core/src/ExtensionProvider.ts
tma66: sounds good

PR #2398, opened 2022-02-10
chore(refactor): Migrate backlinkspanel to use EngineEventsEmitter

tma66:
*Background*
Refactor data sourcing for Backlinks Panel to use the new EngineEventsEmitter interface. 

Instead of refreshing the backlinkspanel manually from different sources, we want the backlinks panel to subscribe to an event and refresh upon event fire. The particular event we're interested in ,`onEngineNoteStateChanged`, gets fired for various note changes (add/delete/update).

*Changes*
1. Have BacklinksTreeDataProvider subscribe to `onEngineNoteStateChanged`
2. Remove calls to `refreshBacklinks` from fileWatcher
3. Move exposed functions to class methods for `BacklinksTreeDataProvider`. Only expose whats needed

*Testing*
1. Update `BacklinksTreeDataProvider.test.ts` to test for callbacks upon note changes
2. Manually verified backlinks refresh for update, create, delete notes

This change will help address cyclic dependencies as well as misc tech debt

Madge before: 53
Madge after: 23

For more information:[[Migrate Backlinkspanel to Engineeventsemitter|dendron://private/task.2022.02.09.migrate-backlinkspanel-to-engineeventsemitter]]

### Basics

### Extended

- General
- Operations
- Architecture

## Instrumentation

N/A

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

N/A

### Basics

N/A

## 

### Basics

## 

### Extended


packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
```diff
@@ -6,239 +6,66 @@ import {
...
-    this.updateSortOrder(BacklinkSortOrder.PathNames);
+    this.onDidChangeTreeData = this._onDidChangeTreeDataEmitter.event;
+    this._engineEvents = engineEvents;
+    this.setupSubscriptions();
   }
 
-  refresh(): void {
-    this._onDidChangeTreeData.fire(undefined);
+  private setupSubscriptions(): void {
+    this._engineEvents.onEngineNoteStateChanged(() => {
```
SeriousBug, line 55: This returns a disposable, the data provider should also be a disposable and should dispose this when itself is disposed.

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
```diff
@@ -283,17 +110,202 @@ export default class BacklinksTreeDataProvider
...
+        }
+
+        const ref2Updated = ref2[0].note.updated;
+        const ref1Updated = ref1[0].note.updated;
+
+        // We want to sort in descending order by last updated
+        return ref2Updated - ref1Updated;
+      });
+    } else {
+      Logger.error({
```
SeriousBug, line 263: Looks like the `sortOrder` is exhaustively checked already, so this could be replaced with `assertUnreachable(sortOrder)`.

SeriousBug (changes requested):
I have one concern with disposables and a nitpick. The PR otherwise looks good, and works in manual testing.

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
```diff
@@ -6,239 +6,66 @@ import {
...
+      const ctx = "refreshBacklinks";
+      Logger.info({ ctx });
+      this.refreshBacklinks();
+    });
   }
 
+  /**
+   * Tells VSCode to refresh the backlinks view. Debounced to fire every 100 ms
+   */
+  public refreshBacklinks = _.debounce(() => {
```
jonathanyeung, line 65: make private

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
jonathanyeung: Good call - once this class is disposable, then in `workspace.ts`, let's add it to the extension context subscriptions, which should dispose of the resource automatically when our extension is closed:

```typescript
  private setupBacklinkTreeView() {
    const ctx = "setupBacklinkTreeView";
    Logger.info({ ctx, msg: "init:backlinks" });

    const backlinksTreeDataProvider = new BacklinksTreeDataProvider(
      this.getEngine(),
      this.getDWorkspace().config.dev?.enableLinkCandidates
    );

    this.context.subscriptions.push(backlinksTreeDataProvider);
    ```

**Review by jonathanyeung — CHANGES_REQUESTED** [review]

File: packages/plugin-core/src/features/BacklinksTreeDataProvider.ts [hunk 4]
```diff
@@ -6,239 +6,66 @@ import {
...
+    this._onDidChangeTreeDataEmitter.fire();
+  }, 100);
```
jonathanyeung, line 67: maybe bump up to 250 ms.

kevinslin (changes requested):
for the PR: can we add the new pr checklist here? feel free to add a **NA** in sections that don't apply

[[Export|dendron://dendron.docs/dev.process.review.checklist.extended.export]]

tma66:
> for the PR: can we add the new pr checklist here? feel free to add a **NA** in sections that don't apply
> 
> [[Export|dendron://dendron.docs/dev.process.review.checklist.extended.export]]

Done

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
jonathanyeung: Ah gotcha - let's still make it private though and encapsulate the subscription to `onDidChangeActiveTextEditor` within `BacklinksTreeDataProvider`.

So basically move this logic: 

```typescript
    vscode.window.onDidChangeActiveTextEditor(() =>
      backlinksTreeDataProvider.refreshBacklinks()
    );
```

into here:
```typescript
  private setupSubscriptions(): void {
  ...}
 ```
 
 And make sure to add the dispose logic of that subscription into the `dispose` method of `BacklinksTreeDataProvider` as well.

tma66: Gotcha makes sense

kevinslin:
looks like the ci tests are failing for plugin in backlinks emitter

tma66:
> looks like the ci tests are failing for plugin in backlinks emitter

^This is what happens when you update a PR at midnight...
Tests should be passing with latest commit

jonathanyeung (approved):
let's merge this on Friday after the release branch snap.

Issue #2341, opened 2022-02-02
Journal notes frontmatter differs between lookup and new daily note


Bassmann:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [X] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

There is a difference in the frontmatter for a journal note depending on how it is created
- using the new journal command (ctrl+shift+j) adds the full date as title as well as adding a trait entry
- creating it using lookup doesn't add the trait and also just adds the day as title

## To Reproduce

- Initialize a new workspace
- Create a new daily journal note. I tested both with the shortcut ctrl-shift-i and with the command Create Daily Journal Note, This creates a new daily note with this frontmatter
```
id: nxLJRRvAYJh5dX2mtmj7Y
title: '2022-02-01'
desc: ''
updated: 1643744888344
created: 1643744888344
traitIds:
  - journalNote
```
- Then do a lookup for daily.journal.2022.02.02 and create a new note.
This creates a note with this frontmatter
```
id: 7IHhHk1fLk8PITfCzD4BG
title: '02'
desc: ''
updated: 1643744647012
created: 1643744647012
```
And interestingly if I use Ctrl-Shift-j to create a note it has this frontmatter
```
id: UUDvruacLmkg1ojRhUNYT
title: '2022-02-01'
desc: ''
updated: 1643745128268
created: 1643745128268
```
i.e. without the traitsId

## Expected behavior

The frontmatter should be the same independent of how the note is created. In particular the title should be the full date 
## Desktop (please complete the following information)

- OS: Windows and Mac
- Version: 0.79

aleksey-rowan:
#2242 is tangentially related to this as it asks for functionality to create daily notes for future dates and these notes should also use the proper frontmatter (the full date as the title in particular).

ScriptAutomate:
Potential duplicate of https://github.com/dendronhq/dendron/issues/2211

Bassmann:
#2211 is mostly the same but beside the traitsId also the title is different depending on the creation method. 

[...]

Issue #2211, opened 2022-01-19
Journal trait is not activated from lookup


chmac:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [x] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

Creating a journal entry from lookup doesn't trigger the `journalNote` trait. The `traitIds` key is missing in frontmatter and the title is not automatically generated.

## To Reproduce

1. Click on a new day in the calendar (for which a note does not yet exist)
2. Note the title of the note and the `traitIds` frontmatter
3. Open lookup
4. Type a non existent journal date, press enter
5. Note the title and lack of `traidIds` front matter

## Expected behavior

I would expect templates and traits and the inbuilt mysterious secret sauce journal system to all work from lookup & calendar. #1911 :-)

hikchoi:
Thank you for the issue. I believe @jonathanyeung would be a much better person to answer your queries.

@jonathanyeung Can you take a look? I am unfortunately not up to date with how note traits work at the moment 🥲

jonathanyeung:
Thanks @chmac - adding to my backlog for investigation

Harshita-mindfire:
Related https://github.com/dendronhq/dendron/issues/2341

jonathanyeung:
A fix for this went out with v83.  

Some notes from the PR: 

---

After this change, when creating a Journal note from the following scenarios below should result in consistent frontmatter

Dendron: Create Daily Journal Note
Dendron: Create Journal Note
Creating a note via Lookup with the Journal button selected
Clicking on a new date in the Calendar webview
Frontmatter should look like

```yaml
---
id: IRbGt7WJKEsHNOcfYE2lS
title: '2022-02-11'
desc: ''
updated: 1644575670992
created: 1644575670993
traitIds:
  - journalNote
---
```
[...]

Issue #1911, opened 2021-12-16
Note Traits - Missing JS file for Journal Notes


micharris42:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [X] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

As of [version 0.73](https://wiki.dendron.so/notes/fRAyMtRmwkCwdRBj24eb0/). the **Note Traits** feature has been enabled in Dendron. As of the same release, journal notes are now automatically created with the frontmatter field "traitIds" and content of "journalNote". 

According to the [Note Trait Quickstart](https://wiki.dendron.so/notes/EQoaBI8A0ZcswKQC3UMpO/), there should be a JS file to accompany the note trait. In this case, `journalNote.js`. No such file exists in the Dendron vault. When using the "Create Note With Custom Trait" command, "journalNote" does not appear as an option. 

## To Reproduce

1. Create a new journal note with the "Create Daily Journal Note" command.

2. The frontmatter:

```
traitIds:
  - journalNote
```

Is automatically added.

3. A `journalNote.js` file is not available to edit in the Dendron vault.

## Expected behavior

If journal notes are meant to be a Note Trait, then there should be a JS file for users to modify.

If journal notes are not meant to be a Note Trait, then the frontmatter field "traitIds" should not be added when notes are created.

## Desktop (please complete the following information)

- OS: Windows 10
- Browser: NA
- Version: VSCodium 1.63.1, Dendron 0.73.1

## Additional context

If journal notes are intended to be a Note Trait, then every journal note created prior to Dendron 0.73 will be missing the frontmatter field "traitIds". Will there be an option to allow users to update their older journal notes with the new frontmatter?

jonathanyeung:
Hi @micharris42, thanks for filing the issue.

I realize now that the docs aren't very clear. 

We didn't initially intend for the built-in journal notes feature to be modifiable. But behind the scenes, it uses the same architecture as with user-configurable note traits, which will have JS files created for them.  

[...]

PR #2401, opened 2022-02-11
fix(workspace): Journal Note Consistency with Title and Traits

jonathanyeung:
## fix(workspace): Journal Note Consistency with Title and Traits

Fixes consistency issues where sometimes the titles on Journal notes weren't properly formatted as `yyyy-MM-dd` and the journalNote traitID wasn't being applied.

This also fixes other pre-existing issues where the journal title would not get properly generated if you deviate from the default `y-MM-dd` format for the journal note config.

Related: 
- https://github.com/dendronhq/dendron/issues/2341
- https://github.com/dendronhq/dendron/issues/2211
- https://github.com/dendronhq/dendron/issues/1911

After this change, when creating a Journal note from the following scenarios below should result in consistent frontmatter
- `Dendron: Create Daily Journal Note`
- `Dendron: Create Journal Note`
- Creating a note via Lookup with the Journal button selected
- Clicking on a new date in the Calendar webview

Frontmatter should look like
```yaml
---
id: IRbGt7WJKEsHNOcfYE2lS
title: '2022-02-11'
desc: ''
updated: 1644575670992
created: 1644575670993
traitIds:
  - journalNote
---
```

NOTE: By design (for now), if you create a new note via lookup with a title like `daily.journal.2022.01.01` **without** the journal modifier button clicked - basically you typed in the note name manually - then the title and traitIds will **not** be applied.  We may look to modify this behavior as we re-examine how some of the button behavior works in Lookup, but that's out of scope in this change.

madge before and after is at 46 - I had to make some changes to `JournalNote.ts` to make it work.

---

### Basics

### Extended
- General
  - [NA] check if similar function already exist in the codebase. if so, can it be re-used?
  - [NA] check if this change adversely impact performance
- Operations
- Architecture
  - [NA ] check if code is introducing changes on a foundational class or interface. if so, call for design review if needed

### Basics
- [ NA] if you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Extended
- [NA ] can we track the performance of this change to know if it is _successful_?

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested
- [NA] If your tests changes an existing snapshot, snapshots have been [updated](dev.process.qa.test)

### Extended
- [NA] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), an example is included in the [test workspace](dev.ref.test-workspace)
- [NA] CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs
- [NA] if your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR
- [NA] does this change introduce a new or better way of doing things that others need to be aware of? if so, an async should be created and a process added in [Development](dev) or [Packages](pkg)

### Extended
- [NA]  is this a developer BREAKING change? if another person cloning from this branch will need to adjust their dependencies or mental model of the architecture, then it is. if this is the case, make sure this is communicated according to [Close Loop](dev.process.close-loop)

### DIFF
--- a/packages/plugin-core/src/commands/CreateDailyJournal.ts
+++ b/packages/plugin-core/src/commands/CreateDailyJournal.ts
@@ -10,8 +10,12 @@ import {
 
 export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
   constructor(ext: IDendronExtension) {
-    super(ext, "dendron.journal", new JournalNote());
-
+    super(
+      ext,
+      "dendron.journal",
+      new JournalNote(ext.workspaceService!.config)
+    );
+    ext.getWorkspaceConfig();
     // override the key to maintain compatibility
     this.key = DENDRON_COMMANDS.CREATE_DAILY_JOURNAL_NOTE.key;
   }
--- a/packages/plugin-core/src/commands/CreateDailyJournal.ts
+++ b/packages/plugin-core/src/commands/CreateDailyJournal.ts
@@ -1,4 +1,4 @@
-import { ConfigUtils, VaultUtils } from "@dendronhq/common-all";
+import { ConfigUtils, DendronError, VaultUtils } from "@dendronhq/common-all";
 import _ from "lodash";
 import { DENDRON_COMMANDS } from "../constants";
 import { IDendronExtension } from "../dendronExtensionInterface";
@@ -10,12 +10,12 @@ import {
 
 export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
   constructor(ext: IDendronExtension) {
-    super(
-      ext,
-      "dendron.journal",
-      new JournalNote(ext.workspaceService!.config)
-    );
-    ext.getWorkspaceConfig();
+    const workspaceService = ext.workspaceService;
+
+    if (!workspaceService) {
+      throw new DendronError({ message: "Workspace Service not initialized!" });
+    }
+    super(ext, "dendron.journal", new JournalNote(workspaceService.config));
     // override the key to maintain compatibility
     this.key = DENDRON_COMMANDS.CREATE_DAILY_JOURNAL_NOTE.key;
   }
--- a/packages/plugin-core/src/commands/NoteLookupCommand.ts
+++ b/packages/plugin-core/src/commands/NoteLookupCommand.ts
@@ -65,6 +65,7 @@ import { AutoCompleter } from "../utils/autoCompleter";
 import { parseRef } from "../utils/md";
 import { VaultSelectionModeConfigUtils } from "../components/lookup/vaultSelectionModeConfigUtils";
 import { WSUtilsV2 } from "../WSUtilsV2";
+import { JournalNote } from "../traits/journal";
 
 export type CommandRunOpts = {
   initialValue?: string;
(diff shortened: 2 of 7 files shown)


packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -652,25 +671,19 @@ export class NoteLookupCommand
...
-        const [, ...maybeDatePortion] = quickpick.noteModifierValue.split(".");
-        // we only override y.MM.dd
-        if (maybeDatePortion.length === 3) {
-          const maybeTitleOverride = maybeDatePortion.join("-");
-          if (maybeTitleOverride.match(/\d\d\d\d-\d\d-\d\d$/)) {
-            return maybeTitleOverride;
-          }
+
+  journalTitleOverride(input: string): string | undefined {
+    if (/.*\d{4}\.\d{2}\.\d{2}$/g.test(input)) {
```
kevinslin, line 676: is there a better way to test for the input? this misses the case for example where format is `y.dd.MM`.

can we use luxon parsing for this? https://github.com/moment/luxon/blob/master/docs/parsing.md

packages/plugin-core/src/test/suite-integ/LookupJournal.test.ts
```diff
@@ -46,6 +44,14 @@ suite("Scratch Notes", function () {
           });
           const newNote = getNoteFromTextEditor();
           expect(newNote.fname.startsWith(`${fname}.journal`)).toBeTruthy();
+          // The note title should be in the format yyyy-MM-dd
+          expect(/\d{4}-\d{2}-\d{2}$/g.test(newNote.title)).toBeTruthy();
+
+          // @ts-ignore
+          const traits = newNote.traitIds;
```
kevinslin, line 51: i've done this too but in general, not a fan of having `@ts-ignore` in tests. does it make sense to establish a convention where `ts-ignore` code needs to be encapsulated in a file?

- test/suite-integ/ts-ignore-utils.ts
```tsx
getNoteTrait(note: NoteProps) [
//@ts-ignore
return note.traitIds
}
```

in the future, we want to enable strict typechecks and dis-allow `@ts-ignore` in the code base unless its in a file called `ts-ignore-*`

kevinslin (changes requested):
can we fill out the checkboxes? for the sections where it doesn't apply, feel free to delete and add a **NA**

packages/plugin-core/src/test/suite-integ/LookupJournal.test.ts
jonathanyeung: I bypassed the `ts-ignore` with a cast + a TODO comment.

> Does it make sense to establish a convention where ts-ignore code needs to be encapsulated in a file?

IMO - probably not worth it; I think it'll make things slightly more convoluted (and it may also lead to more circular deps).  Instead, let's add a point in our PR guidelines + coding conventions to not allow adding of additional `ts-ignore`'s going forward.  And then let's eliminate existing `ts-ignore`'s until we get down to 0, and then we can strengthen the ts compiler settings.

I can take a look at the docs and amend.

packages/plugin-core/src/commands/NoteLookupCommand.ts
jonathanyeung: good point!

packages/plugin-core/src/commands/CreateDailyJournal.ts
```diff
@@ -10,8 +10,12 @@ import {
 
 export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
   constructor(ext: IDendronExtension) {
-    super(ext, "dendron.journal", new JournalNote());
-
+    super(
+      ext,
+      "dendron.journal",
+      new JournalNote(ext.workspaceService!.config)
```
hikchoi, line 16: Should we explicitly check to make sure it's there instead of doing an non-null assertion?

hikchoi: The command is gated behind the enablement clauses so we do know the workspace service will exist at that point, but doesn't hurt to be sure and give a proper error message than a stack trace.

hikchoi (changes requested):

kevinslin (approved):

hikchoi (approved):

PR #2403, opened 2022-02-12
enhance(notes): change dendron id format to be alphanumeric lowercase

kevinslin:
This brings more compatibility to the widest possible use cases when synchronizing dendron notes with other sources.

For example, when ids are used in urls, certain hosting providers like netlify will lowercase all urls which will break users that publish using case sensitive ids.

This update makes the following changes:

- update dendron ids to be alphanumeric lowercase
- removes nanoidDictionary dependency
- extend the default length of a dendron id to 32 characters (this is the same length as what other providers like notion and loom use )

Documentation available at https://github.com/dendronhq/dendron-site/pull/402
***

### Basics

- [~] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices).
- [x] sticking to existing conventions instead of creating new ones [^eg-existing-convention]

### Extended

- General
- Operations
- Architecture
  - [x] check if code is introducing changes on a foundational class or interface. if so, call for design review if needed [^eg-check-for-design-review]

## Instrumentation

NA

### Basics

- [~] [Write Tests](dev.process.qa.test)  -> we don't have setup to exhaustively test uuid is lowercase
- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended

- [~  is this a developer BREAKING change? if another person cloning from this branch will need to adjust their dependencies or mental model of the architecture, then it is. if this is the case, make sure this is communicated according to [Close Loop](dev.process.close-loop)

### DIFF
--- a/packages/common-all/src/uuid.ts
+++ b/packages/common-all/src/uuid.ts
@@ -7,7 +7,7 @@ import { customAlphabet as nanoidInsecure } from "nanoid/non-secure";
  */
 const SHORT_ID_LENGTH = 12;
 /** Default length for nanoids. */
-const LONG_ID_LENGTH = 32;
+const LONG_ID_LENGTH = 23;
 
 const alphanumericLowercase = "0123456789abcdefghijklmnopqrstuvwxyz";
 
--- a/packages/common-all/src/uuid.ts
+++ b/packages/common-all/src/uuid.ts
@@ -1,6 +1,5 @@
 import { customAlphabet as nanoid } from "nanoid";
 import { customAlphabet as nanoidInsecure } from "nanoid/non-secure";
-import { alphanumeric } from "nanoid-dictionary";
 
 /** Using this length, according to [nanoid collision calculator](https://zelark.github.io/nano-id-cc/),
  * generating 1000 IDs per hour, it would take around 919 years to have 1 percent chance of a single collision.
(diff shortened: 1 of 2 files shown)


jonathanyeung:
This code looks fine - were we still planning on making the engine interpret id's in a case-insensitive manner?

jonathanyeung (approved):

kevinslin:
> This code looks fine - were we still planning on making the engine interpret id's in a case-insensitive manner?

So I took a look at the work involved and its actually going to be a moderate lift (eg. a full day). to move there iteratively, I'm thinking we should make this the default and then introduce doctor command to format existing ids (ids can already be incompatible due to `-` symbols which were accepted before)

packages/common-all/src/uuid.ts
```diff
@@ -1,14 +1,15 @@
...
-import { alphanumeric } from "nanoid-dictionary";
 
 /** Using this length, according to [nanoid collision calculator](https://zelark.github.io/nano-id-cc/),
  * generating 1000 IDs per hour, it would take around 919 years to have 1 percent chance of a single collision.
  * This is okay for the "insecure" generator, which is used in limited cases where collisions are less likely.
  */
 const SHORT_ID_LENGTH = 12;
 /** Default length for nanoids. */
-const LONG_ID_LENGTH = 21;
+const LONG_ID_LENGTH = 32;
```
SeriousBug, line 10: Using the collision calculator linked in the comment, bumping the ID length from 21 to 23 with the dictionary change in this PR is enough to preserve the collision resistence. Shorter IDs mean shorter links, so I'd suggesst lowering this to 23.

SeriousBug (changes requested):
Looks good, but I'd suggest a shorter ID length. We don't have to go all the way to 32 to preserve the collision resistence.

packages/common-all/src/uuid.ts
kevinslin: updated!

PR #2405, opened 2022-02-12
fix(workspace): Dendron will try to parse non-dendron files in `onFirstOpen`

kevinslin:
This commit makes the following changes:
- `onFirstOpen` should only track files in the workspace
- note specific operations (eg. moving cursor past frontmatter) should only happen for dendron notes
- adds two new methods to `IDendronExtension` (`activateWatchers` and `deactivate`)

Before, opening a big json file (eg. Dendron Logs) would cause vscode to crash because Dendron would try to parse the entire log file. 

This commit also introduces two new conventions:

1. The `WorkspaceWatcher` is hard to test since methods are `private` and require a working VSCode workspace to actually trigger (which we do not have access to in tests). 
Some solutions that were explored:
- ad-hoc dependency injection: resulted in too many code changes, we're better off actually implementing DI properly using one of the frameworks: https://blog.logrocket.com/top-five-typescript-dependency-injection-containers but out of scope for this task
- making private variables public - not great since they were private for a reason
- introduce the `__DO_NOT_USE_IN_PROD_exposePropsForTesting` method/convention to expose internal workspace state for testing

We went with the last option (will update in [[Testing|dendron://dendron.docs/dev.process.qa.test]] after the PR)

2. During testing, we often do typecasts of the form `foo!.bar` or use `@ts-ignore`. This isn't great. In the future, we want to enable strict type checks and dis-allow type assertions.
To make it easier to get to this future, this PR introduces the `UNSAFE_getWorkspaceWatcherPropsForTesting` convention: everytime you need to type cast, wrap it in a function. 
This groups all unsafe assertions in one place (which we can then add to a special unsafe file which we can ignore for strict type checking)

circular dependencies:
- before: 52
- after: 52

---
# Export

### Basics

- [~] sticking to existing conventions instead of creating new ones 
  - this introduces some new testing conventions
  

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

- [B] does this change introduce a new or better way of doing things that others need to be aware of? if so, an async should be created and a process added in [Development](dev) or [Packages](pkg)
- pending successful code review

## 

### Extended

### DIFF
--- a/packages/engine-server/src/workspace/utils.ts
+++ b/packages/engine-server/src/workspace/utils.ts
@@ -1,19 +1,19 @@
 import {
+  ConfigUtils,
   CONSTANTS,
-  IntermediateDendronConfig,
   DendronError,
   DNodeUtils,
   DVault,
   DWorkspaceV2,
   getSlugger,
+  IntermediateDendronConfig,
   isBlockAnchor,
+  isNotUndefined,
   NoteProps,
   VaultUtils,
   WorkspaceFolderCode,
   WorkspaceOpts,
   WorkspaceType,
-  ConfigUtils,
-  isNotUndefined,
 } from "@dendronhq/common-all";
 import {
   findDownTo,
@@ -100,6 +100,28 @@ export class WorkspaceUtils {
       )
     );
     return dendronWorkspaceFolders.filter(isNotUndefined);
+  }
+
+  /**
+   * Check if a file is a dendron note (vs a regular file or something else entirely)
+   */
+  static isDendronNote({
+    wsRoot,
+    vaults,
+    fpath,
+  }: { fpath: string } & WorkspaceOpts) {
+    // check if we have markdown file
+    if (!fpath.endsWith(".md")) {
+      return false;
+    }
+    // if markdown file, check if it is in a dendron vault
+    if (!WorkspaceUtils.isPathInWorkspace({ wsRoot, vaults, fpath })) {
+      return false;
+    }
+
+    // if markdown file, does it have frontmatter?
+    // this is a rough heuristic
+    return fs.readFileSync(fpath, { encoding: "utf8" }).startsWith("---");
   }
 
   static isNativeWorkspace(workspace: DWorkspaceV2) {
--- a/packages/engine-server/src/workspace/utils.ts
+++ b/packages/engine-server/src/workspace/utils.ts
@@ -105,7 +105,7 @@ export class WorkspaceUtils {
   /**
    * Check if a file is a dendron note (vs a regular file or something else entirely)
    */
-  static isDendronNote({
+  static async isDendronNote({
     wsRoot,
     vaults,
     fpath,
@@ -121,7 +121,7 @@ export class WorkspaceUtils {
 
     // if markdown file, does it have frontmatter?
     // this is a rough heuristic
-    return fs.readFileSync(fpath, { encoding: "utf8" }).startsWith("---");
+    return (await fs.readFile(fpath, { encoding: "utf8" })).startsWith("---");
   }
 
   static isNativeWorkspace(workspace: DWorkspaceV2) {
--- a/packages/engine-server/src/workspace/utils.ts
+++ b/packages/engine-server/src/workspace/utils.ts
@@ -16,6 +16,7 @@ import {
   WorkspaceType,
 } from "@dendronhq/common-all";
 import {
+  FileUtils,
   findDownTo,
   findUpTo,
   genHash,
@@ -109,7 +110,7 @@ export class WorkspaceUtils {
     wsRoot,
     vaults,
     fpath,
-  }: { fpath: string } & WorkspaceOpts) {
+  }: { fpath: string } & WorkspaceOpts): Promise<boolean> {
     // check if we have markdown file
     if (!fpath.endsWith(".md")) {
       return false;
@@ -119,9 +120,10 @@ export class WorkspaceUtils {
       return false;
     }
 
-    // if markdown file, does it have frontmatter?
-    // this is a rough heuristic
-    return (await fs.readFile(fpath, { encoding: "utf8" })).startsWith("---");
+    // if markdown file, does it have frontmatter? check for `---` at beginning of file
+    return (
+      (await FileUtils.matchFilePrefix({ fpath, prefix: "---" })).data || false
+    );
   }
 
   static isNativeWorkspace(workspace: DWorkspaceV2) {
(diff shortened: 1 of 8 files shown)


packages/plugin-core/src/dendronExtensionInterface.ts
```diff
@@ -70,6 +72,8 @@ export interface IDendronExtension {
   noteLookupProviderFactory: INoteLookupProviderFactory;
   schemaLookupProviderFactory: ISchemaLookupProviderFactory;
 
+  activateWatchers(): Promise<void>;
+  deactivate(): Promise<void>;
```
jonathanyeung, line 76: Let's make it clear in the docstring that this method is meant to deactivate the entire extension, not just the watchers.  The way it's sandwiched between `activateWatchers` and `pauseWatchers` is a little bit misleading.

packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -104,6 +104,13 @@ export class WorkspaceWatcher {
     this._windowWatcher = windowWatcher;
   }
 
+  // eslint-disable-next-line camelcase
+  __DO_NOT_USE_IN_PROD_exposePropsForTesting() {
+    return {
+      onFirstOpen: _.bind(this.onFirstOpen, this),
```
jonathanyeung, line 110: Given that this is pseudo-public now (at least via testing), can you add a docstring to `onFirstOpen` on what the return boolean value represents?  It's not very clear what the ret value should be from a function called "onFirstOpen".  It should also help a reader to understand what the test is supposed to do.

packages/engine-server/src/workspace/utils.ts
```diff
@@ -102,6 +102,28 @@ export class WorkspaceUtils {
...
+      return false;
+    }
+    // if markdown file, check if it is in a dendron vault
+    if (!WorkspaceUtils.isPathInWorkspace({ wsRoot, vaults, fpath })) {
+      return false;
+    }
+
+    // if markdown file, does it have frontmatter?
+    // this is a rough heuristic
+    return fs.readFileSync(fpath, { encoding: "utf8" }).startsWith("---");
```
jonathanyeung, line 124: should we prefer async versions for perf reasons?

jonathanyeung (changes requested):

jonathanyeung:
> ad-hoc dependency injection: resulted in too many code changes, we're better off actually implementing DI properly using one of the frameworks: https://blog.logrocket.com/top-five-typescript-dependency-injection-containers but out of scope for this task

Yeah, I'm for this - but obviously, for it to work it'll take a lot more refactoring of our existing classes.  Generally I think DI libraries help once your injected objects become more complex (multi-tier injections).  Where we're at today with the circ dep refactoring I haven't had to inject a multi-tiered object yet, but when we do I think it'll make sense to adopt a library such as TSyringe.

packages/engine-server/src/workspace/utils.ts
kevinslin: changed it to `async`. something I do worry about is that its not immediately clear when functions are sync/async and we already have places in the code where we will either `await` a non-async function or fail to `await` an `async function`.

we should look into either:
- adding a convention (eg. isDendronNoteAsync) to be explicit about the return
- add stricter linting rules about asyncs

packages/plugin-core/src/test/suite-integ/WorkspaceWatcher.test.ts
```diff
@@ -208,8 +215,55 @@ suite("WorkspaceWatcher: GIVEN the dendron extension is running", function () {
...
+          const { engine, vaults, wsRoot } = ExtensionProvider.getDWorkspace();
+          await FileTestUtils.createFiles(wsRoot, [{ path: "sample" }]);
+          const wsutils = new WSUtilsV2(ext);
+          const notePath = path.join(wsRoot, "sample");
+          const editor = await VSCodeUtils.openFileInEditor(
+            vscode.Uri.file(notePath)
+          );
+          const { onFirstOpen } =
+            UNSAFE_getWorkspaceWatcherPropsForTesting(ext);
+          expect(await onFirstOpen(editor)).toBeFalsy();
```
SeriousBug, line 242: minor nitpick, but this could also check that `foldFrontmatter` is not called since the main issue we're trying to fix is that Dendron tries to fold the frontmatter in non-note files. Or that could go into a different test if we want to keep the concerns of each test separate.

packages/engine-server/src/workspace/utils.ts
```diff
@@ -102,6 +102,28 @@ export class WorkspaceUtils {
...
+      return false;
+    }
+    // if markdown file, check if it is in a dendron vault
+    if (!WorkspaceUtils.isPathInWorkspace({ wsRoot, vaults, fpath })) {
+      return false;
+    }
+
+    // if markdown file, does it have frontmatter?
+    // this is a rough heuristic
+    return (await fs.readFile(fpath, { encoding: "utf8" })).startsWith("---");
```
SeriousBug, line 124: How about reading only the first part of the file here for further optimization in very large files? Looking it up a bit, looks like it is possible to read only some of the file using read streams. For example: https://stackoverflow.com/questions/70707646/reading-part-of-file-in-node

SeriousBug (approved):
Left a comment about a possible improvement.

packages/engine-server/src/workspace/utils.ts
jonathanyeung: I'm for an async convention, but additionally, how about explicitly stating its return value? `isDendronNote(...): Promise<string> {}`

jonathanyeung (approved):

packages/engine-server/src/workspace/utils.ts
kevinslin: updated. nice tip!

packages/plugin-core/src/test/suite-integ/WorkspaceWatcher.test.ts
kevinslin: yeah, that would be a better test. unfortunately not sure how to do this since in order to spy on a method, you need to either 1) pass in an object and the method or 2) spy on the method and inject the method

1) isn't feasible because the method is private so we can't spy on the object
2) isn't feasible because we don't have dependency inejction so we can't make `onFirstOpen` call a spied method

https://sinonjs.org/releases/latest/spies/

Issue #1710, opened 2021-11-14
Expose functionality for renaming/moving notes through the CLI


benhsm:
## Please select if your request is either something new or an enhancement ##

- [ ] Enhancement of an existing Feature.
- [x] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are possible) ##

- [ ] Onboard. Initial vault setup and import/export pods
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

Currently, the [Dendron CLI](https://wiki.dendron.so/notes/23a1b942-99af-45c8-8116-4f4bb7dccd21/) does not have exposed functionality for renaming/moving existing notes.

This means that CLI users will have to resort to opening VSCode to change where a note is in a hierarchy / what its name is while preserving internal references to it.

## Describe the solution you'd like ##

I would want to be able to issue a command like:
```
dendron note (move|rename|mv) --wsRoot <dendron workspace> --query <note name> --newName <new name> [--fromVault <vault> --toVault <vault>]
```

Which would have the same effects as running [Dendron: rename note](https://wiki.dendron.so/notes/eea2b078-1acc-4071-a14e-18299fc28f47/#rename-note) or [Dendron: move note](https://wiki.dendron.so/notes/eea2b078-1acc-4071-a14e-18299fc28f47/#move-note) in the VSCode extension.

## Describe alternatives you've considered ##

I'm considering writing a shell script that would parse notes in a Dendron vault for references to a note to be renamed/moved and alter them accordingly along with the target note. This seems unnecessary when the logic for this is already in the engine, and it also wouldn't be as integrated with the rest of Dendron functionality.

## Additional context ##

I mainly want this for purposes of using Dendron through scripts and editors like Vim and Neovim without having to use VSCode. I'd like to eventually be able to do most of what is currently possible in the VSCode interface from the terminal, scripts and other editors, and I gather that this is already part of the long-term vision for Dendron. Searching around, I found one prior issue [#291](https://github.com/dendronhq/dendron/issues/291#issue-727622541) that this is possibly subsumed under.

The functionality for refactoring whole hierarchies is possibly related enough to be tackled alongside this.

[...]

PR #2408, opened 2022-02-13
feat(cli): Add rename functionality to CLI

benhsm:
# Community PR Review Checklist

## First Time Specifics

- [X] if its your first pull request to Dendron, watch out for the [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement) bot that will ask you to agree to Dendron's CLA
- [] if its your first pull request and you're on our Discord, make sure that Kevin gives you the [horticulturalist](https://wiki.dendron.so/notes/7c00d606-7b75-4d28-b563-d75f33f8e0d7.html#horticulturalist) role by adding your discord as part of the description  👨‍🌾👩‍🌾
- [X] (optional) ping `@Dendron Team` in the `#dev` channel of our discord - we usually respond to PRs within 24h

## Commit

- [X] make sure your branch names adhere to our [branch style](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e.html#branch-style)
- [X] make sure the commit message follows Dendron's [commit style](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e.html#commit-style)
- [ ] if this pull request is addressing an existing issue, make sure to link this PR to the issue that it is resolving.
  - NOTE: make sure to read [Addressing open issues](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e.html#addressing-open-issues)

## Code

- [X] code should follow [Code Conventions](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e.html#branch-style) -> It does, I think?
- [ ] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e.html#commit-style).
- [X] sticking to existing conventions instead of creating new ones

## Tests

- [X] [Write Tests](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#writing-tests) 
- [X] [Confirm existing tests pass](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#executing-tests)
- [x] Common cases tested
- [ ] 1-2 Edge cases tested
- [ ] If your tests changes an existing snapshot, snapshots have been [updated](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html#updating-test-snapshots)

## Docs
- [ ] if your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR -> Will need to do this later
- [ ] does this change introduce a new or better way of doing things that others need to be aware of? if so, an async should be created and a process added in [Development](https://docs.dendron.so/notes/3489b652-cd0e-4ac8-a734-08094dc043eb.html) or [Packages](https://docs.dendron.so/notes/32cdd4aa-d9f6-4582-8d0c-07f64a00299b.html)

## Comments

Related to #1710 

Submitting this as a draft request because I've been working on this for a while and I think I'm quite close; I'm just having some trouble with testing. 

I'm not completely sure what the expected behaviour of the rename command should be; what I'm assuming is that a new note should be created, and the old one should be deleted. Right now, either 1) I don't know how to check that the note is deleted, or 2) My implementation isn't deleting the old note properly. I haven't figured out how to do manual testing using the CLI yet.


packages/dendron-cli/src/commands/notes.ts
```diff
@@ -18,6 +18,8 @@ type CommandCLIOpts = {
   query?: string;
   cmd: NoteCommands;
   output?: NoteCLIOutput;
+  newName?: string;
```
kevinslin, line 21: instead of `newName` or `newVault`, would use the following:
- `destFname`
- `destVaultName`

Rationale behind the proposal:
- we already have a convention for using `dest` as the target in some other cli commands
- we prefer options to be more specific (eg. vault is ambiguous, vault name makes it clear what we are looking for)

packages/dendron-cli/src/commands/notes.ts
```diff
@@ -76,6 +80,14 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
       choices: Object.values(NoteCLIOutput),
       default: NoteCLIOutput.JSON,
     });
+    args.option("newName", {
```
kevinslin, line 83: see above comment about naming

packages/dendron-cli/src/commands/notes.ts
```diff
@@ -145,6 +157,23 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
           this.print(`deleted ${note.fname}`);
           return { data: { payload: note.fname, rawData: resp } };
         }
+        case NoteCommands.MOVE:
+        case NoteCommands.RENAME: {
```
kevinslin, line 161: semantically, rename is a convenience function on top of `move` that executes `move` while keeping the vault name the same. this CLI argument doesn't actually do that (you can change the vault name using a flag). 
I would recommend simplifying this PR to just add the `move` command since its a superset of the functionality (and because we prefer for a PR to introduce one change)

packages/dendron-cli/src/commands/notes.ts
```diff
@@ -145,6 +157,23 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+          const { query, vault } = checkQueryAndVault(opts);
+          const note = NoteUtils.getNoteOrThrow({
+            fname: query,
```
kevinslin, line 164: semantically, `query` is meant to be a lookup query which means that it doesn't have to match the path exactly. the way its being used here, we are actually looking for an exact match. 
in order to be explicit, I would use a new argument here  (eg. `srcFname`). this also means we want to remove the `checkQueryAndVault` check and create new function `eg. checkArgsForMoveCommand`

kevinslin (commented):
heads up that @joshi has a docs PR for being able to debug tests using vscode here -> https://github.com/dendronhq/dendron-docs/pull/17/files

kevinslin (approved):

kevinslin:
this looks good - can you also add documentation for this in https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.topic.cli.notes.md#L8:L8?

chmac:
It doesn't look like the docs for this have landed. As per #2727, I think it's a real tragedy to have awesome new features that users (myself included!) are potentially eager to try, but there's no way to easily figure out how they work. 😢

benhsm:
@chmac there's a couple examples in the docs here: <https://wiki.dendron.so/notes/xI97d5GKk7qU6EwstIAXe/#move>
As well as output you get from running `dendron note --help`, but yeah, not much more than that at the moment.

PR #2429, opened 2022-02-15
fix(schema): Apply schema template for goto-note-command if template is in different vault

tma66:
**Background**
Previously, when we run Goto note command, we would only apply a template if it was in same vault as the note. Now we will apply the first template match. If there are multiple template matches across multiple vaults, pick first one to apply. This is temporary until we support xvault support for this command.

**Testing**
Verified manually and added new test case

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -444,12 +444,23 @@ export class NoteUtils {
       const maybeTemplate = schema.data.template;
       if (maybeTemplate) {
         // TODO: Support xvault with user prompting for this flow
-        // Get first valid template note. If there are multiple template matches, apply first one.
-        // This is temp until we get xvault support
-        const tempNote = NoteUtils.getNotesByFnameFromEngine({
+        /*
+         * Get first valid template note.
+         * First look for template in same vault as note. Otherwise, look across all vaults.
+         * If there are multiple template matches, apply first one.
+         * This is temp until we get xvault support
+         */
+        const tempInSameVault = NoteUtils.getNoteByFnameFromEngine({
           fname: maybeTemplate.id,
+          vault: note.vault,
           engine,
-        })[0];
+        });
+        const tempNote =
+          tempInSameVault ||
+          NoteUtils.getNotesByFnameFromEngine({
+            fname: maybeTemplate.id,
+            engine,
+          })[0];
 
         if (tempNote) {
           SchemaUtils.applyTemplate({
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -444,11 +444,12 @@ export class NoteUtils {
       const maybeTemplate = schema.data.template;
       if (maybeTemplate) {
         // TODO: Support xvault with user prompting for this flow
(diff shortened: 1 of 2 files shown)


packages/common-all/src/dnode.ts
```diff
@@ -444,11 +444,12 @@ export class NoteUtils {
...
         // TODO: Support xvault with user prompting for this flow
-        const tempNote = NoteUtils.getNoteByFnameFromEngine({
+        // Get first valid template note. If there are multiple template matches, apply first one.
+        // This is temp until we get xvault support
+        const tempNote = NoteUtils.getNotesByFnameFromEngine({
           fname: maybeTemplate.id,
-          vault: note.vault,
           engine,
-        });
+        })[0];
```
jonathanyeung, line 452: How are the results sorted?  Should we check whether any of the results returned have a vault that matches the current `note.vault`, and if so return that one?

packages/plugin-core/src/test/suite-integ/GotoNote.test.ts
```diff
@@ -131,6 +131,29 @@ suite("GotoNote", function () {
       });
     });
 
+    test("go to new note with template in a different vault", (done) => {
+      runLegacyMultiWorkspaceTest({
```
SeriousBug, line 135: Recommend avoiding the deprecated harness going forward, even if the other tests are using it. Refactoring this for the new harness would look like this:
```ts
    describeMultiWS(
      "GIVEN a new note and a template in different vaults",
      {
        ctx,
        preSetupHook: ENGINE_HOOKS_MULTI.setupBasicMulti,
        postSetupHook: async ({ wsRoot, vaults }) => {
          await ENGINE_HOOKS.setupSchemaPreseet({ wsRoot, vaults });
        },
      },
      () => {
        test("THEN new note uses that template", async () => {
          // Note is in different vault from template
          const { vaults } = ExtensionProvider.getDWorkspace();
          const vault = vaults[1];
          await createGoToNoteCmd().run({
            qs: "bar.ch1",
            vault,
          });
          expect(getActiveEditorBasename()).toEqual("bar.ch1.md");
          const content =
            VSCodeUtils.getActiveTextEditor()?.document.getText() as string;
          expect(content.indexOf("ch1 template") >= 0).toBeTruthy();
        });
      }
    );
```

tma66: Sounds good, thanks for suggestion

packages/common-all/src/dnode.ts
tma66: Internally notes are stored in a map so I don't believe there is an order (unless js does things differently or we use a special map). Makes sense to find one that matches current vault though.

jonathanyeung (approved):

kevinslin (approved):

SeriousBug (approved):

PR #2452, opened 2022-02-18
chore/master branch fixups

kevinslin:
This started as an investigation to fix the weekly release and evolved into a bigger cleanup. This commit touches a lot of files that come from the following changes:

- we never deleted `yarn.lock` when adding dependencies. this has caused us not to pick up new versions via semver and old dependencies to pile up. as part of this change, we've cleared yarn.lock to generate from scratch which removed as well as upgraded many existing dependencies
- as part of the dependency upgrades, `comment-json` changed a type definition which led to a refactoring of all methods that relied on `comment-json`
- due to above migration, found a series of potential runtime bugs due to liberal casting which have now been patched 
- found bug where `CreateJournalNote` would error out when not in dendron workspace (and prevent subsequent commands from being initialized)

***
Circular Dependencies:

- before: 50
- after: 50

***

Changes in this commit:

- activation
    - don't register `CreateDailyJournal` when not in active workspace 
    - split command registration to commands that can be registered without active workspace and those that require an active workspace

- dependencies:
    - wiped the yarn.lock and rebuild dependencies from current package.json
    - excluded conflicting types from being hoisted to the workspace
    - lock nextjs-sitemap to "1.8.11" (next version drops support for nodejs 12)

- filev2
    - update return type of `readJSONWithCommentsSync`
    - export return types from `comment-json` package

- WorkspaceUtils 
    - get/set/update workspace settings
    - workspace settings renamed for clarity
    - runtime + type checks to see if workspace exists

- KeyboardUtils 
    - can get/set keyboard settings

- Migrations
    - checking if workspaceSettings is missing before running migrations
    - comment out `checkAndMigrateLookupKeybindingIfExists` 
        - had some errors but instead of fixing, decided to comment out for following reasons
            - this is now a dozen versions old
            - keyboard migrations are flaky and we don't have telemetry about how this is being applied
            - we don't have sentry tuned to detect for this
    - don't apply keybinding fix on upgrade, comment out instead
        - same reasoning as above

- misc
    - remove `updateWorkspace` method in `_extension.ts`, replaced with `WorkspaceUtils.update` method

***

### Basics

### Extended

- General
- Operations
- Architecture

## Instrumentation

NA

## 

### Basics

- [.] [Confirm manual testing](dev.process.qa.test) 
- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

- [.] does this change introduce a new or better way of doing things that others need to be aware of? if so, an async should be created and a process added in [Development](dev) or [Packages](pkg)

## 

### Extended

- [.]  is this a developer BREAKING change? if another person cloning from this branch will need to adjust their dependencies or mental model of the architecture, then it is. if this is the case, make sure this is communicated according to [Close Loop](dev.process.close-loop)

Some pending next items:
- [ ] update instructions for removing dependencies (need to reset yarn.lock, update dependencies, make sure any upgrades haven't broken the build)
- [ ] start a pinned depndency file (eg.  nextjs-sitemap has to be "1.8.11", next version doesn't support node 12 )
- [ ] update on registering commands (distinction btw  workspace active commands that require workspace service vs regular commands)


packages/plugin-core/src/commands/base.ts
```diff
@@ -50,6 +51,11 @@ export abstract class BaseCommand<
 
   static showInput = window.showInputBox;
 
+  /**
+   * Does this command require an active workspace in order to function
+   */
+  static requireActiveWorkspace = false;
```
SeriousBug, line 57: I think this is fine for this PR, but I think we should eventually switch this to true. A lot of our code relies on an active workspace existing, so it would be safer to only set this to false for commands we know are safe without a workspace.

SeriousBug (approved):
Looks good to me, just one comment about changing the default for `requireActiveWorkspace` in the future.

Issue #2306, opened 2022-01-28
Clicking on local pdf link in preview doesn't open it


Gorzen:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [x] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

I have a pdf in `assets` folder. In one note, I link the pdf like this : `[Article title](./assets/article.pdf)`. If I click on the link in the preview nothing happens. I would expect it to open the PDF in my default app or in VSCode.

Note: If I do `cmd+click` on the editing pane, it opens the pdf in VSCode.

## To Reproduce

1. Add link to a local pdf in a note
2. Open preview
3. Click on the link to pdf
4. Witness that nothing happens

## Expected behavior

The PDF should open in default app or VSCode in editing pane.

## Desktop (please complete the following information)

- OS: macOS 12.1

SeriousBug:
Thanks for the report! I added it to our backlog and will make sure to prioritize this within the next few weeks.

SeriousBug:
I believe this is a duplicate of #1695.

Issue #1695, opened 2021-11-11
Clicking links to pasted files in Dendron Preview doesn't work


sidwellr:
## Please select the area your bug applies to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods
- [ ] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [X] Misc (Choose this if your not sure)

## Describe the bug

Pasting pdf files (and other types too; I've replicated this with and odg file) into a Dendron note displays them properly as links in the Dendron Preview, but nothing happens when you click them.

## To Reproduce

1. Create a scratch note ((shift-control-S)
2. Execute Dendron: Paste File, type path to a pdf file, Enter
3. The file is copied to the Assets directory and a link is added to the note: [test.pdf](assets\test.pdf)
4. Execute Dendron: Show Preview; the preview appears containing a colored link "test.pdf"
5. Click the link. Nothing happens.

## Expected behavior

When clicking the link, I expected the pasted file to open in Acrobat Reader, the program I've configured on my system to open pdf files. The same thing happened when I tried an odg file; I expected LibreOffice Draw to open the file. Both of these work as expected when I double-click the file in a File Explorer window.

## Desktop (please complete the following information)

- OS: Windows 10 Home, version 20H2, build 19042.1348
- Browser: not involved here, but Firefox is my default browser
- Version: vscodium 1.62.1, dendron 0.68.1

## Additional context

The Dendron Runtime Status shows two uncaught errors, both "Cannot read property 'fname' of undefined".

## Dendron Log file

{"level":30,"time":1636655285925,"pid":12256,"hostname":"Threadripper","name":"dendron","ctx":"Logger:configure","msg":"exit","logLevel":"info"}
[...]

PR #2455, opened 2022-02-19
fix(views): clicking preview links for non-note files

SeriousBug:
This PR improves the handling of assets and other non-note files when clicking on links in the preview.
- Dendron can now detect text files vs binary files. Text files are opened inside VSCode, while binary files are opened with the default app.
- Dendron can find asset files inside the `assets` folders of vaults, relative to workspace root, or as absolute paths.

The PR also simplifies some of the code by using `vscode.Uri` to parse the code instead of ad-hoc parsing code, and adds a more comprehensive test suite.

May help #2306 and #1695

## Dependency Callout

enhance(common.views): Added a dependency to detect if a file extension is a text file or not.

- DEPENDENCY_UPDATE: @dendronhq/plugin-core add textextensions@^5.14.0

The dependency is simply a list of file extensions that are well known to be text files. This is not the best way to detect if a file is a text file or not (the "ideal" way involves also reading the file to check encoding and look for file signatures), but since those dependencies are heavier I chose to go with this since it's a relatively lightweight and simple addition.

### Basics

### Extended

- General
    - Clicking links now requires more work, but it is an infrequent enough operation and the increase in workload is a small, constant amount.
- Operations

### Basics

  - Windows and Linux
- [x] Common cases tested
- [x] 1-2 Edge cases tested


packages/plugin-core/src/components/views/PreviewLinkHandler.ts
```diff
@@ -50,103 +58,96 @@ export class PreviewLinkHandler implements IPreviewLinkHandler {
...
-      href &&
-      (href.startsWith("http://") || href.startsWith("https://"))
-    ) {
-      return LinkType.WEBSITE;
-    } else {
-      return LinkType.UNKNOWN;
-    }
-  }
-
+  /** Returns a note if one was found, undefined if no notes were found, and null if the link was ambiguous and user cancelled the prompt to pick a note. */
```
jonathanyeung, line 140: nit: comment runs long. 

As an aside, I like this vs code ext which can help auto-flow comments to fit inside a specified length: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

packages/plugin-core/src/test/suite-integ/PreviewLinkHandler.test.ts
```diff
@@ -1,247 +1,416 @@
...
+          });
+        });
 
-          await factory.createForFName("foo");
+        describe("AND the link is ambiguous", () => {
+          test("THEN it prompts for a note", async () => {
+            const { engine } = ExtensionProvider.getDWorkspace();
+            const showChooseNote = sinon
+              .stub(QuickPickUtil, "showChooseNote")
+              .returns(Promise.resolve(engine.notes["reprehenderit"]));
```
jonathanyeung, line 173: can we store these `reprehenderit` strings into a const variable and reuse?

jonathanyeung (changes requested):
Good job with the tests!

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -233,7 +233,7 @@ export class GotoNoteCommand extends BasicCommand<
     return opts;
   }
 
-  private async trySelectRevealNonNoteAnchor(
+  public static async trySelectRevealNonNoteAnchor(
```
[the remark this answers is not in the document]
SeriousBug: This is also used in `PreviewLinkHandler.ts`

packages/plugin-core/src/components/views/PreviewLinkHandler.ts
```diff
@@ -50,103 +58,96 @@ export class PreviewLinkHandler implements IPreviewLinkHandler {
...
-            // different vault which had ambiguous vault and upon being prompted
-            // to select a note in quick user cancelled.
-            return;
-          }
-
-          this._ext.commandFactory.goToNoteCmd().execute({
-            qs: noteData.note.fname,
-            vault: noteData.note.vault,
+    } catch (err) {
+      Logger.error({ ctx, error: ErrorFactory.wrapIfNeeded(err) });
```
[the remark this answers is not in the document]
SeriousBug: `getNavigationTargetNoteForWikiLink` can throw if it fails to read the note id from the link.

I think `GotoNote` can also throw if the note doesn't exist, because it grabs a note from ID then casts it:
```ts
const note = data?.note as NoteProps;
```

I could refactor both functions to avoid throwing, but I also like having the try block as a fallback. Attempting to process it as an asset is a safe operation because we first check if the file actually exists.

packages/plugin-core/src/components/views/PreviewLinkHandler.ts
SeriousBug: Thanks for the suggestion, I do have it installed actually! It was something I missed a lot moving from Emacs to VSCode.

packages/plugin-core/src/test/suite-integ/PreviewLinkHandler.test.ts
SeriousBug: Do you mean just adding `const showChooseNoteTargetId = "reprehenderit";` to the top of this test file? It is only used here and when creating the note, and the test would fail if someone changed something so we couldn't find the note. I don't think there's really anything to reuse, I'd rather leave this as-is.

jonathanyeung: It's more of a nit, but yes - just for the point of consolidating it between the note creation step and the validation step.  It's more for readability instead of reusability - it makes it a little bit easier to understand what we're trying to verify.

SeriousBug: Got it, I'll go ahead and refactor that then. I can actually grab the entire note, so it just uses the right note instead of looking it up.

packages/plugin-core/src/components/views/PreviewLinkHandler.ts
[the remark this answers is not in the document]
jonathanyeung: I guess to me - either we have an expected error (expected meaning valid code logic but maybe invalid user behavior), in which case it makes to continue, or else it's an unexpected error and we should stop further execution in the function at this point. 

If it's expected, then let's not log it as `Logger.error` (maybe Logger.debug), as this will cause uploads to Sentry, and I'd rather err on less data than more data right now because it's still too noisy.

packages/plugin-core/src/commands/GotoNote.ts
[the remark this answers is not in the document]
jonathanyeung: I see - is it possible to move this to a different class / file then?  GoToNote is involved in a lot of remaining circular dependencies; I'm a bit weary of having more static methods on commands.

packages/plugin-core/package.json
```diff
@@ -1115,6 +1115,7 @@
     "replace-in-file": "^6.2.0",
     "semver": "^7.3.2",
     "string-similarity": "^4.0.4",
+    "textextensions": "^5.14.0",
```
kevinslin, line 1118: can we do a callout in the PR when we add dependencies?

just created an async around it in -> https://github.com/dendronhq/dendron-docs/commit/4998410128feefdf59d108577b100fe88c11c55f

packages/plugin-core/src/commands/GotoNote.ts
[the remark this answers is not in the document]
SeriousBug: Got it, moved this to `WSUtilsV2` instead.

jonathanyeung (approved):

PR #2457, opened 2022-02-20
fix(workspace): dendron can hang when trying to provide hover for large non-dendron file

kevinslin:
fix(workspace): dendron can hang when trying to provide hover for large non-dendron file

This currently happens because `ReferenceHoverProvider` can be triggered for all files. This function calls `getReferenceAtPosition` which will attempt to parse the file using `unified` to find hover positions. This will blow up spectacularly when parsing large files (eg. Dendron logs) and cause the vscode process to hang.

This fix adds a check in `getReferenceAtPosition` to not parse the file if it is not a dendron note or in a dendron workspace. It also provides this check across most of the symbol provider methods in Dendron. List of changes below:

Changes:
- convert `getReferenceAtPosition` to async, do not run for non-dendron note
- ReferenceHoverProvider restrict to dendron note (cause of hanging when hovering over log file)
- ReferenceProvider restrict to dendron note (was causing silent exception)
- updated signatures in the following files which called `getReferenceAtPosition`
  - [[../packages/plugin-core/src/commands/ConvertLink.ts]]
  - [[../packages/plugin-core/src/commands/GotoNote.ts]]
  - [[../packages/plugin-core/src/features/DefinitionProvider.ts]]
  - [[../packages/plugin-core/src/features/RenameProvider.ts]]

New Utilities:
- `Awaited` utility, unwraps `promise` type from type signature
    - loc: packages/common-all/src/types/typesv2.ts
- `extractFMTags`
    - loc: packages/engine-server/src/markdown/remark/utils.ts
- `isActiveAndIsDendronNote`
    - loc: packages/plugin-core/src/ExtensionProvider.ts

***
- before: 47
- after: 49

This change introduces the following circular dependencies. This is caused by:
- [[../packages/plugin-core/src/dendronExtensionInterface.ts]] > [[../packages/plugin-core/src/WorkspaceWatcher.ts]]

To move forward, we should create a `WorkspaceWatcher` interface. Would move this as a next item since this is a high severity fix

```
9) src/ExtensionProvider.ts > src/dendronExtensionInterface.ts > src/WorkspaceWatcher.ts > src/fileWatcher.ts > src/workspace.ts > src/commandFactory.ts > src/commands/GotoNote.ts

23) src/ExtensionProvider.ts > src/dendronExtensionInterface.ts > src/WorkspaceWatcher.ts > src/fileWatcher.ts > src/workspace.ts > src/commandFactory.ts > src/commands/GotoNote.ts > src/utils/md.ts
```

***

### Basics

- [.] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices).

### Extended

- General
- Operations
- Architecture

## Instrumentation
NA

### Basics

- [.] [Confirm existing tests pass](dev.process.qa.test)
- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

- [.] does this change introduce a new or better way of doing things that others need to be aware of? if so, an async should be created and a process added in [Development](dev) or [Packages](pkg)

add new utilities, will be added to docs once this pr is approved

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/ConvertLink.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/ConvertLink.test.ts
@@ -10,10 +10,7 @@ import sinon from "sinon";
 import vscode, { TextEditor } from "vscode";
 import { ConvertLinkCommand } from "../../commands/ConvertLink";
 import { ExtensionProvider } from "../../ExtensionProvider";
-import {
-  getReferenceAtPosition,
-  getReferenceAtPositionResp,
-} from "../../utils/md";
+import { getReferenceAtPosition } from "../../utils/md";
 import { VSCodeUtils } from "../../vsCodeUtils";
 import { WSUtils } from "../../WSUtils";
 import { expect } from "../testUtilsv2";
@@ -33,7 +30,10 @@ const getReference = async ({
     wsRoot,
     vaults,
   });
-  return out as getReferenceAtPositionResp;
+  if (!out) {
+    throw new Error("ref should be truthy");
+  }
+  return out;
 };
 
 suite("ConvertLink", function () {
--- a/packages/plugin-core/src/test/suite-integ/ConvertLink.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/ConvertLink.test.ts
@@ -5,10 +5,11 @@ import {
 } from "@dendronhq/common-test-utils";
 import { LinkUtils, ParseLinkV2Resp } from "@dendronhq/engine-server";
 import _ from "lodash";
-import { beforeEach, afterEach } from "mocha";
+import { afterEach, beforeEach } from "mocha";
 import sinon from "sinon";
-import vscode from "vscode";
+import vscode, { TextEditor } from "vscode";
 import { ConvertLinkCommand } from "../../commands/ConvertLink";
+import { ExtensionProvider } from "../../ExtensionProvider";
 import {
   getReferenceAtPosition,
   getReferenceAtPositionResp,
@@ -17,6 +18,23 @@ import { VSCodeUtils } from "../../vsCodeUtils";
 import { WSUtils } from "../../WSUtils";
 import { expect } from "../testUtilsv2";
 import { describeMultiWS, setupBeforeAfter } from "../testUtilsV3";
+
+const getReference = async ({
+  editor,
+  position,
+}: {
+  editor: TextEditor;
+  position: vscode.Position;
+}) => {
+  const { wsRoot, vaults } = ExtensionProvider.getDWorkspace();
+  const out = await getReferenceAtPosition({
+    document: editor.document,
+    position,
+    wsRoot,
+    vaults,
+  });
+  return out as getReferenceAtPositionResp;
+};
 
 suite("ConvertLink", function () {
   const ctx = setupBeforeAfter(this, {
@@ -71,10 +89,10 @@ suite("ConvertLink", function () {
       test("WHEN broken link with no alias, THEN doesn't show option to use alias text.", async () => {
         const editor = vscode.window.activeTextEditor as vscode.TextEditor;
         const cmd = new ConvertLinkCommand();
-        const reference = getReferenceAtPosition(
-          editor.document,
-          noAliasBrokenLinkPosition
-        ) as getReferenceAtPositionResp;
+        const reference = await getReference({
+          editor,
+          position: noAliasBrokenLinkPosition,
+        });
         const { options, parsedLink } =
           cmd.prepareBrokenLinkConvertOptions(reference);
         expect(parsedLink.alias).toBeFalsy();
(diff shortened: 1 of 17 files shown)


jonathanyeung:
Re: circular deps - i think it's fine.  Already tracking the workspace watcher circ dep via [[Circ Dep Workspace Watcher|dendron://private/task.dev.2022.01.11.circ-dep-workspace-watcher]].  These should be a dupe to the existing cycles.

jonathanyeung (changes requested):

packages/plugin-core/src/test/suite-integ/ReferenceHoverProvider.test.ts
```diff
@@ -0,0 +1,1252 @@
...
+async function provideForNonNote(editor: vscode.TextEditor) {
+  return provide({ editor, pos: new vscode.Position(0, 2) });
+}
+
+suite("GIVEN ReferenceHoverProvider", function () {
+  const ctx = setupBeforeAfter(this);
+
+  describe("AND GIVEN non-dendron file", () => {
+    describeMultiWS(
+      "AND WHEN try to find references",
```
SeriousBug, line 50: nitpick: I think the messages here could be clearer, like "WHEN used in a non-dendron file", "AND the file is empty", "THEN doesn't provide for this file", so that it reads like a sentence when put together.

packages/plugin-core/src/utils/md.ts
```diff
@@ -237,11 +235,22 @@ export type getReferenceAtPositionResp = {
   refText: string;
 };
 
-export const getReferenceAtPosition = (
-  document: vscode.TextDocument,
-  position: vscode.Position,
-  opts?: { partial?: boolean; allowInCodeBlocks: boolean }
-): getReferenceAtPositionResp | null => {
+export const getReferenceAtPosition = async ({
```
SeriousBug, line 238: could we use a named function here like `export async function getReferenceAtPosition` so the name will show up in backtraces?

packages/plugin-core/src/utils/md.ts
```diff
@@ -313,18 +325,19 @@ export const getReferenceAtPosition = (
...
+    if (
+      !(await WorkspaceUtils.isDendronNote({
+        wsRoot,
+        vaults,
+        fpath: document.uri.fsPath,
+      }))
+    ) {
+      return null;
+    }
+    const maybeTags = RemarkUtils.extractFMTags(document.getText());
```
SeriousBug, line 338: No need to add block this PR for it, but we could also limit how much of the text we grab here, like `document.getText(new vscode.Range(0, 0, 300, 0))`. This could cause confusion if someone does legitimately write a note that has more than 300 lines of frontmatter, but it would also help in actual notes that are very long.

packages/plugin-core/src/test/suite-integ/ConvertLink.test.ts
```diff
@@ -18,6 +19,23 @@ import { WSUtils } from "../../WSUtils";
...
+  position: vscode.Position;
+}) => {
+  const { wsRoot, vaults } = ExtensionProvider.getDWorkspace();
+  const out = await getReferenceAtPosition({
+    document: editor.document,
+    position,
+    wsRoot,
+    vaults,
+  });
+  return out as getReferenceAtPositionResp;
```
SeriousBug, line 36: It might be good to add a `expect(out).toBeTruthy()` here so the test will fail early if there's an issue.

SeriousBug (commented):
Left some comments. The fix looks good to me.

packages/plugin-core/src/utils/md.ts
kevinslin: updated. and good point. can you write an async about this and update [[Best Practices|dendron://dendron.docs/dev.process.code.best-practices]]?

packages/plugin-core/src/utils/md.ts
kevinslin: in terms of parsing text, vscode and a lot of editors have a hard limit of what they're willing to parse (eg. don't pass if file is larger than 1MB). we should do something similar and have a metric for how many times we hit that. will add as a next item

packages/plugin-core/src/test/suite-integ/ConvertLink.test.ts
kevinslin: added a check

packages/plugin-core/src/workspace.ts
```diff
@@ -304,6 +304,10 @@ export class DendronExtension implements IDendronExtension {
     return false;
   }
 
+  isActiveAndIsDendronNote(_fpath: string): boolean {
+    throw new Error("not implemented");
```
jonathanyeung, line 308: So to follow through - let's move the logic from the current impl in `ExtensionProvider` to here, and then replace the impl in ExtensionProvider with 

```typescript
  static isActiveAndIsDendronNote(fpath: string) {
    return ExtensionProvider.getExtension().isActiveAndIsDendronNote(fPath);
  }
```

jonathanyeung (changes requested):

jonathanyeung (approved):

PR #2458, opened 2022-02-21
fix: don't call reload index if action is findIncompatibleExtension

hikchoi:
# fix: don't call reload index if action is findIncompatibleExtension
This PR:
- fixes Doctor command so that `Reload Index` is not called before and after if the selected doctor action is `findIncompatibleExtensions`. (not needed, and significantly slows down operation in large workspaces.)

madge reports no diffs.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/Doctor.ts
+++ b/packages/plugin-core/src/commands/Doctor.ts
@@ -288,6 +288,14 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
     return { installStatus, contents };
   }
 
+  async reload(opts: CommandOpts) {
+    const engine =
+      opts.action !== PluginDoctorActionsEnum.FIND_INCOMPATIBLE_EXTENSIONS
+        ? ((await new ReloadIndexCommand().execute()) as DEngineClient)
+        : (undefined as unknown as DEngineClient);
+    return engine;
+  }
+
   async execute(opts: CommandOpts) {
     const ctx = "DoctorCommand:execute";
     window.showInformationMessage("Calling the doctor.");
@@ -313,10 +321,7 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
       await document.save();
     }
     this.L.info({ ctx, msg: "pre:Reload" });
-    const engine =
-      opts.action !== PluginDoctorActionsEnum.FIND_INCOMPATIBLE_EXTENSIONS
-        ? ((await new ReloadIndexCommand().execute()) as DEngineClient)
-        : (undefined as unknown as DEngineClient);
+    const engine = await this.reload(opts);
 
     let note;
     if (opts.scope === "file") {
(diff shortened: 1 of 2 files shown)


packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -288,6 +288,14 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
     return { installStatus, contents };
   }
 
+  async reload(opts: CommandOpts) {
```
kevinslin, line 291: can we avoid casting as per [[Going outside of the typescript compiler|dendron://dendron.docs/dev.process.code.best-practices#going-outside-of-the-typescript-compiler]]?

in this case, I don't think we're doing anything with the output so there's no need to cast.

also, can we add a docstring here?

kevinslin (changes requested):

packages/plugin-core/src/commands/Doctor.ts
jonathanyeung: can we also make it private? 

I see that it's exposed so that we can do a `sinon.spy(cmd, "reload");` in the test - my take (others may disagree), it's not really worth having that test to verify this functionality that's internal to this class.

kevinslin: i think its worhtwile because this is what this case is fixing (that we don't reload for this check). we can use [[Exposing private methods to tests|dendron://dendron.docs/dev.process.qa.test#exposing-private-methods-to-tests]] to expose it for testing

SeriousBug: We could also keep it private, then do a cast when using sinon. Since public/private is resolved at type level only, the spy will still work.
```ts
const reloadSpy = sinon.spy(cmd, "reload" as keyof DoctorCommand);
```

hikchoi: Casting was done because `ReloadIndexCommand` can return undefined, and all places that needed it down the line required `engine` to not be undefined. 
(and also didn't want to duplicate code by type checking on every single switch case, but that was me being a little lazy 😄)


I changed it so that `this.reload` checks for undefined before it returns, and throws if it is. 

Also just called reload on each switch case where it needed it. (I think typescript is not smart enough to narrow `DEngineClient | undefined` to `DEngineClient` based on the switch statement it lands on)

> We could also keep it private, then do a cast when using sinon. Since public/private is resolved at type level only, the spy will still work.

Nice tip! never thought of this.

jonathanyeung (approved):

kevinslin:
looks like the plugin test are failing

hikchoi:
Same problem with #2460 happening, waiting for #2478 to get merged. I'll rebase again after that.

hikchoi:
rebased, checks are all passing

SeriousBug (approved):
Looks good to me.

kevinslin (approved):

PR #2460, opened 2022-02-21
chore: separate tag configs

hikchoi:
# chore: separate tag configs
This PR:
- Splits up `enableFrontmatterTags` and `enableHashesForFMTags` to be in `preview` and `publishing` namespace
  - originally (before publishing namespace migration), both preview and publishing behavior was set by one config under `site.{enableFrontmatterTag|enableHashesForFMTags}`

No diff in madge output

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- [x] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), an example is included in the [test workspace](dev.ref.test-workspace)

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/engine-test-utils/src/__tests__/engine-server/markdown/__snapshots__/dendronPub.spec.ts.snap
+++ b/packages/engine-test-utils/src/__tests__/engine-server/markdown/__snapshots__/dendronPub.spec.ts.snap
@@ -189,16 +189,10 @@ VFile {
 }
 `;
 
-exports[`dendronPub frontmatter tags WHEN configured with enableHashesForFMTags option THEN tags are rendered with a # symbol 1`] = `
+exports[`dendronPub frontmatter tags GIVEN enableFrontmatterTags: false THEN not rendered: PREVIEW 1`] = `
 VFile {
   "contents": "<h1 id=\\"fmtags\\">Fmtags</h1>
-<p>has fm tags</p>
-<hr>
-<h2 id=\\"tags\\">Tags</h2>
-<ol>
-<li><a class=\\"color-tag\\" style=\\"--tag-color: #9e3623;\\" href=\\"tags.first\\">first</a></li>
-<li><a class=\\"color-tag\\" style=\\"--tag-color: #b1d1fc;\\" href=\\"tags.second\\">second</a></li>
-</ol>",
+<p>has fm tags</p>",
   "cwd": "<PROJECT_ROOT>",
   "data": Object {},
   "history": Array [],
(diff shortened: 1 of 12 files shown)


packages/engine-test-utils/src/__tests__/engine-server/markdown/__snapshots__/dendronPub.spec.ts.snap
```diff
@@ -294,6 +333,72 @@ VFile {
...
+`;
+
+exports[`dendronPub frontmatter tags GIVEN enableFrontmatterTags: true THEN rendered when available: single tag, PUBLISHING 2`] = `
+VFile {
+  "contents": "<h1 id=\\"fmtags\\"><a aria-hidden=\\"true\\" class=\\"anchor-heading\\" href=\\"#fmtags\\"><svg aria-hidden=\\"true\\" viewBox=\\"0 0 16 16\\"><use xlink:href=\\"#svg-link\\"></use></svg></a>Fmtags</h1>
+<p>has fm tags</p>
+<hr>
+<h2 id=\\"tags\\"><a aria-hidden=\\"true\\" class=\\"anchor-heading\\" href=\\"#tags\\"><svg aria-hidden=\\"true\\" viewBox=\\"0 0 16 16\\"><use xlink:href=\\"#svg-link\\"></use></svg></a>Tags</h2>
+<ol>
+<li><a title=\\"Private\\" style=\\"color: brown\\" href=\\"https://wiki.dendron.so/notes/hfyvYGJZQiUwQaaxQO27q.html\\" target=\\"_blank\\">first (Private)</a></li>
```
SeriousBug, line 359: Is the fact that these links are rendered private unrelated to this change? The snap before the PR has them rendered non-private.

hikchoi: Not sure why the snap before PR had them as non-private, but in the test `THEN rendered when available: multiple tags: publishing`, I only create the note `has.fmtags`, and the tags that this note has are `first and `second`. Since both `tags.first` and `tags.second` don't exist, I think they _should_ be private?

hikchoi: Just checked. all proc tests that have `ProcFlavor.PUBLISHING` are correctly rendering them to be private since I don't create the tag notes in any of the tests.

All test cases for `ProcFlavor.PREVIEW` won't have them rendered as private because we only do that for publishing.

hikchoi: > Not sure why the snap before PR had them as non-private, 

Looks like before the PR, the proc that was used to test didn't pass in the proc flavor as publishing, thus making it default to `ProcFlavor.REGULAR`.

I've confirmed that rolling back to before the PR, passing in `ProcFlavor.PUBLISHING` to the proc, and running it as-is will update the snapshot so that it correctly renders tags `first` and `second` as private.

hikchoi:
hmm not sure why CI is failing. it's not coming from any files that touched. is it caching the packages?

kevinslin:
> hmm not sure why CI is failing. it's not coming from any files that touched. is it caching the packages?

looks like a compiler issue in Gotonote: https://github.com/dendronhq/dendron/runs/5303577105?check_suite_focus=true#step:9:49

unused variable import

hikchoi:
> > hmm not sure why CI is failing. it's not coming from any files that touched. is it caching the packages?
> 
> looks like a compiler issue in Gotonote: https://github.com/dendronhq/dendron/runs/5303577105?check_suite_focus=true#step:9:49
> 
> unused variable import

I'm aware of that. Just that I am not touching that file and what the CI is complaining isn't present.
(see: https://github.com/dendronhq/dendron/blob/chore/separate-tag-configs/packages/plugin-core/src/commands/GotoNote.ts#L27:L27)

hikchoi:
The tests are failing because an unused import was re-introduced during a merge in https://github.com/dendronhq/dendron/commit/8d359d39c54672fdebbd647cd8bff1867b6000d2

On this branch, I rebased just be for the merge mentioned above happened, so on my end it looked like `GoToNote.ts` didn't have any problem and I was fine building locally.

Suspecting that CI is caching the artifacts from another run before which is affecting subsequent runs. (though the cache action is saying that they didn't find any cached libs, so I'm a bit confused.)

hikchoi:
rebased and checks are all passing now

kevinslin (approved):

PR #2424, opened 2022-02-15
chore: publishing config migration cleanup

hikchoi:
# chore: publishing config migration cleanup

This PR
- fixes migration and its tests' version to be `0.83.0`, which is the actual version that it will be introduced for general release.
- exposes latest migration in `RunMigrationCommand` quickpick (was omitted)
- updates dendron.yml for test-workspace
- deprecates `publishing.previewPort`, `publishing.siteRepoDir`, and `publishing.hierarchy.*.noIndexByDefault` 

### Basics

- [x] code should follow [Code Conventions](dev.process.code)
- [x] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices).
- [x] sticking to existing conventions instead of creating new ones 
  - eg: [if configuration for utilities are already in one module or package, add future utilities there as well](https://github.com/dendronhq/dendron/pull/1960#discussion_r786228021)

### Extended

- General
  - [x] check whether code be simplified
  - [x] check if similar function already exist in the codebase. if so, can it be re-used?
  - [x] check if this change adversely impact performance
- Operations
  - [x] when shipping this change, will it just work or will it introduce additional operational overhead due to complicated interface or known bugs?
- Architecture
  - [x] check if code is introducing changes on a foundational class or interface. if so, call for design review if needed 
    - eg: [making changes to DNode](https://github.com/dendronhq/dendron/pull/2158#pullrequestreview-854689586)

### Basics

- [~] if you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Extended

- [~] can we track the performance of this change to know if it is _successful_? 
  - eg: [see usage for export pod](https://github.com/dendronhq/dendron/pull/2190#pullrequestreview-855715612)

## 

### Basics

- [~] [Write Tests](dev.process.qa.test) 
- [x] [Confirm existing tests pass](dev.process.qa.test)
- [x] [Confirm manual testing](dev.process.qa.test) 
- [~] Common cases tested
- [~] 1-2 Edge cases tested
- [~] If your tests changes an existing snapshot, snapshots have been [updated](dev.process.qa.test)

### Extended

- [~] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), an example is included in the [test workspace](dev.ref.test-workspace)

- CSS
  - [~] display is correct for following dimensions
[...]

PR #2461, opened 2022-02-21
chore: allow config client soft mapping

hikchoi:
# chore: allow config client soft mapping
This PR:
- readds config-client version mapping for v5 (config v5 to client version 0.83.0)
   - In https://github.com/dendronhq/dendron/pull/2424 I removed the compat mapping for v5, which introduced a bug where users who are on v5 will hit an unexpected error because the validation logic was looking for a mapping that didn't exist. This PR fixes this regression.
- allow "soft mapping" of config-client version
  - this is used to relax the config-client compatibility validation in dendron-cli when the newly introduced config version is backwards compatible.
  - If a config is outdated (e.g. config v4 on version 0.83.0), the same mismatch notice will be printed, but will not exit the process since we still accept the outdated config.

madge reports no diffs

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/constants/configs/compat.ts
+++ b/packages/common-all/src/constants/configs/compat.ts
@@ -1,7 +1,24 @@
-export const CONFIG_TO_MINIMUM_COMPAT_MAPPING: { [key: number]: string } = {
-  1: "0.0.0",
-  2: "0.63.0", // config consolidation (commands namespace)
-  3: "0.65.0", // config consolidation (workspace namespace)
-  4: "0.70.0", // config consolidation (preview namespace)
-  // 5: "0.83.0", // config consolidation (publishing namespace): commented this out because adding this compat mapping would prevent users from keeping the v4 config and still use the cli for publishing. re-enable once we remove backward compatibility.
+import _ from "lodash";
+
+export type ConfigMapping = {
+  clientVersion: string;
+  softMapping?: boolean; // config version is mapped to minimum client version, but it's backward's compatible for now.
 };
+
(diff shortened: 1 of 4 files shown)


packages/common-all/src/constants/configs/compat.ts
```diff
@@ -1,7 +1,24 @@
...
 };
+
+export const CONFIG_TO_MINIMUM_COMPAT_MAPPING: {
+  [key: number]: ConfigMapping;
+} = {
+  1: { clientVersion: "0.0.0" },
+  2: { clientVersion: "0.63.0" }, // config consolidation (commands namespace)
+  3: { clientVersion: "0.65.0" }, // config consolidation (workspace namespace)
+  4: { clientVersion: "0.70.0" }, // config consolidation (preview namespace)
+  5: { clientVersion: "0.83.0", softMapping: true }, // config consolidation (publishing namespace): commented this out because adding this compat mapping would prevent users from keeping the v4 config and still use the cli for publishing. re-enable once we remove backward compatibility.
```
kevinslin, line 15: shouldn't this be valid for versions prior to 0.83.0 as well?

kevinslin (approved):

PR #2469, opened 2022-02-22
chore(treev2): Migrate from notesyncservice to engine events

tma66:
**Background**
As part of initiative to deprecate notesyncservice#onnotechanged, we will first migrate the dendron treeviewv2 to the engine events. See part 1 of [[Deprecate Notesyncservice Onnotechanged|dendron://private/task.2022.02.14.deprecate-notesyncservice-onnotechanged]]

**Changes**
1. Update `DendronTreeViewV2` to use engine events instead of notesyncservice
2. Added optional param `activeNote` to `OnDidChangeActiveTextEditorData`. This param lets the frontend know which note to focus on. Sometimes users can make changes to notes outside the active note but we do not want the focus to change.
3. Avoid TypeError due to indexoutofboundsexception in frontend and engine
4. If engine cannot find a note for a given query, return empty array instead of an error

*Madge*
plugin-core before: 20
plugin-core after: 19

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/dendron-plugin-views/src/components/DendronApp.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronApp.tsx
@@ -66,7 +66,7 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
     switch (msg.type) {
       case DMessageEnum.ON_DID_CHANGE_ACTIVE_TEXT_EDITOR:
         const cmsg = msg as OnDidChangeActiveTextEditorMsg;
-        const { sync, note, syncChangedNote } = cmsg.data;
+        const { sync, note, syncChangedNote, activeNote } = cmsg.data;
         if (sync) {
           // skip the initial ?
           logger.info({
@@ -87,7 +87,9 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
           await ideDispatch(engineSlice.syncNote({ ...workspace, note }));
         }
         logger.info({ ctx, msg: "setNoteActive:pre" });
-        ideDispatch(ideSlice.actions.setNoteActive(note));
+        // If activeNote is in the data payload, set that as active note. Otherwise default to changed note
+        const noteToSetActive = activeNote ? activeNote : note;
+        ideDispatch(ideSlice.actions.setNoteActive(noteToSetActive));
         logger.info({ ctx, msg: "setNoteActive:post" });
         break;
       case LookupViewMessageEnum.onUpdate:
--- a/packages/dendron-plugin-views/src/components/DendronApp.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronApp.tsx
@@ -64,6 +64,7 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
     const ctx = "useVSCodeMsg";
     logger.info({ ctx, msgType: msg.type });
     switch (msg.type) {
+      // TODO: Handle case where note is deleted. This should be implemented after we implement new message type to denote note state changes
       case DMessageEnum.ON_DID_CHANGE_ACTIVE_TEXT_EDITOR:
         const cmsg = msg as OnDidChangeActiveTextEditorMsg;
         const { sync, note, syncChangedNote, activeNote } = cmsg.data;
--- a/packages/engine-server/src/enginev2.ts
+++ b/packages/engine-server/src/enginev2.ts
@@ -506,9 +506,9 @@ export class DendronEngineV2 implements DEngine {
       originalQS,
     });
 
-    //if (items.length === 0) {
-    //  return { error: null, data: [] };
-    //}
+    if (items.length === 0) {
+      return { error: null, data: [] };
+    }
 
     const item = this.notes[items[0].id];
     if (createIfNew) {
--- a/packages/engine-server/src/enginev2.ts
+++ b/packages/engine-server/src/enginev2.ts
@@ -505,6 +505,11 @@ export class DendronEngineV2 implements DEngine {
       onlyDirectChildren,
       originalQS,
     });
+
+    //if (items.length === 0) {
+    //  return { error: null, data: [] };
+    //}
+
     const item = this.notes[items[0].id];
     if (createIfNew) {
       let noteNew: NoteProps;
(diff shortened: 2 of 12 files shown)


packages/plugin-core/src/views/DendronTreeViewV2.ts
```diff
@@ -9,41 +9,57 @@ import {
   VSCodeEvents,
 } from "@dendronhq/common-all";
 import { getDurationMilliseconds } from "@dendronhq/common-server";
-import { WorkspaceUtils } from "@dendronhq/engine-server";
+import { EngineEventEmitter, WorkspaceUtils } from "@dendronhq/engine-server";
 import _ from "lodash";
 import path from "path";
 import * as vscode from "vscode";
+import { Disposable } from "vscode-languageclient";
```
jonathanyeung, line 16: This is a bit confusing, but this is the technically the wrong Disposable (although functionally it _probably_ doesn't make a diff). 

`import { Disposable } from "vscode";`

We should do a quick sweep for other places that import it from `vscode-languageclient`, I see that there are several.

packages/plugin-core/src/views/DendronTreeViewV2.ts
```diff
@@ -146,6 +162,10 @@ export class DendronTreeViewV2 implements vscode.WebviewViewProvider {
     });
   }
 
+  /**
+   * Notify webview to sync given note and to focus on active note
+   * @param note to sync
+   */
   public refresh(note: NoteProps) {
```
jonathanyeung, line 169: can we make this one private now?

packages/engine-server/src/enginev2.ts
```diff
@@ -505,6 +505,11 @@ export class DendronEngineV2 implements DEngine {
       onlyDirectChildren,
       originalQS,
     });
+
+    if (items.length === 0) {
```
jonathanyeung, line 509: I think this is the correct fix / behavior - but I wonder if this may cause any regressions in existing code that calls this API and expects a `500` kind of error / exception to be happening.

packages/dendron-plugin-views/src/components/DendronApp.tsx
```diff
@@ -87,7 +87,9 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
           await ideDispatch(engineSlice.syncNote({ ...workspace, note }));
         }
         logger.info({ ctx, msg: "setNoteActive:pre" });
-        ideDispatch(ideSlice.actions.setNoteActive(note));
+        // If activeNote is in the data payload, set that as active note. Otherwise default to changed note
```
jonathanyeung, line 90: Can you move (or copy) this comment into `typesv2` for `activeNote` - basically to explain the logic to someone who is going to call the API of when they should set activeNote and not. 

Also - could you add a `TODO:` to that comment as well that we should split this up into a separate command beyond `OnDidChangeActiveTextEditor`? I.e. `onNoteStateChanged` - I think we should make that change soon.

packages/plugin-core/src/views/DendronTreeViewV2.ts
```diff
@@ -9,41 +9,57 @@ import {
...
-      }
-    });
+    this._onEngineNoteStateChangedDisposable =
+      this._engineEvents.onEngineNoteStateChanged((noteChangeEntry) => {
+        const ctx = "refreshDendronTreeViewV2EngineNoteStateChanged";
+        Logger.info({ ctx });
+        noteChangeEntry.map((changeEntry) => this.refresh(changeEntry.note));
+      });
+  }
+  dispose(): void {
```
jonathanyeung, line 59: I don't think we've hooked up the calling of this `dispose` in workspace.ts yet when we construct the tree view?

jonathanyeung (changes requested):

packages/engine-server/src/enginev2.ts
kevinslin: i don't think it does. as long as we have a test to catch further regressions, this looks good

packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
```diff
@@ -173,3 +173,20 @@ describe("engine, config/", () => {
     });
   });
 });
+
+test("WHEN querying for a note that doesn't exist, return empty array in data", async () => {
```
kevinslin, line 177: can you use the engine test harness as described here? [[Testing a new Engine Method|dendron://dendron.docs/pkg.engine-test-utils.qa.test#testing-a-new-engine-method]]

the advantage is that it will test the engine both directly and via api call

kevinslin (changes requested):

packages/plugin-core/src/views/DendronTreeViewV2.ts
tma66: sure fixed

packages/engine-server/src/enginev2.ts
tma66: Looked through all the use cases and aside from `slice.ts`, this shouldn't break.

packages/dendron-plugin-views/src/components/DendronApp.tsx
tma66: Created new task for TODO [[Split up Ondidchangeactivetexteditordata|dendron://private/task.2022.02.23.split-up-ondidchangeactivetexteditordata]]

packages/plugin-core/src/views/DendronTreeViewV2.ts
tma66: ah good catch

packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
tma66: gotcha, looks good

jonathanyeung (approved):

kevinslin (approved):

PR #2472, opened 2022-02-23
fix(views): faster webviews by reducing engine sync operations

jonathanyeung:
## fix(views): faster webviews by reducing engine sync operations

This change makes loading of our webviews faster. 

1. There was what I assume is a copy-paste error in `slice.ts` where `initNotes` and `syncNote` both had the same typePrefix of "engine/init" being passed into `createAsyncThunk`.  This inadvertently caused a big problem - we have `extraReducers` hooked up to trigger off of pending+fulfilled of `initNotes`.  But under the hood in redux, it seems like these are actually keyed off of the registered typePrefix - so the resulting effect is that every time we just wanted to do `syncNote`, we ended up triggering `initNotes` _twice_ (why twice instead of once I'm not too sure, but the logs showed that this was happening).  We set the `syncNote` flag extensively - it's set during tree view state changes and every time preview is changed - including every time you type into the editor with preview open.  So we were doing maybe 4 sets of `initNotes` every time you did a note navigation and had tree view and note preview open.
2. `initNotes` is a very expensive operation - and we don't actually need it except in TreeView.  I've moved `useEngine` from out of the main `DendronApp.tsx` and pushed it into `DendronTreeExplorerPanel.tsx`.  `DendronNotePage.tsx` does  need `engine.config` state to set, so I added a new method in the engine slice to allow it to just retrieve the config from the engine instead of initializing everything.

Results: 
- Loading of Preview is much faster
- Swapping Between Notes while preview is open is very quick
- LookupView loads very fast (although there's some UI glitching/flickering as a result, which we can address later)
- TreeView looks smoother as well.

Rendering of large markdown files is still quite sluggish though.

---

### Basics

- [ N/A - no plugin-core changes] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices).

### Extended
- General
- Operations
- Architecture

### Basics

Not sure - we don't seem to have any coverage on plugin-views yet.
- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended

### DIFF
--- a/packages/dendron-plugin-views/src/components/DendronApp.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronApp.tsx
@@ -25,7 +25,7 @@ import "../styles/scss/main.scss";
 import { Layout } from "antd";
 const { Content } = Layout;
 
-const { useEngineAppSelector, useEngine } = engineHooks;
+const { useEngineAppSelector } = engineHooks;
 
 function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
   const ctx = "DendronVSCodeApp";
@@ -36,8 +36,6 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
   const logger = createLogger("DendronApp");
 
   logger.info({ ctx, msg: "enter", workspace });
-  // used to initialize the engine
-  useEngine({ engineState: engine, opts: workspace });
 
   const props = {
     engine,
@@ -100,11 +98,6 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
         break;
     }
   });
-
-  // don't load until active
-  if (_.isEmpty(engine.notes)) {
-    return <div>Loading...</div>;
-  }
 
   return <Component {...props} />;
 }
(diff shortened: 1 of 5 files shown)


packages/dendron-plugin-views/src/components/DendronApp.tsx
```diff
@@ -101,11 +99,6 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
     }
   });
 
-  // don't load until active
-  if (_.isEmpty(engine.notes)) {
```
jonathanyeung, line 105: I'll probably amend the logic around here - it still takes some time to render large notes, and the "Loading..." placeholder is currently gone.

SeriousBug (approved):
The change looks good to me. Good catch! The loading is not that bad (a 1000+ line note loaded in about a second for me), so I think it's fine to merge this without re-adding the loading placeholder for now.

tma66 (approved):
Nice find. Looks good overall

kevinslin (approved):

PR #2480, opened 2022-02-24
fix: properly render mermaid and katex when published

hikchoi:
# fix: properly render mermaid and katex when published
This PR:
- Properly handles v5 configs for mermaid and katex (`publishing.enableMermaid` and `publishing.enableKatex`) when publishing, and fixes an issue that cause mermaid diagrams and katex expressions not rendering when the note is published.
- Properly sync katex fonts to plugin during build in `dendron dev sync_assets`

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-frontend/src/components/DendronNote.tsx
+++ b/packages/common-frontend/src/components/DendronNote.tsx
@@ -17,7 +17,7 @@ export const useMermaid = ({
 }) => {
   React.useEffect(() => {
     const logger = createLogger("useMermaid");
-    if (config && ConfigUtils.getProp(config, "mermaid")) {
+    if (config && ConfigUtils.getEnableMermaid(config, true)) {
       // @ts-ignore
       const mermaid = (window as any)._mermaid;
       logger.info("mermaid created");
--- a/packages/dendron-cli/src/utils/build.ts
+++ b/packages/dendron-cli/src/utils/build.ts
@@ -360,6 +360,8 @@ export class BuildUtils {
       "common-assets"
     );
 
+    const commonAssetsBuildRoot = path.join(commonAssetsRoot, "build");
+
     // destination for assets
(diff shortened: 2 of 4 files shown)


hikchoi:
looks like before playwright tests are run, the note data for nextjs-template is pulled from S3.
This data doesn't contain a note that uses katex / mermaid blocks so I deferred writing tests to later.

Do we have a way to update the content of `dendron-site.zip` so that we can add note data that we want to test with the playwright tests?

jonathanyeung:
> looks like before playwright tests are run, the note data for nextjs-template is pulled from S3. This data doesn't contain a note that uses katex / mermaid blocks so I deferred writing tests to later.
> 
> Do we have a way to update the content of `dendron-site.zip` so that we can add note data that we want to test with the playwright tests?

Code side lgtm - re: playwright; @tma66  - any ideas?

kevinslin:
> looks like before playwright tests are run, the note data for nextjs-template is pulled from S3. This data doesn't contain a note that uses katex / mermaid blocks so I deferred writing tests to later.
> 
> Do we have a way to update the content of `dendron-site.zip` so that we can add note data that we want to test with the playwright tests?

When testing in the github actions pipeline, we use the contents of `dendron-site.zip` to run tests. This is currently created by zipping output of https://github.com/dendronhq/template.publish.github. 
Moving forward, I think we should have a `test-workspace` that we bundle with `nextjs-template` to run tests with. This makes it easier to update and avoids downloading an extra dependency from s3

kevinslin (approved):

PR #2484, opened 2022-02-25
enhance(workspace): automatically update configuration for vaults that have remotes, but aren't configured as such

SeriousBug:
If a vault has a remote configured, but the vault configuration did not have a remote set, then this PR automatically updates the configuration to mark the vault as a remote vault.

The PR also adds telemetry for both this feature, and also the existing features that fix missing root note and schemas.

Finally, the PR improves our git wrapper when detecting the current remote. Before we would just use the first remote available, but that might not be the remote for the current branch. The new code tries to parse the remote from the upstream of the current branch, then falls back to the first available remote if that fails.

Docs: https://github.com/dendronhq/dendron-site/pull/418

### Basics

  - No change

### Extended

- General
    - Minimal, in the most common case this will only check if a single file exists during initialization
- Operations
    - Considered edge cases, and the action only works if we're sure that the user has a repo with a remote set. Fully automatic so no confusing interface.
- Architecture

### Basics

  - linked above

### Extended

    - telemetry above tracks usage

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- [~] CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

  - linked above

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -98,6 +98,10 @@ export enum ContextualUIEvents {
   ContextualUICodeAction = "ContextualUI_CodeAction",
 }
 
+export enum WorkspaceEvents {
+  AutoFix = "AutoFix",
+}
+
 export enum NativeWorkspaceEvents {
   DetectedInNonDendronWS = "Native_Workspace_Detected_In_Non_Dendron_WS", // watcher has detected a Dendron workspace getting created inside a non-Dendron workspace
 }
@@ -111,4 +115,5 @@ export const DendronEvents = {
   ConfigEvents,
   ContextualUIEvents,
   NativeWorkspaceEvents,
+  WorkspaceEvents,
 };
--- a/packages/dendron-cli/src/commands/doctor.ts
+++ b/packages/dendron-cli/src/commands/doctor.ts
@@ -71,6 +71,8 @@ export class DoctorCLICommand extends CLICommand<CommandOpts, CommandOutput> {
 
   async execute(opts: CommandOpts): Promise<CommandOutput> {
     const ds = new DoctorService();
-    return ds.executeDoctorActions(opts);
+    const out = await ds.executeDoctorActions(opts);
+    ds.dispose();
+    return out;
   }
 }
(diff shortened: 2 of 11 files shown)


kevinslin (changes requested):
can we make this a doctor command and show a dialogue box to ask people to fix it? a little nervous about auto-fixes as we've been having issues with them in the past. 

in general, a design philosophy we might adapt from python: "explicit is better than implicit"

jonathanyeung:
Echo @kevinslin's comments - rather than having this be a part of `ReloadIndex`, let's put this under doctor - @hikchoi and I were discussing last week the prospect of having a scan-for-potential-issues that runs on ext upgrade that runs multiple doctor checks (not fixes) and then provides a summary of issues to users, it makes sense to add this remote check to be a part of that conversation (we were starting to get a bit weary of notification bombing our users on start up).  

Does having an incorrect remote config block the user from properly upgrading to a standalone vault setup?

hikchoi:
Agree on both @kevinslin and @jonathanyeung's points. We have been running into issues of running automatic fixes during activation. Our current way to tackle that is to be explicit about what is about to change (if the user accepts), and I think this is generally a good way forward. We will at some point run into _too many toast notifications_ during the activation lifecycle, but for now what we have will not be overcrowding; most of these potential issue scans twon't happen all at the same time.

@jonathanyeung and I briefly discussed on using a startup welcome view (or the dedicated dendron side panel we have been discussing) to consolidate all detected issues and presenting it nicely in one place, but this is a bit over-engineered for what we have currently. For now, the 80/20 solution would be making this a doctor command, and warning them on activation if detected. The keybinding resolution that we currently do during activation will be changed to follow this pattern as well.

See [findIncompatibleExtensions](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/commands/Doctor.ts#L341:L352) for an example of this

SeriousBug:
Thanks for the feedback everyone. Automatic conversion was something that was discussed briefly months ago that I had in my backlog. I converted this to a Doctor command instead.

I left the existing "create missing root note or schema" behavior, but also kept my changes which make it concurrent and add analytics for them. I can also revert those changes if desired.

Having a welcome screen where we can display issues and more sounds like a good idea, and I agree a doctor action is good until then.

jonathanyeung (approved):

kevinslin (approved):

PR #2488, opened 2022-02-25
chore(analytics): fix session id logic

kevinslin:
chore(analytics): fix session id logic

To track session id, segment expects the field to be passed in as `track({integrations: ...})`. We currently pass in all metadata as `properties`.
This change exposes all of segment's track properties to our analytic wrappers and also changes the input parameters to be object based (vs positional)

***
no change to circular dep

***

### Basics

### Extended

- General
- Operations
- Architecture

## Instrumentation

NA

## 

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -1,3 +1,4 @@
+/* eslint-disable camelcase */
 export enum VSCodeEvents {
   ServerCrashed = "ServerCrashed",
   InitializeWorkspace = "InitializeWorkspace",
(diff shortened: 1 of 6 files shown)


jonathanyeung (approved):
Can we double check that the telemetry in ide_dev looks good prior to the GA release on the telem changes?  Just wanna avoid having a change that causes us to need to reformat all of our charts.

PR #2499, opened 2022-02-28
fix(internal): write note orphans child notes

hikchoi:
# fix: write note orphans child notes

This PR:
- fixes a bug where overwriting an existing note's id will result in:
  - stale children props of the parent of the note
  - orphaned children of the written note
- generally refactors code that this edge case covers in `_writeNewNote` and add omitted relevant note change entries.
- adds the test that was skipped in `EngineEvents.spec.ts` that this bug was blocking.

No added circ dependencies.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -352,7 +352,6 @@ export class NoteUtils {
   }): NoteChangeEntry[] {
     const changed: NoteChangeEntry[] = [];
     const { childToDelete, notes } = opts;
-    // const parent = DNodeUtils.getParent(childToDelete, { nodeDict: notes });
     let parent;
     if (childToDelete.parent) {
       parent = notes[childToDelete.parent];
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -346,6 +346,35 @@ export class NoteUtils {
     });
   }
 
+  static deleteChildFromParent(opts: {
+    childToDelete: NoteProps;
+    notes: NotePropsDict;
+  }): NoteChangeEntry[] {
+    const changed: NoteChangeEntry[] = [];
+    const { childToDelete, notes } = opts;
+    // const parent = DNodeUtils.getParent(childToDelete, { nodeDict: notes });
+    let parent;
+    if (childToDelete.parent) {
+      parent = notes[childToDelete.parent];
+    } else {
+      throw new DendronError({
+        message: `No parent found for ${childToDelete.fname}`,
+      });
+    }
+
+    const prevParentState = { ...parent };
+    parent.children = _.reject<string[]>(
+      parent.children,
+      (ent: string) => ent === childToDelete.id
+    ) as string[];
+    changed.push({
+      status: "update",
+      prevNote: prevParentState,
+      note: parent,
+    });
+    return changed;
+  }
+
   /**
    * Add node to parents up the note tree, or create stubs if no direct parents exists
    * @param opts
--- a/packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
+++ b/packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
@@ -175,6 +175,11 @@ describe("GIVEN a DendronEngineClient running on client-side", () => {
 
                 expect(createEntries[0].note.fname).toEqual("foo");
                 expect(createEntries[0].note.id).toEqual("updatedID");
+                expect(createEntries[0].note.children).toEqual(["foo.ch1"]);
+                expect(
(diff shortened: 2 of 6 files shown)


hikchoi:
@Harshita-mindfire This PR touches on some of the `GithubIssuePod.spec.ts` tests that were breaking from the changes. Can you check if these changes are fine? 🙏🏼

packages/common-all/src/dnode.ts
```diff
@@ -346,6 +346,35 @@ export class NoteUtils {
     });
   }
 
+  static deleteChildFromParent(opts: {
+    childToDelete: NoteProps;
+    notes: NotePropsDict;
+  }): NoteChangeEntry[] {
+    const changed: NoteChangeEntry[] = [];
+    const { childToDelete, notes } = opts;
+    // const parent = DNodeUtils.getParent(childToDelete, { nodeDict: notes });
```
jonathanyeung, line 355: nit: Delete commented line.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -155,19 +153,52 @@ describe("GIVEN a DendronEngineClient running on client-side", () => {
...
+                "update"
+              ) as NoteChangeUpdateEntry[];
+
+              testAssertsInsideCallback(() => {
+                expect(createEntries.length).toEqual(1);
+                expect(updateEntries.length).toEqual(2);
+                expect(deleteEntries.length).toEqual(1);
+
+                expect(createEntries[0].note.fname).toEqual("foo");
+                expect(createEntries[0].note.id).toEqual("updatedID");
```
jonathanyeung, line 177: Can we add 2 more checks on the foo note? 
- That its parent is correct (root I think here)
- That its children are correct (foo.ch1)

Although it does look like you added these checks to the other Engine test, so maybe we're already covered.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
```diff
@@ -155,19 +153,52 @@ describe("GIVEN a DendronEngineClient running on client-side", () => {
...
                 done();
-              } catch (err) {
-                done(err);
-              }
+              }, done);
             }
           );
 
-          await engineClient.writeNote(fooUpdated, { updateExisting: true });
+          await engineClient.writeNote(fooUpdated);
```
jonathanyeung, line 201: It's still not very clear to me on whether the `{updateExisting: true}` flag is necessary to be set here or not to trigger this behavior. 

IMO - this _should_ be the correct behavior regardless of whether this flag is set or not - @kevinslin thoughts?

jonathanyeung (changes requested):

jonathanyeung:
Looking good. 

There's a potentially related failure in the plugin CI pass for `MoveNoteCommand`.

packages/engine-test-utils/src/__tests__/engine-server/EngineEvents.spec.ts
hikchoi: I don't think for this case setting `updateExisting: true` is right. we want to overwrite.

in `engine.writeNote` if `updateExisting: true` is set, we don't go into the logic at all where the overwrite happens.

hikchoi: for anything other than changing the note's id, `updateExisting: true` makes sense.

hikchoi:
> There's a potentially related failure in the plugin CI pass for MoveNoteCommand.

Looking into it now. will ping when it's fixed.

hikchoi:
@jonathanyeung 

update: 

- cleaned up comments
- added mentioned cases to test
- cleaned up `_writeNewNote` logic so that it's updating _existing note`s parent_ instead of _new note's parent_. They will be the same since we are overwriting them in place, but semantically this will be more natural.
- move note command's test was failing because the deprecated `createNote` was not ensuring the testing note's existence in engine. swapped it out to the new `createNoteWithEngine` so the test works as intended.

hikchoi:
also updated the fixed tests in `GithubIssuePod.spec.ts` to not use non-null assertions as per our recent effort to remove all type assertions and casting.

Harshita-mindfire (approved):
lgtm!

jonathanyeung (approved):

PR #2505, opened 2022-03-02
fix(views): md parsing and preview perf improvements

jonathanyeung:
## fix(views): md parsing and preview perf improvements

I was doing some perf testing on preview, and I found that for notes with a lot of links and tags, the render time was O(n), where n is the number of notes in the workspace.  This was true even if all of the notes or tags were the exact same link, just repeated hundreds of times. 

I tracked this down to some code in `dendronPub.ts` which was using the legacy `getNoteByFnameV5` instead of the new `getNoteByFnameFromEngine`.  The legacy `getNoteByFnameV5` and `getNotesByFnameV5` are O(n) for note lookup, whereas the new versions are ~O(1) due to using a dictionary/hash-map implementation. 

Given the pervasive use of the legacy calls, I started to scrub through to upgrade many existing instances to the new ones, which should result in performance improvements across many different kinds of operations, from preview, publishing, to many other commands as well, including writing notes.  In this PR, I've fully scrubbed out the plural version `getNotesByFnameV5`.  `getNoteByFnameV5` is more pervasive, so that'll be deferred to a later change.

Some results: rendering our super heavy [[Backlog|dendron://private/dendron.tasks.backlog]] note, which has about 300 links/tags, is down from 1500-2000 ms range to 50-200 ms range, which is a notable difference.

We should also add some perf test suites, but I'll defer that to a separate change.

---

### Basics

- [x] -1 circular dependency. (18->17) circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices).

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended

### DIFF
--- a/packages/engine-server/src/markdown/remark/dendronPub.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPub.ts
@@ -155,7 +155,7 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
 
       // Special Logic for 403 Error Static Page:
       if (fname === "403") {
-        note = SiteUtils.addSiteOnlyNotes({ engine })[0];
+        note = SiteUtils.create403StaticNote({ engine });
       } else {
         note = NoteUtils.getNoteByFnameFromEngine({
           fname,
--- a/packages/engine-server/src/markdown/remark/dendronPub.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPub.ts
@@ -151,11 +151,19 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
           )}`,
         });
       }
-      const note = NoteUtils.getNoteByFnameFromEngine({
-        fname,
-        vault,
-        engine,
-      });
+      let note;
+
+      // Special Logic for 403 Error Static Page:
+      if (fname === "403") {
+        note = SiteUtils.addSiteOnlyNotes({ engine })[0];
+      } else {
+        note = NoteUtils.getNoteByFnameFromEngine({
+          fname,
+          vault,
+          engine,
+        });
+      }
+
       if (!note) {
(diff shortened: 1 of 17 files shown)


packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -152,7 +151,19 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
           )}`,
         });
       }
-      const note = MDUtilsV5.getNoteByFname(proc, { fname });
+      let note;
+
+      // Special Logic for 403 Error Static Page:
+      if (fname === "403") {
+        note = SiteUtils.addSiteOnlyNotes({ engine })[0];
```
kevinslin, line 158: can we refactor addSiteOnlyNotes since this is fairly brittle? eg. `SiteUtils.create403Note` and use this instead

packages/engine-server/src/markdown/utilsv5.ts
```diff
@@ -431,7 +417,7 @@ export class MDUtilsV5 {
     });
 
     // add additional plugin for publishing
-
+    //maybe something here is interfering
```
SeriousBug, line 420: Is this left over from debugging, or is there still an issue here?

SeriousBug (approved):
🎉

Looks good, just left one comment.

packages/engine-server/src/markdown/utilsv5.ts
jonathanyeung: good catch - it's leftover from debugging, will remove.

PR #2517, opened 2022-03-03
feat: self contained vaults (seed)

SeriousBug:
This PR makes the first step towards self contained vaults. The changes added are:
- A `dev` configuration flag has been added. When set to true, all vaults will be created as self contained vaults.
- Dendron can create self contained local vaults.
- Dendron can add self contained remote vaults (except ones in workspace vaults).
- Dendron can read the notes in self contained vaults, and open self contained vaults directly like workspaces
- Fixes a regression where vaults would could get added to the gitignore file multiple times, and re-enables the test for this regression

I wanted to get feedback & get the initial step merged while I continue working on the remaining basic features, and the wider feature set.

Docs: https://github.com/dendronhq/dendron-site/pull/435

### Basics

  - Does not affect circular dependency count

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended

- [x] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), an example is included in the [test workspace](dev.ref.test-workspace)

- [~] CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

  - Holding off on the documentation until the core features are done, at which point I'll add the documentation and we can announce the dev flag for folks to test if they want to

## 

### Basics

## 

### Extended


packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -133,6 +133,83 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
     };
   }
 
+  async gatherVaultSelfContained(
+    sourceType: VaultRemoteSource
+  ): Promise<CommandOpts | undefined> {
+    // If the vault name already exists, creating a vault with the same name would break things
+
+    if (sourceType === "local") {
```
kevinslin, line 141: Instead of using naked strings, can we use enums? This makes refactoring easier when its used in many places

packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -133,6 +133,83 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+
+  async gatherInputs(): Promise<CommandOpts | undefined> {
+    const sourceTypeSelected = await VSCodeUtils.showQuickPick([
+      { label: "local", picked: true },
+      { label: "remote" },
+    ]);
+    if (!sourceTypeSelected) {
+      return;
+    }
+    const sourceType = sourceTypeSelected.label as VaultRemoteSource;
```
kevinslin, line 202: can we use a type guard to do a check for this instead of a cast? we're moving to not [[Go outside of the typescript compiler|dendron://dendron.docs/dev.process.code.best-practices#going-outside-of-the-typescript-compiler]]

packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -182,8 +259,85 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+        const localUrl = path.join(wsRoot, opts.path);
+        if (!remoteUrl) {
+          throw new DendronError({
+            message:
+              "Remote vault has no remote set. This should never happen, please send a bug report if you encounter this.",
+          });
+        }
+
+        await fs.ensureDir(localUrl);
+        const git = new Git({ localUrl, remoteUrl });
```
kevinslin, line 287: is there a reason we don't use simplegit as in `handleRemoteRepo`? also, does it make sense to encapsulate common clone logic across handle remote functions?

packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -182,8 +259,85 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+            vaults.length === 1
+              ? "adding vault"
+              : `adding ${vaults.length} vaults`,
+          increment,
+        });
+        const wsService = new WorkspaceService({ wsRoot });
+
+        if (workspace) {
+          // TODO: To be implemented for backwards compatibility until workspace vaults are phased out
+          throw new DendronError({
```
kevinslin, line 311: is it worth adding compatibility to workspace vaults? another approach is to, when we convert from workspace to self contained vaults, to also convert over existing workspace vaults to self contained vaults.
this also reminds me - what is the migration plan for the dendron team? how will we dogfood self contained vaults with `org-workspace`?

packages/plugin-core/src/test/suite-integ/VaultAddCommand.test.ts
```diff
@@ -428,3 +430,160 @@ suite("VaultAddCommand", function () {
...
+          vault: vault!,
+          engine,
+        });
+        expect(note).toBeTruthy();
+        expect(note?.vault.name).toEqual(vaultName);
+      });
+    }
+  );
+
+  describeSingleWS(
```
kevinslin, line 507: we should have a test case for adding a self contained vault that has other vaults in it

test-workspace/dendron.yml
```diff
@@ -3,6 +3,7 @@ dev:
     enableWebUI: true
     enablePreviewV2: true
     enableExportPodV2: true
+    enableSelfContainedVaults: true
```
kevinslin, line 6: reminder to update the json schema for the dendron.yml

test-workspace/dendron.yml
```diff
@@ -20,12 +21,14 @@ commands:
     insertNoteIndex:
         enableMarker: false
     randomNote: {}
+    copyNoteLink: {}
 workspace:
     dendronVersion: 0.82.0
```
kevinslin, line 26: update json schema

test-workspace/dependencies/localhost/vault3/.gitignore
```diff
@@ -0,0 +1,9 @@
+node_modules
+.dendron.*
+build
+seeds
+.next
+pods/service-connections
+
```
kevinslin, line 7: we currently generate new lines for the gitignore, can we remove that?

kevinslin (commented):
looks great. some initial comments. courtesy of kevin and hikchoi :)

packages/common-server/src/git.ts
```diff
@@ -140,6 +141,63 @@ export class GitUtils {
...
+    url?: string;
+  }): string {
+    // If no remote URL exists, then it's a local vault. We keep these in a
+    // local-only folder.
+    if (url === undefined) return FOLDERS.LOCAL_DEPENDENCY;
+    // Check if it matches any web URLs like
+    // https://github.com/dendronhq/dendron-site.git This may also look like
+    // http://example.com:8000/dendronhq/dendron-site.git, we skip the port
+    const webMatch =
+      /^(https?:\/\/)(?<domain>[^/:]+)(:[0-9]+)?\/(?<path>.+?)(\.git)?$/.exec(
```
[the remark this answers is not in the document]
hikchoi: I suggested this during a pairing session, but I am now realizing that js doesn't natively support verbose regex 😅 

I was imagining something like what python has ([re.VERBOSE](https://docs.python.org/3/library/re.html#re.VERBOSE))

Looks like there are a few packages that implement this but not sure if it will be worth for us to add another dependency for this. We might get away with documenting them in a way we would do if verbose flag was available.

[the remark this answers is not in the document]
SeriousBug: It's still kinda possible by using strings and disabling prettier formatting for that block. I'll go ahead and make that change.

packages/engine-server/src/workspace/service.ts
```diff
@@ -356,6 +359,72 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+      await note2File({ note, vault, wsRoot: this.wsRoot });
+    }
+    if (
+      !fs.existsSync(SchemaUtils.getPath({ root: notesPath, fname: "root" }))
+    ) {
+      await schemaModuleOpts2File(schema, notesPath, "root");
+    }
+
+    // Create the config and code-workspace for the vault, which make it self contained
+    DConfig.getOrCreate(vaultPath);
```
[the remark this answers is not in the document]
SeriousBug: Ah good idea, hadn't thought of that.

> currently, changing into a self contained vault

My description of the PR might not have been clear, but that's not supported in this PR. I'll be tackling that next.

packages/plugin-core/src/test/suite-integ/VaultAddCommand.test.ts
SeriousBug: This is not supported yet, I was planning to tackle that in a next step.

packages/plugin-core/src/commands/VaultAddCommand.ts
SeriousBug: My preference here would be to add support for them, because adding support for these is pretty similar to supporting the transitive dependencies for self contained vaults.

> migration plan for the dendron team

My plan was to convert one vault (probably `org-private`, or something else that's private to us) manually to begin with once this PR hits nightly (or general release if we have members who don't use nightly)

Once we have a general release out that has this PR + the ability to open self contained vaults by themselves (which is what I'll be working on next), then I'll go ahead and convert the remaining Dendron vaults as well (hopefully with an automatic convert if I have that implemented by then, manually otherwise)

packages/plugin-core/src/commands/VaultAddCommand.ts
SeriousBug: Yeah my bad, this issue and the naked strings one. I did not change them since it was existing code, but I'll refactor them to leave the code better than I found it (I think that's a value we have, right? :smile: )

test-workspace/dependencies/localhost/vault3/.gitignore
SeriousBug: Are you talking about the empty line at the end?

packages/plugin-core/src/commands/VaultAddCommand.ts
SeriousBug: I think we should either deprecate our custom git wrapper and move towards simplegit, or deprecate simplegit and move towards our custom wrapper. I avoided simeplegit because I'm leaning more towards the custom wrapper, but that's mostly because I wrote a lot of code for it :sweat_smile: 

I don't think we need to encapsulate the common logic, because we're planning to deprecate and remove support for old style vaults (at the very least creating of old style vaults). So the duplicated code will eventually be removed.

packages/plugin-core/src/commands/VaultAddCommand.ts
kevinslin: pass it forward :)

test-workspace/dependencies/localhost/vault3/.gitignore
kevinslin: yep

jonathanyeung:
Nice! - will try to take a look by end of week

kevinslin (changes requested):
looks good.  can we add analytics around:
- if user has self contained vaults enabled (can be prop of initialize workspace event)
- if the user is opening a self contained vault (also prop of initialize workspace event)

SeriousBug:
Thanks for the feedback, I added the telemetry, and added docs to document the feature and the new telemetry.

packages/common-server/src/files.ts
```diff
@@ -70,6 +70,12 @@ export function readYAML(fpath: string): any {
   });
 }
 
+export async function readYAMLAsync(fpath: string): Promise<any> {
```
jonathanyeung, line 73: nit: I think a better naming pattern is to have `readYAML` -> `readYAMLSync`.  The async version you can always tell by the return type.  But the Sync version with any an `any` return type can possibly be either sync or async in nature.  I'm guessing that's why fs-extra follows the pattern of calling out `sync` versions explicitly (and not the async versions). 

Maybe something for later since the other methods in this file already follow the `fooAsync` pattern.

packages/engine-server/src/workspace/service.ts
```diff
@@ -356,6 +359,87 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+   * allow the type to match.
+   * @param addToConfig If true, the created vault will be added to the config
+   * for the current workspace.
+   * @param addToCodeWorkspace If true, the created vault will be added to the
+   * `code-workspace` file for the current workspace.
+   */
+  async createSelfContainedVault(opts: {
+    addToConfig?: boolean;
+    addToCodeWorkspace?: boolean;
+    // Must be created with a self-contained vault
```
jonathanyeung, line 375: cool

packages/engine-server/src/workspace/service.ts
```diff
@@ -356,6 +359,87 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+    await fs.mkdirp(notesPath);
+    await fs.mkdirp(path.join(notesPath, "assets"));
+
+    // Create root note and schema
+    const note = NoteUtils.createRoot({
+      vault,
+      body: ROOT_NOTE_TEMPLATE,
+    });
+    const schema = SchemaUtils.createRootModule({ vault });
+    if (!fs.existsSync(NoteUtils.getFullPath({ note, wsRoot: this.wsRoot }))) {
```
jonathanyeung, line 393: nit: can use async versions.

packages/plugin-core/src/commands/VaultAddCommand.ts
```diff
@@ -133,6 +138,83 @@ export class VaultAddCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+    if (!sourceTypeSelected) {
+      return;
+    }
+    const sourceType = sourceTypeSelected.label;
+
+    const { config } = ExtensionProvider.getDWorkspace();
+    if (config.dev?.enableSelfContainedVaults) {
+      return this.gatherVaultSelfContained(sourceType);
+    } else {
+      // A "standard", non self contained vault
```
jonathanyeung, line 213: super nit: maybe move this comment (or copy it) to the method header of `gatherVaultStandard` - just makes it a little clearer while reading it what 'standard' refers to.

packages/plugin-core/src/test/suite-integ/VaultAddCommand.test.ts
```diff
@@ -428,3 +430,158 @@ suite("VaultAddCommand", function () {
     });
   });
 });
+
+function enableSelfCOntainedVaults(config: IntermediateDendronConfig) {
```
jonathanyeung, line 434: super nit: capitilization on `COntained`

jonathanyeung (changes requested):
looks good - just a few nitpicks.

packages/common-server/src/files.ts
SeriousBug: Agreed with the naming. I'll go ahead and rename those functions too, if it causes any conflicts I can fix them before merging.

jonathanyeung (approved):

kevinslin:
looks good, there's an issue with the merge conflict. feel free to resolve and merge

kevinslin (approved):

PR #2521, opened 2022-03-03
enhance: provide a standard way to create and read backup files

hikchoi:
# enhance: provide a standard way to create and read backup files

This PR:
- implements `BackupService` which can be generally used for various parts of Dendron where we need to back up a file.
  - backup root is created when if it doesn't exist, and added to gitignore
  - one directory for each key defined in `BackupKeyEnum` is created in the backup root, and holds all backups that were done under that key.
  - provides a generic way of creating backup file names, with an option to add timestamp and / or a custom infix right before the file extension.
- implements `OpenBackupCommand` which can be used to open individual backup files grouped by keys stored in the backup root directory.
- fixes an issue wigh `GitUtils.addToGitignore`, where dotfile/dotdirs were not checked for existence properly because the regular expression was not escaped and was using the `g` flag.
- Refactor `DConfig.createBackup` to use the backup service

Madge reports no diffs compared to master

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-server/src/git.ts
+++ b/packages/common-server/src/git.ts
@@ -387,14 +387,17 @@ export class GitUtils {
       // if the .gitignore was missing, ignore it
       if (err?.code !== "ENOENT") throw err;
     }
+
+    // gitignore is missing but we are allowed to create it
+    const shouldCreate = contents === undefined && !noCreateIfMissing;
+
+    // gitignore exists, and the path is not in it yet
     // Avoid duplicating the gitignore line if it was already there
-    const pathExists = contents?.split("\n").indexOf(addPath) !== -1;
-    if (
-      // gitignore is missing but we are allowed to create it
-      (contents === undefined && noCreateIfMissing !== true) ||
-      // gitignore exists, and the path is not in it yet
-      (contents !== undefined && !pathExists)
-    ) {
+    const pathExists =
+      contents !== undefined &&
+      contents.match(new RegExp(`^${_.escapeRegExp(addPath)}/?$`, "m"));
+
+    if (shouldCreate || !pathExists) {
       await fs.appendFile(gitignore, `\n${addPath}`);
     }
   }
--- a/packages/common-server/src/git.ts
+++ b/packages/common-server/src/git.ts
@@ -388,12 +388,12 @@ export class GitUtils {
       if (err?.code !== "ENOENT") throw err;
     }
     // Avoid duplicating the gitignore line if it was already there
+    const pathExists = contents?.split("\n").indexOf(addPath) !== -1;
     if (
       // gitignore is missing but we are allowed to create it
       (contents === undefined && noCreateIfMissing !== true) ||
       // gitignore exists, and the path is not in it yet
-      (contents !== undefined &&
-        !contents.match(new RegExp(`^${addPath}/?$`, "g")))
+      (contents !== undefined && !pathExists)
     ) {
       await fs.appendFile(gitignore, `\n${addPath}`);
     }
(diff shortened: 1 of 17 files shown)


packages/common-server/src/git.ts
```diff
@@ -388,12 +388,12 @@ export class GitUtils {
       if (err?.code !== "ENOENT") throw err;
     }
     // Avoid duplicating the gitignore line if it was already there
```
hikchoi, line 390: noticed dotfile and dotdirs were getting duped because the dot was not in a characer class and not escaped. I briefly thought about fixing this to support all gitignore expressions but that is probably out of scope for this PR 😄 

went the simple way of splitting them by line and checking if it exists in the list.

this PR is still in draft but just wanted to ask if this looks okay or if you would have done it differently @SeriousBug

SeriousBug: I'd actually do this:

```ts
contents.match(new RegExp(`^${_.escapeRegExp(addPath)}/?$`, "m"))
```

The differences are the escape and the "m" flag (accidentally must have used the "g" flag which is not what we want here)

hikchoi: ah didn't think to check if lodash has it already 😅

hikchoi:
note:
- docs for `OpenBackupCommand` pending.
- docs for backup service in `dendron-docs` pending.

packages/plugin-core/src/commands/OpenBackupCommand.ts
```diff
@@ -0,0 +1,103 @@
...
+  private backupService: IBackupService;
+
+  constructor(ext: IDendronExtension) {
+    super();
+    this.extension = ext;
+    const ws = this.extension.getDWorkspace();
+    this.backupService = new BackupService({ wsRoot: ws.wsRoot });
+  }
+
+  async promptBackupEntrySelection(opts: { backups: string[] }) {
```
jonathanyeung, line 26: can we make these methods private?

jonathanyeung (changes requested):

packages/engine-server/src/backup/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./service";
```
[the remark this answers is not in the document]
hikchoi: No real reason it _has_ to be in `engine-server`. `filesv2` is in `common-server` and backup is pretty adjacent to files so that could be one argument for moving it there.

all service classes we implement are in `engine-server` so I followed that convention here.

packages/engine-server/src/backup/service.ts
```diff
@@ -0,0 +1,155 @@
...
+      fs.ensureDirSync(path.join(this.backupRoot, key));
+    });
+  }
+
+  /**
+   * Given some options and a file name, generate a new file name to use for the backup.
+   * @param fileName file name to generate backup name for. Assumes existence of extension.
+   * @param timestamp flag to enable adding timestamp to name.
+   * @param infix optional custom infix to append right before the extension.
+   * @returns generated backup file name.
```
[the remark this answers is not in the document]
hikchoi: added. and also added a way to completely bypass `generateBackupFileName` and just pass in a custom name.

jonathanyeung (approved):

PR #2526, opened 2022-03-04
enhance: Dendron still loads if there's a missing local vault

SeriousBug:
If there's a missing vault, instead of throwing an error and aborting the initialization, Dendron will now skip that vault and continue loading. This includes both the plugin and the CLI.

This is especially useful when publishing since the CLI does not clone missing remote vaults, which breaks publishing even if those vaults are not actually published.

Remove vaults that couldn't be cloned will still abort the initialization in the plugin. As part of self contained vaults, I'll later allow initialization to continue even if cloning a remote vault fails.

### Basics

  - Does not affect circular dependency count

### Extended

- General
  - [~] check if this change adversely impact performance
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

- [~] CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended


packages/common-server/src/files.ts
```diff
@@ -86,33 +91,51 @@ export function globMatch(patterns: string[] | string, fname: string): boolean {
...
+        })
+        .filter(isNotNull),
+      error: null,
+    };
+  } catch (err) {
+    return {
+      error: new DendronError({
+        message: "Error when reading the vault",
+        payload: err,
+        // Marked as minor to avoid stopping initialization. Even if we can't read one vault, we might be able to read other vaults.
```
jonathanyeung, line 126: Can you move this comment to the docstring of the method? That way it's easier for someone using it to know that this is the behavior.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -401,13 +416,22 @@ export class FileStorage implements DStore {
         };
         const vpath = vault2Path({ vault, wsRoot: this.wsRoot });
         // OPT:make async and don't wait for return
-        if (!this.engine.config.noCaching) {
+        // Skip this if we found no notes, which means vault did not initialize
+        if (!this.engine.config.noCaching && notes.length > 0) {
```
jonathanyeung, line 420: the comment doesn't seem to reflect the if statement? `notes.length > 0`

packages/plugin-core/src/commands/ReloadIndex.ts
```diff
@@ -60,10 +61,20 @@ export class ReloadIndexCommand extends BasicCommand<
...
-    this.L.info({ ctx, vaultDir, msg: "creating root schema" });
-    await schemaModuleOpts2File(schema, vaultDir, "root");
-    return AutoFixAction.CREATE_ROOT_SCHEMA;
+    try {
+      const schema = SchemaUtils.createRootModule({ vault });
+      this.L.info({ ctx, vaultDir, msg: "creating root schema" });
+      await schemaModuleOpts2File(schema, vaultDir, "root");
+      return AutoFixAction.CREATE_ROOT_SCHEMA;
+    } catch (err) {
+      this.L.info({
```
jonathanyeung, line 70: should this be error level?

jonathanyeung (changes requested):

packages/engine-server/src/drivers/file/storev2.ts
SeriousBug: We only write notes to cache if there are notes to write, `notes.length > 0`. If `notes.length === 0`, then this block is skipped. Am I misreading the code?

packages/plugin-core/src/commands/ReloadIndex.ts
SeriousBug: This is expected if there are missing vaults. Should I log that at an error level?

packages/engine-server/src/drivers/file/storev2.ts
jonathanyeung: No, I'm being dumb 😀, please ignore.

packages/plugin-core/src/commands/ReloadIndex.ts
jonathanyeung: Maybe keep at info for now.  Besides missing vaults, are there any other potential unexpected errors that may come out of the try block?

jonathanyeung (approved):

kevinslin (approved):

PR #2528, opened 2022-03-04
fix(views): fix race condition in tree view v2 initialization logic

jonathanyeung:
## fix(views): fix race condition in tree view v2 initialization logic

There's a race condition in the Tree View V2 which may cause the Tree View to not initialize properly.  This change makes the data fetching operations on Tree View wait for the engine to be initialized first.

--- 

### Basics

- [N/A] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices).

### Extended
- General
- Operations
- Architecture

### Basics

Tested via loading org-workspace under the debugger.  I'll test again on nightly in case timings are different in the fully packaged extension. 

- [N/A] [Write Tests](dev.process.qa.test) 
- [x] Common cases tested
- [N/A] 1-2 Edge cases tested
- [N/A] If your tests changes an existing snapshot, snapshots have been [updated](dev.process.qa.test)

### Extended
N/A

### DIFF
--- a/packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronTreeExplorerPanel.tsx
@@ -8,6 +8,7 @@ import {
   createLogger,
   TreeViewUtils,
   engineHooks,
+  engineSliceUtils,
 } from "@dendronhq/common-frontend";
 import { Spin, Tree, TreeProps } from "antd";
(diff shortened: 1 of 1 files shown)


SeriousBug (approved):
Did some manual testing, the tree view loads consistently for me now.

kevinslin (approved):

PR #2530, opened 2022-03-04
enhance/enable backlink toggle

kevinslin:
enhance(publish): add configuration to disable backlinks, both across the workspace and on a per note level

- we add `enableBackLinks` in two places:
  - inside `dendron.yml` in `publishing` namespace
  - inside individual notes in `config.global` namespace
- NOTE: despite `enableBackLinks` existing in only `publishing`, it also applies to the preview because both the preview and publishing have the same unified desitnation `HTML` and we currently don't have a way of telling the two apart
  - there's more work to be done for fine tuning this for `preview`. this this isn't something that customers have asked for, we're going to save that work for a next phase
- NOTE: this change is necessary because of issues with backlinks in dendron published pages 

- this also adds a test workspace in `nextjs-tempalte` to facilitate easier testing of `nextjs`. a next item is to create playwright test based on the test workspace inside of `nextjst-template`
- this also refactors common-all so that `constants.ts` is moved under `constants/index.ts` and a bunch of sub exports are consolidated in one place
- this also updates `ci.yml` because of [[Changing File to Folder|dendron://dendron.docs/dev.process.code.gotchas.changing-file-to-folder]]

Docs: https://github.com/dendronhq/dendron-site/pull/425

***

### Basics

### Extended

- General
- Operations
- Architecture

## Instrumentation

NA

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended


hikchoi (commented):
> NOTE: despite enableBackLinks existing in only publishing, it also applies to the preview because both the preview and publishing have the same unified desitnation HTML and we currently don't have a way of telling the two apart

can't we check the ProcFlavor if it's PUBLISHING? (i.e. `MDUtilsV5.shouldApplyPublishingRules(proc)`)

> we add enableBackLinks in two places:
>   inside dendron.yml in publishing namespace
>   inside individual notes in config.global namespace

I mentioned this in https://github.com/dendronhq/dendron/pull/1979#discussion_r775388131 as well, (and it seems like enablePrettyRefs was also added as a note local config) but it's a bit confusing to introduce the global config namespace here for the purpose of per-note configurations.

1. Adding these to the global namespace would complicate things when we decide we want to actually use that namespace for what it was first intended to do (control a certain behavior in both preview and publishing).
1. It is semantically conflicting. 
    - According to the docs, here we are "overriding" the config (https://github.com/dendronhq/dendron-site/pull/425/files#diff-801bbfc119812e4eb022e2b9215f382bc8ecb3bbb721a81efdc91a436fe6aabbR13). There is no `global` namespace in `dendron.yml` to override. It's at this point just a proxy to let us enable per-note behavior.
    - Without full context that doing this is frontmatter is "overriding" a config, the current syntax is suggesting the opposite of what it is doing
        - we put `config.global.enableBackLinks: false` in frontmatter, but we are setting a note _local_ config.
        - this will naturally cause users to raise a question: _why is it called global?_
        - especially since we don't have a global namespace yet.

So these are the two concerns I have with the direction note local configs are going right now. Do you have plans to tackle this? Otherwise, the logic themselves look good.

hikchoi:
I've also left a comment on the docs PR.

kevinslin:
@hikchoi hear you on these concerns. this is why `enableChildLinks` was not formally documented and why this won't be either until we either release a `global` scope in `dendron.yml` or break out `preview` and `publishing` specific config. 

currently, we are doing this to hotfix certain elements of publishing (eg. backlinks on published pages with collections: https://www.kevinslin.com/notes/r4RrLx4xTM2Sc0yOouLkb/)

hikchoi:
> @hikchoi hear you on these concerns. this is why `enableChildLinks` was not formally documented and why this won't be either until we either release a `global` scope in `dendron.yml` or break out `preview` and `publishing` specific config.
> 
> currently, we are doing this to hotfix certain elements of publishing (eg. backlinks on published pages with collections: https://www.kevinslin.com/notes/r4RrLx4xTM2Sc0yOouLkb/)

Got it. Then it looks good 👍🏼

jonathanyeung (changes requested):

jonathanyeung (approved):

PR #2543, opened 2022-03-09
fix(basics): improve perf around reference rendering including hover

jonathanyeung:
## fix(basics): improve perf around reference rendering including hover

This change improves perf / reduces mem usage whenever a reference is being rendered. This includes hover preview, which uses the note reference rendering implementation.

The issue is with `proc()`, which is a copy constructor - this is actually an expensive call when you have a lot of notes in your workspace, because it does a deep copy of `proc.data` - in our case, this includes the entirety of `DEngineClient`, that is, all notes, schemas, vaults, etc.  We should make it a principle to not use `proc()` when we're in full mode and have `DEngineClient` in the data payload.

For org-private, this reduces hover preview time and reference preview rendering down from 500-1200ms range to < 50 ms.  From some profiling, the memory impact is not as significant as I expected (probably due to string interning), but this may also help a bit with some stability issues in the editing experience.

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested
- [ N/A] If your tests changes an existing snapshot, snapshots have been [updated](dev.process.qa.test)

### Extended
N/A
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs
N/A

### Extended
N/A

### DIFF
--- a/packages/engine-server/src/markdown/remark/noteRefsV2.ts
+++ b/packages/engine-server/src/markdown/remark/noteRefsV2.ts
@@ -845,13 +845,17 @@ function convertNoteRefHelperAST(
   } else {
     // Otherwise, just clone the existing proc instead of creating a new one.
     // This will largely preserve existing opts, we just need to change a few.
-    noteRefProc = proc();
-    // proc is the parser that was parsing the note the reference was in, so need to update fname to reflect that we are parsing the referred note
-    MDUtilsV5.setProcData(noteRefProc, {
-      fname: note.fname,
-      insideNoteRef: true,
-      vault: note.vault,
-    });
+
+    noteRefProc = MDUtilsV5.procRemarkFull(
+      {
+        ...MDUtilsV5.getProcData(proc),
+        insideNoteRef: true,
+        fname: note.fname,
+        vault: note.vault,
+        engine,
+      },
+      MDUtilsV5.getProcOpts(proc)
+    );
   }
   const wsRoot = engine.wsRoot;
   noteRefProc = noteRefProc.data("fm", MDUtilsV5.getFM({ note, wsRoot }));
--- a/packages/engine-server/src/markdown/remark/noteRefsV2.ts
+++ b/packages/engine-server/src/markdown/remark/noteRefsV2.ts
@@ -821,42 +821,27 @@ function convertNoteRefHelperAST(
   opts: ConvertNoteRefHelperOpts & { procOpts: any }
 ): Required<RespV2<Parent>> {
   const { proc, refLvl, link, note } = opts;
-  const { dest } = MDUtilsV5.getProcData(proc);
   let noteRefProc: Processor;
   // Workaround until all usages of MDUtilsV4 are removed
   const engine =
     MDUtilsV5.getProcData(proc).engine ||
     MDUtilsV4.getEngineFromProc(proc).engine;
-  if (dest === DendronASTDest.HTML) {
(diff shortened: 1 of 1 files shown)


packages/engine-server/src/markdown/remark/noteRefsV2.ts
```diff
@@ -845,13 +845,17 @@ function convertNoteRefHelperAST(
   } else {
     // Otherwise, just clone the existing proc instead of creating a new one.
```
jonathanyeung, line 846: @SeriousBug, @kevinslin - I'm not as familiar w/ the MD code, but this change may put noteRefProc in a slightly different state than before.  Can you think of any issues with the logic?  I tested in test-workspace, particularly [[Links|dendron.ref.links]], and it looks to be parity in preview + hover from what I can tell.

Another question; zooming out to view the code right above this change:
```typescript
  if (dest === DendronASTDest.HTML) {
    // For HTML, we need to make sure that we don't use a processor with HTML
    // target. Otherwise as we process recursive references, the HTML output
    // from deeper levels is broken (everything gets converted into `<div>`s for
    // some reason)
    noteRefProc = MDUtilsV5.procRemarkFull(
      {
        ...MDUtilsV5.getProcData(proc),
        insideNoteRef: true,
        fname: note.fname,
        vault: note.vault,
        engine,
      },
      MDUtilsV5.getProcOpts(proc)
    );
  } else {
    // Otherwise, just clone the existing proc instead of creating a new one.
    // This will largely preserve existing opts, we just need to change a few.
    noteRefProc = proc();
    // proc is the parser that was parsing the note the reference was in, so need to update fname to reflect that we are parsing the referred note
    MDUtilsV5.setProcData(noteRefProc, {
      fname: note.fname,
      insideNoteRef: true,
      vault: note.vault,
    });
  }
  ```
  
After my change, the logic is actually the same within the if and else branch, so we can collapse it.  But I wanted to check prior to deleting the `// For HTML, we need to make sure...` comment.  How does this work?  I'm not sure what state in `proc` has been altered / is necessary to alter to support the `dest === DendronASTDest.HTML` scenario.

Comment (SeriousBug) [reply packages/engine-server/src/markdown/remark/noteRefsV2.ts:846]: Creating a new proc used to cause bugs because some plugins would not be created in the new proc, which is something I had to fix by reverting to this. The issues are generally caused by MDUtilsV4/V5 mismatch, and by any code that attaches a plugin to the processor outside of `procRemark...` functions. I don't quite remember exactly what bug it was fixing though, perhaps it's not a concern anymore if we already refactored those other code that was giving problems.

Although, create a new processor instead of cloning the existing one?

For the code above, the rest of the comment explains it:
```
    // ... Otherwise as we process recursive references, the HTML output
    // from deeper levels is broken (everything gets converted into `<div>`s for
    // some reason)
```
If you do remove it, you should check that nested note references work correctly. You'll likely have to update this branch to override the dest if the dest is `HTML`.

kevinslin: I tested this the if/else collapsed and can confirm recursive references work fine

kevinslin: feel free to collapse and merge

kevinslin (approved):

Issue #2531, opened 2022-03-05
Reference anchor shows up the in preview but it's not supposed to


Non-J:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [x] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When trying to reference an item of a list, the anchor used for reference shows up in the preview. See the attached screenshot below.

## To Reproduce

1. Create a new note.
2. Write the markdown as shown in the screenshot. 
3. Show the preview
4. See the problem

## Expected behavior

The anchor shouldn't show up in the preview, or at least be styled differently so as not to be confused with the actual content.

## Screenshots

![image](https://user-images.githubusercontent.com/45286033/156861928-5980cf61-2df7-4b46-850c-29d9cbe155f1.png)

## Desktop (please complete the following information)

- OS: Windows
- Browser (if with published websites; include version): N/A
- Versions (all that apply to bug):
  - Dendron Extension: v0.84.0
  - VS Code version: v1.65.0
  - Dendron CLI: N/A

## Additional context

I've asked this on discord where I've been told this is a regression and to submit a bug report.

https://discord.com/channels/717965437182410783/735365126227493004/949277439371141191

## Dendron Log file

https://gist.github.com/Non-J/88f00e03876b0f2c2e4645ab48be9b7e

jonathanyeung:
Thanks for the suggestion @Non-J.

SeriousBug:
Hi, this should be fixed in Dendron 0.86. Please let us know if it continues to happen after you update, otherwise this issue will automatically close in a few days.

Non-J:
Yes, this issue has been fixed in the latest version. Thank you.

PR #2548, opened 2022-03-09
fix: block anchors showing up in the preview

SeriousBug:
Fixes an issue where the block anchors would show up in the preview.

#2531

Co-authored-by: tma66 <tma66@users.noreply.github.com>

### Basics

  - Same number

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- [~] CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/engine-server/src/markdown/remark/dendronPub.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPub.ts
@@ -395,11 +395,7 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
       }
       if (node.type === DendronASTTypes.BLOCK_ANCHOR) {
         // no transform
-        if (
-          dest === DendronASTDest.MD_ENHANCED_PREVIEW ||
-          dest === DendronASTDest.MD_REGULAR ||
-          dest === DendronASTDest.MD_DENDRON
-        ) {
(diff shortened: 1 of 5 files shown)


tma66 (approved):

packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -397,7 +397,8 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
         // no transform
         if (
           dest === DendronASTDest.MD_ENHANCED_PREVIEW ||
```
jonathanyeung, line 399: Is it better to change this now to `if(dest !== DendronASTDest.HTML)`?

jonathanyeung:
I notice there's an extra column in a table, but we can worry about that later: 

image

jonathanyeung (approved):

SeriousBug:
The table issue must have been a problem beforehand as well. I'll add a backlog item to fix that.

kevinslin (approved):

PR #2555, opened 2022-03-10
fix: Re-enable inactive user survey and store prompt status in filesystem for prompt reliability.

hikchoi:
# fix: Re-enable inactive user survey and store prompt status in filesystem for prompt reliability.
This PR:
- Removes the use of state service when storing prompt status of inactive user survey.
  - Use metadata service instead
- Re-enables inactive user survey prompt during activation.
- Change re-prompt cadence to 4 weeks from 2 weeks.
- Ignores a rare case where first install time is later than first lookup time
  - this has caused issues in around 4% of the observed prompted users
- Handle case for recently-prompted users that have `inactiveUserMsgSendTime` set, but does not yet have the new metadata `inactiveUserMsgStatus` set.
  - In a similar case, where we introduced the global state and didn't backfill the in-between cases, 12% of the total observed prompted users were prompted prematurely.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -30,6 +30,7 @@ import {
 import {
   FileAddWatcher,
   HistoryService,
+  InactvieUserMsgStatusEnum,
   MetadataService,
   MigrationChangeSetStatus,
   MigrationUtils,
@@ -981,7 +982,7 @@ export function shouldDisplayInactiveUserSurvey(): boolean {
   const metaData = MetadataService.instance().getMeta();
 
   const inactiveSurveyMsgStatus = metaData.inactiveUserMsgStatus;
-  if (inactiveSurveyMsgStatus === "submitted") {
+  if (inactiveSurveyMsgStatus === InactvieUserMsgStatusEnum.submitted) {
     return false;
   }
 
@@ -1034,7 +1035,7 @@ export function shouldDisplayInactiveUserSurvey(): boolean {
     CUR_TIME.minus(LAST_LOOKUP_TIME) >= FOUR_WEEKS;
 
   // if they have cancelled last time, we should be waiting another four weeks.
-  if (inactiveSurveyMsgStatus === "cancelled") {
+  if (inactiveSurveyMsgStatus === InactvieUserMsgStatusEnum.cancelled) {
     const shouldSendAgain =
       INACTIVE_USER_MSG_SEND_TIME !== undefined &&
       CUR_TIME.minus(INACTIVE_USER_MSG_SEND_TIME) >= FOUR_WEEKS &&
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -979,12 +979,10 @@ export async function showLapsedUserMessage(assetUri: vscode.Uri) {
 }
 
 export async function shouldDisplayInactiveUserSurvey(): Promise<boolean> {
-  const inactiveSurveySubmitted = await StateService.instance().getGlobalState(
-    GLOBAL_STATE.INACTIVE_USER_SURVEY_SUBMITTED
-  );
+  const metaData = MetadataService.instance().getMeta();
 
-  // don't display if they have submitted before.
-  if (inactiveSurveySubmitted === "submitted") {
+  const inactiveSurveyMsgStatus = metaData.inactiveUserMsgStatus;
+  if (inactiveSurveyMsgStatus === "submitted") {
     return false;
   }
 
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -976,21 +975,29 @@ export async function showLapsedUserMessage(assetUri: vscode.Uri) {
...
-  // don't display if they have submitted before.
-  if (inactiveSurveySubmitted === "submitted") {
+  const inactiveSurveyMsgStatus = metaData.inactiveUserMsgStatus;
+  if (inactiveSurveyMsgStatus === "submitted") {
     return false;
   }
 
+  // rare case where global state has been reset (or a reinstall) may cause issues with
+  // the prompt logic. ignore these cases and don't show the
+  if (
```
kevinslin, line 988: can you create a followup task to fix this in the future? syouldn't be part of this pr but we do want to fix these cases

packages/plugin-core/src/survey.ts
```diff
@@ -607,19 +608,13 @@ export class SurveyUtils {
             "https://airtable.com/shry4eLgvVE6WR0Or?prefill_SurveyName=InactiveFeedback";
           VSCodeUtils.openLink(AIRTABLE_URL);
 
-          await StateService.instance().updateGlobalState(
-            GLOBAL_STATE.INACTIVE_USER_SURVEY_SUBMITTED,
-            "submitted"
-          );
+          MetadataService.instance().setInactiveUserMsgStatus("submitted");
```
kevinslin, line 611: can we also deprecate `StateService`? i'm suspect of all data from it - we should really be moving everything into `MetadataService`

kevinslin (approved):

packages/plugin-core/src/_extension.ts
hikchoi: [[Metadata Service Is Unaware of Reinstalls|dendron://private/task.2022.03.14.metadata-service-is-unaware-of-reinstalls]]

packages/plugin-core/src/survey.ts
hikchoi: I'll mark it as deprecated for now, and add it to [[Impactful Change Notice|dendron://dendron.docs/dev.ref.impactful-change-notice]] (+async)

hikchoi: - notice: [[Deprecation of StateService|dendron://dendron.docs/dev.ref.impactful-change-notice#deprecation-of-stateservice]]
- task: [[Deprecate Stateservice and Migrate to Metadataservice|dendron://private/task.2022.03.14.deprecate-stateservice-and-migrate-to-metadataservice]]

packages/plugin-core/src/survey.ts
```diff
@@ -607,19 +608,13 @@ export class SurveyUtils {
...
           vscode.window.showInformationMessage(
             "Thanks for helping us make Dendron better 🌱"
           );
           AnalyticsUtils.track(SurveyEvents.InactiveUserSurveyAccepted);
         } else {
-          await StateService.instance().updateGlobalState(
-            GLOBAL_STATE.INACTIVE_USER_SURVEY_SUBMITTED,
-            "cancelled"
-          );
+          MetadataService.instance().setInactiveUserMsgStatus("cancelled");
```
SeriousBug, line 617: nitpick: should "submitted" and "cancelled" here be an enum?

SeriousBug (commented):
Not sure I understand the issue to judge whether the PR solves it or not. The code itself is fine.

~~I wasn't able to trigger the prompt during manual testing by manually editing `medata.json`, but I might not have been setting it to the right values to trigger it.~~ Ah nevermind, it triggered when I launched a new regular VSCode window (not development). I guess we just don't trigger it during development which makes sense.

packages/plugin-core/src/survey.ts
hikchoi: `Metadata.inactiveUserMsgStatus` has a union type of `"submitted" | "cancelled"`. I can't think of any immediate benefits of making this an enum, but I don't have any strong opinions on this. Do we have a convention around this? I did add some conventions around using types and enums in the context of things we expose with config, but I don't think we can apply those generally.

Open to suggestions.

kevinslin: so the *right* thing to do for all strings is to have them be localized instead of using raw strings. since its a user facing prompt, would not use an enum. for now, would just keep this as is

hikchoi: This is not a user facing prompt. it is the metadata values that we set to store the state of the past prompts.

kevinslin: ah, i see. in that case, enum makes sense

hikchoi: switched them to an enum

kevinslin (approved):

kevinslin (approved):

PR #2563, opened 2022-03-12
chore(analytics): add release and session to sentry

kevinslin:
chore(analytics): add release and session to sentry

- this adds `sessionId` as a tag in sentry 
- this also adds `release` to sentry so we can start tracking [release health](https://docs.sentry.io/product/releases/health/)
- cleanup
    - converts app targets like "vscode" and "cli" into an enum
    - since we bumped the min version, updated the following
        - remove check for old version of vscode when checking for workspace trust
        - remove `VSCodeEvents.UserOnOldVSCodeVerUnblocked` event 
    - remove commented code

Next Items
- [ ] SOP on how to make use of release health data
- [ ] track `sessionId` in server (this is a bigger task and out of scope for this pr)
- [ ] add sentry to cli
***

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-server/src/errorReporting.ts
+++ b/packages/common-server/src/errorReporting.ts
@@ -1,6 +1,7 @@
 import { DendronError, Stage } from "@dendronhq/common-all";
 import { RewriteFrames } from "@sentry/integrations";
 import * as Sentry from "@sentry/node";
+import { CaptureContext } from "@sentry/types";
 import _ from "lodash";
 
 // Extracted to make testing easy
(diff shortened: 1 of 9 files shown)


packages/common-server/src/errorReporting.ts
```diff
@@ -39,10 +40,29 @@ export function isBadErrorThatShouldBeSampled(
...
+  environment: Stage;
+  sessionId?: number;
+  release: string;
+}): void {
   const dsn =
     "https://bc206b31a30a4595a2efb31e8cc0c04e@o949501.ingest.sentry.io/5898219";
 
+  const initialScope: CaptureContext = {};
+  if (sessionId) {
+    initialScope.tags = { sessionId };
```
SeriousBug, line 63: For future reference: if I wanted to add more properties to the user in Sentry (for example whether they have self contained vaults enabled or not), would this be where to add it?

SeriousBug (approved):
Looks good to me, just curious about one thing.

packages/common-server/src/errorReporting.ts
kevinslin: yep, just wrote docs as part of an async

PR #2569, opened 2022-03-15
enhance: initialize workspace can create self contained vaults

SeriousBug:
This PR:
- Enhances Initialize Workspace command to create self contained vaults
  - The feature is behind a VSCode config `dendron.enableSelfContainedVaultWorkspace`. Had to use a VSCode config, since you have to be able to activate this without having a workspace.
  - The first vault in the workspace is used as the top level vault
  - If there are any other vaults (say, part of a template), they are added as local dependencies.
  - The created workspace has `selfContainedVaults` dev config enabled, so any new vaults added will also be self contained vaults
- Refactors existing initialization tests to use the new test harnesses
- Adds a test for self contained vault workspaces
- Adds an option for plugin tests to create self contained vaults in `describeSingleWS` and `describeMultiWS`
- Adds self contained vault support to the CLI vault create

Docs: https://github.com/dendronhq/dendron-site/pull/452

### Basics

  - Unchanged

### Extended

- General
- Operations
  - [~] when shipping this change, will it just work or will it introduce additional operational overhead due to complicated interface or known bugs?
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended


packages/common-server/src/filesv2.ts
```diff
@@ -484,11 +484,16 @@ export const vault2Path = ({
...
 
-export function writeJSONWithComments(fpath: string, data: any) {
+export function writeJSONWithCommentsSync(fpath: string, data: any) {
   const payload = stringify(data, null, 4);
   return fs.writeFileSync(fpath, payload);
 }
 
+export async function writeJSONWithComments(fpath: string, data: any) {
+  const payload = stringify(data, null, 4);
+  await fs.writeFile(fpath, payload);
```
kevinslin, line 494: nitpick: can we just return the promise instead of awaiting?

packages/engine-server/src/workspace/service.ts
```diff
@@ -859,6 +877,84 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+      wsVault.fsPath = ".";
+      wsVault.selfContained = true;
+
+      // Mutate vault objects to convert them to self contained vaults. The
+      // first vault will be skipped because the conversion is a no-op for
+      // vaults that are already self contained.
+      const selfContainedVaults = vaults.map(
+        WorkspaceService.standardToSelfContainedVault
+      );
+      for (const vault of selfContainedVaults) {
```
kevinslin, line 939: can we use [[asyncLoopOneAtATime|dendron://dendron.docs/pkg.common-server#asynclooponeatatime]] to avoid the `eslint-disable`?

packages/plugin-core/src/ExtensionProvider.ts
```diff
@@ -24,6 +24,15 @@ export class ExtensionProvider {
     return ExtensionProvider.extension;
   }
 
+  /** Returns true if the extension is registered.
+   *
+   * This should be true most of the time, but you may need to check first in
+   * commands that run before or during initialization.
+   */
+  static isRegistered() {
```
kevinslin, line 32: can we add this to [[Cook|dendron://dendron.docs/pkg.plugin-core.dev.cook]] and also note it in an async?

packages/plugin-core/src/test/suite-integ/Extension.test.ts
```diff
@@ -169,77 +168,73 @@ suite("GIVEN SetupWorkspace Command", function () {
...
-      userConfigDirStub = mockUserConfigDir();
-      wsFoldersStub = stubWSFolders(undefined);
-    },
-    afterHook: async () => {
-      homeDirStub.restore();
-      userConfigDirStub.restore();
-      wsFoldersStub.restore();
-    },
-    noSetInstallStatus: true,
+  this.timeout(6 * 1000);
```
kevinslin, line 171: can we pass this into `describreMultiWS` as a constant instead of setting it here?

packages/plugin-core/src/ExtensionProvider.ts
SeriousBug: Documented and created a PR for it. Will make an async once this is merged.

packages/plugin-core/src/commands/SetupWorkspace.ts
```diff
@@ -184,11 +192,20 @@ export class SetupWorkspaceCommand extends BasicCommand<
 
   async execute(opts: CommandOpts): Promise<DVault[]> {
     const ctx = "SetupWorkspaceCommand extends BaseCommand";
+    // This command can run before the extension is registered, especially during testing
```
jonathanyeung, line 195: Maybe this is less of a problem since this is just a feature flag and eventually we're saying this will be the default so we don't need to check settings anymore - but since the setting is in VSCode settings, we don't actually need the extension up in order to read the setting right? Can we just read the VSCode setting rather than our own workspace config file?

packages/plugin-core/src/ExtensionProvider.ts
jonathanyeung: I'm a bit weary of us using this in the wrong way - I think we may eventually get into initialization race conditions if we use this broadly and in the wrong way.  I also think of this ExtensionProvider static class as a temporary solution until we can have a better dependency injection model. 

I'm ok to leave it, but is it really necessary to have this per the other comment around accessing vs code settings w/o the Extension being activated?

packages/plugin-core/src/commands/SetupWorkspace.ts
SeriousBug: You're right, it should just look at the VSCode setting. Let me refactor, which will also remove the need for `isRegistered` since VSCode already has the config ready before the extension even initializes.

packages/plugin-core/src/ExtensionProvider.ts
SeriousBug: Makes sense. See my other comment, removing this interface. I instead moved the function to `VSCodeUtils`, and deprecated `ExtensionProvider.getWorkspaceConfig()` to avoid the dependency on the extension.

kevinslin (approved):

PR #2578, opened 2022-03-16
feat: add doctor command for a more reliable keybinding resolution

hikchoi:
# feat: add doctor command for a more reliable keybinding resolution

This PR:
- Removes automatic keybinding conflict resolution in favor of manual resolution.
- The main goal of this PR is to give back user agency while also providing a streamlined way of manually resolving conflicts.

- adds doctor command `fixKeybindingConflicts`
  - detects known extensions that have keybinding conflicts
  - provide detailed instructions on how to resolve them
- removes automatic keybinding conflict resolution that happens during extension activation

## Notes

- `dendron.copyToClipboard`
  - This is a command that is added to provide a way to copy arbitrary text to system clipboard within a webview.
  - This command is not accessible from any user-facing VSCode UI. (`when` clause is `false`)

- `KNOWN_KEYBINDING_CONFLICTS`
  - This holds all the mappings between a known pair of conflicting keybindings.

- `VSCodeUtils.getCodeUserConfigDir`
  - Currently there isn't a straightforward way to know for sure what the exact user data directory is when given a vscode session. (We are making a best guess)
  - As such, moving forward _we should not try to automatically create a `keybindings.json` file if it doesn't exist._
    - This used to be the case but not anymore since we are not automatically resolving keybinding conflicts anymore.
    - If we absolutely have to, we should do so by doing `vscode.commands.executeCommand("workbench.action.openGlobalKeybindingsFile")`, and delegating the creation to VSCode. 
      - At this point we can figure out the exact path of user-data-dir, but unfortunately the file has to be opened in the editor. This would be fine if we incorporate the file open in the user experience flow.
 
## Flow

> To manually test this, pull this branch in and `Run Extension: Remote (plugin-core)` with an extension that has conflicts installed. Currently the only one is `vscodevim`. If you are on a mac there won't be any conflicts, but you can uncomment [this](https://github.com/dendronhq/dendron/pull/2578/files#diff-248f919f41b45daa5d918842ab7dc039c336356b37665bc387826e4048922fe5R1048) to try it out.

1. During activation (only on initial install), we try to detect known keybinding conflicts. This takes into account both default and user defined keybindings.
2. If a conflict is detected, a toast message will appear with a button that will trigger the doctor command.
  image
3. When the button is clicked, the default keybinding (JSON), and a webview will be opened. (_**note that `extension.vim_tab` is not actually a conflict. This is for demonstration purposes only**_)
  image
4. Instructions on how to resolve the detected conflicts is described on the top of the webview (above the thematic break)
  image
5. The link `keybindings.json` is linking to a command uri that will open `keybindings.json`
6. List of detected conflicts and a set of links that will copy a JSON block is presented below
  image
7. When clicked, a toast message will be shown
  image
8. This could be pasted into `keybindings.json` directly as the necessary disable (or remap) entry is already in the clipboard.
   image

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/KeybindingUtils.ts
+++ b/packages/plugin-core/src/KeybindingUtils.ts
@@ -1,4 +1,9 @@
-import { DendronError, ERROR_SEVERITY } from "@dendronhq/common-all";
+import {
+  DendronError,
+  ERROR_SEVERITY,
+  ExtensionEvents,
+  KeybindingConflictDetectedSource,
+} from "@dendronhq/common-all";
 import {
   CommentJSONValue,
   readJSONWithCommentsSync,
@@ -22,6 +27,7 @@ import {
   CopyToClipboardCommandOpts,
   CopyToClipboardSourceEnum,
 } from "./commands/CopyToClipboardCommand";
+import { AnalyticsUtils } from "./utils/analytics";
 
 type Keybindings = Record<string, string>;
 export class KeybindingUtils {
@@ -256,7 +262,10 @@ export class KeybindingUtils {
       .showWarningMessage(message, action)
       .then(async (resp) => {
         if (resp === action) {
+          AnalyticsUtils.track(ExtensionEvents.ShowKeybindingConflictAccepted);
           await this.showKeybindingConflictPreview(opts);
+        } else {
+          AnalyticsUtils.track(ExtensionEvents.ShowKeybindingConflictRejected);
         }
       });
   }
@@ -266,6 +275,9 @@ export class KeybindingUtils {
       knownConflicts: KNOWN_KEYBINDING_CONFLICTS,
     });
     if (conflicts.length > 0) {
+      AnalyticsUtils.track(ExtensionEvents.KeybindingConflictDetected, {
+        source: KeybindingConflictDetectedSource.activation,
+      });
       await KeybindingUtils.showKeybindingConflictConfirmationMessage({
         conflicts,
       });
--- a/packages/plugin-core/src/KeybindingUtils.ts
+++ b/packages/plugin-core/src/KeybindingUtils.ts
@@ -184,6 +184,8 @@ export class KeybindingUtils {
       `1. Open [keybindings.json](${keybindingJSONCommandUri})`,
       `1. Paste the clipboard content to \`keybindings.json\``,
       "",
+      `[Video Guide](https://www.loom.com/embed/82d3fbccd126446bac7a4d16027c07aa)`,
+      "",
       "For more information on how to set keyboard rules in VSCode, visit [Keyboard Rules](https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-rules)",
       "",
       `Use [Default Keybindings](${defaultKeybindingJSONCommandUri}) to reference all default keybindings.`,
--- a/packages/plugin-core/src/KeybindingUtils.ts
+++ b/packages/plugin-core/src/KeybindingUtils.ts
@@ -63,9 +63,8 @@ export class KeybindingUtils {
       return conflicts;
     }
 
-    const userKeybindingConfig = readJSONWithCommentsSync(keybindingConfigPath);
+    // const userKeybindingConfig = readJSONWithCommentsSync(keybindingConfigPath);
 
-    // TODO: check for existing user keybinding items
     return conflicts;
   }
 
@@ -208,6 +207,8 @@ export class KeybindingUtils {
       }
     );
     panel.webview.html = md.render(contents);
+
+    // TODO: add proper tracking
     // AnalyticsUtils.track(
     //   ExtensionEvents.IncompatibleExtensionsPreviewDisplayed
     // );
@@ -247,49 +248,6 @@ export class KeybindingUtils {
     return true;
   }
 
-  static getKeybindingConfig = ({
-    createIfMissing = false,
-  }: {
-    createIfMissing: boolean;
-  }): RespV3<CommentJSONArray<Keybindings>> => {
-    const ctx = "getKeybindingConfig";
-    const { keybindingConfigPath } = this.getKeybindingConfigPath();
-
-    if (!fs.existsSync(keybindingConfigPath) && createIfMissing) {
-      fs.ensureFileSync(keybindingConfigPath);
-      fs.writeFileSync(keybindingConfigPath, "[]");
-      Logger.info({ ctx, keybindingConfigPath, msg: "creating keybindings" });
-    }
-
-    if (!fs.existsSync(keybindingConfigPath)) {
-      return {
-        error: DendronError.createFromStatus({
-          status: ERROR_STATUS.INVALID_STATE,
-          message: "no keybinding config found",
-        }),
-      };
-    }
-    try {
-      const keybindings = readJSONWithCommentsSync(keybindingConfigPath);
-      if (!KeybindingUtils.checkKeybindingsExist(keybindings)) {
-        return {
-          error: DendronError.createFromStatus({
-            status: ERROR_STATUS.INVALID_CONFIG,
-            message: "keybinding config invalid",
-          }),
-        };
-      }
-      return { data: keybindings };
-    } catch (err) {
-      return {
-        error: DendronError.createFromStatus({
-          status: ERROR_STATUS.INVALID_CONFIG,
-          message: "keybinding config invalid",
-        }),
-      };
-    }
-  };
-
   static getKeybindingConfigPath = () => {
     const { userConfigDir, osName } = VSCodeUtils.getCodeUserConfigDir();
     return {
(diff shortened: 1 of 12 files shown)


kevinslin:
llooks like theres a bunch of tests failing

hikchoi:
Confirmed working as intended on Windows thanks to @Harshita-mindfire

hikchoi:
Update: Slightly improved the presentation of the webview

1. Fixed a few typos
2. Slightly shortened instructions
3. Now opens user keybinding (`keybindings.json`) on the active editor column.
4. Webview is open to the side so that `Default Keybindings` and `keybindings.json` is visible right away
5. Condensed the presentation of `disable` and `remap` command uri links to be beside respective commands

image

packages/plugin-core/src/_extension.ts
```diff
@@ -735,85 +733,14 @@ export async function _activate(
       }
     }
 
-    const backupPaths: string[] = [];
-    let keybindingPath: string;
-
-    // check vim keybindings ^j7zygfkbxjnq
```
jonathanyeung, line 741: any docs we need to clean up with this block anchor removal?

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,322 @@
...
     }
+    await KeybindingUtils.openGlobalKeybindingFileAndGetJSON({
+      close: false,
+    });
+    const keybindingJSONCommandUri = `command:workbench.action.openGlobalKeybindingsFile`;
+    const defaultKeybindingJSONCommandUri = `command:workbench.action.openDefaultKeybindingsFile`;
+    const contents = [
+      "# Known Keybinding Conflicts",
+      "",
+      "The keybindings listed below are known to have conflicts with default keybindings for Dendron commands.",
```
jonathanyeung, line 170: nit: maybe instead of 'below', replace with 'at the bottom' - When I first read this my eyes wandered to the #### 1. / #### 2. section.

packages/plugin-core/src/test/suite-integ/Extension.test.ts
```diff
@@ -762,71 +762,71 @@ suite("GIVEN a native workspace", function () {
...
-        beforeAllSymbol
-      );
-      expect(Object.getOwnPropertySymbols(newKeybindings[0])[0]).toEqual(
-        beforeKeySymbol
-      );
-      done();
-    });
-  });
-});
+// suite("keybindings", function () {
```
jonathanyeung, line 765: Can we delete the code instead of commenting it out?

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,322 @@
...
+
+          const copyCommandUri = (args: { text: string }) =>
+            `command:dendron.copyToClipboard?${encodeURIComponent(
+              JSON.stringify(args)
+            )}`;
+
+          const out = [
+            `### \`${conflict.commandId}\``,
+            `- key: \`${conflictKeybindingEntry.key}\``,
+            `- command: \`${conflict.commandId}\` [Disable](${copyCommandUri({
```
jonathanyeung, line 242: nit: maybe expand the description here.  Right now it reads like

```
[Disable extension.vim_tab]
[Remap extension.vim_tab]
[Remap dendron.lookupNoteAutoComplete]
```

But maybe expand to something like 

```
Disable extension.vim_tab - Click to generate JSON.
...
```

So to avoid the confusion that a user may think clicking on the link is enough to fix the issue (in case they didn't properly read the instructions, like myself the first time. 😄 

And maybe in the pop-up that comes up, we can expand it from 'text copied to clipboard' to something like 'text copied, now add to your keybindings.json', and maybe have a button that takes them there as a cherry-on-top.  These are just nits though.

jonathanyeung (changes requested):

packages/plugin-core/src/test/suite-integ/Extension.test.ts
hikchoi: This is temporarily commented out and I'll be replacing these with new tests :)

packages/plugin-core/src/KeybindingUtils.ts
hikchoi: > maybe expand the description here

I went the other way of condensing it and sticking them next to the command ids to make it less busy 😅
Any ideas on how to make it more explicit  and also not too crowded? It's a bit hard to format these with markdown.

> maybe have a button that takes them there as a cherry-on-top.

I could make `CopyToClipboard` accept callbacks to make this work (or just make a separate command), but not sure if we can pass the callback after encoding it as a URI component. I think having a link in the webview like right now would be enough for visibility.

packages/plugin-core/src/_extension.ts
hikchoi: https://github.com/dendronhq/dendron-docs/pull/22

packages/plugin-core/src/KeybindingUtils.ts
jonathanyeung: As I was saying - mostly a nit.  I think your less busy approach is also fine.

hikchoi:
update:

- tests added
- docs added
  - wiki: https://github.com/dendronhq/dendron-site/pull/448
  - docs: https://github.com/dendronhq/dendron-docs/pull/22
    - created an async for this

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
-            message: "keybinding config invalid",
-          }),
-        };
-      }
-      return { data: keybindings };
-    } catch (err) {
+    return globalKeybindingJSON;
+  }
+
+  static getInstallStatus() {
```
kevinslin, line 56: would rename: `eg. getInstallStatusForIncompatibleExtensions`

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
+      const resolvedDendronConflict = conflicts.find(
+        (conflict) => conflict.conflictsWith === command
+      );
+
+      if (resolvedDendronConflict) {
+        alreadyResolved.push(resolvedDendronConflict);
+      }
+    });
+
+    return _.differenceBy(conflicts, alreadyResolved);
```
kevinslin, line 125: this seems very expensive to run on each loop. why can't we just add all conflicts to a set and report if the set is non empty?

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
+      command = keybinding.command;
+      if (command.startsWith("-")) {
+        command = command.substring(1);
+      }
+      const resolvedDendronConflict = conflicts.find(
+        (conflict) => conflict.conflictsWith === command
+      );
+
+      if (resolvedDendronConflict) {
+        alreadyResolved.push(resolvedDendronConflict);
```
kevinslin, line 121: can we add a reason  to `alreadyResolved` for easier debugging? (eg. {cause: "DendronKeyRemap|IncompatibleExtensionKeyRemap", conflict})

also there's an edge case where I remap Dendron key to itself (same keybinding, diff arguments) that this doesn't catch

kevinslin: also more comments here would be nice. took me a while to understand this logic

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
+    const panel = vscode.window.createWebviewPanel(
+      "keybindingConflictPreview",
+      "Keybinding Conflicts",
+      vscode.ViewColumn.Beside,
+      {
+        enableCommandUris: true,
+      }
     );
-    return { data: { keybindingConfigPath, newKeybindings } };
+    panel.webview.html = md.render(contents);
```
kevinslin, line 272: instead of using `md.render`, can we use the dendron preview and render a note? see [[Using the Dendron Preview on arbitrary markdown|dendron://dendron.docs/pkg.plugin-core.dev.cook#using-the-dendron-preview-on-arbitrary-markdown]]

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
+    const contents = [
+      "# Known Keybinding Conflicts",
+      "",
+      "The keybindings listed at the bottom are known to have conflicts with default keybindings for Dendron commands.",
+      "",
+      "Neither Dendron nor the extension may function properly if the keybinding conflict is not resolved.",
+      "",
+      "Consider resolving the keybinding conflicts throught the following methods:",
+      "",
+      "#### 1. Disable conflicting keybindings",
```
kevinslin, line 182: can we also add a link to a loom video demonstrating this?

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
+      "If you don't use the listed keybinding, consider disabling it.",
+      "",
+      `1. Click on the link \`Disable\` next to each conflicting keybinding listed below.`,
+      `    - This will copy the necessary keybinding entry to your clipboard.`,
+      `1. Open [keybindings.json](${keybindingJSONCommandUri})`,
+      `1. Paste the clipboard content to \`keybindings.json\``,
+      "",
+      "#### 2. Remap conflicting keybindings",
```
kevinslin, line 190: see above

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
+              entry: conflictKeybindingEntry,
+              disable: false,
+            });
+
+          const copyCommandUri = (args: { text: string }) =>
+            `command:dendron.copyToClipboard?${encodeURIComponent(
+              JSON.stringify(args)
+            )}`;
+
+          const out = [
```
kevinslin, line 245: can we add telemetry around the clicking of these command Uris?

packages/plugin-core/src/_extension.ts
```diff
@@ -735,85 +733,15 @@ export async function _activate(
...
-              const backupUri = vscode.Uri.file(backupPath);
-              await VSCodeUtils.openFileInEditor(backupUri, {
-                column: vscode.ViewColumn.Beside,
-              });
-            });
-          }
-        });
-    }
-
     if (userOnOldVSCodeVer) {
```
kevinslin, line 736: if you rebase, this block should now be removed

packages/plugin-core/src/KeybindingUtils.ts
```diff
@@ -1,206 +1,330 @@
...
-  //   if (!needsMigration) return { keybindingConfigPath };
-
-  //   migratedKeybindings = assign(keybindings, migratedKeybindings);
-  //   return { keybindingConfigPath, migratedKeybindings };
-  // }
+  static async maybePromptKeybindingConflict() {
+    const conflicts = KeybindingUtils.getConflictingKeybindings({
+      knownConflicts: KNOWN_KEYBINDING_CONFLICTS,
+    });
+    if (conflicts.length > 0) {
```
kevinslin, line 294: telemetry if we do this

packages/plugin-core/src/commands/CopyToClipboardCommand.ts
```diff
@@ -0,0 +1,41 @@
...
+ *   text: "some text",
+ *   message: "copied!"
+ * })}`
+ *
+ * ...
+ *
+ * content = `[click this](${commandUri})`
+ *
+ */
+export class CopyToClipboardCommand extends BasicCommand<
```
kevinslin, line 29: can you add this to [[Cook|dendron://dendron.docs/pkg.plugin-core.dev.cook]]?

packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -357,6 +363,17 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
         await this.showIncompatibleExtensionPreview({ installStatus });
         break;
       }
+      case PluginDoctorActionsEnum.FIX_KEYBINDING_CONFLICTS: {
+        const conflicts = KeybindingUtils.getConflictingKeybindings({
+          knownConflicts: KNOWN_KEYBINDING_CONFLICTS,
+        });
+        if (conflicts.length > 0) {
+          await KeybindingUtils.showKeybindingConflictPreview({ conflicts });
```
kevinslin, line 371: telemetry if we find conflicts

packages/plugin-core/src/KeybindingUtils.ts
hikchoi: Last time we discussed this we established that we currently can't do this as described in that note.

The method described can be used to render arbitrary markdown from a file that exists outside of a Dendron vault.

I can create another task to figure out a way to make pseudo-NoteProps for this purpose, but I would say it is out of scope for this PR.

packages/plugin-core/src/commands/CopyToClipboardCommand.ts
hikchoi: If you check the [docs PR](https://github.com/dendronhq/dendron-docs/pull/22), I have them in there. I also created an async.

packages/plugin-core/src/_extension.ts
hikchoi: I don't think it's removed in master right now. https://github.com/dendronhq/dendron/blob/ca708505d62927f28614ce8cd83fcc515b68496a/packages/plugin-core/src/_extension.ts#L813

packages/plugin-core/src/KeybindingUtils.ts
hikchoi: @SeriousBug I remember you worked on the non-note rendering logic for preview. Is there something I'm missing here about passing arbitrary markdown to our preview? My understanding is that we still need a file to create a pseudo-note for the purpose of rendering.

hikchoi: Assuming passing in arbitrary markdown without having it anywhere in the filesystem works, another problem is that we only have one instance of the preview view in any given session, and we don't want to reuse an existing preview panel to render this content. I'm not sure if we can dynamically register multiples of the same views once the extension is activated.

packages/plugin-core/src/KeybindingUtils.ts
hikchoi: Now we count strict disabling of conflicting command as the only resolution. Also revised the preview descriptions and links accordingly.

packages/plugin-core/src/KeybindingUtils.ts
hikchoi: remap no longer part of the resolution method.

packages/plugin-core/src/commands/CopyToClipboardCommand.ts
hikchoi: I can add a bit more about `CopyToClipboardCommand` itself if you think the docs as-is is a bit lacking.

kevinslin:
looks good - only outstanding comment is now just the loom video

packages/plugin-core/src/KeybindingUtils.ts
hikchoi: Link to [Video Guide](https://www.loom.com/embed/82d3fbccd126446bac7a4d16027c07aa) added

PR #2581, opened 2022-03-16
fix(workspace): race condition when backing up configuration 

kevinslin:
fix(workspace): race condition when backing up configuration 

When ensuring that a backup folder exists, actually await the return result

***

### Basics

### Extended

- General
- Operations
- Architecture

## Instrumentation

NA

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

***

Doc updates: https://github.com/dendronhq/dendron-docs/commit/84ae197d0729315979d54920f0d9b52a3cc13d7e

### DIFF
--- a/packages/engine-server/src/backup/service.ts
+++ b/packages/engine-server/src/backup/service.ts
@@ -1,4 +1,5 @@
 import {
+  asyncLoopOneAtATime,
   DendronError,
   Disposable,
   ERROR_STATUS,
@@ -68,8 +69,8 @@ export class BackupService implements Disposable, IBackupService {
       root: this.wsRoot,
     });
     await fs.ensureDir(this.backupRoot);
-    Object.keys(BackupKeyEnum).forEach(async (key) => {
-      await fs.ensureDir(path.join(this.backupRoot, key));
+    await asyncLoopOneAtATime<string>(Object.values(BackupKeyEnum), (key) => {
+      return fs.ensureDir(path.join(this.backupRoot, key));
     });
   }
 
--- a/packages/engine-server/src/backup/service.ts
+++ b/packages/engine-server/src/backup/service.ts
@@ -21,6 +21,7 @@ export type BackupServiceOpts = {
 
 /**
  * Predefined keys to be used for backups
+ * Loop over values when extracting keys
  * ^6ao9nojre6ai
  */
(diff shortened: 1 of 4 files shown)


packages/engine-server/src/backup/service.ts
```diff
@@ -68,8 +69,8 @@ export class BackupService implements Disposable, IBackupService {
       root: this.wsRoot,
     });
     await fs.ensureDir(this.backupRoot);
-    Object.keys(BackupKeyEnum).forEach(async (key) => {
-      await fs.ensureDir(path.join(this.backupRoot, key));
+    await asyncLoopOneAtATime<string>(Object.values(BackupKeyEnum), (key) => {
+      return fs.ensureDir(path.join(this.backupRoot, key));
```
jonathanyeung, line 73: In this case, do we actually need the logic in the loop to be sequential though?

i.e. versus
```typescript
    await Promise.all(
      Object.keys(BackupKeyEnum).map((key) =>
        fs.ensureDir(path.join(this.backupRoot, key))
      )
    );
```

kevinslin: that's fair. updated

hikchoi:
Thanks for the catch. The ensureDir calls were originally sync, but I changed it last minute to async and forgot to stick a promise.all on them 😓

jonathanyeung (approved):

packages/engine-server/src/backup/service.ts
```diff
@@ -68,8 +69,8 @@ export class BackupService implements Disposable, IBackupService {
       root: this.wsRoot,
     });
     await fs.ensureDir(this.backupRoot);
-    Object.keys(BackupKeyEnum).forEach(async (key) => {
-      await fs.ensureDir(path.join(this.backupRoot, key));
+    await asyncLoop<string>(Object.values(BackupKeyEnum), (key) => {
```
SeriousBug, line 72: Could we add a docstring to `BackupKeyEnum` to say whether it's the key or the value that's supposed to be the path? The switch from the key to value confused me when reviewing at least

SeriousBug (approved):
Looks good, one comment for documentation

hikchoi (approved):

PR #2582, opened 2022-03-16
enhance(pods): support exporting a note to multiple Airtable destination

Harshita-mindfire:
```
enhance(pods): support exporting a note to multiple Airtable destination
enhance(workspace):  Doctor command 'fixAirtableMetadata' to convert airtableId from single scalar value to a hashmap
```
This PR aims to support multiple Airtable Ids in note frontmatter on Airtable export v2 by storing airtable ids in a hashmap
[Airtable Docs](https://github.com/dendronhq/dendron-site/pull/447)

### Basics

- [x] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](dev.process.code.best-practices). No new dependencies added(17)

### Extended
- General
  - [~] check if similar function already exist in the codebase. if so, can it be re-used?
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/engine-server/src/doctor/service.ts
+++ b/packages/engine-server/src/doctor/service.ts
@@ -1,4 +1,5 @@
 import {
+  assertInvalidState,
   DendronError,
   DEngineClient,
   Disposable,
@@ -29,6 +30,7 @@ export enum DoctorActionsEnum {
   REGENERATE_NOTE_ID = "regenerateNoteId",
   FIND_BROKEN_LINKS = "findBrokenLinks",
   FIX_REMOTE_VAULTS = "fixRemoteVaults",
+  FIX_AIRTABLE_METADATA = "fixAirtableMetadata",
 }
 
 export type DoctorServiceOpts = {
@@ -40,6 +42,7 @@ export type DoctorServiceOpts = {
   exit?: boolean;
   quiet?: boolean;
   engine: DEngineClient;
+  podId?: string;
 };
 
 /** DoctorService is a disposable, you **must** dispose instances you create
@@ -133,7 +136,7 @@ export class DoctorService implements Disposable {
   }
 
   async executeDoctorActions(opts: DoctorServiceOpts) {
-    const { action, engine, query, candidates, limit, dryRun, exit } =
+    const { action, engine, query, candidates, limit, dryRun, exit, podId } =
       _.defaults(opts, {
         limit: 99999,
         exit: true,
@@ -329,6 +332,31 @@ export class DoctorService implements Disposable {
           })
         );
         return { exit: true };
+      }
+      case DoctorActionsEnum.FIX_AIRTABLE_METADATA: {
+        // Converts the airtable id in note frontmatter from a single scalar value to a hashmap
+        if (!podId) {
+          assertInvalidState(
+            "Please provide pod Id to perform Doctor operation"
+          );
+        }
+        doctorAction = async (note: NoteProps) => {
+          //get airtable id from note
+          const airtableId = _.get(note.custom, "airtableId") as string;
+          const pods = {
+            airtable: {
+              [podId]: airtableId,
+            },
+          };
+          delete note.custom["airtableId"];
+          const updatedNote = {
+            ...note,
+            custom: { ...note.custom, pods },
+          };
+          // update note
+          engine.writeNote(updatedNote, { updateExisting: true });
+        };
+        break;
       }
       default:
         throw new DendronError({
--- a/packages/engine-server/src/doctor/service.ts
+++ b/packages/engine-server/src/doctor/service.ts
@@ -43,6 +43,8 @@ export type DoctorServiceOpts = {
   quiet?: boolean;
   engine: DEngineClient;
   podId?: string;
+  hierarchy?: string;
+  vault?: DVault | string;
 };
 
 /** DoctorService is a disposable, you **must** dispose instances you create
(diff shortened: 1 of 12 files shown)


packages/engine-server/src/doctor/service.ts
```diff
@@ -330,6 +333,31 @@ export class DoctorService implements Disposable {
...
+      case DoctorActionsEnum.FIX_AIRTABLE_METADATA: {
+        // Converts the airtable id in note frontmatter from a single scalar value to a hashmap
+        if (!podId) {
+          assertInvalidState(
+            "Please provide pod Id to perform Doctor operation"
+          );
+        }
+        doctorAction = async (note: NoteProps) => {
+          //get airtable id from note
+          const airtableId = _.get(note.custom, "airtableId") as string;
```
SeriousBug, line 345: I think this should check if the note has an airtable ID already, and skip notes that don't have them. Looks likewe do that filtering in the plugin code, but I think we can do it here instead so the CLI code also benefits from it.

packages/pods-core/src/builtin/AirtablePod.ts
```diff
@@ -402,28 +426,50 @@ export class AirtableUtils {
...
-        if (!noteAirtableId) {
-          const updatedNote = {
-            ...note,
-            custom: { ...note.custom, airtableId },
+        let pods: AirtablePodMetadata = _.get(note.custom, "pods");
+        // return if the note is already exported ie. update query is ran.
+        if (pods && pods.airtable && pods.airtable[podId]) return undefined;
+
+        // if this is the first time a pod metadata is added to the note, add airtable pod metadata under pods namespace
+        if (!pods) {
```
SeriousBug, line 447: Could we add some backwards compatibility here? If the old airtable frontmatter without the pods namespace exists, use the existing frontmatter. That way we don't have to immediately fix all notes with the doctor, just ones where we want to export to multiple targets.

SeriousBug (changes requested):
Looks good, left a couple comments for things I think could be improved

packages/plugin-core/src/components/pods/PodControls.ts
```diff
@@ -533,4 +533,18 @@ export class PodUIControls {
         assertUnreachable(type);
     }
   }
+  // Prompt user to select custom pod Id
+  public static async promptToSelectCustomPodId(): Promise<string | undefined> {
+    const configs = PodV2ConfigManager.getAllPodConfigs(
+      path.join(getExtension().podsDir, "custom")
```
kevinslin, line 539: can we use `ExtensionProvider` here?

packages/engine-test-utils/src/__tests__/pods-core/v2/AirtableExportPodV2.spec.ts
```diff
@@ -629,3 +634,118 @@ describe("GIVEN export note with date ", () => {
...
+        };
+        const resp = await setupTest(preSetupHook);
+        await checkString(
+          resp.error!.message,
+          "The following notes are missing airtable ids: dendron://vault1/task.alpha (task.alpha)"
+        );
+      });
+    });
+
+    describe("AND linked note has airtable id", () => {
```
kevinslin, line 678: can we also test the case where the linked note is associated with multiple records? linking to the note should get the record from the matching podId

packages/plugin-core/src/components/pods/PodControls.ts
```diff
@@ -533,4 +533,18 @@ export class PodUIControls {
         assertUnreachable(type);
     }
   }
+  // Prompt user to select custom pod Id
```
jonathanyeung, line 536: nit: can you convert this to ts-doc style so it shows up in intellisense? 

```typescript
/**
* comment
*/
```

packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -431,6 +437,35 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+        const candidates = Object.values(notes).filter(
+          (value) =>
+            value.fname.startsWith(hierarchy) &&
+            value.stub !== true &&
+            VaultUtils.isEqualV2(value.vault, vault) &&
+            value.custom.airtableId
+        );
+        this.L.info({
+          ctx,
+          msg: `${DoctorActionsEnum.FIX_FRONTMATTER} candidates: ${candidates.length}`,
```
jonathanyeung, line 458: maybe instead of `candidates: 7`, how about `7 Notes will be Affected`?

packages/engine-server/src/doctor/service.ts
```diff
@@ -330,6 +333,31 @@ export class DoctorService implements Disposable {
         );
         return { exit: true };
       }
+      case DoctorActionsEnum.FIX_AIRTABLE_METADATA: {
+        // Converts the airtable id in note frontmatter from a single scalar value to a hashmap
+        if (!podId) {
+          assertInvalidState(
+            "Please provide pod Id to perform Doctor operation"
```
jonathanyeung, line 340: How about "Please provide the pod Id that was used to export the note(s)."

packages/plugin-core/src/commands/pods/AirtableExportPodCommand.ts
```diff
@@ -186,14 +186,15 @@ export class AirtableExportPodCommand extends BaseExportPodCommand<
         records: records.created,
         engine,
         logger,
+        podId: config.podId,
```
jonathanyeung, line 189: Do we need to make a corresponding change on the CLI side to include the podId?

jonathanyeung (changes requested):

packages/plugin-core/src/commands/pods/AirtableExportPodCommand.ts
Harshita-mindfire: updated files

packages/engine-server/src/doctor/service.ts
Harshita-mindfire: updated doctor/service.ts to filter candidates.

packages/pods-core/src/builtin/AirtablePod.ts
Harshita-mindfire: This method is only called when a new record is created. If `airtableId` is present in the note fronmatter, Airtable export v2 will update the following record and this method is not called.

SeriousBug (approved):

jonathanyeung (approved):

PR #2587, opened 2022-03-17
chore: fix circular dependencies in Lookup Controller Code

jonathanyeung:
## chore: fix circular dependencies in Lookup Controller Code

Unwinds circular dependencies in the Lookup Controller Code: 
- Uses a pseudo MVVM pattern.  LookupControllerV3 becomes the 'model' / controller.  A view model represents the data state for the lookup options.  2 views subscribe to the view model - a `QuickPickView,` which represents the UI state of the buttons in the quick pick, and a `LookupPanelView`, which represents the view for the LookupPanel webview on the side. So essentially:

```mermaid
graph LR
    A(Lookup Controller V3) <-->|Responds to State Changes| B(Lookup View Model)
    B<-->|Two Way Binding| C(Quick Pick View)
    B<-->|Two Way Binding| D(Lookup Panel View)
```
                                                                        
- To facilitate the ViewModel, introduces a `TwoWayBinding` class that utilizes the `EventEmitter` class to propagate UI changes to subscribers (in this case, the two views) as well as propagate UI actions to the controller.
- Moves button business logic into the controller.  All refresh logic on UI state is separated into the `QuickPickView`. 
- Additionally, did some shifting of types between `Types.ts` and `ButtonTypes.ts` to remove some more circular dependencies.  The positioning may not make the most semantic sense, but we can rename and reorganize later; the immediate goal is to eliminate the circular dependencies.

There's more cleanup in how the controller vs provider relationship works, but since this change is already big I decided to focus mostly on the circular aspects here and defer those other changes for later.

Goes from 14 circular dependencies -> 2 in plugin-core.  Almost there!

Dendron-Docs Update: https://github.com/dendronhq/dendron-docs/pull/24 

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [In Progress] [Confirm manual testing](dev.process.qa.test) 
- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended


packages/plugin-core/src/test/suite-integ/NoteLookupCommand.test.ts
```diff
@@ -2225,619 +2114,367 @@ suite("NoteLookupCommand", function () {
...
-          done();
-        },
+            cmd.cleanUp();
+            done();
+          },
+        });
       });
     });
 
-    test("toggling scratch modifier off will only leave selection2link applied", (done) => {
```
jonathanyeung, line 2399: @hikchoi - there's a series of tests here around "toggling * only leaves selection2link applied".  Do you remember the rationale behind this behavior or what the scenario was about?  

I can re-add corresponding tests in the controller test, but I had to remove these particular ones because `onTriggerButton()` is no longer exposed.

hikchoi: these are tests to verify that toggling a modifier button of different type back and forth will revert back the quickpick values to the original value.

e.g.)

start with with selection2link and scratch modifier applied. At this point the quickpick value would be something like `scratch.yyyy.mm.dd.some-selection`

I was testing here that toggling scratch modifier off would leave the quickpick value as `{current-note}.some-selection`.

These were added because there was a regression during the lookup v3 migration where quickpick values were not being restored or becoming some garbage value.

I went a little extra with covering every single combination here, so it may be a good idea to trim them down if you find them a bit redundant.

hikchoi:
Sorry about the delay with the review. I'll have the review in soon.

packages/plugin-core/src/types/TwoWayBinding.ts
```diff
@@ -0,0 +1,54 @@
...
+   * @param callback
+   * @param thisArg
+   * @returns
+   */
+  bind(
+    callback: (newValue: T, previous: T) => void,
+    thisArg?: any
+  ): Disposable {
+    return this._emitter.event((data) => {
+      const binded = callback.bind(thisArg);
```
hikchoi, line 50: nit: `bound`?

packages/plugin-core/src/components/views/LookupV3QuickPickView.ts
```diff
@@ -0,0 +1,303 @@
...
+        if (ToLinkBtn) buttons.push(ToLinkBtn);
+        if (ExtractBtn) buttons.push(ExtractBtn);
+        if (ToItemsBtn) buttons.push(ToItemsBtn);
+
+        this.updateButtonsOnQuickPick(...buttons);
+      })
+    );
+
+    // Vault Selection is mapped to 'other' for some reason; Fix Later.
+    const vaultSelectionBtn = this.getButton("other");
```
hikchoi, line 101: Can't really recall why we left it at "other", but I don't think we have a reason not to give it a better name at this point haha.

packages/plugin-core/src/components/views/LookupV3QuickPickView.ts
```diff
@@ -0,0 +1,303 @@
...
+          this.updateButtonsOnQuickPick(directChildBtn);
+        })
+      );
+    }
+
+    const journalBtn = this.getButton(LookupNoteTypeEnum.journal);
+    const scratchBtn = this.getButton(LookupNoteTypeEnum.scratch);
+    const taskBtn = this.getButton(LookupNoteTypeEnum.task);
+
+    if (journalBtn && scratchBtn && taskBtn) {
```
hikchoi, line 146: `CreateJournalNoteCommand` and `CreateScratchNoteCommand` both creates their own custom controller and uses those. I think with this block, the controller for `CreateJournalNoteCommand` and `CreateScratchNoteCommand` will not function correctly.

hikchoi: To be fair, I think the implementation of both is a bit hacky at best. Would they need their own separate `QuickPickView`s?

packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -266,10 +268,7 @@ export class NoteLookupCommand
       lc.fuzzThreshold = copts.fuzzThreshold;
     }
 
-    if (!_.isUndefined(this.controller.view)) {
```
hikchoi, line 269: I think removing this would cause the lookup panel view (most likely fails to render) to pop up during `CreateJournal/ScratchNoteCommand`, which will make the normal lookup view not function as well

packages/plugin-core/src/commands/SchemaLookupCommand.ts
```diff
@@ -265,7 +265,16 @@ export class SchemaLookupCommand extends BaseCommand<
       );
     } finally {
       opts.controller.onHide();
+      // if (this._controller) {
```
hikchoi, line 268: nit: remove comment

packages/plugin-core/src/components/lookup/ButtonTypes.ts
```diff
@@ -79,7 +103,6 @@ export type IDendronQuickInputButton = QuickInputButton & {
   type: ButtonType;
   description: string;
   pressed: boolean;
-  onLookup: (payload: any) => Promise<void>;
```
hikchoi, line 82: Where was this used?

packages/plugin-core/src/components/lookup/LookupControllerV3.ts
```diff
@@ -230,96 +235,507 @@ export class LookupControllerV3 implements ILookupControllerV3 {
...
+  public onHide() {
     const ctx = "LookupControllerV3:onHide";
-    this._quickpick?.dispose();
-    this._quickpick = undefined;
+    this._quickPick?.dispose();
+    this._quickPick = undefined;
     this._cancelTokenSource?.dispose();
-    this._view?.deregisterController();
+
+    this._disposables.forEach((disposable) => disposable.dispose());
```
hikchoi, line 244: If we do this, does the views dispose themselves as well?

packages/plugin-core/src/components/lookup/LookupProviderV3.ts
```diff
@@ -1,41 +1,6 @@
...
-  ILookupProviderV3,
-  NoteLookupProviderSuccessResp,
-  OnAcceptHook,
-  OnUpdatePickerItemsOpts,
-  ProvideOpts,
-  SchemaLookupProviderSuccessResp,
-} from "./LookupProviderV3Interface";
-import { transformQueryString } from "./queryStringTransformer";
-import { DendronQuickPickerV2, DendronQuickPickState } from "./types";
-import { OldNewLocation, PickerUtilsV2 } from "./utils";
```
hikchoi, line 4: Any reason to leave these two function here? I think the file name is a bit confusing as well since now it looks like this file only contains some utils.

packages/plugin-core/src/components/lookup/LookupViewModel.ts
```diff
@@ -0,0 +1,30 @@
+import { TwoWayBinding } from "../../types/TwoWayBinding";
+import { VaultSelectionMode } from "./types";
+
+export enum NameModifierMode {
```
hikchoi, line 4: Just curious. Are we moving away from calling these "note type modifiers" since it clashes with note traits?

hikchoi (changes requested):
Left a few first-pass comments. Also looks like there's some problem trying to register views (I'll post repro step in a bit)

hikchoi:
repro for view registering problem:

1. Try `Create Journal Note` command
2. Lookup View gets a status bar and hangs without rendering
3. cancel
4. `Toggle Developer Tools` to open the console
5. try `Create Journal Note` command again
6. see error:

```sh
ERR View already awaiting revival: Error: View already awaiting revival
```

7. Now try regular lookup
8. See error:

```sh
ERR Webview already registered: Error: Webview already registered
```

Same thing happens for `Craete Scratch Note` command.

hikchoi:
forgot to mention: deferred reviewing test codes until the next pass.

packages/plugin-core/src/types/TwoWayBinding.ts
jonathanyeung: 😓

packages/plugin-core/src/components/lookup/LookupProviderV3.ts
jonathanyeung: Moved these to the `utils.ts` file in the corresponding lookup folder

packages/plugin-core/src/components/lookup/LookupControllerV3.ts
jonathanyeung: It should - both the views implement Disposable, and they get added to this `_disposables` array:

```typescript
      this._disposables.push(new LookupPanelView(this._viewModel));
      ...
      this._disposables.push(
      new LookupV3QuickPickView(quickpick, this._viewModel, this._provider.id)
    );
 ```

File: packages/plugin-core/src/components/views/LookupV3QuickPickView.ts (hunk 4 above)
Comment (jonathanyeung) [reply packages/plugin-core/src/components/views/LookupV3QuickPickView.ts:146]: Good call - yeah I just amended this object to do a bunch of `if(journalBtn) then set/unset it`. 

I agree it's hacky - I think ideally we add args to this view's constructor to instantiate the view with different options based on which buttons are active.  But right now, that button status is only exposed via a prop in `DendronQuickPickerV2`, so it's a bit messy.  I think I'll keep it for now - there are changes I do wanna make later to clean up how the Controller/Provider interfaces work, but to reduce scope for this particular change and focus on circ dep removal, I'll defer that for now.

File: packages/plugin-core/src/commands/NoteLookupCommand.ts (hunk 5 above)
Comment (jonathanyeung) [reply packages/plugin-core/src/commands/NoteLookupCommand.ts:269]: I moved everything involving `VSCodeUtils.setContext(DendronContext.SHOULD_SHOW_LOOKUP_VIEW, *)` into `LookupPanelView` so that it can also own this part of the panel lifecycle.

File: packages/plugin-core/src/components/lookup/ButtonTypes.ts (hunk 7 above)
Comment (jonathanyeung) [reply packages/plugin-core/src/components/lookup/ButtonTypes.ts:82]: There weren't any references to this.

File: packages/plugin-core/src/components/views/LookupV3QuickPickView.ts (hunk 3 above)
Comment (jonathanyeung) [reply packages/plugin-core/src/components/views/LookupV3QuickPickView.ts:102]: Updated to "selectVault"

File: packages/plugin-core/src/components/lookup/LookupViewModel.ts (hunk 10 above)
Comment (jonathanyeung) [reply packages/plugin-core/src/components/lookup/LookupViewModel.ts:4]: No good reason - I realize now that we've duplicated some of these enums in 3 places - I just did a pass at consolidating to the ones defined in common-all.

**Comment by jonathanyeung** [comment]
@hikchoi - with the latest changes, I think the issue you mentioned should be resolved now.  I've also added a test for the selection2link mod -> journal -> no journal modifier scenario.

**Review by kevinslin — COMMENTED** [review]
nice work! just a few minor comments

File: packages/plugin-core/src/components/lookup/NoteLookupProvider.ts [hunk 11]
```diff
@@ -115,16 +67,13 @@ export class NoteLookupProvider implements ILookupProviderV3 {
...
         } as OnUpdatePickerItemsOpts);
         Logger.debug({ ctx, msg: "exit" });
         return out;
       },
       100,
       {
-        // Use trailing to make sure we get the latest letters typed by the user
-        // before accepting.
-        leading: false,
+        leading: true,
```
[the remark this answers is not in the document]
jonathanyeung: good catch - I did rebase already, but the merge was difficult since I moved a bunch of stuff.  Let me do a closer inspection

packages/plugin-core/src/components/lookup/LookupControllerV3.ts
```diff
@@ -168,20 +160,27 @@ export class LookupControllerV3 implements ILookupControllerV3 {
...
     // invoke button behaviors
-    PickerUtilsV2.refreshButtons({ quickpick, buttons, buttonsPrev });
-    await PickerUtilsV2.refreshPickerBehavior({ quickpick, buttons });
-    quickpick.onDidTriggerButton(this.onTriggerButton);
+    this._quickPick.buttons = this._initButtons;
+    this.setupViewModelCallbacks();
+
+    // Now Create the Views:
+    this._disposables.push(
+      new LookupV3QuickPickView(quickpick, this._viewModel, this._provider.id)
```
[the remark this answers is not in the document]
jonathanyeung: Yeah the flicker is quite distracting, I was going to look into it later.  I'll add a TODO comment.

hikchoi (approved):
Looks good! Manually tested the view registration bug and it works fine now 👍🏼

packages/plugin-core/src/test/suite-integ/LookupProviderV3.test.ts
```diff
@@ -187,6 +187,8 @@ suite("selection2Items", () => {
         );
 
         expect(expectedItemLabels).toEqual(actualItemLabels);
+
+        cmd.cleanUp();
```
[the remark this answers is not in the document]
jonathanyeung: Added Docs-PR to the PR description.

PR #2592, opened 2022-03-18
chore: add pipeline for pub dendron patch

jonathanyeung:
## chore: add pipeline for pub dendron patch

Creates a pipeline for publishing Dendron as a patch release, meant for 'emergency' or bug-fixing type releases. This also renames the existing publish Dendron pipeline to `publish-extension-dendron-minor` to avoid pipeline confusion.


kevinslin:
can we do a test release for this? eg. publish a patch release sometime next week to verify?

PR #2599, opened 2022-03-20
enhance(basics): getting started tutorial updates

kevinslin:
enhance(basics): getting started tutorial updates

- if default folder for Dendron already exist, generate new folder with an incrementing suffix to avoid overriding original folder
- add telemetry event `ClickStart` to track if getting started gets clicked

This also refactors some of the workspace activation logic into a separate file

***

### Basics

- [ ] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](https://wiki.dendron.so/notes/pMS27sHxbWeKMoPRrWEzs.html).

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -32,6 +32,7 @@ export enum CLIEvents {
 
 export enum TutorialEvents {
   WelcomeShow = "WelcomeShow",
+  ClickStart = "ClickStart",
   Tutorial_0_Show = "Tutorial_0_Show",
   Tutorial_1_Show = "Tutorial_1_Show",
   Tutorial_2_Show = "Tutorial_2_Show",
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -32,7 +32,7 @@ export enum CLIEvents {
 
 export enum TutorialEvents {
   WelcomeShow = "WelcomeShow",
-  ClickStart = "ClickStart",
+  ClickStart = "Getting_Started_Clicked",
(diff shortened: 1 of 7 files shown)


packages/common-server/src/filesv2.ts
```diff
@@ -569,6 +569,26 @@ export async function findNonNoteFile(opts: {
...
+   * @param param0
+   */
+  static genFilePathWithSuffixThatDoesNotExist({
+    fpath,
+    sep = "-",
+  }: {
+    fpath: string;
+    sep?: string;
+  }) {
+    // Try to put into a eefault '~/Dendron' folder first. If path is occupied, create a new folder with an numbered suffix
```
jonathanyeung, line 583: nit: this function doesn't specify anything about a `~/Dendron` folder.  Also, sp on 'default'

packages/common-all/src/analytics.ts
```diff
@@ -32,6 +32,7 @@ export enum CLIEvents {
 
 export enum TutorialEvents {
   WelcomeShow = "WelcomeShow",
+  ClickStart = "ClickStart",
```
jonathanyeung, line 35: nit: how about we start using past tense on the events?  Maybe "Getting_Started_Clicked" - may also be good to be slightly more specific here, in case in the future we have various 'start' buttons.

jonathanyeung (approved):

PR #2604, opened 2022-03-22
fix(workspace): race condition when loading native workspace

kevinslin:
fix(workspace): race condition when loading native workspace

To load up a workspace, we call the `reloadWorkspace` function which calls Dendron's `reloadIndex` function.
`reloadIndex` is registered after `reloadWorkspace`. If initialization of workspace finishes too quickly, Dendron will try to call `reloadIndex` before it is registered.
Fixed by registering `reloadIndex` ahead of calling `reloadWorkspace`

Also fix an issue where `getCodeWorkspaceSettingsSync` would be called in native workspaces

### DIFF
--- a/packages/common-server/src/filesv2.ts
+++ b/packages/common-server/src/filesv2.ts
@@ -580,7 +580,7 @@ class FileUtils {
     fpath: string;
     sep?: string;
   }) {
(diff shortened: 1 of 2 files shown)


jonathanyeung (approved):
lgtm - on a broader note, I do think at some point (not now) we should re-work how our initialization logic works to make it more robust.  I'm worried that we're gradually going to get more race condition type bugs.

PR #2608, opened 2022-03-22
fix: Backlinks will no longer disappear in preview upon editing

tma66:
**Background**
Previously, backlinks would disappear in the preview after any edit. This is due to the backlinks being overridden during the `findLinks` process as well as not being propagated during the `hydrate` process

**Changes**
1. Keep existing backlinks when processing a note body to find links
2. Keep existing backlinks when hydrating from one note to another

TODO: This pr addresses missing backlinks. However, backlinks are still not being updated. This will be addressed during a backlinks overhaul

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/engine-server/src/markdown/remark/utils.ts
+++ b/packages/engine-server/src/markdown/remark/utils.ts
@@ -378,9 +378,7 @@ export class LinkUtils {
   }
   /**
    * Get all links from the note body
-   * Currently, just look for wiki links and keep existing backlinks
-   *
-   * TODO: Fix backlinks not being updated when adding new reference to another note or renaming old reference
+   * Currently, just look for wiki links
    *
    * @param opts.filter - {type, loc
    *
@@ -412,8 +410,7 @@ export class LinkUtils {
       filter: { loc: filter?.loc },
       note,
     });
-    const backlinks = note.links.filter((link) => link.type === "backlink");
-    return links.concat(backlinks);
+    return links;
   }
 
   static findHashTags({ links }: { links: DLink[] }) {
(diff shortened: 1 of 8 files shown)


packages/engine-server/src/markdown/remark/utils.ts
```diff
@@ -378,7 +378,10 @@ export class LinkUtils {
   }
   /**
    * Get all links from the note body
-   * Currently, just look for wiki links
+   * Currently, just look for wiki links and keep existing backlinks
```
jonathanyeung, line 381: Are the changes in this file necessary for the bug to be fixed, or are the changes in `dnode` / `storev2` sufficient? 

The modifications here feel a little bit odd to me, because this logic is part of a utils around Markdown/Remark, i.e. note parsing.  So it feels a bit strange to also return some data from `NoteProps` metadata in the return value, versus just data parsed from the note body.

jonathanyeung (changes requested):
Had a question, otherwise lgtm

kevinslin (commented):
looks good to me, pending existing comments. also, fyi - the reason the snapshot files updated is because in this test, you updated the backlinks which we currently have a snapshot test for.

packages/engine-server/src/markdown/remark/utils.ts
tma66: I can move it out of `LinkUtils` no problem. I'll move it to `getLinks` of `DendronEngineV2` instead as there are multiple places where we override node.links with the results of `getLinks`

jonathanyeung (approved):

PR #2622, opened 2022-03-23
fix(internal): Engine updateNote not properly firing update events

tma66:
Fix bug in `DendronEngineClient.updateNote` where it searches by note id instead of note fname. This is causing events to be "create" instead of "update"

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended


packages/engine-server/src/engineClient.ts
```diff
@@ -426,7 +426,7 @@ export class DendronEngineClient implements DEngineClient, EngineEventEmitter {
     const existing = await this.api.engineGetNoteByPath({
       vault: note.vault,
       ws: this.ws,
-      npath: note.id,
+      npath: note.fname,
```
SeriousBug, line 429: Would you mind adding a docstring for `npath` on 'GetNoteOptsV2` that it should be a note name? That might help avoid this type of bugs in the future.

kevinslin (approved):

jonathanyeung (approved):
Nice catch! - also @hikchoi as an FYI in case this affects tree view work.

packages/engine-server/src/engineClient.ts
tma66: sure

kevinslin (approved):

Issue #2550, opened 2022-03-10
Lookup stops working properly after pressing tab


chmac:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [x] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

After pressing the tab key in the lookup box it stops working properly.

## To Reproduce

1. Create notes `foo.foo`, and `foo.bar` (without having a `foo`)
2. Open lookup
3. Type `fo` and press tab
4. Now type `.baz` 
5. Note that your lookup now reads `foo.baz` yet the stub `foo` is still selected in the drop down list and "Create new" is not auto selected
    1. EDIT: Also note that the dropdown still shows `foo.foo` and `foo.bar` while if you had typed `foo.baz` without using tab completion the drop down would only show "Create new" as no existing notes match your filter
6. Try selecting "Create new" and observe that you're prompted to select a vault even if you have the "automatically create in the same vault as currently" option set

I'm not sure exactly what's going on, but I've noticed it a few times in different contexts, mostly when trying to auto complete and create new notes in existing hierarchies. It seems like things start to behave differently after the tab completion.

## Expected behavior

In the example above, I expect to be able to tab complete parts of existing hierarchies, type a new suffix, and then press enter to create a new note with that name.

Harshita-mindfire:
Thanks @chmac for reporting the issue. I can repro it. It seems lookup is only considering the value that is populated with pressing `Tab`, and everything typed after it is ignored. We will let you know once this is fixed 🙏

aleksey-rowan:
Would this be related to or a duplicate of 2579?

- https://github.com/dendronhq/dendron/issues/2579

chmac:
I think #2579 is a duplicate of this.

kevinslin:
should be fixed when #2623 is shipped next week

aleksey-rowan:
[...]

PR #2623, opened 2022-03-23
fix(lookup): autocomplete causes notes to be created in wrong vault 

kevinslin:
fix(lookup): autocomplete causes notes to be created in wrong vault 

NoteLookupAutocomplete pre-emptively cancelling updatePicker results

***

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/NoteLookupCommand.ts
+++ b/packages/plugin-core/src/commands/NoteLookupCommand.ts
@@ -164,6 +164,7 @@ export class NoteLookupCommand
     return this._provider;
   }
 
+  //  ^1h1dr08geo6c
   async onAutoComplete() {
     if (this._quickPick) {
       this._quickPick.value = AutoCompleter.getAutoCompletedValue(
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -172,7 +173,6 @@ export class NoteLookupCommand
 
       await this.provider.onUpdatePickerItems({
         picker: this._quickPick,
-        token: this.controller.createCancelSource().token,
```
jonathanyeung, line 175: What was triggering the cancel?

Harshita-mindfire: on pressing `tab`, this new Cancel source was created resulting in hijacking the token of Controller

Harshita-mindfire (commented):
This might also resolve another related issue: { https://github.com/dendronhq/dendron/issues/2550 }

SeriousBug (approved):
Seems to fix the issue for me. Also the other issue harshita linked.

PR #2631, opened 2022-03-24
fix(copynotelink): Allow user to run copyNoteLink without needing to save first

tma66:
**Bug**
If user updates note title without saving and runs CopyNoteLink command, it grabs the previous note title. This is because the engine note has not been updated

**Changes**
1. Save document if dirty before executing rest of command
2. Subscribe to engine on`onEngineNoteStateChanged` to make sure execution happens after engine is updated
3. Dispose after command is executed
4. Refactored `CopyNoteLink.test.ts` to use latest test suites
5. Expanded test util to allow creation of preset notes after engine is running in test harness

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/CopyNoteLink.ts
+++ b/packages/plugin-core/src/commands/CopyNoteLink.ts
@@ -4,15 +4,16 @@ import {
   genUUIDInsecure,
   isBlockAnchor,
   isLineAnchor,
+  NoteChangeEntry,
   NoteProps,
   NoteUtils,
   VaultUtils,
 } from "@dendronhq/common-all";
 import { isInsidePath } from "@dendronhq/common-server";
-import { AnchorUtils } from "@dendronhq/engine-server";
+import { AnchorUtils, EngineEventEmitter } from "@dendronhq/engine-server";
 import _ from "lodash";
 import path from "path";
-import { TextEditor, window } from "vscode";
+import { Disposable, TextEditor, window } from "vscode";
 import { DendronClientUtilsV2 } from "../clientUtils";
 import { PickerUtilsV2 } from "../components/lookup/utils";
 import { DENDRON_COMMANDS } from "../constants";
@@ -23,17 +24,27 @@ import { VSCodeUtils } from "../vsCodeUtils";
 import { BasicCommand } from "./base";
 
 type CommandOpts = {};
-type CommandOutput = {
-  link: string;
-  type: string;
-  anchorType?: string;
-};
+type CommandOutput =
+  | {
+      link: string;
+      type: string;
+      anchorType?: string;
+    }
+  | undefined;
 
 export class CopyNoteLinkCommand extends BasicCommand<
   CommandOpts,
   CommandOutput
 > {
   key = DENDRON_COMMANDS.COPY_NOTE_LINK.key;
+  private engineEvents: EngineEventEmitter;
+  private _onEngineNoteStateChangedDisposable: Disposable | undefined;
+
+  constructor(engineEvents: EngineEventEmitter) {
+    super();
+    this.engineEvents = engineEvents;
+  }
+
   async sanityCheck() {
     if (_.isUndefined(VSCodeUtils.getActiveTextEditor())) {
       return "No document open";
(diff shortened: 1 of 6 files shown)


packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -175,13 +186,46 @@ export class CopyNoteLinkCommand extends BasicCommand<
...
+            if (savedNote.length > 0) {
+              await this.executeCopyNoteLink(savedNote[0].note, editor);
+              this.dispose();
+            }
+          }
+        );
+      await editor.document.save();
+      return;
+    } else if (this._onEngineNoteStateChangedDisposable) {
+      // If this is not disposed, it means we are still listening on engine state change from previous CopyNoteLink.execute command
```
jonathanyeung, line 208: Hmm is this true though?  This command isn't implemented as a singleton, so I think concurrent calls of CopyNoteLink will result in multiple subscriptions.  When the command is run, what's invoked by vscode isn't this `execute()`, rather, it's what we wrote in `_extension.ts`:

```typescript
        sentryReportingCallback(async (args) => {
          if (args === undefined) {
            args = {};
          }
          await new CopyNoteLinkCommand(ws.getEngine()).run(args); // We'll have multiple new calls with multiple invocations.
        })
```

packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -175,13 +186,46 @@ export class CopyNoteLinkCommand extends BasicCommand<
...
+      this._onEngineNoteStateChangedDisposable =
+        this.engineEvents.onEngineNoteStateChanged(
+          async (noteChangeEntries: NoteChangeEntry[]) => {
+            const savedNote = noteChangeEntries.filter(
+              (entry) => entry.note.fname === fname && entry.status === "update"
+            );
+            // Received event from engine about successful save
+            if (savedNote.length > 0) {
+              await this.executeCopyNoteLink(savedNote[0].note, editor);
+              this.dispose();
```
jonathanyeung, line 201: I think we should also use `setTimeout` to dipose of the listener at a certain timeout and also in case anything goes wrong. Otherwise, if something goes wrong in this situation and the user runs the command repeatedly, we may end up with a bunch of zombie subscriptions.

jonathanyeung (changes requested):

packages/plugin-core/src/commands/CopyNoteLink.ts
tma66: Oh you're right. In that case, do we need to create new each time the key is invoked? If we instantiate outside the `registerCommand`, we can use the same instance. Otherwise multiple commands might be trying to save if the first command isn't "done" yet.

packages/plugin-core/src/commands/CopyNoteLink.ts
```diff
@@ -175,13 +186,46 @@ export class CopyNoteLinkCommand extends BasicCommand<
...
+      const note = NoteUtils.getNoteByFnameFromEngine({
+        fname,
+        vault,
+        engine,
+      }) as NoteProps;
+      return this.executeCopyNoteLink(note, editor);
+    }
+  }
+
+  dispose(): void {
```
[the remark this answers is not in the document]
tma66: Sure

packages/plugin-core/src/commands/CopyNoteLink.ts
tma66: Good call

packages/plugin-core/src/commands/CopyNoteLink.ts
jonathanyeung: I think that should work well

kevinslin (approved):

packages/plugin-core/src/test/testUtilsV3.ts
```diff
@@ -654,7 +655,10 @@ export function cleanupWorkspaceStubs(ctx: ExtensionContext): void {
 export function subscribeToEngineStateChange(
   callback: (noteChangeEntries: NoteChangeEntry[]) => void
 ): Disposable {
-  const engine = ExtensionProvider.getEngine();
-  const engineClient = engine as unknown as DendronEngineClient;
+  const engineClient = toDendronEngineClient(ExtensionProvider.getEngine());
   return engineClient.onEngineNoteStateChanged(callback);
 }
+
+export function toDendronEngineClient(engine: IEngineAPIService) {
```
jonathanyeung, line 662: Hrm - outside of the scope of this change, but this casting shouldn't be necessary if we fix our interfaces.  `IEngineAPIService` should extend `DendronEngineClient`

jonathanyeung (approved):

packages/plugin-core/src/test/testUtilsV3.ts
tma66: Makes sense to me, although we'll need to talk about the howtos, especially since the constructors are pretty different

Issue #2603, opened 2022-03-21
Add feedforward UI hints to command sequences (e.g. "Move note")


fnurl:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [x] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [x] Lookup - Dendron's Lookup Command
- [ ] Views - Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [ ] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

When running multi step commands in Dendron I find myself wondering things like *"Did I choose the correct command?"*, *"Oh, this does not look right, I better cancel."*.

E.g. when moving a note to another vault using `Dendron: Move note` the next step in the process shows the current name of the note without any prompt regarding what I should do.

![CleanShot 2022-03-21 at 13 16 04](https://user-images.githubusercontent.com/106794/159259485-612cb32c-8748-4353-875c-0c54e5f66f2e.png)

## Describe the solution you'd like ##

I would be of great help if feedforward hints were available for all choices in a command sequence.

E.g. for `Dendron: Move note`, messages like *"Press enter to keep the current name"* and *"Choose destination vault next"* should be displayed.

Some commands, such as "Create New" have descriptions below the command name. This would be a good place to put the feedforward hints.

## Describe alternatives you've considered ##

N/A

## Additional context ##

N/A

SeriousBug:
Thanks for the suggestion! Adding a small description to move note seems easy enough, so I'll go ahead and add one. Are there any other commands where you noticed missing descriptions? Let me know and I can add those too.

SeriousBug:
Hey, I have a PR coming to add some descriptions. What do you think?

![image](https://user-images.githubusercontent.com/1008124/160064956-8dd6c9ea-b577-426d-8413-2c47360cf779.png)

I opted not to add all the description to the create new item because I think that would be more confusing if you type a new name. I think the title saying that you're picking which notes to move makes it clearer that you'll be moving these notes, what do you think?

PR #2632, opened 2022-03-25
enhance(refactor): improve title and "create new" description for the move note command

SeriousBug:
In the "Move Note" command, changes the title and the description for the "Create New" item to be more descriptive.

image

The "Create New" label is hard to change because our code uses it to recognize that the create new option is selected, so I opted to leave that as it is.

Not exactly what #2603 asked for, but I think it's a step in the right direction.

Also added analytics to check for how many items have been actually moved.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

  - we can track usage analytics for move note to see if there's any uptick after this is released

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/MoveNoteCommand.ts
+++ b/packages/plugin-core/src/commands/MoveNoteCommand.ts
@@ -310,6 +310,16 @@ export class MoveNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
     );
     panel.webview.html = md.render(contentLines.join("\n"));
   }
+
+  addAnalyticsPayload(_opts?: CommandOpts, out?: CommandOutput) {
+    return {
+      movedNotesCount:
+        // `out.changed` also includes the notes that had to be renamed in the
+        // process, but the created notes are the ones we moved
+        out?.changed?.filter((change) => change?.status === "create").length ||
+        0,
+    };
+  }
 }
 
 async function closeCurrentFileOpenMovedFile(
--- a/packages/plugin-core/src/commands/MoveNoteCommand.ts
+++ b/packages/plugin-core/src/commands/MoveNoteCommand.ts
@@ -105,6 +105,7 @@ export class MoveNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
     const provider = extension.noteLookupProviderFactory.create("move", {
       allowNewNote: true,
       forceAsIsPickerValueUsage: true,
+      newNoteDetail: "Move current note and rename it",
     });
     provider.registerOnAcceptHook(ProviderAcceptHooks.oldNewLocationHook);
     const initialValue = path.basename(
@@ -136,7 +137,7 @@ export class MoveNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
         },
       });
       lc.show({
-        title: "Move note",
+        title: "Move note - select the note to move",
         placeholder: "foo",
         provider,
         initialValue: opts?.initialValue || initialValue,
--- a/packages/plugin-core/src/components/lookup/LookupProviderV3.ts
+++ b/packages/plugin-core/src/components/lookup/LookupProviderV3.ts
@@ -425,7 +425,7 @@ export class NoteLookupProvider implements ILookupProviderV3 {
       if (shouldAddCreateNew) {
         const entryCreateNew = NotePickerUtils.createNoActiveItem({
           fname: queryOrig,
-          detail: CREATE_NEW_NOTE_DETAIL,
+          detail: this.opts.newNoteDetail ?? CREATE_NEW_NOTE_DETAIL,
         });
 
         const bubbleUpCreateNew = ConfigUtils.getLookup(ws.config).note
--- a/packages/plugin-core/src/components/lookup/LookupProviderV3Interface.ts
+++ b/packages/plugin-core/src/components/lookup/LookupProviderV3Interface.ts
@@ -53,6 +53,9 @@ export type ILookupProviderOptsV3 = {
    * when (and only when) nothing is queried.
    */
   extraItems?: DNodePropsQuickInputV2[];
+  /** If given, override the default "Note does not exists, create?" detail for
+   * the create new note option. */
+  newNoteDetail?: string;
 };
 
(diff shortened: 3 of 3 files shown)


packages/plugin-core/src/commands/MoveNoteCommand.ts
```diff
@@ -136,7 +137,7 @@ export class MoveNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
         },
       });
       lc.show({
-        title: "Move note",
+        title: "Move note - select the note to move",
```
jonathanyeung, line 140: I don't think this is accurate though? If I'm not mistaken, this lookup is actually to select the destination - the note to move is the one that's currently in focus. 

Hence the confusion - the user will almost always want to use the `Create New` option - so it's confusing that we show existing notes - but we also want to show existing notes to indicate existing hierarchies.

Maybe what we can do is to have a dynamic title.  I.e. something like 

Moving ${editor.activeDocuments.fname} note - select its destination

jonathanyeung (changes requested):

packages/plugin-core/src/commands/MoveNoteCommand.ts
SeriousBug: Hmm, actually selecting an existing note seems to have a weird behaviour: it will create a new note and copy the contents of the selected note. But the selected note is not deleted, and the new note is named as whatever you had typed in the quickpick. For example:

- Type `examp`
- Select `example`
- It will prompt to pick a vault, select another vault
- A new note `examp` is created with the contents of `example` in the selected vault. Both notes will still exist.

My bad, must have missed how it actually works when I was experimenting!

I think what it should actually do is:
- If you pick an existing note, it moves that note to a new vault (proposed, or intended but buggy?)
- If you pick multiple notes with the multi-select, it should move those notes to a new vault (current)
- If you pick "create new", it moves the current note to a new vault and renames it (current)

jonathanyeung: I can repro the `examp` scenario - that definitely can't be intentional 😅 

@hikchoi - since I think we've discussed similar things with refactor - I forgot where we landed on it though.

Maybe it'd be better to explicitly make this a two step operation: 
1. picking the note(s) to move
2. picking the destination

Right now, the single quickpick can kinda function as both depending on whether you're in multi-select mode or not. This is a bit confusing. 

So maybe something like the following:
Quick Pick 1: "Select Note(s) to Move" - if a document is open, we pre-fill the quick pick text with the current note name, otherwise, leave blank. User must select a valid note, otherwise we error out and stop.
Quick Pick 2: "Select Destination" - if multi-select, only allow the user to pick the vault destination.  If single select, then allow user to pick both the destination fname + destination vault (behavior right now is also to have always-prompt automatically selected here).  Furthermore, for single select, if we pick a destination that already exists, then we error out (as we do today already I think).

hikchoi: The last change I remember with move note was to allow selecting existing note name in the quickpick, and then letting the user choose the vault other than the vault that has the note with the same name. So with that I think the current behavior is what we intended (just not intuitive).

The `examp` case is definitely not an experience we want the users to have. I like @jonathanyeung's explicit step proposal.

I also remember a github issue asking for feed-forward labels that give users information of next steps. That would also be good to implement here.

For now though, I think we can be more explicit with what the quickpick does. The quickpick option is escaping me atm. did quickpick have a `prompt?: ` option to add more descriptive information? If so that would be a good place to elaborate.

SeriousBug: @hikchoi No, there doesn't seem to be one. That's why I ended up adding more text to the title. We only have the title, the placeholder (which won't show up since we prefill), and the descriptions of individual items.

SeriousBug: @jonathanyeung Should I implement what you're suggesting then, or leave it to someone else to implement with a future task? I'm trying to figure out what to do with this PR and linked issue.

kevinslin: i think there's few distinct but related issues that come together for the confusing ux:
- a) the "create new" is confusing since usually "moving" a note "doesn't create a new note"
- b) move note has different behavior whether you move a single note vs moving multiple notes
- c) move note does 3 things - it can rename a note, it can "move a note" under an existing hierarchy, it can move a note to a different vault
- d) moving a note inside a vault is equivalent to rename operation, moving a note between vaults is actually changing what directory it is under

one solution is separating move note into two commands: Move note (inside vault). Move note (between vaults)
instead of deciding this in this pr, how about creating an async with the things we discussed?

this might be a good chance to do some "learn & adapt" about our product design principles -> [[Design Principles|dendron://dendron.handbook/handbook.company.design-principles]]
ideally, these principles should help us answer this questions, and if not, we should come up with new ones.

kevinslin: continuing of the above, i think there's certain elements from apple's design principles[^1][^2] that are good for us to adapt. specifically:
- consistency: there should be a consistent mental models across similar interfaces
- affordance: it should be clear what actions are possible
- progressive disclosure: only reveal complex controls when that complexity is

currently, move note is not consistent (different toggles end in different behaviors), there is no good affordance (move multiple notes is not discoverable), but it is progressive (in the sense that the simple thing - moving a note within vault - is "simple")

[^1]: https://www.crazyegg.com/blog/better-web-experience/
[^2]: https://www.cultofmac.com/486340/apple-design-principles-sketchnotes/

SeriousBug: Got it. I'll create an async then, and close this PR since it seems like we need a more comprehensive change. I'll re-add the analytics I was introducing here in a different PR.

Issue #2564, opened 2022-03-14
The ()s in the url's image cause error with Dendron's extended image parser


andrey-jef:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [x] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [x] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When appending the css properties to an image, instead of applying the properties, Dendron shows the css text next to image. This error happens only with a specific url.

## To Reproduce

1. Create a markdown note with embedded image url
```markdown
Example
![example-current-liabilities](https://www.investopedia.com/thmb/aEVczlh4Rlnoc7CD-tyvgOKJW0g=/660x0/filters:no_upscale():max_bytes(150000):strip_icc():format(webp)/dotdash_Final_Current_Liabilities_Sep_2020-01-6515e265cfd34787ae2b0a30e9f1ccc8.jpg){max-width: 300px, display: block, margin: 0 auto}
```
- image url: 
    ```markdown
    https://www.investopedia.com/thmb/aEVczlh4Rlnoc7CD-tyvgOKJW0g=/660x0/filters:no_upscale():max_bytes(150000):strip_icc():format(webp)/dotdash_Final_Current_Liabilities_Sep_2020-01-6515e265cfd34787ae2b0a30e9f1ccc8.jpg
    ```
- appended css properties:
    ```css
    {max-width: 300px, display: block, margin: 0 auto}
    ```
2. Open `Dendron Preview` to show the note

## Expected behavior

The css properties should be applied to image display.

## Screenshots

![Screenshot 2022-03-14 023815](https://user-images.githubusercontent.com/81280428/158090413-99d8d85f-fca0-4e62-81d1-13e037145d0d.jpg)

## Desktop (please complete the following information)

- OS: Win 10
- Browser (if with published websites; include version): Opera, Version:84.0.4316.31
- Versions (all that apply to bug):
  - Dendron Extension: v0.85.0
  - VS Code version: VS Codium v1.65.2
  - Dendron CLI:

## Additional context

[...]

PR #2634, opened 2022-03-25
fix(markdown): support parenthesis in the image URL

SeriousBug:
Adds support for `()` within extended image URLs, which is supported in regular image URLs.

#2564


Harshita-mindfire (commented):
lgtm!

kevinslin (approved):

PR #2635, opened 2022-03-25
chore: identify users with role when background survey is answered

hikchoi:
# chore: identify users with role when background survey is answered
This PR:
- Adds an identify call when the background survey is answered. This allows us to have the user role as an identified user trait.
- Modifies `AnalyticUtils.identify` to optionally accept a partial `VSCodeIdentifyProps`

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -118,6 +118,8 @@ export class BackgroundSurvey extends DendronQuickPickSurvey {
       });
     }
 
+    const role = maybeOtherResult || result.label;
+    AnalyticsUtils.identify({ role });
     AnalyticsUtils.track(SurveyEvents.BackgroundAnswered, {
       results: [result.label], // passing as array because this used to be a multi-select survey
       other: maybeOtherResult,
--- a/packages/plugin-core/src/survey.ts
+++ b/packages/plugin-core/src/survey.ts
@@ -118,8 +118,7 @@ export class BackgroundSurvey extends DendronQuickPickSurvey {
       });
     }
 
-    const role = maybeOtherResult || result.label;
-    AnalyticsUtils.identify({ role });
+    AnalyticsUtils.identify({ role: result.label });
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/survey.ts
```diff
@@ -118,6 +118,8 @@ export class BackgroundSurvey extends DendronQuickPickSurvey {
       });
     }
 
+    const role = maybeOtherResult || result.label;
+    AnalyticsUtils.identify({ role });
```
kevinslin, line 122: if we call identify here with the role and in the future call identify without it, just checking that future calls without the role won't reset the property?

hikchoi: yes. once a user property is set, it will persist. they will then be included with every event captured until another identify call happens, at which point any number of user properties can update.

hikchoi: I'll DM you a user lookup to verify.

SeriousBug: Do you mean by that once we get an identify with `role` set once, we can set the cohort to "users who identified with as this role ever?"

If that's the case, looking at amplitude, it looks like you can at most ask for users who had a property within the last 90 days. We might need to save the role in meta, and keep identifying the user with that role.

kevinslin: > Do you mean by that once we get an identify with role set once, we can set the cohort to "users who identified with as this role ever?" 
yes

> (I'm assuming that the role will be unset once the user restarts VSCode)
nope, not setting a property persists the property

kevinslin (approved):

kevinslin:
can we update this to airtable analytics doc?

PR #2654, opened 2022-03-30
chore: add analytics for how many notes have been moved

SeriousBug:
Adds analytics on how many notes have been moved with a "Move Note" command. If the user cancels the prompt, that is also tracked as 0 notes being moved.


jonathanyeung (approved):

jonathanyeung:
Heads up that we pushed this one into `release/0.89.0` as well.

PR #2665, opened 2022-03-31
fix: tree item sort order in treeview v1 to be on par with v2 in preparation for v2 deprecation

hikchoi:
# fix: fix item sorting in treeview v1 to be on par with v2

This PR:
- Fixes an issue with tree view v1 not sorting notes properly.
  - children were being sorted, but not being set (`_.sortBy` is does not mutate the object.)
  - when the tree view is first caculated, the sort order is _sorta_ right (ordered by title, presumably by the engine elsewhere when the children were added / retrieved / indexed), but a rename operation would have broke the ordering because of this issue.
- Fixes an issue with `TreeUtils.sortNotesAtLevel` when bubbling down `tags` hierarchy.
  - If a note with title that comes after `T` exists, the `tags` hierarchy will not be bubbled down.
  - This fixes the alphabetical ordering, and it now respects the `nav_order` property if it exists in `tags.md`.
- Uses `TreeUtils.sortNotesAtLevel` for sorting items
  - will respect `nav_order`
  - will then sort by title, then by updated timestamp
  - and push down tags to the bottom of the tree if `nav_order` is not set.
- Explicitly sets the label of the tree view item.
  - before, this was not set unless it was a root note and in this case the label is derived from the resource URI
  - This caused some confusion since the ASCII ordering uses the title which is capitalized most of the time, but the label shows the uncapitalized basename of the file.
    -  (i.e. notes with title `A` and file name `a.md` came before a note with title `_b` and file name `_b.md` because underscore is _after_ uppercase alphabets in the ASCII chart.)
- ~~Adds a test wrapper `testAsyncWithCallback`, that allows to overspecify the test resolution.~~
  - ~~Can be used with `describeMulti/SingleWS` harness when you need to run an async test and also explicitly call `done`~~
  - ~~daily async created for this~~
- Refactor one existing test in `EngineNoteProvider.test.ts`
  - to use the new test harness
  - ~~to use the `testAsyncWithCallback` wrapper instead~~
  - ~~and prevent `done` from being called twice when failed~~

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended


hikchoi:
@jonathanyeung Trying to add tests for this in [[../packages/plugin-core/src/test/suite-integ/EngineNoteProvider.test.ts]] but I'm a bit stuck here.

The existing tests are using the legacy harness, so we can pass in `done` and also have `onInit` async and call `done` with an error.

With our new harness (`describeMultiWS`), we can't overspecify the resolution. We either make it async and pass `done` or keep it sync and pass `done`. How would you change the tests here if we were to refactor them to use the new test harness?

jonathanyeung:
@hikchoi - I think something like this should work? 

```typescript
suite("EngineNoteProvider Tests", function testSuite() {
  // Set test timeout to 2 seconds
  this.timeout(2000);

  describeMultiWS(
    "WHEN workspace type is not specified",
    {
      preSetupHook: ENGINE_HOOKS.setupBasic,
    },
    () => {
      test("THEN initializes correctly", async () => {
        const { wsRoot, vaults } = getDWorkspace();

        const testNoteProps = await NoteTestUtilsV4.createNote({
          fname: "alpha",
          vault: vaults[0],
          wsRoot,
        });

        const mockEvents = new MockEngineEvents();
        const provider = new EngineNoteProvider(mockEvents);

        const promise = new Promise<void>((resolve) => {
          provider.onDidChangeTreeData(() => {
            // If you comment line below, test should fail.
            resolve();
          });
        });

        const entry: NoteChangeEntry = {
          note: testNoteProps,
          status: "create",
        };

        mockEvents.testFireOnNoteChanged([entry]);

        return promise;
      });
    }
  );
  ```

**Comment by SeriousBug** [comment]
I think you could also use the new harnesses with a test function that uses the `done` callback.

**Comment by hikchoi** [comment]
> @hikchoi I think you could also use the new harnesses with a test function that uses the `done` callback.

yes, but we want the tests to be async here, and we can't use the done callback if it's async. Am I understanding that wrong? 😕

**Comment by SeriousBug** [comment]
@hikchoi You could use a trick like this:
```ts
test("THEN something", (done) => {
  (async () => {
    // ...
    done();
  })();
});
```
It's a bit ugly, but functionally equivalent to how the old test harnesses worked. We can add a helper function like `testAsyncWithCallback` to make this a little nicer.

packages/plugin-core/src/test/suite-integ/EngineNoteProvider.test.ts
```diff
@@ -1,169 +1,317 @@
...
+                  done(error);
                 }
-                break;
-              }
-              default: {
-                done({ message: "Note with unexpected vault found!" } as Error);
               }
+            });
+
+            if (!error) done();
```
hikchoi, line 191: `done` was being called twice when it failed which made the error message hard to read.

jonathanyeung (approved):
lgtm other than a merge conflict with `yarn.lock`

kevinslin (approved):

Issue #2534, opened 2022-03-07
Tree View is Empty


jgmpjgmp:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [X] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug
My Tree View section is completely empty. After reading other issues, I searched for notes with double dots in the title (none) and for reloaded index. without any change. Problem has actually affected me for several months.

I am also seeing the "Cannot read property 'title' of Undefined error (https://github.com/dendronhq/dendron/issues/2451) 

## To Reproduce

Sorry - unknown

## Desktop (please complete the following information)

- OS: Windows 10
- Versions (all that apply to bug):
  - Dendron Extension: 0.84.0
  - VS Code version: 1.65.0

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here
[Log.txt](https://github.com/dendronhq/dendron/files/8199622/Log.txt)

jgmpjgmp:
No journal files (for better or for worse)

SeriousBug:
This will likely be solved with #2528.

jonathanyeung:
Hi @jgmpjgmp - are you using Tree View V1 or V2? 

Tree View V1 is the default setting, but if your `dendron.yml` file (command to open is `Dendron: Configure (yaml)` ) has this in it, then it's V2:

```yaml
dev:
    enableWebUI: true
```

jgmpjgmp:
It was V1. Adding the enableWebUI: true fixed it -- showing the treeview. Thanks!

jonathanyeung:
Glad that V2 is working for you - however, let's leave this issue open as the fix that @SeriousBug  mentioned above only applies to V1, and V1 should be working as well.

SeriousBug:
@jonathanyeung On that matter, do you know when the V1 tree view will be dropped (or V2 will become default)? Are there missing features we're waiting on?

jonathanyeung:
[...]

Issue #2451, opened 2022-02-18
VSCode error notification "Cannot read property 'title' of undefined" when actively typing/editing in any document


SotoDucani:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When actively typing or editing a document, a error notification will continuously pop up with the text "Cannot read property 'title' of undefined". The notification will rapidly continue to pop up and refresh as typing continues, and stop when typing no longer continues. This behavior persists through restarting VSCode, clearing the dendron cache file, and reloading the index as described in the Troubleshooting page of the wiki.

I've also taken some time to dig through my files to see if I broke any of the frontmatter unintentionally, however I haven't been able to find anything.

## To Reproduce

1. Find a Dendron workspace that exhibits the issue. I only use one work-related workspace heavily that's showing the problem. Other personal workspaces have not as of yet shown the issue.
2. Open up any document and begin typing.
3. Error notifications will begin to populate in VSCode

## Expected behavior

No error notifications are populated.

## Screenshots

![image](https://user-images.githubusercontent.com/8963875/154712546-84754134-ef03-443c-98ba-13c2b0d73d5e.png)

## Desktop (please complete the following information)

- OS: Windows 10 20H2
- Browser: Chrome
- Version: 97.0.4692.71

### VSCode version info
Version: 1.60.2 (system setup)
Commit: 7f6ab5485bbc008386c4386d08766667e155244e
Date: 2021-09-22T12:00:31.514Z
Electron: 13.1.8
Chrome: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Windows_NT x64 10.0.19042

Dendron version: v0.82.0

## Dendron Log file

[ServerLogs.txt](https://github.com/dendronhq/dendron/files/8098414/ServerLogs.txt)
[WorkspaceSettings.txt](https://github.com/dendronhq/dendron/files/8098415/WorkspaceSettings.txt)
[...]

PR #2668, opened 2022-04-01
chore(views): report exceptions for treeview

jonathanyeung:
## chore(views): report exceptions for treeview 

There's code execution done by vs code in the native tree view that wasn't covered by Sentry - this change fixes that and also adds reporting for the backlinks view as well.

This should point us at the issue behind the following:
- https://github.com/dendronhq/dendron/issues/2534
- https://github.com/dendronhq/dendron/issues/2451

---

### Basics

### Extended
- General
  - [N/A] check if this change adversely impact performance
- Operations
  - [N/A] when shipping this change, will it just work or will it introduce additional operational overhead due to complicated interface or known bugs?
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
+++ b/packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
@@ -1,6 +1,8 @@
 import { assertUnreachable, DateFormatUtil } from "@dendronhq/common-all";
-import path from "path";
+import { EngineEventEmitter } from "@dendronhq/engine-server";
+import * as Sentry from "@sentry/node";
 import _, { Dictionary } from "lodash";
+import path from "path";
 import {
   Disposable,
   Event,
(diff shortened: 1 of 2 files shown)


jonathanyeung:
@SeriousBug for awareness - I remember we had chatted about this issue in the past.

jonathanyeung:
@kevinslin, @hikchoi - how about pushing this one up to `release/0.89.0` so that we can get the data 1 week sooner?  This should help us get the data to fix an issue with tree view v1.

jonathanyeung:
The issue reported in https://github.com/dendronhq/dendron/issues/2534 (cannot read prop `title` of undefined) looks like it originates in one of two places:
- https://github.com/dendronhq/dendron/blob/ca286279733b02d9376524f0987fdc774a5420df/packages/plugin-core/src/views/EngineNoteProvider.ts#L148
- https://github.com/dendronhq/dendron/blob/ca286279733b02d9376524f0987fdc774a5420df/packages/plugin-core/src/views/EngineNoteProvider.ts#L89

Instead of fixing it right now however, I'd rather stagger that fix to first see if we see this error show up in Sentry to confirm our suspicions.

hikchoi:
> The issue reported in #2534 (cannot read prop `title` of undefined) looks like it originates in one of two places:
> 
> * https://github.com/dendronhq/dendron/blob/ca286279733b02d9376524f0987fdc774a5420df/packages/plugin-core/src/views/EngineNoteProvider.ts#L148
> * https://github.com/dendronhq/dendron/blob/ca286279733b02d9376524f0987fdc774a5420df/packages/plugin-core/src/views/EngineNoteProvider.ts#L89
> 
> The root cause looks to be that the engine state is not correct for node.children in some cases.
> 
> Instead of fixing it right now however, I'd rather stagger that fix to first see if we see this error show up in Sentry to confirm our suspicions.

I also noticed this while working on https://github.com/dendronhq/dendron/pull/2665 I was going to try to fix it with this PR but didn't look like a tree view problem.

also, unrelated, but `_.sortBy` doesn't mutate the object so the sort isn't applied here back to the children. this is fixed with my PR.

hikchoi (approved):

PR #2673, opened 2022-04-01
fix publish issues

kevinslin:


packages/nextjs-template/package.json
```diff
@@ -31,7 +31,8 @@
     "fs-extra": "^10.0.0",
     "html-react-parser": "^1.3.0",
     "lodash": "^4.17.21",
-    "next": "12",
+    "luxon": "^1.25.0",
```
hikchoi, line 34: is fixing luxon also necessary here?

hikchoi (commented):
a question, but otherwise lgtm.

PR #2676, opened 2022-04-02
fix(workspace): preserve wikilink metadata on export

kevinslin:
fix(workspace): preserve wikilink metadata on export

Issue where wikilinks were not being preserved in the engine after saving. This would manifest when trying to export via a pod. 
***
### Pass it Forward
- [[Async Test Presets|dendron://private/user.kevin.journal.2022.04.02.m.async-presets]]: https://github.com/dendronhq/dendron-docs/commit/fee9b1100c1f3d3bc61eaab3b10fe7a923c769c3
***

### Basics

### Extended

- General
- Operations
- Architecture

## Instrumentation

NA

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended


hikchoi (approved):
other than the naming trip, looks good.

packages/common-all/src/dnode.ts
```diff
@@ -1082,29 +1082,49 @@ export class NoteUtils {
   }
 
   /**
-   * Add {@link DNodeProps["parent"]}, {@link DNodeProps["children"]}, and backlinks from {@link noteHydrated} to {@link noteRaw}
+   * Add derived metadata from `noteHydrated` to `noteRaw`
+   * By default, include the following properties:
+   *  - parent
+   *  - children
+   *  - backLinks
```
tma66, line 1089: Would default include just parent/children since we have to explicitly keep backlinks?

packages/common-all/src/dnode.ts
```diff
@@ -1082,29 +1082,49 @@ export class NoteUtils {
...
+      keepAnchors: boolean;
+    }>;
   }) {
     const hydrateProps = _.pick(noteHydrated, ["parent", "children"]);
-    if (noteHydrated) {
-      noteRaw.links = noteHydrated.links.filter(
-        (link) => link.type === "backlink"
+
+    // check if we hydrate with links
+    if (opts?.keepAllLinks) {
```
tma66, line 1110: Couple questions. 
1. Are `keepAllLinks` and `keepAnchors` being used anywhere?
2. Any particular reason why we concat backlinks to noteRaw.links?

tma66 (commented):
Works with manual testing. Just couple clarifying questions

packages/common-all/src/dnode.ts
```diff
@@ -1082,29 +1082,49 @@ export class NoteUtils {
   }
 
   /**
-   * Add {@link DNodeProps["parent"]}, {@link DNodeProps["children"]}, and backlinks from {@link noteHydrated} to {@link noteRaw}
+   * Add derived metadata from `noteHydrated` to `noteRaw`
```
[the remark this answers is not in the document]
kevinslin: i think naming wise, we could be better. whats happening is that metadata from the hydrated note is added to the raw note

packages/common-all/src/dnode.ts
kevinslin: your right - updated

packages/common-all/src/dnode.ts
kevinslin: not right now. originally, thought we could skip a call to update metadata by just keeping all links but inspected the code and realize that wasn't the case. kept the options here in case we wanted to use it down the line

as for why we concat - the idea is so we can chain links (eg. imagine hydrating with backlinks and frontmatter links) in which case we wouldn't want to wipe out what was in links

tma66: Gotcha. The only thing I'd advise keeping on eye out is in case `noteRaw` already contains backlinks in which case we could have duplicates of the same thing. I remember running into it awhile back but I forgot which scenario caused it (or even if its possible anymore)

kevinslin: got it. i removed the un-used options and reverted the logic to simplify things

tma66 (approved):

PR #2679, opened 2022-04-03
chore: fast rebuild script when building dendron plugin locally

kevinslin:
chore: fast rebuild script when building dendron plugin locally

This adds the steps in [[Build|dendron://dendron.docs/pkg.plugin-core.dev.build]] into a script

### DIFF
--- a/bootstrap/scripts/fastRebuild.sh
+++ b/bootstrap/scripts/fastRebuild.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+echo "reset repo"
+git checkout package.json
+
+echo "remove compiled assets"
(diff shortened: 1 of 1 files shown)


hikchoi (approved):
LGTM. remind me in what situation did we want to do this instead of a full cleanup / then yarn setup?

kevinslin:
> LGTM. remind me in what situation did we want to do this instead of a full cleanup / then yarn setup?


This is if you've already build the plugin locally and need to re-built it (you ran a `yarn build:{version}:local`).
The regular [[cleanup script|#cleanup]] will remove all `node_modules` across all packges.
The fast rebuild will only remove `node_modules` from packages where it needs to be removed.

PR #2687, opened 2022-04-05
chore: telemetry for number of remote vaults & repo contributor count

SeriousBug:
This telemetry is needed to measure the success of the self contained vaults project.

### Basics

### Extended

- General
    - The telemetry runs async while Dendron initialization is happening, and is only done during initialization.
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- [x] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), an example is included in the [test workspace](https://wiki.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)

- CSS
    - [x] lg: screen ≥ 992px
    - [x] safari
    - [x] firefox
    - [x] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -496,6 +496,8 @@ export async function _activate(
       }
 
       ws.workspaceService = wsService;
+      // Start early to do this in background while Dendron initializes
+      const numContributors = wsService.getAllReposNumContributors();
 
       // check for vaults with same name
       const vaults = ConfigUtils.getVaults(dendronConfig);
@@ -649,6 +651,9 @@ export async function _activate(
         numSelfContainedVaults: ws
           .getDWorkspace()
           .vaults.filter(VaultUtils.isSelfContained).length,
+        numRemoteVaults: ws.getDWorkspace().vaults.filter(VaultUtils.isRemote)
+          .length,
+        numContributors: await numContributors,
       };
 
       if (siteUrl !== undefined) {
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -10,6 +10,7 @@ import {
   DendronError,
   ExtensionEvents,
   getStage,
+  GitEvents,
   InstallStatus,
   MigrationEvents,
   NativeWorkspaceEvents,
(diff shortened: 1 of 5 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -496,6 +496,8 @@ export async function _activate(
       }
 
       ws.workspaceService = wsService;
+      // Start early to do this in background while Dendron initializes
+      const numContributors = wsService.getAllReposNumContributors();
```
kevinslin, line 500: would be good to profile this command and report this to figure out if we need to cache these results or run into pathalogical cases for big repos. i would have the method return number of `ms` it executed for and then report it here

kevinslin (changes requested):

packages/plugin-core/src/_extension.ts
jonathanyeung: I think we should also not `await` the `wsService.getAllReposNumContributors()` - let's put it in the `Promise.then(() => AnalyticsUtils.track())`.

SeriousBug: Thanks for the feedback, I:
- added a new event to track the contributor counts
- refactored tracking so initialization doesn't await it
- started tracking the duration of how long it took to compute

The actual computation is done by git, so even for large repos it shouldn't take too long (main Dendron repo takes about 42ms for me), but it's good to not have this block initialization anyway.

packages/plugin-core/src/_extension.ts
```diff
@@ -653,9 +651,23 @@ export async function _activate(
...
 
+      // Track contributors to repositories, but do so in the background so
+      // initialization isn't delayed.
+      const startGetAllReposNumContributors = process.hrtime();
+      wsService
+        .getAllReposNumContributors()
+        .then((numContributors) => {
+          AnalyticsUtils.track(GitEvents.ContributorsFound, {
+            maxNumContributors: _.max(numContributors),
+            duration: getDurationMilliseconds(startGetAllReposNumContributors),
```
kevinslin, line 664: don't need to do this now but moving forward, might make sense to shift tracking performance related metrics to sentry since they deliberately do apm (application performance monitoring) which amplitude is not great at. an async for another time

kevinslin (approved):

jonathanyeung (approved):

PR #2696, opened 2022-04-05
fix(airtable): Exporting to airtable automatically saves current document

tma66:
**Background**
Similar to https://github.com/dendronhq/dendron/pull/2631, this PR fixes the issue where user needs to save first before exporting to airtable.

**Changes**
1. Before exporting, we will first save the current document if it is dirty. After the save, listen to the engine for an `onEngineNoteStateChanged` event before continuing the command.
2. Expose EngineEmitter as part of EngineAPIService. This makes it easier to listen to events since the extension is readily available
3. Remove updating the engine as part of `onWillSaveNote`. 
- The TextDocument entering onDidSave for the TextDocumentService contains the proper updated timestamp
- `note.updated = now` actually updates the engine note for us so we can leave this in in case there's a bug that causes  onDidSave to never be triggered
- Ideally onWillSaveNote does not trigger an engine event. Otherwise every save action will cause 2 events to be fired which could be confusing

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/WorkspaceWatcher.ts
+++ b/packages/plugin-core/src/WorkspaceWatcher.ts
@@ -334,7 +334,6 @@ export class WorkspaceWatcher {
       // eslint-disable-next-line  no-async-promise-executor
       const p = new Promise(async (resolve) => {
         note.updated = now;
-        await engine.updateNote(note);
         return resolve(changes);
       });
       event.waitUntil(p);
--- a/packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
+++ b/packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
@@ -277,14 +277,14 @@ export abstract class BaseExportPodCommand<
       const fname = NoteUtils.uri2Fname(editor.document.uri);
       this._onEngineNoteStateChangedDisposable = this.extension
         .getEngine()
-        .getEngineEmitter()
-        .onEngineNoteStateChanged(
+        .engineEventEmitter.onEngineNoteStateChanged(
           async (noteChangeEntries: NoteChangeEntry[]) => {
             const updateNoteEntries = noteChangeEntries.filter(
               (entry) => entry.note.fname === fname && entry.status === "update"
             );
             // Received event from engine about successful save
             if (updateNoteEntries.length > 0) {
+              this.dispose();
               const savedNote = updateNoteEntries[0].note;
               // Remove notes that match saved note as they contain old content
               const filteredPayload = opts.payload.filter(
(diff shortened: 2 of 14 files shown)


packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -334,7 +334,6 @@ export class WorkspaceWatcher {
       // eslint-disable-next-line  no-async-promise-executor
       const p = new Promise(async (resolve) => {
         note.updated = now;
-        await engine.updateNote(note);
```
jonathanyeung, line 337: Don't we still need this?

packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
```diff
@@ -240,19 +249,83 @@ export abstract class BaseExportPodCommand<
...
+                payload: filteredPayload.concat(savedNote),
+              });
+              this.dispose();
+            }
+          }
+        );
+      await editor.document.save();
+      // Dispose of listener after 1 sec (if not already disposed) in case engine events never arrive
+      setTimeout(() => {
+        this.dispose();
```
jonathanyeung, line 304: we should notify the user of an error in this case and tell them that their export didn't happen.

jonathanyeung (changes requested):

Harshita-mindfire (approved):

Harshita-mindfire (approved):
LGTM saves the document before Airtable Export. 

Although this is out of scope of this PR but sometimes on activation, `engine not set` is thrown from the [setup command](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/_extension.ts#L1128:L1128).

tma66:
> LGTM saves the document before Airtable Export.
> 
> Although this is out of scope of this PR but sometimes on activation, `engine not set` is thrown from the [setup command](https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/_extension.ts#L1128:L1128).

Ah yes, I was just looking at that. Looks like its due to setting up the commands before the engine is ready. I will address it in a followup PR

packages/plugin-core/src/WorkspaceWatcher.ts
tma66: 1. Since vscode fires 2 events for a save event (onWillSave and onDidSave), we only want to update the engine once (otherwise we will also fire 2 engine events). The second event that is handled by TextDocumentService will contain all the info needed to update the engine.
2. This is more of a side effect, but `note.updated = now` actually updates the engine note already

packages/plugin-core/src/services/EngineAPIService.ts
```diff
@@ -308,4 +308,8 @@ export class EngineAPIService
   getAnchors(opts: GetAnchorsRequest): Promise<GetNoteAnchorsPayload> {
     return this._internalEngine.getAnchors(opts);
   }
+
+  getEngineEmitter(): EngineEventEmitter {
```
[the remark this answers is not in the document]
tma66: sure

packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
tma66: done

packages/plugin-core/src/WorkspaceWatcher.ts
jonathanyeung: Derp - we discussed this already haha.  Got it, makes sense.

kevinslin (approved):

PR #2697, opened 2022-04-05
fix: upgrade `textextensions` version so links to `XAML` files in preview open in editor

SeriousBug:
- DEPENDENCY_UPDATE: @dendronhq/plugin-core updated textextensions from 5.14.0 to 5.15.0

This change includes my contribution adding `XAML` as a text extension to the list, which will allow the Dendron preview to open XAML files when a linked to one is clicked in the preview.

The upgrade command also pulled in minor version upgrades to `nextjs`, but I left them in since they should be harmless.

#1695

### DIFF
--- a/packages/dendron-next-server/package.json
+++ b/packages/dendron-next-server/package.json
@@ -62,7 +62,7 @@
     "gulp-postcss": "^9.0.0",
     "less-plugin-npm-import": "^2.1.0",
     "lodash": "^4.17.20",
-    "next": "^11.1.2",
+    "next": "11.1.4",
     "postcss": "^8.2.13",
     "react": "^17.0.2",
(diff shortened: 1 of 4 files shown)


kevinslin:
actually had to downgrade nextjs because the swc module we're causing issues in github pipelines. could we revert the nextjs changes and add "12.1.0" in nextjs-template `package.json` without the `^`?

SeriousBug:
Got it, sure! I fixed the versions to what they were in the lockfile, there's no change in the lockfile now. Git still shows changes, but you can see it's because I changed `^11.1.2` to `11.1.4` since that's what it already was in the lockfile.

kevinslin (approved):

PR #2700, opened 2022-04-06
chore: log anonymous id at startup

SeriousBug:
Logs the user anonymous id at startup, which may be helpful in debugging when a user sends us extension logs.

### DIFF
--- a/packages/common-server/src/analytics.ts
+++ b/packages/common-server/src/analytics.ts
@@ -256,6 +256,7 @@ export class SegmentClient {
       this._anonymousId = genUUID();
       fs.writeFileSync(uuidPath, this._anonymousId);
     }
+    this.logger.info({ msg: "anonymous id", anonymousId: this._anonymousId });
   }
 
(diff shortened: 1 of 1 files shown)


packages/common-server/src/analytics.ts
```diff
@@ -256,6 +256,7 @@ export class SegmentClient {
       this._anonymousId = genUUID();
       fs.writeFileSync(uuidPath, this._anonymousId);
     }
+    this.logger.info({ msg: "anonymous id", anonymousId: this._anonymousId });
```
jonathanyeung, line 259: super nit: maybe "anonymousId" as one-word will be easier to find via Ctrl+F since we're more accustomed to having it camel case (in the code anyway)

jonathanyeung (approved):

packages/common-server/src/analytics.ts
SeriousBug: You can still `Ctrl+F` it since it's also the property name (`anonymousId: this._anonymousId`)

PR #2711, opened 2022-04-07
chore: ab testing framework

SeriousBug:
Adds an A/B testing framework to Dendron. To demonstrate this framework, also adds an A/B test for displaying a toast with a link to the changelog, versus displaying the changelog page in a webview.

Docs: https://github.com/dendronhq/dendron-docs/pull/25

The webview looks like this: 
image

Co-authored-by: jonathanyeung <jonathanyeung@users.noreply.github.com>

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-server/src/abTesting.ts
+++ b/packages/common-server/src/abTesting.ts
@@ -1,3 +1,4 @@
+import { DendronServerError } from "@dendronhq/common-all";
 import _ from "lodash";
 import SparkMD5 from "spark-md5";
 
@@ -34,13 +35,23 @@ const MAX_B16_INT = 0xffffffff;
  *
  * EXAMPLE_TEST.getUserGroup("anonymous user UUID");
  * ```
+ CURRENT_AB_TESTS|* ^85lbm3148c1a
  */
 export class ABTest<GroupNames> {
-  public name: string;
+  private _name: string;
+  public get name(): string {
+    return this._name;
+  }
   private groups: ABTestGroup<GroupNames>[];
 
   constructor(name: string, groups: ABTestGroup<GroupNames>[]) {
-    this.name = name;
+    this._name = name;
+
+    if (groups.length < 2)
+      throw new DendronServerError({
+        message:
+          "An A/B test is created with less than 2 groups. Each test must have at least 2.",
+      });
 
     const sumWeights = _.sumBy(groups, (group) => group.weight);
     this.groups = groups.map((group) => {
@@ -50,7 +61,7 @@ export class ABTest<GroupNames> {
 
   /** Given the user ID, find which group of the AB test the user belongs to. */
   public getUserGroup(userId: string) {
-    const hash = SparkMD5.hash(`${this.name}:${userId}`);
+    const hash = SparkMD5.hash(`${this._name}:${userId}`);
     const hashedInt = parseInt(`0x${hash.slice(0, 8)}`, 16);
     const userRandom = hashedInt / MAX_B16_INT;
 
--- a/packages/common-server/src/abTesting.ts
+++ b/packages/common-server/src/abTesting.ts
@@ -1,0 +1,69 @@
+import _ from "lodash";
+import SparkMD5 from "spark-md5";
+
+/** One group in an A/B test. Describes one group of users. */
+type ABTestGroup<GroupNames> = {
+  /** Name for the group. */
+  name: GroupNames;
+  /** The likelihood of a user to be in this group. The number should be
+   * positive and nonzero. The weight is calculated in relation to the weight of other groups
+   * in the A/B test. */
+  weight: number;
+};
+
+const MAX_B16_INT = 0xffffffff;
+
+/** One A/B test.
+ *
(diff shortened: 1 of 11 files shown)


packages/common-server/src/abTesting.ts
```diff
@@ -0,0 +1,80 @@
+import { DendronServerError } from "@dendronhq/common-all";
```
jonathanyeung, line 1: looks like there's an empty file artifact in `packages/common-all/src/constants/abTests.ts`

packages/plugin-core/src/abTests.ts
```diff
@@ -0,0 +1,25 @@
+import { ABTest } from "@dendronhq/common-server";
+
+// ^xi5t1r2j51ot
+
+export enum UpgradeToastOrViewTestEnum {
```
jonathanyeung, line 5: maybe we should have a convention around these to suffix these enums with "Groups", so `UpgradeToastOrViewTestGroups`, or `UpgradeNotificationTypeTestGroups`

packages/plugin-core/src/abTests.ts
```diff
@@ -0,0 +1,25 @@
...
+  upgradeToast = "upgradeToast",
+  upgradeView = "upgradeView",
+}
+
+export const UPGRADE_TOAST_OR_VIEW_TEST = new ABTest("UpgradeToastOrViewTest", [
```
jonathanyeung, line 10: maybe having a quick 1-2 liner description of what the experiment is about would be helpful.

jonathanyeung (changes requested):
Excited for this!

packages/common-server/src/abTesting.ts
```diff
@@ -0,0 +1,80 @@
...
+import _ from "lodash";
+import SparkMD5 from "spark-md5";
```
kevinslin, line 3: this is a dev-dependency in common-server and only a direct dependency in engine-server. we need to add it as a direct dependency in common-server

packages/plugin-core/src/_extension.ts
```diff
@@ -796,25 +813,51 @@ async function showWelcomeOrWhatsNew({
         previousExtensionVersion,
       });
       await StateService.instance().setGlobalVersion(version);
+
+      // ^t6dxodie048o
+      const toastOrView = UPGRADE_TOAST_OR_VIEW_TEST.getUserGroup(
```
kevinslin, line 818: doesn't have to be in this pr but moving forward, would be good to encapsulate this in a funciton

```ts
ACTIVE_TESTS: [key: AbTestEnumKey]: AbTest = {}

treatment = getTreatment<T extends AbTest>(testname: AbTestEnumKey): T
```

packages/plugin-core/src/_extension.ts
```diff
@@ -796,25 +813,51 @@ async function showWelcomeOrWhatsNew({
...
+                  vscode.Uri.parse(
+                    "https://dendron.so/notes/9bc92432-a24c-492b-b831-4d5378c1692b.html"
+                  )
+                );
+              }
+            });
+          break;
+        }
+        case UpgradeToastOrViewTestEnum.upgradeView: {
+          showUpgradeView();
```
kevinslin, line 855: can we track a "success action" for the upgrade view? for the toast, success is if they clicked the button to see whats changed. what is success when showing the view? (eg. did user scroll on the iframe or click a link)

packages/plugin-core/src/abTests.ts
kevinslin: +1 to Group Suffix

packages/plugin-core/src/_extension.ts
SeriousBug: I don't think we can track links clicked since we have the page loaded inside a webview (I think the cross-origin policy blocks that). Scrolling is also not interesting because the changelog in most weeks fits within the screen, so the user can read everything without scrolling.

I added a time measurement instead which measures how long the webview was visible for. I added an event for the view closing, and attached the duration to that. We can decide when we're looking at the telemetry how long of an interaction means the view was successful.

kevinslin (approved):

packages/plugin-core/src/_extension.ts
jonathanyeung: @kevinslin - yeah I think understanding 'success' for this particular feature will be slightly difficult.  I like Kaan's approach here, with the caveat that the webview will stay open if someone has Dendron upgraded, but then walks away from their computer.

Another thing we can look at is if the A/B group starts interacting with the new features outlined in the changelog in different ways (i.e. are they more likely to use features from the new week, etc.)

jonathanyeung:
merging and pushing forward to v90

PR #2712, opened 2022-04-07
chore(internal): add telemetry for note graph view usage

Harshita-mindfire:
This PR aims to add telemetry for graph view. It adds the following properties for the existing event `dendron.showNoteGraph`. 
1. A new property `message`  to measure the % of users that uses Graph View to navigate their workspace.
2. A new property `timeOpen` is added to measure the duration for which Graph View was used.

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
+++ b/packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
@@ -10,7 +10,6 @@ import {
   NoteUtils,
   OnDidChangeActiveTextEditorMsg,
 } from "@dendronhq/common-all";
-import { getDurationMilliseconds } from "@dendronhq/common-server";
 import { EngineEventEmitter, WorkspaceUtils } from "@dendronhq/engine-server";
 import _ from "lodash";
 import path from "path";
@@ -36,7 +35,6 @@ export class NoteGraphPanelFactory {
     ext: DendronExtension,
     engineEvents: EngineEventEmitter
   ): vscode.WebviewPanel {
-    const start = process.hrtime();
     if (!this._panel) {
       const { bundleName: name, label } = getWebEditorViewEntry(
         DendronEditorViewKey.NOTE_GRAPH
@@ -86,7 +84,7 @@ export class NoteGraphPanelFactory {
               column: ViewColumn.One,
             });
             AnalyticsUtils.track(DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key, {
-              message: GraphViewMessageType.onSelect,
+              message: GraphViewMessageEnum.onSelect,
             });
             break;
           }
@@ -153,10 +151,6 @@ export class NoteGraphPanelFactory {
         if (this._onEngineNoteStateChangedDisposable) {
           this._onEngineNoteStateChangedDisposable.dispose();
         }
-        const profile = getDurationMilliseconds(start);
-        AnalyticsUtils.track(DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key, {
-          timespan: profile,
-        });
       });
     }
     return this._panel;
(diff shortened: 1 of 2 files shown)


Harshita-mindfire:
Updated Docs https://github.com/dendronhq/dendron-site/pull/467

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
```diff
@@ -136,6 +143,10 @@ export class NoteGraphPanelFactory {
         if (this._onEngineNoteStateChangedDisposable) {
           this._onEngineNoteStateChangedDisposable.dispose();
         }
+        const profile = getDurationMilliseconds(start);
+        AnalyticsUtils.track(DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key, {
+          timespan: profile,
```
kevinslin, line 148: we already have a precendent for using [duration](https://airtable.com/appVUixSC6jZbSUI1/tblHzTJJL5aVhJMYq/viwgxTbxGhjlLuBSh/recxnNkjek9LvLRZ5?blocks=hide) to measure profile information. lets use that

kevinslin (changes requested):
can you make sure these metrics are recorded in the "next" section of airtable > https://airtable.com/appVUixSC6jZbSUI1/tblUAoSy7SrLzF9Nq/viwx63jVGWp3cHNCY/rechYL7Hkyo5wUsI8?blocks=hide

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
Harshita-mindfire: Hmm..I was actually trying to measure how long did user had the graph view open.

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
```diff
@@ -28,6 +31,7 @@ export class NoteGraphPanelFactory {
     ext: DendronExtension,
     engineEvents: EngineEventEmitter
   ): vscode.WebviewPanel {
+    const start = process.hrtime();
```
SeriousBug, line 34: This should be within the `if (!this._panel)` bracket right? Otherwise it will reset if someone uses the show note graph command again while they already have the note graph open?

SeriousBug (commented):
Just left one comment about an edge case, otherwise it looks good.

Another edge case is if the user closes VSCode without closing the panel I think we won't get the usage duration, but it might not be possible to work around that.

It also might be a bit more work to use this, but I added a function to measure the time that a webview spends open. It also handles if the user switches away from the view. I'll let you decide if you want to switch over to using that command. https://github.com/dendronhq/dendron/pull/2711/files#diff-8159cc0d95f739f2e73c81bdfb0bdc2c2d15654da306e4984e39974294cdf9e2R288

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
Harshita-mindfire: Used [OpenWebviewAndMeasure](https://github.com/dendronhq/dendron/blob/e3e7f15a503d222c68bef9e0f6b6c8973b0f58b2/packages/plugin-core/src/views/utils.ts#L288) utility to measure timeOpen.

kevinslin (approved):

jonathanyeung:
lgtm

PR #2713, opened 2022-04-07
fix(publish): add luxon as dev dependency

kevinslin:
fix(publish): add luxon as dev dependency

Required for clean build of nextjs-template


jonathanyeung (approved):

SeriousBug (approved):
Builds fine for me.

PR #2714, opened 2022-04-07
chore: properly track OnClicked event on tutorial launch

kevinslin:
chore: properly track OnClicked event on tutorial launch

OnClicked event wasn't being tracked since `Reload Window` is being called almost immediately afterwards.
Initially tried a synchronous track call but that introduced noticable lag (~6s) so instead, opted to write a payload to metadata store to be uploaded on initialization.

Also ended up refactoring and updating tests alongside this change, detailed below

### homeDir not stubbed

The homeDir was only being stubbed for the first call in the test but not subsequent tests when `setupBeforeAfter` is used in conjunction

```ts
ctx = setupBeforeAfter()
...
setupBeforeAndAfter({
    ctx,
    preActivateHook: async () => { 
        mockHomeDirStub = TestEngineUtils.mockHomeDir();
    }
},() => {
    test("test1") {
        // home dir is stubbed
    }
    test("test2") {
        // home dir is no longer stubbed
    }
});
```

This is because when initialized using `setupBeforeAfter`, stubs are restored after every call in an `afterEach` call. 
`homeDir`, in the above situation, is only stubbed during the frist call to `test` and then removed for subsequent `test` calls.

Fixed by removing calls to `setupBeforeAfter`

### test updates

The `ExtensionContext` is now made available to test functions inside of `describeMultiWS` and `describeSingleWS`.

```ts
fn: (ctx: ExtensionContext) => void
```

This allows test functions to interact and stub the `ctx` without the need of creating it externally (eg. using `setupBeforeAfter`)

### temporary workaround

As part of cleaning up `setupBeforeAfter`, `setupWorkspaceStubs` is now called everytime when `describeMultiWS` is called (previously, it was only called if a `ctx` object was NOT passed in - ie. when we called `setupBeforeAfter`). 

In both `setupBeforeAfter` as well as in `setupWorkspaceStubs`, we wrap `VSCodeUtils`. This means that test cases that call **both** `setupBeforeAfter` and `setupWorkspaceStbs` (via `describeMultiWS`), will run into the [wrap an object that is already wrapped error](https://stackoverflow.com/questions/36074631/sinon-error-attempted-to-wrap-function-which-is-already-wrapped). 

As a workaround, we add a `try/catch` in `setupBeforeAfter` that checks for this and restores sinon before trying to re-wrap


packages/engine-server/src/metadata/service.ts
```diff
@@ -40,6 +41,10 @@ type Metadata = Partial<{
    * When the user last used lookup
    */
   lastLookupTime: number;
+  /**
+   * Time when the welcome button was clicked
+   */
+  welcomeClicked: number;
```
hikchoi, line 47: nit: 
```suggestion
  welcomeClickedTime: number;
```

hikchoi: as well as the getter naming

hikchoi (commented):
Nice catch on homeDir not being stubbed properly. This definitely have caused some head scratching for me in the past. I notice you refactored some tests but we still have some in other places. Are the ones you fixed the major ones that fail our CI? Do we have plans to remove the rest?

On using the metadata store to save the welcome click event:
I can guess that this pattern may become something we do more in the future. When the time comes, do you think it's worth creating a separate store destination instead of using `meta.json`? Since `meta.json` is something we want to keep around and it seems like you are cleaning up `welcomeClicked` after we call the track? (definitely 80/20 to keep it as is right now since we are only doing it for one place.)

packages/plugin-core/src/_extension.ts
```diff
@@ -176,6 +177,23 @@ class ExtensionUtils {
...
+  static trackWelcomeClicked() {
+    const welcomeClicked = MetadataService.instance().getWelcomeClicked();
+    // check if we have a welcome click message
+    // see [[../packages/plugin-core/src/WelcomeUtils.ts#^z5hpzc3fdkxs]] where this property is set
+    if (welcomeClicked) {
+      AnalyticsUtils.trackSync({
+        event: TutorialEvents.ClickStart,
+        timestamp: welcomeClicked,
+      }).then(() => {
+        MetadataService.instance().deleteMeta("welcomeClicked");
```
jonathanyeung, line 193: Why are we deleting? Just to clean up something that we don't need again? (I think this is fine, but just want to clarify my understanding)

packages/common-all/src/timing.ts
```diff
@@ -18,3 +18,13 @@ export const nanosToMillis = (nanos: number) => {
 export const milliseconds = () => {
   return new Date().getTime();
 };
+
+export class TimeUtils {
+  static async sleep(ms: number) {
```
jonathanyeung, line 23: I feel like I remember seeing @tma66 add something similar recently?  Was it in test-utils though? Maybe we can consolidate that version to this one.

jonathanyeung (changes requested):

packages/plugin-core/src/utils/analytics.ts
```diff
@@ -94,20 +94,37 @@ export class AnalyticsUtils {
...
       integrations: { Amplitude: { session_id: sessionId } },
-    });
+    } as Parameters<typeof SegmentUtils.track>[0];
+  }
+
+  static track(event: string, props?: any) {
+    SegmentUtils.track(this._trackCommon({ event, props }));
+  }
+
+  static trackSync(opts: { event: string; props?: any; timestamp?: Date }) {
```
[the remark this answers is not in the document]
SeriousBug: That's also a bit tricky because if you do await the track call in the middle of an operation, it can block up to a few seconds. That means if we accidentally await it when not needed, we'll be introducing a significant delay for anyone with a slow connection. I think it makes sense to keep the primary function in a form where it's impossible to misuse.

Perhaps we can just do better naming? Instead of `trackSync` it could be `trackAndWait` or `trackWithPromise`.

packages/plugin-core/src/test/testUtilsV3.ts
```diff
@@ -360,9 +360,22 @@ export function setupBeforeAfter(
...
+      // in describeMultiWS > [[../packages/plugin-core/src/test/testUtilsV3.ts#^lk3whwd4kh4k]]
+      // TODO: keep in place until we completely remove `setupBeforeAndAfter`
+      try {
+        // @ts-ignore
+        sinon
+          .stub(VSCodeUtils, "getInstallStatusForExtension")
+          .returns(InstallStatus.NO_CHANGE);
+      } catch (e) {
+        // eat it.
+        sinon.restore();
```
SeriousBug, line 374: This restores everything, so it could restore more than intended. It's temporary, but would it be safer to just skip stubbing it if the first stub failed instead?

SeriousBug (approved):
Looks good to me other than this comment. Tests work fine locally. Thanks for catching the build issue!

packages/plugin-core/src/utils/analytics.ts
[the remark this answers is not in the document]
jonathanyeung: I think we (myself included) have a tendency to think "oh this returns a promise, so let me await it".  But I think the correct mental model should be to always evaluate "Do I need the results of the operation now?  **Only** if the answer is 'yes' should I await it. 

I think the best way to address this is with a warning in the function comments, which I had written in the underlying  track implementation (https://github.com/dendronhq/dendron/blob/20533bdf548b04d3d1c7af09bb7d4f1b520c76d7/packages/common-server/src/analytics.ts#L303), but that has since been masked with the addition of the track methods in the `SegmentUtils.ts` class. 

Another point - do we actually need a `SegmentUtils` static class when `SegmentClient` is a singleton with a static getter?  Seems redundant to me.

packages/engine-server/src/metadata/service.ts
kevinslin: updated

kevinslin:
> Nice catch on homeDir not being stubbed properly. This definitely have caused some head scratching for me in the past. I notice you refactored some tests but we still have some in other places. Are the ones you fixed the major ones that fail our CI? Do we have plans to remove the rest?

The issue happens everywhere `setupBeforeAfter` is used. there are 32 instances of this happening. I fixed 15 of them before running out of time (I had a timebox of 1h to fix tests) which is why I added the try/catch in `setupBeforeAfter` so that remaining tests will still run. We still want to remove the tests - @tma66, let me know if this makes sense for you to do as part of testing refactor

> On using the metadata store to save the welcome click event: I can guess that this pattern may become something we do more in the future. When the time comes, do you think it's worth creating a separate store destination instead of using `meta.json`? Since `meta.json` is something we want to keep around and it seems like you are cleaning up `welcomeClicked` after we call the track? (definitely 80/20 to keep it as is right now since we are only doing it for one place.)

i think it'll make sense but we'll cross that bridge when we get there :)

packages/plugin-core/src/utils/analytics.ts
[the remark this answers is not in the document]
kevinslin: fair enough. we'll also want to fix this in `common-server` though since we use this pattern there as well. will have a separate pr for that

packages/plugin-core/src/_extension.ts
kevinslin: we check for this field at startup to see if we need to send a `welcomeClicked` event. if we don't delete it, we'll send the same event on every startup

packages/plugin-core/src/test/testUtilsV3.ts
kevinslin: it would - unfortunately this still causes tests to fail since there are other sinon stubs underneath that will still throw an error, as well as potential stubs in the `beforeHook`. decided it wasn't worth the trouble

PR #2716, opened 2022-04-08
fix: metadata service respects initial installs from new vscode instances for improved metadata accuracy

hikchoi:
# fix: metadata service respects initial installs from new vscode instances for improved metadata accuracy

This PR:
- fixes an issue where an install event on a fresh vscode instance is recorded as an "initial install" globally.
  - this may have caused problems with prompt logic that rely on the integrity of the metadata service.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -3,6 +3,7 @@ export enum VSCodeEvents {
   ServerCrashed = "ServerCrashed",
   InitializeWorkspace = "InitializeWorkspace",
   Install = "Install",
+  NewInstanceInstall = "NewInstanceInstall",
   SchemaLookup_Show = "SchemaLookup_Show",
   NoteLookup_Gather = "NoteLookup_Gather",
   SchemaLookup_Gather = "SchemaLookup_Gather",
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -3,7 +3,6 @@ export enum VSCodeEvents {
   ServerCrashed = "ServerCrashed",
   InitializeWorkspace = "InitializeWorkspace",
   Install = "Install",
-  NewInstanceInstall = "NewInstanceInstall",
   SchemaLookup_Show = "SchemaLookup_Show",
   NoteLookup_Gather = "NoteLookup_Gather",
   SchemaLookup_Gather = "SchemaLookup_Gather",
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -414,9 +414,22 @@ export async function _activate(
       previousGlobalVersion,
       currentVersion,
     });
+
+    // check if this is an install event, but a repeated one on a new instance.
+    let isNewInstanceInstall = false;
+
     // set initial install ^194e5bw7so9g
     if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
-      MetadataService.instance().setInitialInstall();
+      // even if it's an initial install for this instance of vscode, it may not be for this machine.
+      // in that case, we should skip setting the initial install time since it's already set.
+      const metadata = MetadataService.instance().getMeta();
+      if (metadata.firstInstall === undefined) {
+        MetadataService.instance().setInitialInstall();
+      } else {
+        // we still want to proceed with InstallStatus.INITIAL_INSTALL because we want everything
+        // tied to initial install to happen in this instance of VSCode once for the first time
+        isNewInstanceInstall = true;
+      }
     }
 
     // TODO: temporary backfill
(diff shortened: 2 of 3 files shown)


packages/common-all/src/analytics.ts
```diff
@@ -3,6 +3,7 @@ export enum VSCodeEvents {
   ServerCrashed = "ServerCrashed",
   InitializeWorkspace = "InitializeWorkspace",
   Install = "Install",
+  NewInstanceInstall = "NewInstanceInstall",
```
kevinslin, line 6: instead of creating a new event, can we use "Install" event but with the property, "newInstance" set to true?

kevinslin (changes requested):

packages/common-all/src/analytics.ts
hikchoi: fixed, and also recorded in airtable

packages/plugin-core/src/_extension.ts
```diff
@@ -394,9 +394,22 @@ export async function _activate(
...
+    // check if this is an install event, but a repeated one on a new instance.
+    let isNewInstanceInstall = false;
+
     // set initial install ^194e5bw7so9g
     if (extensionInstallStatus === InstallStatus.INITIAL_INSTALL) {
-      MetadataService.instance().setInitialInstall();
+      // even if it's an initial install for this instance of vscode, it may not be for this machine.
+      // in that case, we should skip setting the initial install time since it's already set.
+      const metadata = MetadataService.instance().getMeta();
+      if (metadata.firstInstall === undefined) {
```
jonathanyeung, line 406: Any chance that during our transition, that someone who is an active user of Dendron doesn't have `metadata.firstInstall` defined yet? 

My preference would also to be check for the presence of the `dendron.uuid` file - if it exists, then also don't set to initial install.  That way, since the uniqueness of `track` is based on .uuid, we should ensure we're 100% aligned and we won't have old users showing up in the onboarding funnel.

packages/plugin-core/src/_extension.ts
```diff
@@ -805,12 +821,20 @@ async function showWelcomeOrWhatsNew({
...
+        msg: `extension, ${
+          isNewInstanceInstall
+            ? "initial install"
+            : "install on new vscode instance"
+        }`,
+      });
       // track how long install process took ^e8itkyfj2rn3
       AnalyticsUtils.track(VSCodeEvents.Install, {
         duration: getDurationMilliseconds(start),
+        isNewInstanceInstall,
```
jonathanyeung, line 835: Can we rename this to something like "isSecondaryInstallation" or something like that?  Because even if this is the first ever installation, it's still technically a "new instance install".

packages/plugin-core/src/test/suite-integ/Extension.test.ts
```diff
@@ -1093,6 +1093,38 @@ suite("GIVEN Dendron plugin activation", function () {
       }
     );
   });
+
+  describe("AND WHEN fresh install on new vscode instance", () => {
```
jonathanyeung, line 1097: for clarity, can we reword this to "on a secondary vscode instance"

jonathanyeung (changes requested):

kevinslin:
can you rebase from master? fixes some build issues

packages/plugin-core/src/_extension.ts
hikchoi: > Any chance that during our transition, that someone who is an active user of Dendron doesn't have metadata.firstInstall defined yet?

That's possible. That's why we have a temporary backfill logic that follows right after.
Though if an active user doesn't have it defined yet, and never went through that backfill, and also has installed a fresh instance of vscode, this would fill it up with the current time.

Your suggestion to check for the uuid would solve the problem though. So I'll go with that.

hikchoi:
Added suggested changes. Took some troubleshooting to figure out when the first Segment client instantiation actually happens. Maybe it's time for a deep cleansing of the activation logic 😄

@kevinslin I noticed that you removed an entire suite of tests here: https://github.com/dendronhq/dendron/pull/2714/files#diff-60b981e53d27fcadc5d898d4d161eccda9235211cc64e308cce08114e08b8fa1L987

I'm adding them back as they are needed to test the behavior of my changes, but were there any particular reason those were removed in your PR?

I did have to go back to your changes to remind myself how to clean up the extension context to make it work properly.

jonathanyeung (approved):

PR #2727, opened 2022-04-11
feat(workspace): Meeting Notes

jonathanyeung:
## feat(workspace): Meeting Notes

Adds a new convenience command to create meeting notes.  When a user creates a meeting note, we will create a pre-canned template for them, as well as a pre-built schema such that the template will automatically apply to their meeting notes (provided they don't drastically modify the note name, see approach discussion below).

As part of this change, adds the following additional things:
- Moved some code around to add a utility to generate a schema file (extracted from the previously implemented `CreateSchemaFromHieararchy` command.  We can also use this utility for the long-discussed, never implemented pre-canned template for Daily Journal Notes for new users.
- Some light infra for modifying the tutorial for an AB test.
- Some light infra for "tip-of-the-day" toast (called `FeatureShowcase` in the code)

### AB Testing for this Feature
1. We will toast an 'A' group with a "Try Meeting Notes?" popup upon activation.
2. We will show additional content around meeting notes for Tutorial page 2 for an 'A' group.

TODO: Add Docs + Telem Docs.  I'll do this after an initial review pass.

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended


jonathanyeung:
Adding prev draft pr comment for historical tracking.  After offline discussion, we have decided to go for approach 2.

---

Mostly looking for feedback on the user flow for meeting notes:

Note traits don't work with schemas yet.  Until those systems integrate well, I'm taking an 80/20 approach for getting a template applied to meeting notes:

### Approach 1: (implemented by this review, barring some TODO's)
- we create a user setting that allows a user to specify which note they want to use as their meeting notes template
- upon running create-meeting-notes cmd for the first time, we will create a pre-defined template at a static location
- we'll give hints to the user that they can edit the template if they like, or also change the user setting to point to a different location altogether for their meeting note if they so choose.

#### Pros
- Should make sure template gets applied, no matter which destination the user picks for their meeting note. This is the desired behavior for allowing note traits to apply a template going forward.

#### Cons
- Not consistent with the way schema templates work today
- Adds an additional user setting.

### Approach 2:
- meeting notes default to `meet.` prefix, so we can auto create a schema matching this hierarchy and make a template apply to it.
- upon running create-meeting-notes cmd for the first time, we will create a pre-defined template at the same static location we created in previous step.

#### Pros
- Consistent with the way schema templates work today

#### Cons
- Schema won't get applied if the user changes the note path when running the command.
- We have to create a schema file for users

packages/common-all/src/SchemaCreationUtils.ts
```diff
@@ -0,0 +1,73 @@
+import YAML from "js-yaml";
+import _ from "lodash";
+import { SchemaData, SchemaTemplate } from "./types";
+
```
kevinslin, line 4: in common-all, all file names are small case. can we stick to the same convention?

this file could just be called `schema` and we can put schema utils inside of it

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -54,14 +61,22 @@ export class TutorialInitializer
...
+
+    const replaceString =
+      ABUserGroup === MeetingNoteTestGroups.show ? MeetingNoteTutorialText : "";
+
     // Tailor the tutorial text to the particular OS and for their workspace location.
     const options = {
       files: [path.join(vpath, "*.md")],
 
-      from: [/%KEYBINDING%/g, /%WORKSPACE_ROOT%/g],
+      from: [/%KEYBINDING%/g, /%WORKSPACE_ROOT%/g, /%MEETING_NOTE_CONTENT%/g],
```
kevinslin, line 75: this works but i think moving forward, we should think of just having different zip files with different tutorials.
instead of modifying tutorial in text, just swap out with completely different tutorial (its okay if 90% of text is the same)

the reason is we still need to sync the updated tutorial with the workspace tutorial. the 80/20 way of doing this is to run `export pod` and dump the contents.
showing native keybindings depending on the operating system is nice but makes this process more complex because we can't just use the exported docs. 

to simplify syncing docs and ab test, i would propose just doing a/b tests with completely different doc sets

packages/plugin-core/src/commands/CreateMeetingNoteCommand.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  /**
+   *
+   * @param ext
+   * @param noConfirm - for testing purposes only; don't set in production code
+   */
+  constructor(ext: IDendronExtension, noConfirm?: boolean) {
+    const workspaceService = ext.workspaceService;
+
+    if (!workspaceService) {
```
kevinslin, line 42: won't this always be the case if `requireActiveWorkspace = true`?

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
```diff
@@ -110,6 +110,11 @@ export class CreateNoteWithTraitCommand extends BaseCommand<
       title = this.trait.OnCreate.setTitle(context);
     }
 
+    let body;
+    if (this.trait.OnCreate?.setBody && this.checkWorkspaceTrustAndWarn()) {
```
kevinslin, line 114: can we be more specific on the error? `checkWorkspaceTrustAndWarn` should take a helper text so we can tell the user which method failed (eg. `onCreate.setBody` vs `onCreate.setTitle`)

packages/plugin-core/src/FeatureShowcase.ts
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Tip to try out meeting notes
+   */
+  private showMeetingNotesTip() {
+    const confirm = "Try It";
+    const reject = "Later";
+
+    vscode.window
+      .showInformationMessage(
+        `Would you like to try out the new Meeting Notes Feature?`,
```
kevinslin, line 45: would change the text to provide more context. 

Dendron now has meeting notes which makes it easy to keep track of your meetings. Would you like to try it out?

"Create a meeting note", "Later"

packages/plugin-core/src/FeatureShowcase.ts
```diff
@@ -0,0 +1,78 @@
...
+ * now, we have only 1 feature - we can expand functionality in this class as we
+ * have more tips.
+ */
+export class FeatureShowcase {
+  /**
+   * Show right now uses vscode.showInformationMessage.  Later, we can have
+   * different display locations, such as the proposed Dendron Side Panel
+   */
+  public show(): void {
+    const ABUserGroup = MEETING_NOTE_FEATURE_SHOWCASE_TEST.getUserGroup(
```
kevinslin, line 22: can we disable this for tutorial workspace? this might be tricky - eg. if this is their first time using dendron and they are on the tutorial, we don't want to show tool tips for new features. maybe limit this to users that have run initialize workspace gt 3 times?

kevinslin (changes requested):
generally looks good. a few thoughts on the ab test harnesses

packages/plugin-core/src/FeatureShowcase.ts
jonathanyeung: One consideration here though is that I _think_ it's a little bit better if the text fits in one line, otherwise the user needs to click another button before they can see the entire text and may lose interest.

packages/plugin-core/src/commands/CreateMeetingNoteCommand.ts
jonathanyeung: Yes - but that depends on the logic of how the static prop `requireActiveWorkspace` works.  So if there isn't this null check / throw here, the usage of `workspaceService.config` will be a ts-error.

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
jonathanyeung: I think the error msg is correct - if workspace isn't trusted, all custom note-trait behavior should be blocked, irrespective of whether its setBody logic, setTitle logic, or something else. 

There is a problem here though where it'll pop the toast multiple times; I've fixed that now.

packages/plugin-core/src/FeatureShowcase.ts
jonathanyeung: I'm inclined to shorten it to "Dendron now has meeting notes - Try it out!"
[ Create a Meeting Note ] [Later]

We can A/B test it later.

packages/plugin-core/src/FeatureShowcase.ts
jonathanyeung: Do we keep track of init workspace > 3 times?  

How about instead, we do `firstWsInitialize` at least 1 week ago?

packages/plugin-core/src/FeatureShowcase.ts
```diff
@@ -0,0 +1,88 @@
...
+          MetadataService.instance().setFeatureShowcaseStatus(
+            ShowcaseEntry.TryMeetingNotes
+          );
+        } else {
+          // Don't set the metadata because the option for rejected is "Later",
+          // let's toast the user again later.
+          userResponse = "rejected";
+        }
+
+        AnalyticsUtils.track(VSCodeEvents.FeatureShowcased, {
```
kevinslin, line 77: can we also track when the toast shows up but then dissapears? this means the user never did anything tieh the prompt. so before we show the information message, we track a `FeatureShowCasePrompted` event and then also track a `FeatureShowCaseAccepted` event

kevinslin (approved):

chmac:
It's cool that this (and many other new features) are landing in dendron. Is this documented somewhere? I can't find anything about "meeting notes" in the wiki, nor on the docs site.

In my opinion, it's a real tragedy to release new features which aren't documented. How are users supposed to discover and use these features without the docs in place?

Harshita-mindfire:
> It's cool that this (and many other new features) are landing in dendron. Is this documented somewhere? I can't find anything about "meeting notes" in the wiki, nor on the docs site.
> 
> In my opinion, it's a real tragedy to release new features which aren't documented. How are users supposed to discover and use these features without the docs in place?

Hey @chmac, Meeting Note is a special note in Dendron. You can find it documented here: https://wiki.dendron.so/notes/5c213aa6-e4ba-49e8-85c5-1bdcb33ce202/#meeting-note

chmac:
@Harshita-mindfire Ah, that's awesome, thanks for the link. I wonder if there's some way to make that show up when typing "meeting note" into the search, which it currently doesn't (at least for me).

Maybe adding links to the docs in the changelog would be cool. Anyway, this is all a bit off topic from this PR, thanks for pointing me at the docs @Harshita-mindfire 👍

PR #2732, opened 2022-04-12
chore: new engagement metrics

jonathanyeung:
## chore: new engagement metrics

Adds new metrics for new KPI's around user engagement:
- Note Update / Delete / Create based off of EngineEvents
- Note View - if a user opens up a new Dendron Note and keeps it as the active text editor for at least 5 seconds.  I did some testing to see how 5 seconds 'feels', and to me it feels like a reasonable grace period, but I'm open to tweaking this as well.  The goal is to reasonably capture the effect of someone looking at a note and getting some info out of it while also filtering out any instances of a user opening a note inadvertently or opening the note they weren't looking for.

I also did some cleanup on the watcher class to remove the legacy `registerActiveTextEditorChangedHandler` pattern.

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -136,9 +136,7 @@ export enum NativeWorkspaceEvents {
 
 export enum EngagementEvents {
   NoteViewed = "NoteViewed",
-  NoteCreated = "NoteCreated",
-  NoteUpdated = "NoteUpdated",
-  NoteDeleted = "NoteDeleted",
+  EngineStateChanged = "EngineStateChanged",
 }
 
 export enum AppNames {
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -1,3 +1,6 @@
+// For all new additions to the telemetry events, follow UpperCamelCasing and
+// use noun+verb for the event name.
+
 /* eslint-disable camelcase */
 export enum VSCodeEvents {
   ServerCrashed = "ServerCrashed",
@@ -131,6 +134,13 @@ export enum NativeWorkspaceEvents {
   DetectedInNonDendronWS = "Native_Workspace_Detected_In_Non_Dendron_WS", // watcher has detected a Dendron workspace getting created inside a non-Dendron workspace
 }
 
+export enum EngagementEvents {
+  NoteViewed = "NoteViewed",
+  NoteCreated = "NoteCreated",
+  NoteUpdated = "NoteUpdated",
+  NoteDeleted = "NoteDeleted",
+}
+
 export enum AppNames {
   CODE = "vscode",
   CLI = "cli",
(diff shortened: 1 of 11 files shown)


packages/dendron-cli/src/commands/launchEngineServer.ts
```diff
@@ -101,9 +106,47 @@ export class LaunchEngineServerCommand extends CLICommand<
...
+
+      // These events will only upload if the upload action completes before the
+      // CLI command completes. They are uploaded on a best effort basis.
+      engine.onEngineNoteStateChanged((entries) => {
+        const createCount = extractNoteChangeEntriesByType(
+          entries,
+          "create"
+        ).length;
+
+        if (createCount > 0) {
```
SeriousBug, line 122: Why not consolidate these into a single event with properties for what kind of change it is? I'm asking since one of the Amplitude review items was that we had a lot of events repeated after one another.

SeriousBug (commented):
Looks good, left some feedback on the use of events.

packages/dendron-cli/src/commands/launchEngineServer.ts
jonathanyeung: At first I was thinking that it'd be easier to build analytical views with them as separate events, but on second thought it should be fine as a consolidated event. These will likely be noisy, so I can make the adjustment.  👍🏼

jonathanyeung:
Airtable Upcoming Release (targeting v91) has been updated with the 2 new events.

packages/dendron-cli/src/commands/launchEngineServer.ts
```diff
@@ -101,9 +106,36 @@ export class LaunchEngineServerCommand extends CLICommand<
...
       ws: wsRoot,
     });
+
     if (init) {
       this.L.info({ ctx, msg: "pre:engine.init" });
       const out = await engine.init();
+
+      // These events will only upload if the upload action completes before the
+      // CLI command completes. They are uploaded on a best effort basis.
+      engine.onEngineNoteStateChanged((entries) => {
```
kevinslin, line 116: fine for now but in the future, if we find this happens a lot - we might want to build a system around writing these metrics to disk and having a background thread that uploads entries when it can basis

kevinslin (approved):
looks good. no blockers, just a few thoughts and a reminder to schedule an async about the new convention around metrics

packages/common-all/src/analytics.ts
```diff
@@ -1,3 +1,6 @@
+// For all new additions to the telemetry events, follow UpperCamelCasing and
```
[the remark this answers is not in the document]
jonathanyeung: @hikchoi actually already added it here a couple of weeks ago.

PR #2746, opened 2022-04-13
chore: clean up state service dependencies

hikchoi:
# chore: clean up state service dependencies

This PR:
- Removes all use of state service to prefer the use of metadata service instead
  - except for `MRUDocs` in global state.
- All remaining references to the state service are there to backfill past state into metadata.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [~] [Confirm manual testing](https://wiki.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html) 
- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended


hikchoi:
@Harshita-mindfire This PR moves all global state to metadata service. I left `MRUDocs` untouched as I'm not sure if this should be moved to the metadata service (not quite sure how this is used in the pods). Any thoughts? Maybe we should keep the state service around just for the pods 😄

Harshita-mindfire:
> @Harshita-mindfire This PR moves all global state to metadata service. I left `MRUDocs` untouched as I'm not sure if this should be moved to the metadata service (not quite sure how this is used in the pods). Any thoughts? Maybe we should keep the state service around just for the pods 😄

For Pods that interact with the global state, we add a wrapper function in `plugin-core/pods.ts` and pass them as `utility methods` to pods. If it is not a blocker we can keep the state service, and meanwhile I can look into this.

Harshita-mindfire:
> @Harshita-mindfire This PR moves all global state to metadata service. I left `MRUDocs` untouched as I'm not sure if this should be moved to the metadata service (not quite sure how this is used in the pods). Any thoughts? Maybe we should keep the state service around just for the pods 😄

For Pods that interact with the global state, we add a wrapper function in `plugin-core/pods.ts` and pass them as `utility methods` to pods. If it is not a blocker we can keep the state service, and meanwhile I can look into this.

packages/engine-server/src/metadata/service.ts
```diff
@@ -98,6 +122,21 @@ export class MetadataService {
...
+  }
+
+  getActivationContext() {
+    return (
+      this.getMeta().workspaceActivationContext ??
+      WorkspaceActivationContext.normal
+    );
+  }
+
   setMeta(key: keyof Metadata, value: any) {
```
Harshita-mindfire, line 140: Just a Q:
How can we add metadata if the key is dynamic? 
eg:  to store hierarchy destination for imported google doc. (key: note.fname, value: user input)

kevinslin: we can't (or if we do, it would no longer be typed). for this use case, would have the key be "pods.gdoc" and the value be an object that can have dynamic keys

packages/engine-server/src/workspace/service.ts
```diff
@@ -1504,4 +1512,15 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
       },
     });
   }
+
+  resetMeta() {
```
kevinslin, line 1516: can we also add a flag indicating that the metadata has been reset? can be a new property that we should also report on

kevinslin:
looks good, just a comment about adding one more variable to track resets

packages/engine-server/src/workspace/service.ts
hikchoi: I've added this so I can make `ResetConfigCommand` to clean this up, but I ended up keeping that command intact for now. (so this method is not used anywhere yet. I might just remove it)

I think it serves very little use cases for the users and only for dev purposes.

Even then, since we are now dealing with a simple json files (`.dendron.ws` and `meta.json`) instead of a sqlite database (vscode's `globalStorage` and `workspaceStorage`), I think it's better to directly modify / backup the actual files rather than have a command that nukes it.

At the very least we should have the enablement set to false in production. What do you think? (analytics show that this command has very little use. I suspect they are not intentional uses as well)

hikchoi:
> > @Harshita-mindfire This PR moves all global state to metadata service. I left `MRUDocs` untouched as I'm not sure if this should be moved to the metadata service (not quite sure how this is used in the pods). Any thoughts? Maybe we should keep the state service around just for the pods 😄
> 
> For Pods that interact with the global state, we add a wrapper function in `plugin-core/pods.ts` and pass them as `utility methods` to pods. If it is not a blocker we can keep the state service, and meanwhile I can look into this.

Sounds good. I'll leave it as-is and hand it over to you for the remaining pod related refactoring 👍🏼

packages/engine-server/src/metadata/service.ts
hikchoi: something like

```js
type Metadata {
  ...
  "pods.gdoc": Record<string, string>
}
```

or

```js
type PodsGDoc = {[key: string]: string};
type Metadata {
  ...
  "pods.gdoc": PodsGDoc
  ...
}
```

not entirely sure of the implications of using a `Record` type vs a regular index signature, but both seem identical on the surface.

packages/engine-server/src/workspace/service.ts
hikchoi: also, most internal use of the command were in tests (and I suspect it will be moving forward as well), and now this being tied to a workspace service kinda makes it useless for those cases. We can't use it in hooks that run before workspace initialization. I would probably opt for creating a utility function for tests that directly grabs the files filesystem and modifies them.

I'll remove them from the workspace service until we find the need to reintroduce them

packages/engine-server/src/metadata/service.ts
Harshita-mindfire: Got it. Thank you

packages/engine-server/src/workspace/service.ts
kevinslin: for a separate PR, can we introduce a new flag? (eg. enableDevContext). if set, we switch to development context which enables developer only commands like resetConfig

kevinslin:
looks good, can you rebase from master to solve conflicts?

PR #2756, opened 2022-04-13
enhance(workspace): Remove stale entries from cache upon initialization

tma66:
**Background**
Currently the cache is constantly growing as we never remove entries. This PR will do a diff between the files read and the current cache and remove entries that are no longer valid

See [[Remove deleted notes from cache|dendron://private/task.2022.03.10.cache-backlinks-at-boot#remove-deleted-notes-from-cache]] for more information

**Changes**
1. Delete entries in cache if not found in fileMetaDict
2. Update `NotesCacheEntry` to include optional status. This will be later used in a backlinks cache implementation

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/types/index.ts
+++ b/packages/common-all/src/types/index.ts
@@ -63,9 +63,19 @@ export type NotesCache = {
   notes: NotesCacheEntryMap;
 };
 export type NotesCacheEntryMap = { [key: string]: NotesCacheEntry };
-export type NotesCacheEntry = {
+export type NotesCacheEntry =
+  | {
+      hash: string;
+      data: Omit<NoteProps, "body">;
+      status?: "create" | "delete";
+    }
+  | NotesCacheUpdateEntry;
+
+export type NotesCacheUpdateEntry = {
   hash: string;
+  prevData: Omit<NoteProps, "body">;
   data: Omit<NoteProps, "body">;
+  status?: "update";
 };
 
 export type DendronSiteFM = {
--- a/packages/common-all/src/types/index.ts
+++ b/packages/common-all/src/types/index.ts
@@ -63,19 +63,9 @@ export type NotesCache = {
   notes: NotesCacheEntryMap;
 };
 export type NotesCacheEntryMap = { [key: string]: NotesCacheEntry };
-export type NotesCacheEntry =
-  | {
-      hash: string;
-      data: Omit<NoteProps, "body">;
-      status?: "create" | "delete";
-    }
-  | NotesCacheUpdateEntry;
-
-export type NotesCacheUpdateEntry = {
+export type NotesCacheEntry = {
   hash: string;
-  prevData: Omit<NoteProps, "body">;
   data: Omit<NoteProps, "body">;
-  status?: "update";
 };
 
 export type DendronSiteFM = {
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -556,12 +556,25 @@ export class FileStorage implements DStore {
     const {
       notes,
       cacheUpdates,
+      fileMetaDict,
       errors: parseErrors,
     } = await new NoteParser({
       store: this,
       cache,
       logger: this.logger,
     }).parseFiles(noteFiles, vault);
+    const seenFiles = new Set(
+      _.values(fileMetaDict).flatMap((items) => {
+        return items.map((items) => {
+          return path.parse(items.fpath).name;
+        });
+      })
+    );
+    _.keys(cache.notes).map((key) => {
+      if (!seenFiles.has(key)) {
+        delete cache.notes[key];
+      }
+    });
     errors = errors.concat(parseErrors);
     this.logger.info({ ctx, msg: "parseNotes:fin" });
 
@@ -623,7 +636,7 @@ export class FileStorage implements DStore {
             return;
           }
           try {
-            const anchors = await AnchorUtils.findAnchors({
+            const anchors = AnchorUtils.findAnchors({
               note: n,
             });
             cacheUpdates[n.fname].data.anchors = anchors;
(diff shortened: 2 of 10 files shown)


packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -556,12 +556,25 @@ export class FileStorage implements DStore {
...
       notes,
       cacheUpdates,
+      fileMetaDict,
       errors: parseErrors,
     } = await new NoteParser({
       store: this,
       cache,
       logger: this.logger,
     }).parseFiles(noteFiles, vault);
+    const seenFiles = new Set(
```
kevinslin, line 566: instead of constructing it from scratch, can we build this set in `getAllFiles` and return it so we're not iterating over all notes twice?

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -556,12 +556,25 @@ export class FileStorage implements DStore {
...
+      _.values(fileMetaDict).flatMap((items) => {
+        return items.map((items) => {
+          return path.parse(items.fpath).name;
+        });
+      })
+    );
+    _.keys(cache.notes).map((key) => {
```
kevinslin, line 573: instead of removing cache entries here, we already loop through all notes below. can we delete notes from this set as we process them - any notes not deleted should be considered stale and need to be deleted from the cache

kevinslin (changes requested):

packages/engine-server/src/drivers/file/storev2.ts
tma66: We call `fs.readdir` which returns an array. Converting to set will require iterating twice. However, I can reuse the data from `getAllFiles` so it makes it more readable

packages/engine-server/src/drivers/file/storev2.ts
tma66: good point, fixed

packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
```diff
@@ -54,6 +58,55 @@ describe("engine, cache", () => {
     );
   });
 
+  test("delete stale entries from cache", async () => {
```
kevinslin, line 61: can we have some more test cases?
ideally, add this to the `delete` preset > https://github.com/dendronhq/dendron/blob/chore/migrate-cal-widget-kevin/packages/engine-test-utils/src/presets/engine-server/delete.ts#L38:L38

make sure the cache is correct after the deletion

kevinslin (changes requested):

packages/engine-test-utils/src/__tests__/engine-server/enginev2.spec.ts
tma66: Moved and updated other presets to test for cache

packages/common-all/src/types/index.ts
```diff
@@ -63,9 +63,19 @@ export type NotesCache = {
   notes: NotesCacheEntryMap;
 };
 export type NotesCacheEntryMap = { [key: string]: NotesCacheEntry };
-export type NotesCacheEntry = {
+export type NotesCacheEntry =
+  | {
+      hash: string;
+      data: Omit<NoteProps, "body">;
+      status?: "create" | "delete";
```
jonathanyeung, line 70: Are we actually using this `status` prop anywhere? If not, can we remove it?

jonathanyeung (changes requested):

jonathanyeung:
Also looks like there's a test failure: 

```
@dendronhq/plugin-core:   1 failing
@dendronhq/plugin-core:   1) MoveNoteCommand
@dendronhq/plugin-core:        DOMAIN_NO_CHILDREN
@dendronhq/plugin-core:          THEN correct results:
@dendronhq/plugin-core:      ReferenceError: expect is not defined
@dendronhq/plugin-core:   	at TestPresetEntryV4.NOTES.DOMAIN_NO_CHILDREN.common_test_utils_1.TestPresetEntryV4.preSetupHook.async.vaults.vaults (/home/runner/work/dendron/dendron/packages/engine-test-utils/lib/presets/engine-server/rename.js:424:9)
```

packages/common-all/src/types/index.ts
tma66: I have a followup pr that will use it for backlinks caching. Mostly put it here so people can see the schema changes before I do the implementations. Hope you don't mind if I leave it here

tma66:
> Also looks like there's a test failure:
> 
> ```
> @dendronhq/plugin-core:   1 failing
> @dendronhq/plugin-core:   1) MoveNoteCommand
> @dendronhq/plugin-core:        DOMAIN_NO_CHILDREN
> @dendronhq/plugin-core:          THEN correct results:
> @dendronhq/plugin-core:      ReferenceError: expect is not defined
> @dendronhq/plugin-core:   	at TestPresetEntryV4.NOTES.DOMAIN_NO_CHILDREN.common_test_utils_1.TestPresetEntryV4.preSetupHook.async.vaults.vaults (/home/runner/work/dendron/dendron/packages/engine-test-utils/lib/presets/engine-server/rename.js:424:9)
> ```
> 
> By the way the tests are written, I don't think you can add these statements in the body of the test. The validation has to be done on the plugin-core side of the test.
> 
> ```ts
>       expect(_.size(cacheVault.notes)).toEqual(3);
>       expect(_.find(cacheVault.notes["beta"])).toBeTruthy();
> ```

Ah that makes sense. I'll rework the tests

packages/common-all/src/types/index.ts
tma66: Actually I'm going to remove. Turns out my next pr will do things a little differently anyways

kevinslin (approved):
nice job adding tests to the other cases as well :)

jonathanyeung (approved):

PR #2768, opened 2022-04-15
enhance: detect misconfigured self contained vaults and offer to fix them

SeriousBug:
If the user has a self contained vault that was incorrectly configured as a self contained vault, then this PR will detect that issue and offer to fix it with a prompt. This PR also adds the doctor action that will fix the misconfigured vault.

image

I also disabled the "create missing root note/schema" automatic fix for when it's a misconfigured self contained vault, as otherwise it would create a root note in the wrong folder. That means the initialization fails, but Dendron will unbreak itself once the user fixes the config or accepts out prompt.

### DIFF
--- a/packages/plugin-core/src/commands/ReloadIndex.ts
+++ b/packages/plugin-core/src/commands/ReloadIndex.ts
@@ -200,7 +200,7 @@ export class ReloadIndexCommand extends BasicCommand<
     const { wsRoot, engine } = ws;
 
     // Check if there are any misconfigured self contained vaults.
-    // Deliberately now awaiting this to avoid blocking the reload
+    // Deliberately not awaiting this to avoid blocking the reload
     ReloadIndexCommand.checkAndPromptForMisconfiguredSelfContainedVaults({
       engine: ExtensionProvider.getEngine(),
     });
--- a/packages/plugin-core/src/commands/ReloadIndex.ts
+++ b/packages/plugin-core/src/commands/ReloadIndex.ts
@@ -14,12 +14,15 @@ import {
   schemaModuleOpts2File,
   vault2Path,
 } from "@dendronhq/common-server";
+import { DoctorActionsEnum, DoctorService } from "@dendronhq/engine-server";
 import fs from "fs-extra";
 import _ from "lodash";
 import path from "path";
 import { ProgressLocation, window } from "vscode";
 import { DENDRON_COMMANDS } from "../constants";
 import { ExtensionProvider } from "../ExtensionProvider";
+import { Logger } from "../logger";
+import { IEngineAPIService } from "../services/EngineAPIServiceInterface";
 import { AnalyticsUtils } from "../utils/analytics";
 import { BasicCommand } from "./base";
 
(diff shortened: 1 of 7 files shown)


packages/plugin-core/src/commands/ReloadIndex.ts
```diff
@@ -103,6 +119,66 @@ export class ReloadIndexCommand extends BasicCommand<
...
+  /** Checks if there are any self contained vaults that aren't marked correctly, and prompts the user to fix the configuration. */
+  static async checkAndPromptForMisconfiguredSelfContainedVaults({
+    engine,
+  }: {
+    engine: IEngineAPIService;
+  }) {
+    const ctx = "checkAndPromptForMisconfiguredSelfContainedVaults";
+    const { wsRoot, vaults } = ExtensionProvider.getDWorkspace();
+    const doctor = new DoctorService();
+    const vaultsToFix = await doctor.findMisconfiguredSelfContainedVaults(
```
kevinslin, line 131: can we track this? ideally the doctor tracking can take an extra property that specifies where it was being invoked from. 
values:

- "manual-trigger" vs "init-check"

kevinslin (changes requested):
just one comment about tracking doctor changes

packages/plugin-core/src/commands/ReloadIndex.ts
SeriousBug: Thanks for the reminder!

It's actually pretty hard to track the doctor command execution, because we need to reload the window afterwards for this command. That means the window reloads before segment sends out the command execution.

I'll add it anyway, and also add tracking for the detection which should be more reliable.

packages/plugin-core/src/commands/ReloadIndex.ts
```diff
@@ -116,6 +199,12 @@ export class ReloadIndexCommand extends BasicCommand<
     let initError: DendronError | undefined;
     const { wsRoot, engine } = ws;
 
+    // Check if there are any misconfigured self contained vaults.
+    // Deliberately now awaiting this to avoid blocking the reload
```
kevinslin, line 203: nitpick: deliberately not awaiting?

packages/plugin-core/src/commands/ReloadIndex.ts
kevinslin: do we need to reload window or can we get away with reloading index?

also, did we implement a source property to measure whether this was automatically triggered?

SeriousBug: I tried with just reloading the index at first, but it wasn't enough. I don't remember exactly where, I think the engine or the plugin `getVaults`, loads the vaults config at startup and doesn't update it after. If it's important, I can refactor the engine code to fix that and avoid the window reload.

We can look at the doctor command analytics which tracks the type of action that was run to understand the number of manual runs, and we can look at the auto-detect analytics to understand the automatically triggered runs. I can also add a property to `MissingSelfContainedVaultsMessageAccept` whether it originated from the automatic or manual run, but:
- The telemetry, most of the time, is not sent because of the window reload
- The other telemetry I mentioned already roughly tracks that information

kevinslin: got it. long term, we should refactor engien to be able to refresh its state without reload window. but not necessary for this pr. 

in terms of telemetry, two thoughts here:

1. we can try making it synchronous. i believe the latency is only incurred because segment sdk has a cache flush interval. we could force a flush when awaiting (would need to benchmark to see how long this takes)
2. we can use the same technique as used for the `GetWelcomeClicked` -> write data points to disk and then check on startup. see https://github.com/dendronhq/dendron/pull/2714 

Not a blocker on this PR but lets track this as a followup task

PR #2770, opened 2022-04-15
chore/fix tracking

kevinslin:
- add os as regular identify prop
- simplify logic
- rename vscode extension cmd
- get and track install status

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -8,11 +8,13 @@ import {
   ConfigEvents,
   ConfigUtils,
   CONSTANTS,
+  DendronConfig,
   DendronError,
   DENDRON_VSCODE_CONFIG_KEYS,
   getStage,
   GitEvents,
   InstallStatus,
+  IntermediateDendronConfig,
   isDisposable,
   MigrationEvents,
   NativeWorkspaceEvents,
@@ -41,10 +43,10 @@ import * as Sentry from "@sentry/node";
 import { ExecaChildProcess } from "execa";
 import fs from "fs-extra";
 import _ from "lodash";
+import os from "os";
 import path from "path";
 import semver from "semver";
 import * as vscode from "vscode";
-import os from "os";
 import {
   CURRENT_AB_TESTS,
   SelfContainedVaultsTestGroups,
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/settings.ts
```diff
@@ -24,7 +24,10 @@ import { DendronExtension, getDWorkspace } from "./workspace";
 export { Snippets };
 
 export class Extensions extends EngineExtension {
-  static getVSCodeExtnsion() {
+  /**
+   * Get Dendron recommended extensions
+   */
+  static getDendronVSCodeExtnsion() {
```
jonathanyeung, line 30: nit, carry over sp on 'Extension'.  Also, if this gets recommended extensions, should we just name the function that way? `getDendronRecommendedExtensions`

packages/plugin-core/src/_extension.ts
```diff
@@ -126,6 +128,87 @@ class ExtensionUtils {
     }
   };
 
+  static async runMigrationsIfNecessary({
```
jonathanyeung, line 131: Should we move some of these to the new `StartupUtils`? @hikchoi

jonathanyeung (changes requested):

packages/plugin-core/src/_extension.ts
hikchoi: that makes sense.

packages/plugin-core/src/_extension.ts
```diff
@@ -188,6 +385,41 @@ class ExtensionUtils {
       });
     }
   }
+
+  // run this when we don't have an active workspace
+  static watchForNativeWorkspace(ext: DendronExtension) {
```
SeriousBug, line 390: We could just drop this if you'd prefer, it has been deprecated and is not needed anymore. Feel free to keep it in this PR if that's easier though.

SeriousBug (approved):
Looks good to me, initializes fine in manual testing. Left one comment.

packages/plugin-core/src/_extension.ts
```diff
@@ -645,6 +826,13 @@ export async function _activate(
       }
       const reloadSuccess = await reloadWorkspace();
       const durationReloadWorkspace = getDurationMilliseconds(start);
+
+      await ExtensionUtils.trackWorkspaceInit({
+        durationReloadWorkspace,
+        activatedSuccess: !!reloadSuccess,
```
hikchoi, line 832: nit: maybe `activationSucceeded` as per noun + verb convention.

hikchoi (changes requested):
Looks good other than some nits and comments from others.

packages/plugin-core/src/settings.ts
kevinslin: renamed. this does recommended extensions to install AND remove

packages/plugin-core/src/_extension.ts
kevinslin: moved

packages/plugin-core/src/_extension.ts
kevinslin: updated

packages/plugin-core/src/_extension.ts
kevinslin: updated

PR #2774, opened 2022-04-16
fix plugin views browser mode

kevinslin:


maxkrieger:
looks like this is already in master via https://github.com/dendronhq/dendron/commit/bad48bea80e28a88c3b900347ad2caba55e13e19#diff-631c11fedfb5a47b0722141e278d351d4459921bb1265366cbbeae8e5e56bf90

Issue #2644, opened 2022-03-28
Wrong preview/linking when hierarchy segment starts with "md"


gbieging:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [x] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When creating a link to a note and one of the hierarchy segments starts with "md" the preview shows the link with only the part before that segment, clicking it also opens a note with only the fast part as the name.

## To Reproduce

1. Open a note
2. Write a link with a ".md" somewhere (e.g.: [[lorem.ipsum.mdolor.sit.amet]])
3. Open preview
5. Click link
6. Wrong note opens

## Expected behavior

The full link name to appear and clicking to go to the correct note

## Screenshots

![image](https://user-images.githubusercontent.com/14062649/160456209-38f70988-eb0c-4ef2-b584-5164196f8d53.png)

## Desktop (please complete the following information)

- OS: Windows
- Browser: N/A
- Versions (all that apply to bug):
  - Dendron Extension: 0.87.0
  - VS Code version: tested with 1.63.2 and 1.65.2
  - Dendron CLI:

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here

SeriousBug:
Thanks for reporting this problem, I can reproduce it. We'll fix this as soon as possible.

aleksey-rowan:
A very similar issue was reported last, but it was closed as it referred to using external links inside a wiki link:

- https://github.com/dendronhq/dendron/issues/2594

Also linking two tangentially-related issues that were fixed before (they both involve having `md` in filename):

- https://github.com/dendronhq/dendron/issues/1972
- https://github.com/dendronhq/dendron/issues/856

gbieging:
I've found two more details that might be relevant: 
- ctrl-clicking the link in the edit pane works correctly; 
- deleting a note created by clicking the broken link gives a warning that the link needs to be updated

Harshita-mindfire:
This should be fixed after Dendron v0.92. Feel free to comment here if that is not the case.

PR #2781, opened 2022-04-18
fix(view): broken preview for links with sub-hierarchy starting with .md

Harshita-mindfire:
Related issue: https://github.com/dendronhq/dendron/issues/2644

### Basics

### Extended
- General
  - [~] check if this change adversely impact performance
- Operations
  - [~] when shipping this change, will it just work or will it introduce additional operational overhead due to complicated interface or known bugs?
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
NA

## Docs
NA

### Extended
NA

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1200,10 +1200,11 @@ export class NoteUtils {
 
   static normalizeFname(nodePath: string) {
     // remove md extension
+    nodePath = _.trim(nodePath);
     if (nodePath.endsWith(".md")) {
       nodePath = nodePath.slice(undefined, -3);
     }
-    return _.trim(nodePath);
+    return nodePath;
   }
 
   static isNoteProps(props: Partial<NoteProps>): props is NoteProps {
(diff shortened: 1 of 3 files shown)


packages/common-all/src/dnode.ts
```diff
@@ -1200,9 +1200,8 @@ export class NoteUtils {
 
   static normalizeFname(nodePath: string) {
     // remove md extension
-    const idx = nodePath.lastIndexOf(".md");
-    if (idx > 0) {
-      nodePath = nodePath.slice(0, idx);
+    if (nodePath.endsWith(".md")) {
+      nodePath = nodePath.slice(undefined, -3);
     }
     return _.trim(nodePath);
```
SeriousBug, line 1206: Should probably trim before checking `endsWith`, in case `nodePath` was `note.md `.

SeriousBug (changes requested):
Looks good aside from a small nitpick. This seems to fix the issue for me.

I'd also suggest:
- adding a note with `md` in the middle to the test workspace
- adding a test case that would fail without this fix to prevent regressions in the future

SeriousBug (approved):

PR #2783, opened 2022-04-19
fix(view): views don't update for new notes with self contained vaults

SeriousBug:
The issue was caused by the API stringifying properties passed to it. This meant `selfContained: true` would become `selfContained: "true"`, and the engine would think that these are different vaults.

This issue was first discovered for the calendar view, but likely affected other views.

This fix solves the problem by parsing the stringified property. Also included are a few small refactors to use `VaultUtils.isEqualV2`. 

Depends on #2747.


Harshita-mindfire (approved):
lgtm. Works fine on testing

packages/common-frontend/src/features/engine/slice.ts
```diff
@@ -92,7 +92,7 @@ export const syncNote = createAsyncThunk(
       return resp;
     }
     const data = resp.data!;
-    logger.info({ state: "pre:setNotes" });
+    logger.info({ state: "pre:setNotes", notes: data.map(NoteUtils.toLogObj) });
```
kevinslin, line 95: this could be a lot of notes and might have performance impact. can we check perf of mapping 10k notes like this? 

also would change this to a debug statement

kevinslin (changes requested):
can you set the origin commit to calendar view? doesn't look like this is based off of master

jonathanyeung:
> can you set the origin commit to calendar view? doesn't look like this is based off of master

Let's try to merge the calendar-view-migration shortly after we fix the Ubuntu run (expand mem for node in CI): https://github.com/dendronhq/dendron/pull/2747

Afterward, let's rebase this change off master.

packages/common-frontend/src/features/engine/slice.ts
SeriousBug: This doesn't get used for note initialization I think, but I added a check to only log the notes if there's less than 10 notes. I think that does what I wanted, which is being able to see if the notes updated during development, and shouldn't affect performance for release.

SeriousBug:
It was a little hard to rebase this PR, so I decided to create a new PR instead. See here: #2790

PR #2790, opened 2022-04-20
fix(view): views don't update for new notes with self contained vaults

SeriousBug:
The issue was caused by the API stringifying properties passed to it. This meant selfContained: true would become selfContained: "true", and the engine would think that these are different vaults.

This issue was first discovered for the calendar view, but likely affected other views.

This fix solves the problem by parsing the stringified property. Also included are a few small refactors to use VaultUtils.isEqualV2.

This is the same as #2783, but there were a lot of conflicts when rebasing that PR so I opted to cherry pick it into a new PR.


kevinslin:
looks like a bunch of tests are failing?

kevinslin (approved):

PR #2791, opened 2022-04-20
disable fast refresh

kevinslin:


jonathanyeung (approved):

Harshita-mindfire (commented):
Can we update the valid views names with Calendar and graph panels:
```
  "DendronCalendarPanelView",
  "DendronGraphPanelView"
```
https://github.com/dendronhq/dendron/blob/571468c247429f65c605ab2e0a3e420d507cee20/packages/dendron-plugin-views/src/index.tsx#L2

Currently this gives an error of invalid name in dev-console.

PR #2794, opened 2022-04-20
feat(chore): germ stage implementation of config overrides

kevinslin:
feat(chore): germ stage implementation of config overrides

This is an early build of introducing [[46 Config Overrides|dendron://dendron.docs/rfc.46-config-overrides]]

It allows a custom `dendronrc.yml` file, located either inside the `$WS_ROOT` or the `$HOME` folder to be merged with the values in `dendron.yml`.

Most immediately, the usecase is to support dendron team members to easily add their own local vaults to the team workspace.

Docs: https://github.com/dendronhq/dendron-docs/pull/27

### Other Changes
- the `Extensions.test` file is getting quite large and conflates activating the extension and testing `SetupWorkspace`
    - split `Extensions.test` into 4 tests
        - Extension.test: test regular activation logic
        - Extension-PostInstall: test plugin install logic
        - Extension-PostUpgrade: test version upgrade logic
    - add `SetupWorkspace.test`
    - refactor some modules into shared test utilities

### Metrics
- adds the following metrics (will be synced with airtable once approved)

- EventProperties (part of Initialize Workspace):
    - hasLocalConfig, boolean, indicates if a local config is present
    - numLocalConfigVaults, number, number of vaults that are loaded from local config

### Post Approval tasks
- [ ] async about extensions.test
- [ ] update metrics

***

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended


packages/engine-server/src/workspace/service.ts
```diff
@@ -874,9 +879,8 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
   }
 
   createConfig() {
-    // This line actually does something: it will create a config if one doesn't exist.
     // eslint-disable-next-line no-unused-expressions
```
jonathanyeung, line 882: nit: in this case, we can also remove the eslint line right?

jonathanyeung (approved):
Looks good for a dogfood test.  Just minor comments / q's.

packages/engine-server/src/config.ts
```diff
@@ -174,6 +188,73 @@ export class DConfig {
...
+    wsRoot: string
+  ): RespV3<IntermediateDendronConfig> {
+    const wsPath = path.join(wsRoot, CONSTANTS.DENDRON_LOCAL_CONFIG_FILE);
+    const globalPath = path.join(
+      os.homedir(),
+      CONSTANTS.DENDRON_LOCAL_CONFIG_FILE
+    );
+    let foundPath: string | undefined;
+
+    if (fs.existsSync(globalPath)) {
```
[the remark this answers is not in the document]
kevinslin: yep, its one or the other. its in line with how rc files currently work in tools like `eslint`. can change in the future but opt for simple behavior for now

packages/engine-server/src/workspace/service.ts
kevinslin: you're right. fixed!

PR #2796, opened 2022-04-21
enhance(workspace): warn at startup and reload if there are notes with duplicate IDs

SeriousBug:
A somewhat common issue that can happen when people copy and paste notes is notes with duplicate IDs. This PR adds a check during initialization and Reload Index to detect and warn if there are notes with duplicate IDs.

**Q**: I wanted to add telemetry for this check, but I'm not sure where to add it. It looks like analytics are only available in plugin and CLI, so I can't track it from the engine. Detecting the error from the plugin is also difficult because I can't attach a special `error.status` or `error.code` to look for because these are deprecated, and I'm not sure how else to detect this error happened. Let me know if you have any thoughts.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [ ] [Confirm existing tests pass](https://wiki.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html)
- [ ] [Confirm manual testing](https://wiki.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html) 
- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended


kevinslin:
> A somewhat common issue that can happen when people copy and paste notes is notes with duplicate IDs. This PR adds a check during initialization and Reload Index to detect and warn if there are notes with duplicate IDs.
> 
> **Q**: I wanted to add telemetry for this check, but I'm not sure where to add it. It looks like analytics are only available in plugin and CLI, so I can't track it from the engine. Detecting the error from the plugin is also difficult because I can't attach a special `error.status` or `error.code` to look for because these are deprecated, and I'm not sure how else to detect this error happened. Let me know if you have any thoughts.

i think this makes a good case of not deprecating the `code` field in `IDendronError`. it can be a no-op in cases when its not needed but its useful to have an error table to map to for custom actions like this

its also a good example of `learn & adapt` - http return codes already provide a rich syntax for describing errors and its valid not just for http. for example, this is a good example of a 400, a client error 

```
The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).[30]
```

given a code + status, we can also have a lookup table to do custom actions based on errors

```ts

if code = 400 and status = ERROR_STATUS.INVALID_ID {
  track ...
}
```

kevinslin:
@SeriousBug can you create an async to bring this up to the team?

SeriousBug:
Thanks, I unmarked them as deprecated and added some support functions for it. I'm taking a note to write an async once this is merged.

jonathanyeung:
Why can't we just code off a special message? 

```
if message = ERROR_MESSAGE.INVALID_ID {
  track ...
}

File: packages/engine-server/src/drivers/file/storev2.ts [hunk 1]
```diff
@@ -130,6 +131,24 @@ export class FileStorage implements DStore {
...
+            new DendronError({
+              message: `Note ${ent.fname} in ${VaultUtils.getName(
+                ent.vault
+              )} has the same ID as ${duplicate.fname} in ${VaultUtils.getName(
+                duplicate.vault
+              )}`,
+              payload: [NoteUtils.toLogObj(ent), NoteUtils.toLogObj(duplicate)],
+              // Duplicate IDs break some features, but we shouldn't stop initialization completely
+              severity: ERROR_SEVERITY.MINOR,
+              code: StatusCodes.BAD_REQUEST,
```
jonathanyeung, line 147: @kevinslin - on the topic of code, status.  I'm now confused - what HTTP status code is the server sending back to the client when we have multiple errors with different codes? 

500 `INTERNAL_SERVER_ERROR` seems more appropriate than `BAD_REQUEST` for this situation.  A client calling `init()` is not a bad request; the problem is that the note store is in a corrupted state.

jonathanyeung (changes requested):

SeriousBug:
@jonathanyeung 

> Why can't we just code off a special message that's part of the general error payload?

Because the message is used for the error display already. The plugin will display the message in the error, and the message is not a static string but gets formatted based on the actual error (it's formatted to include the names and vaults of the notes that have conflicting IDs).

While we could override that and use the message field as the error code, I don't see the point why we would want that. We already have a way of passing additional data in errors in the status and code fields, and the message means the error message to be displayed to the user.

I don't have strong feelings towards the error codes (400, 500 etc.), but I think we should have a way to say "I'm looking for an error of this type" which is why I used the status field

packages/engine-server/src/drivers/file/storev2.ts
kevinslin: the reason its a client error is because the store/workspace is "client side"/user input. 

if a user was modifying a file and had bad frontmatter and we tried to update the engine, we would presumably respond with a 4xx response. 

that being said, its also fair to say this is a server error because loading existing notes "server side" failed when the user called "init".  

i'm good with classifying this as a 500

kevinslin:
can we add a test for this?

jonathanyeung:
@SeriousBug , @kevinslin The reason I don't want us to add these fields back is because it makes it hard as a caller to understand what I'm supposed to do or check when I get a Dendron Error.  When am I supposed to check the status code? If that's not set, do I check the status string? What if both of these are set? What if none are set? These fields are both redundant for non HTTP API calls because payload has type `any`.

Strictly speaking for HTTP calls though, the Status Code field is important to have, and I think it's important that we set it correctly since this is a universal protocol for HTTP - this will save us work down the line as we have varied clients talking to the server.  There's a big difference from a client's perspective between INTERNAL_SERVER_ERROR and BAD_REQUEST.  The former means, something went wrong and there's probably not much I can do about it beyond retry.  The latter means, _I_ did something wrong and I need to modify my request in order for it to work.

I think error handling on the server vs only within the client is sufficiently different that we shouldn't attempt to use the same class for both (I guess that's the motivation behind `DendronServerError`, but we haven't really been using that properly so far. Conflating properties between the two makes error handling harder.

For this particular situation, I think the best way to pass back the information to the client is to expand `DEngineInitPayload` to include a `warnings` field.  And on further thought, I think 200 S_OK is the appropriate HTTP code for this init() call - It shouldn't be `BAD_REQUEST` from the perspective of this client, because the only thing the client tried was a parameterless init call.   Also, I think we should probably reserve 500 INTERNAL_ERROR for fatal exceptions - which is also not the case here, since all of the notes still initialized fine with the exception of the ones with duped ID's.

SeriousBug:
@jonathanyeung I'm open to alternatives if you have any suggestions, let me describe my goal first.

I want to be able to handle errors depending on what the error is. In particular, I want to recognize the "note ID duplicated" error and handle it by doing analytics tracking.

Whether you call this a "minor severity error" as we do right now, or introduce a special "warnings" type is tangential to what I'm trying to do here. Whether it's an error or a warning, I need to determine if a specific kind of error (or warning) happened and take action.

Whether we should put a 200 or 400 or 500 code on this error is also tangential to this issue. Right now we do absolutely nothing with the code. We could just not set it. Or leave it deprecated. The code is not specific enough to determine what the exact error (or warning) was anyway, so I still need the status.

The caller never *has to* check the status. The only time you would check the status is If you want to take an action when an error happens. In that case you generally need to know what the error was. You might want to handle a missing file error, but propagate the error if it's something unexpected. You need to know what type of error it was.

---

In conclusion, I'm suggesting we keep the status field so we can track specific error types. I also suggest the discussion on the error code be left for later.

jonathanyeung:
@SeriousBug - understand that you had a relatively simple objective with this change, and so I apologize that this has turned into a larger discussion 😅 

I do want to iron out the 'error' part because this change as it stands right now unmarks fields from being deprecated, which can have snowball consequences for us later.

What do you think of my earlier suggestion? 

> pass back the information to the client is to expand `DEngineInitPayload` to include a warnings field

If we do it like this, then we can have all the resolution we need in sending back the info of which note ID's are duplicated, and we also will restrict the change to just modify the `init` API without having to touch any of the error code logic.

SeriousBug:
@jonathanyeung I absolutely wouldn't say "a pain in the butt", I understand the need to fix things up.

For your suggestions about warnings, we have an established way of reporting warnings: errors with the severity set to minor. These are already considered non-fatal errors by the initialization, so the initialization will continue if minor errors are present.

That being said I'm not opposed to introducing a warnings field, but if we do so I think we should include richer type information about the warning. Right now errors are just `any`, which is the right type for them since the API may end up returning any error if something unexpected happens. For warnings I think each endpoint should have a well-defined list of warnings it could have. For example initialization could return `VaultMissingWarning` and `DuplicateNoteIDWarning`. What do you think about that? If that sounds good, I'll roll back the un-deprecation and add the warnings instead.

kevinslin:
So want to pre-face this by saying that error handling, like organizing information, has different convetions and there isn't **one right way** to do it. We just need to find **a way** that works for us.

@jonathanyeung, what i'm hearing from you is the following:

- error handling on the server vs only within the client is sufficiently different that we shouldn't attempt to use the same class for both
- adding these codes back will add complexity on the caller side 
- These fields are redundant for non HTTP API calls because payload has type any

My thoughts:

> error handling on the server vs only within the client is sufficiently different that we shouldn't attempt to use the same class for both

I think the only difference is the addition of a `code` which I think would be useful in client side errors as well - its just that we don't bother with it

>  adding these codes back will add complexity on the caller side 

I don't think it has to be. As long as these fields are optional for client errors, we can do the same check we do today (we don't do much with error body anyways)
```ts
const resp = someMethod()
if (resp.error) {
    throw resp.error
}
// continue with ahppy case
```

Adding `code` means that we **can** do something more useful if we want to 

> - These fields are redundant for non HTTP API calls because payload has type any

Sorry, could you elaborate here? Not sure I follow

In general, I think errors should have the following properties:
- human readable message ( maps to `message`)
- a way of adding arbitrary metadata depending on the error (maps to `payload`)
- a way of identifying what `type` of error is happening (maps to ??? - this is ambiguous today)

I'm a fan of the [stripe error property](https://stripe.com/docs/api/errors) (generally held as the pillar of good api design). 

they have the following signature around errors (i've omitted payment specific properties): 

- main
    - `type` string
        - The type of error returned. One of api_error, card_error, idempotency_error, or invalid_request_error
    - `code` string
        - For some errors that could be handled programmatically, a short string indicating the error code reported.
    `message` string
        - A human-readable message providing more details about the error. For card errors, these messages can be shown to your users.
    - `param` string
        - If the error is parameter-specific, the parameter related to the error. For example, you can use this to display a message near the correct form field.
- additional attributes:
    - `doc_url` string
        - A URL to more information about the error code reported.
    - `source` hash
        - The source object for errors returned on a request involving a source.

What do you think of adopting the above?

***

```ts
type DendronClientError = {
    type: string
    message: string
    payload?: any
    code?: string
    ...
}

type DendronServerError = DendronClientError {
    code: string
}
```

jonathanyeung:
My key point is that all of the things you described are relevant to server-client HTTP calls, but not _all_ types of errors.  Keep in mind that stripe's API design is a REST API, which comes with certain constraints that they must work under:
1. It needs to follow HTTP protocols (i.e., mapping to a universal set of HTTP_STATUS_CODES).
2. The data needs to be serializable
3. Same as point 1, but it needs to work with all types of clients (JS, Python, C, etc.)

We have the same constraints for our express server REST API.  But we decidedly do **not** have any of the above constraints for our TS to TS code.

For example, say we're looking at some plugin-core code (TS) talking to one of our markdown parsing utility functions (also TS). The parsing hits an error and throws a `DendronError`.  If we were to adopt the proposal and have all these fields in the `DendronError`, I, as the developer of either the plugin-core code or the markdown parsing code, have a lot of questions:
- What am I supposed to do with this code field? It's set to HTTP_500 INTERNAL_SERVER_ERROR, but there's no server here?  I'm not sure what that means.
- As the markdown parsing library author, I shouldn't be concerned with creating a polished human readable message to display to an end user.  That's not the responsibility of the library.  All I'm concerned about is that the program invoking my function can understand what's going on.  That can be an enum (not technically human readable, since it usually maps to an int).

So I'm arguing that DendronError shouldn't contain these fields, because it's the base class of errors.  The proper design is to use composition / inheritance to create more complex error types that work well for our different scenarios, be it express-server, markdown-parsing lib, or whatever.  

By adding all of these fields into the base class as optional, we're losing the benefits of having a type system and conflating functionality.

kevinslin:
after some offline discussion with @jonathanyeung , we decided to go for a hybrid approach, outlined in https://docs.google.com/document/d/1F6yyHGMjf-behwhMAWyfq_I2XNu5WJCqFZ4b1IP-t1w/edit

tldr: all errors have a common interface (code/message/payload) but we will rely on subclassing to parameterize payload/code

apologies @SeriousBug for the churn

SeriousBug:
@kevinslin That's alright, so should I implement the system proposed in the document in this PR?

kevinslin:
> @kevinslin That's alright, so should I implement the system proposed in the document in this PR?

yep, to be safe, you can wait until after monday to see if anyone has amendments

SeriousBug:
I'm closing this PR and replacing it with #2879, because starting from scratch was easier.

Issue #1907, opened 2021-12-16
GoToSiblingCommand: You found a bug! We didn't think this could happen but you proved us wrong


viddo:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When navigating using _"Dendron: Go Next Sibling"_ command I get the following popup:
![Screenshot 2021-12-16 at 07 26 40](https://user-images.githubusercontent.com/978461/146319655-4c0577e4-3eaa-43be-b038-2c30aafda28f.jpg)

The string appears to originate from: https://github.com/dendronhq/dendron/blob/a7227fd4d8991e44729989c821a22560dcb8348b/packages/plugin-core/src/logger.ts#L25

## To Reproduce

1. Create four notes, e.g. `test.a`, `test.b`, `test.c`, `test.d`
2. Rename a note, e.g. `test.c` -> `test.cx` (name doesn't really matter)
3. Try command _Dendron: Go Prev Sibling_ or _Dendron: Go Next Sibling_

## Expected behavior

Going to prev/next sibling per new filename

## Desktop (please complete the following information)

- OS: MacOS Catalina 10.15.7
- ~~Browser: [e.g. chrome, safari]~~ _not relevant_
- Dendron Version: 0.73.1
- VSCodium Version: 1.63.1 + "about VSCodium" yields:
```
Version: 1.63.1
Commit: fe719cd3e5825bf14e14182fddeb88ee8daf044f
Date: 2021-12-15T00:21:21.398Z
Electron: 13.5.2
Chromium: 91.0.4472.164
Node.js: 14.16.0
V8: 9.1.269.39-electron.0
OS: Darwin x64 19.6.0
```

## Additional context

When changing to `test.a`, `test.b` or `test.d` and navigating using aforementioned it skips the renamed file (i.e. next from `test.b` jumps to `test.d` and vice versa on "prev" command). 

As a workaround one can run the command _Dendron: Reload index_ (or _Developer: Reload window_) to resolve the issue.

[...]

PR #2805, opened 2022-04-22
fix: second pass of treeview v1 sync issue

hikchoi:
# fix: second pass of treeview v1 sync issue
- This PR does a second pass of fixing engine / treeview state mismatch issues.
- Case: renaming a note into an already existing stub note
  - Doing this used to leave the + mark in the tree view because the parent / children were not properly updated
  - This used to leave `stub: true` in the renamed note's frontmatter, which caused a lot of problems
    - The resulted renamed note would appear in lookup twice. one would be a stub and one wouldn't.
      - depending on the accepted item, there was a possibility of losing the entire body of the note.
    - Now all notes will lose their `stub: true` property once they are persisted in the store.
      - this, however, doesn't _disallow_ the user from adding the property themselves.
        - a separate PR will be submitted for a doctor command that scrubs existing persisted notes with `stub: true`
        - a separate PR will be submitted for a code action provider that warns users if they have `stub: true` in the frontmatter of their active note.
- Case: renaming a note that has a stub parent and any children
  - This used to simply not work. Tree view would crash since the engine events emitted weren't right, and we ended up with the parent pointing to the renamed note, which is already registered in the tree view.
  - now explicitly handles updating of parent when a rename requires a stub replacement during `delete`.
- Since we are chaining multiple updates to possibly the same note, we don't clean up duplicate update events at the end of `rename`. We preserve them since they aren't truly duplicates. We need each of them to correctly end up with the desired engine state.
  - This change broke a lot of existing tests (particularly the ones that count emitted events). They are all fixed here.
 
### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -1091,6 +1091,7 @@ export class FileStorage implements DStore {
     let noDelete = false;
     if (maybeNote?.stub || opts?.updateExisting) {
       note = { ...maybeNote, ...note };
+      note = _.omit(note, "stub");
       noDelete = true;
     } else {
       changedEntries = await this._writeNewNote({
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -239,7 +239,7 @@ export class FileStorage implements DStore {
         out.push({ note: noteToDelete, status: "update", prevNote });
       } else {
         // the delete operation originated from rename.
-        // since _noteToDelete) in this case is renamed and
+        // since _noteToDelete_ in this case is renamed and
         // moved to another location, simply adding stub: true
         // will not work.
         // we need a fresh stub note that will fill in the old note's place.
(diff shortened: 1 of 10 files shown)


hikchoi:
I'll fill up the PR checklist during the weekend. I need to take a break after banging my head fixing this for the past two days 😅 

Submitting pre-emptively so that it could be reviewed.

hikchoi:
more things to fix. turning to draft.

SeriousBug:
Wow great work, looking forward to this fix.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -902,10 +923,6 @@ export class FileStorage implements DStore {
     notesChangedEntries = changedFromDelete
       .concat(changeFromWrite)
       .concat(notesChangedEntries);
-    // remove duplicate updates
-    notesChangedEntries = _.uniqBy(notesChangedEntries, (ent) => {
```
jonathanyeung, line 906: Was this causing an issue?

jonathanyeung (approved):
Nice tests!

packages/engine-server/src/drivers/file/storev2.ts
hikchoi: yes.

in the past we just updated the existing stub notes so we didn't need multiple updates to the parent / children. but now we are explicitly deleting and creating the replacing stub in this case, we need to emit multiple events.

the `uniqBy` predicate isn't really checking if the change entries are truly duplicate. there are some properties of the entries that are not duplicates and we need to emit them in order to guarantee that plugin state is the same with the engine.

There are also some cases where we emitted delete _after_ create, which caused the tree view to omit newly created notes (because the plugin saw them as deleted). These would be corrected when we reload since the plugin catches up with whatever is persisted in the filesystem during init.

I haven't exhaustively checked for everything, but I suspect that a lot of the tree view mismatch is caused because we are not emitting the events in the right order. I'll fix them as I find them moving forward.

SeriousBug (approved):
Looks good to me, did some manual testing as well and the tree view updates after a rename.

hikchoi:
related: https://github.com/dendronhq/dendron/issues/1907

This fix resolved the issue reported in #1907 
This was happening because after the rename of a note that has stub parent, the extension was not receiving proper events emitted from the engine, which caused a drift in state.

PR #2806, opened 2022-04-22
enhance(view): graph themes

Harshita-mindfire:
This PR aims to add a new graph configuration `Graph Theme` to change the graph styling.
1. Three graph themes are introduced: `Block` and `Classic` and `Monokai`(open to suggestions for a better name). `Classic` graph theme has the present graph styling we offer. 
2. Adds a new field `graphTheme` to the MetadataService to store the graph theme selected by the user. Next time, the graph view is loaded by this graph theme.
3. If custom styling is present, adds `Custom` to the Graph Theme list.
A button is added to create new styling/modify existing CSS in the dropdown
Docs: https://github.com/dendronhq/dendron-site/pull/485

TODO: update telemetry in airtable.

- NOTE: the links don't work. you'll need to go into the wiki and use the lookup to find the note until we fix some issues in the markdown export

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
NA

## Docs

### Extended


SeriousBug:
Looks good! If I had to suggest names:
- "default" -> "classic", since that's the classic version of what we offer
-  "classic" -> "monokai", and I'd tweak the colors to match the colors used by the monokai theme. See [here](https://github.com/Brackets-Themes/Monokai) for example images and [here](https://gist.github.com/r-malon/8fc669332215c8028697a0bbfbfbb32a) for colors.

I'd also suggest removing the "block" theme. Maybe it's just me but the graphs it produces reminds me of swastikas. If we're trying to show off that you can display the lines differently, maybe we could go for a curved or wavy look for only certain types of links?

Harshita-mindfire:
> Looks good! If I had to suggest names:
> 
> * "default" -> "classic", since that's the classic version of what we offer
> * "classic" -> "monokai", and I'd tweak the colors to match the colors used by the monokai theme. See [here](https://github.com/Brackets-Themes/Monokai) for example images and [here](https://gist.github.com/r-malon/8fc669332215c8028697a0bbfbfbb32a) for colors.

Updated! Adjusted colors according to dark and light themes.
> I'd also suggest removing the "block" theme. Maybe it's just me but the graphs it produces reminds me of swastikas. If we're trying to show off that you can display the lines differently, maybe we could go for a curved or wavy look for only certain types of links?

Will add this to `meet.daily` for reviews from the team.
> I would also suggest displaying user readable names instead of all caps, so "Classic" rather than "CLASSIC".

Done!

packages/plugin-core/src/abTests.ts
```diff
@@ -83,6 +84,28 @@ export const MEETING_NOTE_FEATURE_SHOWCASE_TEST = new ABTest(
   ]
 );
 
+export enum GraphThemeTestGroups {
```
jonathanyeung, line 87: Let's go ahead and add all styles as groups.

packages/engine-server/src/metadata/service.ts
```diff
@@ -65,6 +65,10 @@ type Metadata = Partial<{
    *
    */
   workspaceActivationContext: WorkspaceActivationContext;
+  /**
+   * Default Theme for Note Graph View
+   */
+  defaultGraphTheme?: GraphThemeEnum;
```
jonathanyeung, line 71: nit: remove the word 'default' here - it's the user's _current_ theme that we're setting here.

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
```diff
@@ -141,6 +153,35 @@ export class NoteGraphPanelFactory {
...
             break;
           }
+          case GraphViewMessageEnum.onGraphThemeChange: {
+            this.defaultGraphTheme = msg.data.graphTheme;
+            break;
+          }
+          case GraphViewMessageEnum.onRequestDefaultGraphTheme: {
+            let graphTheme = MetadataService.instance().getDefaultGraphTheme();
+            if (!graphTheme) {
+              const ABUserGroup = GRAPH_THEME_TEST.getUserGroup(
```
jonathanyeung, line 163: I think it may be a bit too disruptive to put the A/B Test here - I'm not sure we should change the theme for existing users.  How about we tweak the experiment a little bit? 

When a _new_ user is installing dendron, we have the A/B group decide which default style they should get.  We save that in `MetadataService`. 

For existing users, we can show them a toast message to the effect of "hey - we have different styles for the graph view now, check it out!".  We can maybe put this in a separate A/B experiment group (or maybe worth showing to everyone?).  For the toast, I have some code (wip) that can do this.  See https://github.com/dendronhq/dendron/pull/2832/files#diff-78b40b72529a46738cf6b47c9cc89fb508a2af3ee230f27c250eb8825c5f02eb

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
```diff
@@ -141,6 +153,35 @@ export class NoteGraphPanelFactory {
             }
             break;
           }
+          case GraphViewMessageEnum.onGraphThemeChange: {
+            this.defaultGraphTheme = msg.data.graphTheme;
```
jonathanyeung, line 157: let's add a telemetry marker on this - whenever the user changes their theme + what theme they changed to.

jonathanyeung (changes requested):
Cool!

A few things
- maybe I'm doing something wrong, but locally right now I'm not seeing the graph style change as I flip between the theme options? 
- as a V2, can we make it more obvious to a user how they can create their own theme?  For example, add some button to the 'themes' list that's something like 'Create Your Own', which will then run `Dendron: Configure Graph Styles (css)` command.  On a separate note, we should update the template that gets generated when we run that command.  Right now it says `Full Dendron-specific styling documentation can be found here: [LINK HERE]`.  Maybe as part of the docs, we can have the .css snippets for 'monokai', 'blocks' that a user can copy as a starting point, and then modify individual fields - makes it easier for the user to get started.

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
Harshita-mindfire: Updated A/B test only for new users. 
Todo: Toast for existing users

Harshita-mindfire: Updated A/B test only for new users. 
Todo: Toast for existing users

SeriousBug:
> If custom style is present, it always picks up that style.

@Harshita-mindfire If custom CSS disables the themes, then we should make that clearer in the UI I think. Would it be possible to gray-out the options and add a text like "Custom CSS overrides theme selection"?

kevinslin:
> > * maybe I'm doing something wrong, but locally right now I'm not seeing the graph style change as I flip between the theme options?
> 
> @jonathanyeung Do you have custom styling? If custom style is present, it always picks up that style.

this bit me too. its not obvious and goes against the `zero cost features` principle. can we do the following?
- show custom theme as an option in the dropdown. if the user has a custom theme, apply this, if not, open the wiki link on docs on how to create one
- if the user has a custom theme and switches to a default theme, then we should see the new theme load

kevinslin (changes requested):
this looks good. left some comments about integrating the custom theme in the dropdown.

also, can we include this in the tip of the day? (eg. you can customize your graph with themes)

Harshita-mindfire:
Integrated Custom theme in Graph Theme dropdown: 
- When no custom styling is present
image
- When custom styling is present
image

Harshita-mindfire:
> * as a V2, can we make it more obvious to a user how they can create their own theme?  For example, add some button to the 'themes' list that's something like 'Create Your Own', which will then run `Dendron: Configure Graph Styles (css)` command.  On a separate note, we should update the template that gets generated when we run that command.  Right now it says `Full Dendron-specific styling documentation can be found here: [LINK HERE]`.  Maybe as part of the docs, we can have the .css snippets for 'monokai', 'blocks' that a user can copy as a starting point, and then modify individual fields - makes it easier for the user to get started.

added in this PR

packages/dendron-plugin-views/src/components/GraphFilterView.tsx
```diff
@@ -233,4 +273,59 @@ const FilterViewSection = ({
...
+          </Radio>
+        ))}
+      </Space>
+    </Radio.Group>
+  );
+};
+
+/**
+ * vscode message to update graphTheme selected by User.
+ * When the graph panel is disposed, this value is written back to Metadata Service.
```
jonathanyeung, line 313: nit: "When the graph panel is disposed" - is this accurate? Do we not write back every time the radio button is changed?

packages/plugin-core/src/_extension.ts
```diff
@@ -580,6 +583,25 @@ export async function _activate(
...
+          break;
+        }
+        case GraphThemeTestGroups.block: {
+          graphTheme = GraphThemeEnum.Block;
+          break;
+        }
+        default:
+          graphTheme = GraphThemeEnum.Classic;
+      }
+      MetadataService.instance().setGraphTheme(graphTheme);
```
jonathanyeung, line 604: Hmm - for analytics purposes, it might be easier for us to analyze if we set a marker right here as well.  Maybe we can reuse the same event of changing graph themes, but add a property to the effect of "SetDuringInstall: true" or something like that. 

Otherwise, we'll need to remember that only folks who installed on v92 or greater will have had this altered.

jonathanyeung (changes requested):
looks good! just one more minor telemetry thing

packages/dendron-plugin-views/src/components/GraphFilterView.tsx
Harshita-mindfire: Each time the radio button is changed, we store the value in [this.defaultGraphTheme](https://github.com/dendronhq/dendron/pull/2806/files#diff-c05ef452a749e19d2b71f255ebbd218e4bca7d34abbbfc6aa587e50e266f973cR161) variable, and only write back to the meta file when the graph is disposed with the latest value.

packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
Harshita-mindfire: Added

packages/plugin-core/src/showcase/GraphThemeTip.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+  get showcaseEntry(): ShowcaseEntry {
+    return ShowcaseEntry.GraphTheme;
+  }
+
+  getDisplayMessage(displayLocation: DisplayLocation): string {
+    switch (displayLocation) {
+      case DisplayLocation.InformationMessage:
+        return `New themes for Graph View! Change the appearance...`;
```
jonathanyeung, line 26: Oops - sorry miscommunication here.  I meant for the string to be

"New themes for Graph View! Change the appearance of the note graph by clicking the config button on the top left corner of the graph view and selecting one of the built-in styles. You can even customize the appearance to your liking with css"

PR #2832, opened 2022-04-27
feat(views): Dendron Side Panel

jonathanyeung:
## feat(views): Dendron Side Panel

Adds a dedicated Dendron Side Panel.

- This moves all of our views into this side panel.  A user can still choose to move individual panels to wherever they like though.
- Adds a 'tip of the day' view

Code is ready for review, but leaving in draft mode for the TODO's below:

TODO:
- Before we push this, we should have a set of tips to show in 'tip of the day'.  Otherwise, it's gonna be a meeting notes tip everyday 😄 
- Should we classify this as a breaking change? Since it alters user's UI.
 
---

# Pull Request Checklist


jonathanyeung:
What it looks like right now: 

image

packages/plugin-core/src/features/TipOfTheDayWebview.ts
```diff
@@ -0,0 +1,182 @@
...
+  FeatureShowcaseUserResponse,
+  IFeatureShowcaseMessage,
+} from "../showcase/IFeatureShowcaseMessage";
+import { AnalyticsUtils } from "../utils/analytics";
+
+/**
+ * Side Panel webview that shows the tip of the day
+ * TODO: Add functionality
+ *  - let user rotate tips
+ *  - save which tip they're on. (right now it just shows the first one.)
```
kevinslin, line 15: for mvp, can it show a random tip vs first one?

packages/plugin-core/src/showcase/FeatureShowcaseToaster.ts
```diff
@@ -0,0 +1,106 @@
...
+import {
+  DisplayLocation,
+  FeatureShowcaseUserResponse,
+  IFeatureShowcaseMessage,
+} from "./IFeatureShowcaseMessage";
+
+/**
+ * Class to showcase certain features of Dendron as a toast message
+ */
+export class FeatureShowcaseToaster {
```
kevinslin, line 16: does it make sense to have a toast AND a tip of the day? should we use the toast when the tip isn't visible and vice versa?

packages/plugin-core/src/showcase/MeetingNotesTip.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+  }
+  get showcaseEntry(): ShowcaseEntry {
+    return ShowcaseEntry.TryMeetingNotes;
+  }
+
+  getDisplayMessage(displayLocation: DisplayLocation): string {
+    switch (displayLocation) {
+      case DisplayLocation.InformationMessage:
+        return `Dendron now has meeting notes. Try it out!`;
```
kevinslin, line 40: the `Dendron now has...` prefix shouldn't be used if its a tip. a tip generally lets people know about existing functionality. can e change the text to remove indicator of recency?

packages/plugin-core/src/showcase/MeetingNotesTip.ts
```diff
@@ -0,0 +1,59 @@
...
+        assertUnreachable(displayLocation);
+    }
+  }
+
+  get onConfirm(): string {
+    return "dendron.createMeetingNote";
+  }
+
+  get confirmText(): string {
+    return "Create a meeting note";
```
kevinslin, line 53: a second phase but we should keep track of what tips users have clicked on to avoid showing the same tips

kevinslin (commented):
- can we do an async and get the team to weight in on this? eg. everyone come up with at least 2 lines of features to add here.

- we should do some `learn & adapt` on this (not a blocker but should be a followup). the apple `tips` app - how does it present tips? if we want to be more developer focused, how do popular developer tools (eg. vscode, linear, intellij) manage tips?

packages/plugin-core/src/features/TipOfTheDayWebview.ts
jonathanyeung: Yeah - was planning on adding that logic in this revision.

packages/plugin-core/src/showcase/FeatureShowcaseToaster.ts
jonathanyeung: I think for now yes - I'd prefer to show in both locations until we collect more data on how effective the display locations are.

packages/plugin-core/src/showcase/MeetingNotesTip.ts
jonathanyeung: The Toaster already has this logic.  Tip of the Day doesn't - in other implementations (IntelliJ for example), you can cycle through all the tips and repeat.  It's just that they have a lot of tips, so it'll take many days before you get one you've seen before.

packages/plugin-core/src/showcase/MeetingNotesTip.ts
jonathanyeung: This is carry over from last week's toast experiment.  Given that it was quite effective, I'm inclined to keep the toast message the same, at least for a couple of more weeks.  That toast will not show more than once for users, unless they click "Later".

kevinslin:
can you rebase off master?

packages/plugin-core/src/features/TipOfTheDayWebview.ts
```diff
@@ -0,0 +1,198 @@
...
+      },
+      undefined,
+      undefined
+    );
+
+    this.showNextTip();
+  }
+
+  private showNextTip(): boolean {
+    this._curTipIndex = (this._curTipIndex + 1) % this._tips.length;
```
Harshita-mindfire, line 98: Tip of the day resulted in an error for me: `An error occurred while loading view: dendron.tip-of-the-day`. 
It might be because we get a floating-point number from this and we a undefined tip is returned from [here](https://github.com/dendronhq/dendron/pull/2832/files#diff-6491d89c0e1ed956f06220885819e59144ad233fc28be7afa8702362ab50795bR26)

packages/plugin-core/src/features/TipOfTheDayWebview.ts
```diff
@@ -0,0 +1,198 @@
...
+export default class TipOfTheDayWebview implements vscode.WebviewViewProvider {
+  private _webview: vscode.WebviewView | undefined;
+  private _tips: IFeatureShowcaseMessage[];
+  private _curTipIndex;
+
+  private BUTTON_CLICKED_MSG = "buttonClicked";
+  private TIP_SHOWN_MSG = "loaded";
+
+  private get _currentTip(): IFeatureShowcaseMessage {
+    return this._tips[this._curTipIndex];
```
Harshita-mindfire, line 26: Tip of the day resulted in an error for me: `An error occurred while loading view: dendron.tip-of-the-day`. 
An undefined tip is returned from here with this._curTipIndex being floating-point number.

jonathanyeung: 🤦🏼 - good catch!

kevinslin:
this looks good. can we follow up on this task next week? some top level concerns:
- adding tests
- giving user the ability to cycle through tips
- add more tips
- add a getting started section

kevinslin (approved):

Issue #2842, opened 2022-04-27
Upgrade notification happens every time, and sometimes says upgraded from v0.0.0


AshSimmonds:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

As titled.

## To Reproduce

Unknown - apart from just opening VSCode.

Note: I am using 2-3 different machines with different settings a day, and it's something that's just started after I stopped using Dendron for about a month and came back as of v0.91.0 - and I also use both normal and have just started using Insider's Build.

## Expected behavior

Only display the notification toast when it actually upgrades.

## Screenshots

n/a

## Desktop (please complete the following information)

- Windows 10/11
- Dendron 0.91.0 --> 0.92.0
  - VS Code: 1.66.2
  - Insider's Build: 1.67.0-insider

## Dendron Log file

```json
{"level":30,"time":1651098790176,"pid":13924,"hostname":"Flippy-Turd","name":"dendron","ctx":"Logger:configure","msg":"exit","logLevel":"info"}
{"level":30,"time":1651098790266,"pid":13924,"hostname":"Flippy-Turd","name":"dendron","ctx":"_activate","stage":"prod","isDebug":false,"logLevel":"info","logPath":"c:\\Users\\ashsi\\AppData\\Roaming\\Code - Insiders\\logs\\20220428T080307\\exthost1\\dendron.dendron","extensionPath":"c:\\Users\\ashsi\\.vscode-insiders\\extensions\\dendron.dendron-0.92.0","extensionUri":"c:\\Users\\ashsi\\.vscode-insiders\\extensions\\dendron.dendron-0.92.0"}
{"level":30,"time":1651098790292,"pid":13924,"hostname":"Flippy-Turd","name":"SegmentClient","msg":"No cache path for Segment specified. Failed event uploads will not be retried."}
{"level":30,"time":1651098790293,"pid":13924,"hostname":"Flippy-Turd","name":"SegmentClient","msg":"user telemetry setting: enabled by config"}
[...]

PR #2843, opened 2022-04-28
fix: correctly handle previous global and workspace version

hikchoi:
# fix: correctly handle previous global and workspace version
This PR:
- fixes a regression introduced in the initialization logic where previous global and workspace version was not properly handled
  - With the recent clean-up of the initialization logic the following changes were omitted
    - previousGlobalVersion from state service was not getting backfilled to the corresponding metadata, and was not read from metadata for subsequent use
    - previousWorkspaceVersion from state service was not getting backfilled to the corresponding workspace metadata, and was not read from workspace metadata for subsequent use
  - This resulted in an `InstallStatus.UPGRADED` for every initialization

NOTE: to manually test that this fixes the issue, you may need to reload once after dendron is initialized since that we need the first pass of initialization for the metadata to sync up

# Pull Request Checklist

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -490,6 +490,7 @@ export async function _activate(
   const { workspaceFile, workspaceFolders } = vscode.workspace;
   const logLevel = process.env["LOG_LEVEL"];
   const { extensionPath, extensionUri, logUri } = context;
+  const metadataService = MetadataService.instance();
   const stateService = new StateService({
     globalState: context.globalState,
     workspaceState: context.workspaceState,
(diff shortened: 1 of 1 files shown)


hikchoi:
related to #2842

packages/plugin-core/src/_extension.ts
```diff
@@ -553,8 +554,22 @@ export async function _activate(
...
+    // this used to be handled by StateService (if undefined, 0.0.0).
+    // this will be reassigned if we are in a dendron workspace.
+    let previousWorkspaceVersion = "0.0.0";
+
+    const previousGlobalVersionFromState = stateService.getGlobalVersion();
+
+    // temporarily here to backfill globalState into metadata
+    // this should be removed once we determine that
+    // we have sufficiently backfilled out user.
+    if (previousGlobalVersionFromState) {
```
kevinslin, line 567: won't users that upgraded recently also have a version that is already written? should this logic change so that if no current global version is found in `metadataService`, then try to look for it in state service?

hikchoi: in `showWelcomeOrWhatsNew`, we call `metadataService.setGlobalVersion(version)` with the current extension version.

Everyone encountering this bug would have already have the global version written in metadata. It's just that we didn't read that during initialization thus `IntallStatus` was set to `UPGRADED`.

jonathanyeung (approved):

PR #2845, opened 2022-04-28
fix/upgrade prompt 2

kevinslin:
- Revert "fix: correctly handle previous global and workspace version (#2843)"
- fix prompt


packages/plugin-core/src/_extension.ts
```diff
@@ -620,8 +619,6 @@ export async function _activate(
       // --- Get Version State
       const wsRoot = wsImpl.wsRoot;
       const wsService = new WorkspaceService({ wsRoot });
-      const wsMeta = wsService.getMeta();
-      previousWorkspaceVersion = wsMeta.version;
```
hikchoi, line 624: This is needed. we are tracking `workspaceState` in `wsMeta` since the StateService changes.

PR #2852, opened 2022-04-29
feat: add goto command

kevinslin:
feat: add goto command

This adds a new command `Go to` which supports navigating the `uri` parameter of [[Proxy Notes|dendron://dendron.docs/rfc.38-links-to-non-note-files#proxy-notes]].

See docs in https://github.com/dendronhq/dendron-site/pull/489

In addition to implementing this feature, the following sections were cleaned up: 
- refactor codeActionProvider (to avoid introducing a new circular dependency)
- fix some tests in GotoNote
- add presents for GotoNote
- extract some utilities as common utilities

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

NA

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/GoToNoteInterface.ts
+++ b/packages/plugin-core/src/commands/GoToNoteInterface.ts
@@ -4,6 +4,7 @@ import { Position, ViewColumn } from "vscode";
 export enum TargetKind {
   NOTE = "note",
   NON_NOTE = "nonNote",
+  LINK = "link",
 }
 
 export type GoToNoteCommandOpts = {
@@ -31,4 +32,5 @@ export type GoToNoteCommandOutput =
   | { kind: TargetKind.NOTE; note: NoteProps; pos?: Position; source?: string }
   // When opening a non-note file
   | { kind: TargetKind.NON_NOTE; fullPath: string; type: GotoFileType }
+  | { kind: TargetKind.LINK; fullPath: string }
   | undefined;
--- a/packages/plugin-core/src/commands/GoToNoteInterface.ts
+++ b/packages/plugin-core/src/commands/GoToNoteInterface.ts
@@ -32,5 +32,6 @@ export type GoToNoteCommandOutput =
   | { kind: TargetKind.NOTE; note: NoteProps; pos?: Position; source?: string }
   // When opening a non-note file
   | { kind: TargetKind.NON_NOTE; fullPath: string; type: GotoFileType }
-  | { kind: TargetKind.LINK; fullPath: string }
+  // When opening a link to a non txt-file like resource (eg. pdf, website, etc)
+  | { kind: TargetKind.LINK; fullPath: string; fromProxy: boolean }
   | undefined;
--- a/packages/plugin-core/src/commands/Goto.ts
+++ b/packages/plugin-core/src/commands/Goto.ts
@@ -23,6 +23,10 @@ type CommandOutput = RespV3<GoToNoteCommandOutput>;
 
 const GOTO_KEY = "uri";
 
+/**
+ * Go to the current link under cursor. This command will exhibit different behavior depending on the type of the link.
+ * See [[dendron.ref.commands.goto]] for more details
+ */
 export class GotoCommand extends BasicCommand<CommandOpts, CommandOutput> {
   key = DENDRON_COMMANDS.GOTO.key;
 
--- a/packages/plugin-core/src/commands/Goto.ts
+++ b/packages/plugin-core/src/commands/Goto.ts
@@ -34,6 +34,21 @@ export class GotoCommand extends BasicCommand<CommandOpts, CommandOutput> {
     super();
   }
 
+  addAnalyticsPayload?(_opts?: CommandOpts, out?: CommandOutput) {
+    if (!out?.data) {
+      return {};
+    }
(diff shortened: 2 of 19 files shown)


packages/plugin-core/src/commands/Goto.ts
```diff
@@ -0,0 +1,94 @@
...
+
+type CommandOpts = {};
+
+type CommandInput = {};
+
+type CommandOutput = RespV3<GoToNoteCommandOutput>;
+
+const GOTO_KEY = "uri";
+
+export class GotoCommand extends BasicCommand<CommandOpts, CommandOutput> {
```
jonathanyeung, line 26: Can you add a small class header to indicate this command's relationship to `GoToNoteCommand`?

packages/plugin-core/src/utils/editor.ts
```diff
@@ -236,3 +246,45 @@ export async function isBrokenWikilink(): Promise<boolean> {
...
+  kind?: TargetKind;
+  /**
+   * What {@link vscode.ViewColumn} to open note in
+   */
+  column?: ViewColumn;
+  /** added for contextual UI analytics. */
+  source?: string;
+};
+
+export async function getLinkFromSelectionWithWorkspace() {
```
jonathanyeung, line 264: We discussed this offline, but for others - can you add a function header to clarify what 'WithWorkspace' means here?

jonathanyeung (changes requested):
minor nits on docstrings, otherwise lgtm. 

Left a small comment on the corresponding docs PR.

jonathanyeung (approved):

packages/plugin-core/src/constants.ts
```diff
@@ -246,6 +246,15 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     // no prefix, we don't want to show this command
     title: `${CMD_PREFIX} Contribute`,
   },
+  GOTO: {
+    key: "dendron.goto",
+    // no prefix, we don't want to show this command
```
SeriousBug, line 251: Just a nitpick, but it looks like this comment has been accidentally getting copy-pasted around a bunch, these commands all do have a prefix and we do want to show them

packages/plugin-core/src/commands/GoToNoteInterface.ts
```diff
@@ -31,4 +32,5 @@ export type GoToNoteCommandOutput =
   | { kind: TargetKind.NOTE; note: NoteProps; pos?: Position; source?: string }
   // When opening a non-note file
   | { kind: TargetKind.NON_NOTE; fullPath: string; type: GotoFileType }
+  | { kind: TargetKind.LINK; fullPath: string }
```
SeriousBug, line 35: Could you add a comment for in what case this kind is supposed to be used, similar to the existing comments for the other kinds?

packages/plugin-core/src/commands/Goto.ts
```diff
@@ -0,0 +1,98 @@
...
+      return {
+        error: new DendronError({ message: "selection is not a note" }),
+      };
+    }
+
+    // TODO: for now, get first note, in the future, show prompt
+    const note = notes[0];
+
+    // if note doesn't have url, run goto note command
+    if (_.isUndefined(note.custom[GOTO_KEY])) {
```
SeriousBug, line 76: I'd suggest adding telemetry to check how often this happens vs. how often we get to go through a proxy note

packages/plugin-core/src/utils/editor.ts
```diff
@@ -236,3 +246,48 @@ export async function isBrokenWikilink(): Promise<boolean> {
...
+
+/**
+ * NOTE: this method requires that `ExtensionProvider` be available and can provide a workspace
+ */
+export async function getLinkFromSelectionWithWorkspace() {
+  const { selection, editor } = VSCodeUtils.getSelection();
+  if (
+    _.isEmpty(selection) ||
+    _.isUndefined(selection) ||
+    _.isUndefined(selection.start) ||
```
SeriousBug, line 272: nitpick: you could simplify the above 3 conditions as `selection?.start !== undefined`

packages/plugin-core/src/utils/editor.ts
```diff
@@ -236,3 +246,48 @@ export async function isBrokenWikilink(): Promise<boolean> {
...
+  const reference = await getReferenceAtPosition({
+    document: editor.document,
+    position: selection.start,
+    opts: { allowInCodeBlocks: true },
+    wsRoot,
+    vaults,
+  });
+  if (!reference) return;
+  return {
+    alias: reference?.label,
```
SeriousBug, line 288: nitpick: the `?`s are unnecessary I think?

SeriousBug (commented):
Looks good, other than some nitpicks and one point about telemetry.

packages/plugin-core/src/utils/editor.ts
kevinslin: this is needed so that typescript compiler doesn't complain about selection being undefined in following blocks. added comments

SeriousBug (approved):

PR #2853, opened 2022-04-29
chore: mark `dendron.dendron-markdown-links` as incompatible

SeriousBug:
The extension `dendron.dendron-markdown-links` is unnecessary since the note graph is built into Dendron now, but it also breaks the built-in note graph. This change will warn users if they have the extension installed.

### DIFF
--- a/packages/plugin-core/src/constants.ts
+++ b/packages/plugin-core/src/constants.ts
@@ -980,6 +980,8 @@ export const INCOMPATIBLE_EXTENSIONS = [
   "kortina.vscode-markdown-notes",
   "maxedmands.vscode-zettel-markdown-notes",
   "tchayen.markdown-links",
+  // Note graph is now built into Dendron, and having this extension enabled breaks it.
+  "dendron.dendron-markdown-links",
 ];
 
 export type osType = "Linux" | "Darwin" | "Windows_NT";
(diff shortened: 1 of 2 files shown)


jonathanyeung (approved):

jonathanyeung:
Looks like there's a failing test: 

```log
@dendronhq/plugin-core:   2) FIND_INCOMPATIBLE_EXTENSIONS
@dendronhq/plugin-core: 	at process.topLevelDomainCallback (node:domain:152:15)
@dendronhq/plugin-core:        GIVEN findIncompatibleExtensions selected
@dendronhq/plugin-core: 	at process.callbackTrampoline (node:internal/async_hooks:128:24)
@dendronhq/plugin-core:          THEN List all as not installed if found none:
@dendronhq/plugin-core:       AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
@dendronhq/plugin-core: 11 !== 10
```

jonathanyeung (changes requested):
looks like 1 failing test that expects the exact # of incompatible extensions.  Maybe worth removing that test? (or that check)

hikchoi (changes requested):
```expect(out.installStatus.length).toEqual(10);``` from both tests could be removed.

Also, the doctor command links to a document with a full list and detail about incompatible extensions: `[[Incompatible Extensions|dendron://dendron.dendron-site/dendron.guides.troubleshooting.incompatible-extensions]]`

New additions should be added.

SeriousBug:
Thanks for catching the error in the tests!

@hikchoi I added this extension to that list already, but yes we should add any others we know cause bugs.

hikchoi:
> Thanks for catching the error in the tests!
> 
> @hikchoi I added this extension to that list already, but yes we should add any others we know cause bugs.

ah right. I didn't notice the update to the docs because a docs PR wasn't linked to this 😅

kevinslin (approved):

Issue #2739, opened 2022-04-12
Setting to Restore Earlier Display of Journal Notes in Tree View


micharris42:
## Please select if your request is either something new or an enhancement ##

- [X] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [ ] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup - Dendron's Lookup Command
- [X] Views - Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [ ] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

As of [Dendron 0.90](https://wiki.dendron.so/notes/a3lvzn5hlmnatci49y2ltan/), the presentation of journal notes in the Tree View as changed. Prior to this version, only the "day" portion of the note's title was displayed in the Tree View. For a note with the title frontmatter field of "2022-04-12", this would render in the Tree View as:

```
>2022
  >04
    12
```

As of Dendron 0.90, the full title is displayed:

```
>2022
  >04
    2022-04-12
```

While displaying the full title in the Tree View makes sense for most notes, for journal notes specifically, the information is superfluous, as the year and month is already presented in the above hierarchical levels.

## Describe the solution you'd like ##

That there be an option to change the way journal notes are displayed in the Tree View to only show the day portion of the note's title, and to hide the year and month.

## Describe alternatives you've considered ##

Do not think there is a real alternative here. There is the option for the very dedicated user to go through all of their journal notes and manually change the title frontmatter field, but for anyone with a journal spanning multiple years worth of entries, they would have to be quite the dedicated user indeed.

micharris42:
To be clear, I am not against the note titles being used in the Tree View instead of the file names. I am in agreement that for most cases, the note title is probably more descriptive. 

My issue is—as stated above—that in the case of journal notes specifically, the information is superfluous. The year and month are now displayed twice, first as part of the hierarchy and again as part of the note title.

Looking at your suggestions above, @aleksey-rowan, it occurs to me that option 3:

"an **additional metadata item** in notes (something like `label`) that would control what is displayed in the tree view; if `label` is not set, it could default to either file name or title"

[...]

PR #2858, opened 2022-04-29
feat: allow customization of tree view label / sorting to preserve old tree view behavior

hikchoi:
# feat: allow customization of tree view label / sorting to preserve old tree view behavior

This PR:
- adds a new config namespace `workspace.views` that will hold configurations related to dendron contributed views
- adds a config namespace `treeView` under `workspace.views`, with the key `treeItemLabelType`
- allow customization of the tree item label in tree view

- This change is added to preserve the old labeling behavior of tree view V1 for users who are already used to the tree view labeling notes with their file names.
- However, since the change to labeling was done because of the confusion introduced by the way we sort them, the sorting behavior is also tied to the labeling behavior.
  - If the items are labeled with the note's title, it will also be sorted by the title.
  - If the items are labeled with the note's fname, it will also be sorted by the fname.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended


hikchoi:
DOCS: https://github.com/dendronhq/dendron-site/pull/486

jonathanyeung:
Instead of having this as a setting in `dendron.yml`, how about we add a button to the tree view that allows users to change the options there of displaying title vs fname?  Maybe we can either add a dedicated button or use the ellipses '...' button with menu (for example of what i'm referring to as ellipses, see VS's built-in tree view control for 'Source Control')

We can then store the setting value into our metdata svc so that it persists for the user.

I feel like this will make the setting much more discoverable and make it easier to toggle. 

Sorry, I know this is quite a bit different than what you've coded up - lemme know what you think.

hikchoi:
> Instead of having this as a setting in `dendron.yml`, how about we add a button to the tree view that allows users to change the options there of displaying title vs fname? Maybe we can either add a dedicated button or use the ellipses '...' button with menu (for example of what i'm referring to as ellipses, see VS's built-in tree view control for 'Source Control')
> 
> We can then store the setting value into our metdata svc so that it persists for the user.
> 
> I feel like this will make the setting much more discoverable and make it easier to toggle.
> 
> Sorry, I know this is quite a bit different than what you've coded up - lemme know what you think.

@jonathanyeung 

Yeah I briefly thought about that route as well. Do you think having this setting persist in metadata svc vs other settings controlled in `dendron.yml` won't create any confusion? 

Where do we draw the line of "this setting stays in `dendron.yml`" and "this setting is persisted in the metadata"? 
I didn't have a good answer to this question so I went with the usual route of adding a new config in `dendron.yml`.

Happy to discuss and change the implementation accordingly.

hikchoi:
As per @kevinslin's comment in the daily async, I will keep the current implementation of storing the setting in `dendron.yml`.

For the sake of iteration, I will:
- [ ] add a toggle button that will flip between different label/sort behavior
- [ ] have it persist only for the current session
- [ ] display a toast on toggle that the default could be set in `dendron.yml`

hikchoi:
update:

I still need to add some tests for the newly added features listed below, but this PR is ready for review otherwise.

- toggle button for label type added
- when toggled, a toast will appear that suggests setting default value
  - user will have an option to set the current label type as default in `dendron.yml` (as a button)
    - clicking will back up `dendron.yml`, and then change `workspace.view.treeView.treeItemLabelType` to the current one.
  - user will also have an option to dismiss this toast forever (as a button)
- added a _expand all_ button behind `DendronContext.DEV_MODE`.
  - this works fine, but it's a bit slow. 
    - Unfortunately I don't think the VSCode team has plans to add this feature to the tree view API yet, so we have to loop through expandable items and call `.reveal(...)`.
    - With the depth limited to 3 hierarchies, testing in org-workspace gives us around 7000ms to expand 600-ish tree items initially. After the tree is built, it takes around 2000ms. (if the tree is rebuilt, i.e. label type is changed, it will take 7000ms again)

toggle buttons:
image

toast message:
image

toast message after clicking Set to default:
image

note: if you manually test this against `org-workspace`, these errors will pop up.
these are actual notes that have duplicate ids in our workspace.: 
image

hikchoi:
related to #2739

hikchoi:
Update: UI state is now persisted using the metadata store instead of config. As a result the toast message no longer prompts for setting the default value

packages/plugin-core/src/_extension.ts
```diff
@@ -1255,6 +1258,54 @@ function _setupLanguageFeatures(context: vscode.ExtensionContext) {
...
+      })
+    );
+  }
+
+  /**
+   * This is a little flaky right now, but it works most of the time.
+   * Leaving this for dev / debug purposes.
+   * Enablement is set to be DendronContext.DEV_MODE
+   *
+   * TODO: fix tree item register issue and flip the dev mode flag.
```
jonathanyeung, line 1297: out of curiosity, the flakiness is that sometimes it throws that 'duplicate id registration' error?

jonathanyeung (approved):
Looks great!

Tested locally, seems to work well.  Maybe a few nitpicks UI wise, but we can refine it in a later iteration
- I don't think it's necessary to toast everytime we click the button.  Since every line will at least change capitalization, there's quite a lot of visual feedback that the button click is doing something. 
- my pref would be to align with VS Code's panels in having the sort order be a '...' context menu, but more of a nitpick.  (I'm not sure if they expose that control to extensions?)

hikchoi:
> I don't think it's necessary to toast everytime we click the button. Since every line will at least change capitalization, there's quite a lot of visual feedback that the button click is doing something.

I can fix this. would be a quick change.

> my pref would be to align with VS Code's panels in having the sort order be a '...' context menu, but more of a nitpick. (I'm not sure if they expose that control to extensions?)

Completely missed that we discussed about this. Yes this is available to extension authors. The contribution point is `view/title` without the `group` specification (the individual icons have `group: navigation`): https://code.visualstudio.com/api/extension-guides/tree-view#view-actions

I'll get this in for this PR. I don't like the idea of introducing a visible UI element and taking it away in short succession 😄

packages/plugin-core/src/_extension.ts
hikchoi: That is _mostly_ fixed. it still happens but very rarely it seems. I removed the cause for like 95% of the cases but I'm yet to figure out the rest.

Also not mentioned here, but we are at worst calling n (numNotes) different async callbacks to fetch the file so it's pretty slow. The main reason VSCode doesn't provide this as a standard API.

I would ideally want to change this to be tree item specific (possible, add it as an inline button for each item), or have it only expand visible tree items (not sure if possible). Haven't specced out the journey here yet.

I still would like to have this behind the dev flag even if we make it tree item specific though, as this is pretty useful for visually debugging the tree view.

kevinslin (approved):
lgtm. feel free to merge when you feel is ready

hikchoi:
Update: moved the toggle button here
image

This will cycle through.

@jonathanyeung, fyi: I added `dendron-view` and other contribution points that are currently in `master` that are not set up to automatically generated with `gen:config`.

See https://github.com/dendronhq/dendron/pull/2858/commits/bcbe0768fc25a97f3bc4fff2eadc8995cf4a33d2

packages/plugin-core/src/_extension.ts
jonathanyeung: cool, thanks for the detailed explanation.

PR #2870, opened 2022-05-03
chore: end the MEETING_NOTE_FEATURE_SHOWCASE_TEST a/b test

jonathanyeung:
## chore: end the MEETING_NOTE_FEATURE_SHOWCASE_TEST a/b test

We've collected enough data for this prompt test for Meeting Notes.  The effect of ending this test is that users in the B group will see the toast one time.  Users in the A group will not see the toast a second time.

### DIFF
--- a/packages/plugin-core/src/abTests.ts
+++ b/packages/plugin-core/src/abTests.ts
@@ -11,6 +11,13 @@ export enum UpgradeToastWordingTestGroups {
   openChangelog = "openChangelog",
 }
 
+/**
+ * Section: Tests (Active or soon to be active)
+ *
+ * NOTE: please follow this convention for naming future tests:
+ * YYYY-MM-TEST_NAME.  For example, 2022-04-MEETING_NOTE_FEATURE_SHOWCASE
+ */
+
 /** Test if showing a web view on an upgrade is more successful than showing a toast notification. */
 export const UPGRADE_TOAST_WORDING_TEST = new ABTest(
   "UpgradeToastWordingTest",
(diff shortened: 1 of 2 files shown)


packages/plugin-core/src/abTests.ts
```diff
@@ -122,6 +108,9 @@ export const CURRENT_AB_TESTS = [
   UPGRADE_TOAST_WORDING_TEST,
   SELF_CONTAINED_VAULTS_TEST,
   MEETING_NOTE_TUTORIAL_TEST,
-  MEETING_NOTE_FEATURE_SHOWCASE_TEST,
   GRAPH_THEME_TEST,
 ];
+
+// Concluded Experiments (NOTE: make sure that you don't re-use these strings as
+// the name of a future AB Test to avoid conflicting strings):
+// - MeetingNoteFeatureShowcaseTest
```
kevinslin, line 116: can we have this in an SOP instead of the code? if we introduce these changes: https://github.com/dendronhq/org-private/commit/6d558b90ac2516861dab27d1c5d92f3e21a91b05

we can find all previous tests by filtering via [[abtest|dendron://private/ext.airtable.table.events#abtest]]

kevinslin: actually, thinking about this some more -should ab test have a {year}-{month} prefix so we never have to worry about this?

jonathanyeung: yeah that's a good idea. so for this one, `2022-04-MeetingNoteFeatureShowcase`

packages/plugin-core/src/abTests.ts
```diff
@@ -11,6 +11,13 @@ export enum UpgradeToastWordingTestGroups {
   openChangelog = "openChangelog",
 }
 
+/**
+ * Section: Tests (Active or soon to be active)
+ *
+ * NOTE: please follow this convention for naming future tests:
```
kevinslin, line 17: can we have this also link to the internal SOP doc about a/b tests?

PR #2873, opened 2022-05-03
fix: self contained vaults get cloned into the wrong directory

SeriousBug:
Fixes a bug where self contained vaults would get cloned into the wrong directory at startup if they were missing.

### DIFF
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -1101,13 +1101,13 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
         message: "Internal error: cloning non-git vault",
       });
     }
-    if (vault.workspace !== undefined || vault.seed !== undefined) {
-      throw new DendronError({
-        message:
-          "Internal error: can't clone workspace and seed vaults directly",
-      });
+    let repoPath: string;
+    if (vault.selfContained) {
+      // vault2Path gives the path to the `notes` folder for self contained vaults
+      repoPath = path.join(wsRoot, vault.fsPath);
+    } else {
+      repoPath = vault2Path({ vault, wsRoot });
     }
-    const repoPath = path.join(wsRoot, vault.fsPath);
     this.logger.info({ msg: "cloning", repoPath });
     const git = simpleGit({ baseDir: wsRoot });
     await git.clone(urlTransformer(vault.remote.url), repoPath);
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -32,6 +32,7 @@ import {
   DLogger,
   GitUtils,
   note2File,
+  pathForVaultRoot,
   readJSONWithComments,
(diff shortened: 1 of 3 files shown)


packages/engine-server/src/workspace/service.ts
```diff
@@ -1446,7 +1454,7 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
       await Promise.all(
         _.map(workspaces, async (wsEntry, wsName) => {
           const wsPath = path.join(wsRoot, wsName);
-          if (!fs.existsSync(wsPath)) {
+          if (!(await fs.pathExists(wsPath))) {
```
kevinslin, line 1457: was there a reason this was converted from the sync version?

packages/engine-server/src/workspace/service.ts
```diff
@@ -1503,7 +1511,7 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
     const emptyRemoteVaults = vaults.filter(
       (vault) =>
         !_.isUndefined(vault.remote) &&
-        !fs.existsSync(vault2Path({ vault, wsRoot }))
+        !fs.existsSync(path.join(wsRoot, vault.fsPath))
```
kevinslin, line 1514: why not use vault2path here?

packages/engine-server/src/workspace/service.ts
```diff
@@ -1097,9 +1097,17 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
     const wsRoot = this.wsRoot;
     if (!vault.remote || vault.remote.type !== "git") {
-      throw new DendronError({ message: "cloning non-git vault" });
+      throw new DendronError({
+        message: "Internal error: cloning non-git vault",
+      });
+    }
+    let repoPath: string;
+    if (vault.selfContained) {
+      // vault2Path gives the path to the `notes` folder for self contained vaults
```
kevinslin, line 1106: 1. vault2Path is used across the codebase. are there other places where this issue props up? 
2. since we have self contained vaults going forward, we should encapsulate this logic in a function (eg. vault2SCPath)

kevinslin (changes requested):
some comments about vault2path

packages/engine-server/src/workspace/service.ts
SeriousBug: It's already in an async function, so it makes sense to use the async version to me. Avoiding blocking for file operations makes the engine more efficient.

packages/engine-server/src/workspace/service.ts
SeriousBug: > Are there other places where this issue props up?

I'm not sure, I never tried going through all the invocations. I'll add a task to skim over the invocations and see if any of them need to be updated.

> since we have self contained vaults going forward, we should encapsulate this logic in a function (eg. vault2SCPath)

Will do, sure.

kevinslin (approved):

PR #2879, opened 2022-05-04
enhance(workspace): detect and warn for duplicate note IDs

SeriousBug:
Adds a warning when there are notes with duplicate IDs detected during initialization or reload.

Also refactors error handling to match what was discussed in the previous PR.

### DIFF
--- a/packages/common-all/src/error.ts
+++ b/packages/common-all/src/error.ts
@@ -47,7 +47,7 @@ type ServerErrorProps = {
 export type IDendronError<TCode = StatusCodes | undefined> =
   DendronErrorProps<TCode>;
 
-export abstract class BaseDendronError<TCode = StatusCodes | undefined>
+export class DendronError<TCode = StatusCodes | undefined>
   extends Error
   implements IDendronError<TCode>
 {
@@ -93,14 +93,8 @@ export abstract class BaseDendronError<TCode = StatusCodes | undefined>
     return this.stringifyForLogs();
   }
 
-  /** Returns the HTTP error code equivalent for this error.
-   *
-   * Override this to change
-   */
-  abstract translateToHTTPErrorCode(): StatusCodes | undefined;
-
   /** If false, this error does not necessarily mean the operation failed. It should be possible to recover and resume. */
-  public isFatal() {
+  public get isFatal() {
     return this.severity === ERROR_SEVERITY.FATAL;
   }
 
@@ -159,15 +153,6 @@ export abstract class BaseDendronError<TCode = StatusCodes | undefined>
   }
 }
 
-export class DendronError
-  extends BaseDendronError
-  implements IDendronError<StatusCodes>
-{
-  translateToHTTPErrorCode(): StatusCodes | undefined {
-    return this.code;
-  }
-}
-
 export class DendronCompositeError extends Error implements IDendronError {
   public payload: DendronErrorProps[];
   public severity?: ERROR_SEVERITY;
@@ -218,22 +203,22 @@ export class DendronCompositeError extends Error implements IDendronError {
 
     return (
       _.isArray(error.payload) &&
-      error.payload.every(BaseDendronError.isDendronError)
+      error.payload.every(DendronError.isDendronError)
     );
   }
+}
 
-  /** If the error is a composite error, then returns the list of errors inside it.
-   *
-   * If it is a single error, then returns that single error in a list.
-   *
-   * If this was not a Dendron error, then returns an empty list.
-   */
-  static errorsList(error: any) {
-    if (DendronCompositeError.isDendronCompositeError(error))
-      return error.payload;
-    if (BaseDendronError.isDendronError(error)) return [error];
-    return [];
-  }
+/** If the error is a composite error, then returns the list of errors inside it.
+ *
+ * If it is a single error, then returns that single error in a list.
+ *
+ * If this was not a Dendron error, then returns an empty list.
+ */
+export function errorsList(error: any) {
+  if (DendronCompositeError.isDendronCompositeError(error))
+    return error.payload;
+  if (DendronError.isDendronError(error)) return [error];
+  return [];
 }
 
 export class DendronServerError
--- a/packages/common-all/src/error.ts
+++ b/packages/common-all/src/error.ts
@@ -4,7 +4,7 @@ import { AxiosError } from "axios";
 import { ERROR_SEVERITY, ERROR_STATUS } from "./constants";
 import { RespV3 } from "./types";
 
-type DendronErrorProps = {
+export type DendronErrorProps<TCode = StatusCodes | undefined> = {
   /**
    * Arbitrary payload
    */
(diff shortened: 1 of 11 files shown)


SeriousBug:
@kevinslin @jonathanyeung Please let me know if this wasn't the design you were thinking of for the error handling. You can also take over the PR and make changes to the error handling if that makes sense.

packages/common-all/src/error.ts
```diff
@@ -45,15 +44,70 @@ type ServerErrorProps = {
...
+    }
+    if (this.payload) {
+      payload.payload = this.payloadStringify();
+    }
+    return JSON.stringify(payload);
+  }
+
+  /** The output that may be sent to Sentry, or other telemetry service.
+   *
+   * This function will eventually check that the output is stripped of PII,
```
jonathanyeung, line 89: nit: can we tweak this description slightly? 

```typescript
 /** The output that may be sent to Sentry, or other telemetry service.
   *
   * This return value from this function should be stripped of PII - if your derived error class has PII in its payload, then override this function and provide a payload that has the PII removed.
   */
  public stringifyForTelemetry() {
    // This function will eventually try to do its own checks to strip out PII, but for now the default is to return the same string as `stringifyForLogs()`
    return this.stringifyForLogs();
  }   
```

packages/common-all/src/error.ts
```diff
@@ -45,15 +44,70 @@ type ServerErrorProps = {
...
+  }
+
+  /** Returns the HTTP error code equivalent for this error.
+   *
+   * Override this to change
+   */
+  abstract translateToHTTPErrorCode(): StatusCodes | undefined;
+
+  /** If false, this error does not necessarily mean the operation failed. It should be possible to recover and resume. */
+  public isFatal() {
```
jonathanyeung, line 103: nit: can we use getter syntactic sugar? 

public get isFatal(): boolean {
  ...
}

jonathanyeung, line 103: nit: can we use getter syntactic sugar?

```typescript
public get isFatal(): boolean { ... }
```

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -67,6 +70,60 @@ import { SchemaParser } from "./schemaParser";
 import { InMemoryNoteCache } from "../../util/inMemoryNoteCache";
 import { NotesFileSystemCache } from "../../cache";
 
+export enum EngineInitErrorType {
+  DUPLICATE_NOTE_ID = "duplicate note id",
+}
+
+export class DuplicateNoteError extends BaseDendronError<EngineInitErrorType.DUPLICATE_NOTE_ID> {
```
jonathanyeung, line 77: nit: my pref would be to have this class in a separate file, but I don't feel too strongly about it. 

Could you also add a small class header?

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -67,6 +70,60 @@ import { SchemaParser } from "./schemaParser";
...
+      fname: opts.noteA.fname,
+      vault: opts.noteA.vault,
+    };
+    this.noteB = {
+      fname: opts.noteB.fname,
+      vault: opts.noteB.vault,
+    };
+  }
+
+  public noteA: {
```
jonathanyeung, line 111: rather than noteA/noteB, does it make more sense to have a property of type {fname, vault} array? In case there are 3 notes with the same ID for example.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -67,6 +70,60 @@ import { SchemaParser } from "./schemaParser";
...
+  };
+  public noteB: {
+    fname: string;
+    vault: DVault;
+  };
+
+  static isDuplicateNoteError(
+    error: IDendronError<any>
+  ): error is DuplicateNoteError {
+    return Object.values(EngineInitErrorType).includes(error.code);
```
jonathanyeung, line 123: Can we alter this check? This won't work anymore once we expand the enum values in `EngineInitErrorType`.

packages/common-all/src/error.ts
```diff
@@ -16,10 +16,9 @@ type DendronErrorProps = {
   severity?: ERROR_SEVERITY;
 
   /**
-   * @deprecated - should only used in DendronServerError
    * Optional HTTP status code for error
```
jonathanyeung, line 19: Can you amend this description to "Optional TCode to identify the type of error" or something to that effect? Basically, remove the 'HTTP' reference.  Or, maybe mention that it'll be a Status_Code if you don't specify a generic parameter.

Was `<TCode = StatusCodes | undefined>` added for compatibility purposes?

packages/common-all/src/error.ts
```diff
@@ -45,15 +44,70 @@ type ServerErrorProps = {
...
+   */
+  public stringifyForTelemetry() {
+    return this.stringifyForLogs();
+  }
+
+  /** Returns the HTTP error code equivalent for this error.
+   *
+   * Override this to change
+   */
+  abstract translateToHTTPErrorCode(): StatusCodes | undefined;
```
jonathanyeung, line 100: I don't see any invocations of this yet - do we need it here?  I'd prefer to have anything HTTP related to be a part of a derived class as opposed to in this base class.

jonathanyeung (changes requested):
Overall, looking great - thanks a lot for helping to implement the error scaffolding.  Just a few minor comments

packages/engine-server/src/drivers/file/storev2.ts
SeriousBug: The way it works right now can't detect more than a pair of duplicate note IDs. If there are 3 notes with the same ID then there would be 2 different errors. I'll leave it as-is for now, if someone implements a better check that can capture multi-note conflicts then they can refactor it.

packages/common-all/src/error.ts
SeriousBug: > Was `<TCode = StatusCodes | undefined>` added for compatibility purposes?

Yes, that's what it used to be, so I wanted to keep that backwards compatible.

packages/common-all/src/error.ts
SeriousBug: I must have misunderstood the design docs then, I'll remove this.

packages/common-all/src/error.ts
```diff
@@ -45,15 +44,64 @@ type ServerErrorProps = {
...
+    }
+    return JSON.stringify(payload);
+  }
+
+  /** The output that may be sent to Sentry, or other telemetry service.
+   *
+   * This function will eventually check that the output is stripped of PII,
+   * but for now that's the same as these.
+   */
+  public stringifyForTelemetry() {
```
jonathanyeung, line 92: _[My bad - I had this comment in the review earlier but it somehow got deleted]_

nit: can we tweak this description slightly?

 /** The output that may be sent to Sentry, or other telemetry service.
   *
   * This return value from this function should be stripped of PII - if your derived error class has PII in its payload, then override this function and provide a payload that has the PII removed.
   */
  public stringifyForTelemetry() {
    // This function will eventually try to do its own checks to strip out PII, but for now the default is to return the same string as `stringifyForLogs()`
    return this.stringifyForLogs();
  }

jonathanyeung (approved):
lgtm - I added 1 additional comment that accidentally got deleted earlier, but not a big deal.

kevinslin (approved):

PR #2888, opened 2022-05-06
chore: enhancing meeting note telemetry

jonathanyeung:
## chore: enhancing meeting note telemetry #2888

Adding telemetry to let us know when a user is creating a note from a Meeting Note. This is meant to test the hypothesis that Meeting Notes will lead to users creating more notes.

### DIFF
--- a/packages/plugin-core/src/commands/GotoNote.ts
+++ b/packages/plugin-core/src/commands/GotoNote.ts
@@ -275,7 +275,9 @@ export class GotoNoteCommand extends BasicCommand<
     let pos: undefined | Position;
     const out = await this.extension.pauseWatchers<GoToNoteCommandOutput>(
       async () => {
-        const existing = await client.getNoteByPath({
+        // First query to see if the note exists to determine if we will send
+        // telemetry, createIfNew set to false
+        const respFromExistenceCheck = await client.getNoteByPath({
           npath: qs,
           createIfNew: false,
           vault,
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -274,13 +275,33 @@ export class GotoNoteCommand extends BasicCommand<
     let pos: undefined | Position;
     const out = await this.extension.pauseWatchers<GoToNoteCommandOutput>(
       async () => {
+        const existing = await client.getNoteByPath({
```
kevinslin, line 278: confused about this line. doesn't it get the path of the note that is being created, not the existing note?

jonathanyeung: The variable is named misleadingly - it's basically querying the engine if a note named ${qs} - 'createIfNew' is set to false.  It's used below, if `if (!existing.data?.note)`, i.e, if there isn't an existing note, then send the telem marker.

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -274,13 +275,35 @@ export class GotoNoteCommand extends BasicCommand<
...
           npath: qs,
           createIfNew: true,
           vault,
           overrides,
         });
         const note = data?.note as NoteProps;
+
+        // If a new note is going to be created, check if we should send meeting
+        // note telemetry.
+        if (!respFromExistenceCheck.data?.note) {
```
kevinslin, line 297: you should be able to get this from the `changed` property of `getNoteByPath`

_.filter(changed, ent => ent.status === "created" && note.fname = qs)

jonathanyeung:
Airtable "upcoming" updated with new telem markers

PR #2889, opened 2022-05-06
chore: add a task to automatically sync tutorials

hikchoi:
# chore: add a task to automatically sync tutorials

This PR:
- Adds a VSCode task that can automatically sync tutorials from the `dendron-site` seed added in the monorepo
- Adds a generic way to load differently treated tutorials based on the currently active AB test.
- Adds a newly treated tutorial for refactoring BML (through dendronhq/dendron-site#495)
- Syncs main and treated tutorial from `dendron-site` to `plugin-core`

Tutorial syncing conventions:

- in `dendron-site`, the main tutorial that has all note refs / links resolved to url lives in `tutorial.*` hierarchy
  - this is synced as-is to `plugin-core`
- in `dendron-site`, the treated tutorials go under `tutorial-alt.{treatmentName}.*` hierarchy
  - this is selectively synced to `plugin-core`
  - when synced, `tutorial-alt.{treatmentName}` from the filename will be replaced by `tutorial`, so any internal linking within the treated tutorials should be pointing to `tutorial.*` instead of `tutorial-alt.{treatmentName}`
  - when running the VSCode task, you are prompted for the `treatmentName`, and only that will be synced to `plugin-core`
    - To keep experiment data clean, we always want to have only one treatment to the tutorial. This is the reason why we are selecting one treatment of tutorial to sync.

- note: I have some thoughts on ways to improve this flow: markdown export currently mostly works for us, but there are some things that still needs some manual tweaking after the export. I'll create a separate task / proposal and share later.

General flow of syncing tutorials:
- write treated tutorials in an appropriate hierarchy in `dendron-site`. (in the monorepo, or in `org-workspace` and pull it into the monorepo)
- `Tasks: Run Task` -> select `sync:tutorial` -> type in `treatmentName`

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended

- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

### Basics

## 

### Basics

## 

### Extended


hikchoi:
- Pending tasks
  - [ ] Refactor tutorial notes in `dendron-site` to contain the main and treated tutorials based on the decided convention.
  - [ ] Sync to new tutorial notes

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -60,35 +53,59 @@ export class TutorialInitializer
...
+    let filter: fs.CopyFilterSync;
+    let replacePattern: string;
+    switch (ABUserGroup) {
+      // swap out to appropriate enum case for current AB testing group
+      case MeetingNoteTestGroups.show: {
+        filter = (filePath) => {
+          if (path.extname(filePath) === ".md") {
+            const basename = path.basename(filePath);
+            return (
+              basename === "root.md" ||
```
kevinslin, line 76: do we also need to copy over root.schema.yml?

hikchoi: we do. this block is only concerned with paths that end with `.md`. all other paths (asset path and schema) are always copied over. (see line 79-81 just below)

hikchoi:
diffs between alt / main tutorial for your reviewing convenience:

### tutorial.md
```diff
diff packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.md packages/plugin-core/assets/dendron-ws/tutorial/alt/tutorial.md
2c2
< id: 4u6pv56mnt25d8l2wzfygu7
---
> id: tbilk9to67d0dwgnjanj5ph
5c5
< updated: 1652122742329
---
> updated: 1652126871212
```

### tutorial.user-interface.md
```diff
colordiff packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.user-interface.md packages/plugin-core/assets/dendron-ws/tutorial/alt/tutorial.user-interface.md
2c2
< id: epmpyk2kjdxqyvflotan2vt
---
> id: kl6ndok3a1f14be6zv771c9
5c5
< updated: 1652122928650
---
> updated: 1652126953958
>
```

### tutorial.taking-notes.md
```diff
diff packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.taking-notes.md packages/plugin-core/assets/dendron-ws/tutorial/alt/tutorial.taking-notes.md
2c2
< id: yyfpeq3th3h17cgvj8cnjw5
---
> id: iq3ggn67k1u3v6up8ny3kf5
7c7
< updated: 1652123018292
---
> updated: 1652126933577
122c122
< Read [Why Dendron uses hierarchies](https://wiki.dendron.so/notes/683740e3-70ce-4a47-a1f4-1f140e80b558) for detailed information on Dendron note structure. Dendron also allows you to change the hierarchy at a later date. Read [Refactoring in Dendron](https://wiki.dendron.so/notes/9MZBqhrijEM4QpZRa5t08) in case you change your mind on how to organize your notes.
---
> Read [Why Dendron uses hierarchies](https://wiki.dendron.so/notes/683740e3-70ce-4a47-a1f4-1f140e80b558) for detailed information on Dendron note structure.
```

### tutorial.linking-notes.md
```diff
diff packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.linking-notes.md packages/plugin-core/assets/dendron-ws/tutorial/alt/tutorial.linking-notes.md
2c2
< id: khv6u4514vnvvy4njhctfru
---
> id: rjnqumna1ye82u9u76ni42k
5c5
< updated: 1652123099783
---
> updated: 1652126857106
56,112d55
< ## Refactoring in Dendron
<
< You may wonder what happens to your links if you want to change the name of a note or header. Dendron keeps this in mind and provides commands to update your notes. All the while, Dendron updates any and all wikilinks in your notes.
<
< ### Rename Note
<
< The [Rename Note](https://wiki.dendron.so/notes/srajljj10V2dl19nCSFiC) command is used to change the name of, and all links to, a single note.
<
< > 🌱 **ACTION:** Use the lookup to navigate to `[[recipes.italian.desserts.tiramisu]]`. Then:
< >
< > - Run the command `Dendron: Rename Note` to change the title to `[[recipes.italian.desserts.gelato]]`.
< > - Use the lookup to navigate to `[[recipes.ingredients.shopping-list]]`. You should see the wikilink there has updated from `[[recipes.italian.desserts.tiramisu#ingredients]]` to `[[recipes.italian.desserts.gelato#ingredients]]`, ensuring links are updated!
<
< ### Rename Header
<
< The [Rename Header](https://wiki.dendron.so/notes/srajljj10V2dl19nCSFiC) command is used to change the name of, and all links to, a single header.
<
< > 🌱 **ACTION:** Use the lookup to navigate to `[[recipes.italian.desserts.gelato]]`. Then:
< >
< > - Place the cursor in `## Ingredients`, and run the command `Dendron: Rename Header`
< > - Rename the header to `## Best Ingredients`
< > - Use the lookup to navigate to `[[recipes.ingredients.shopping-list]]`. You should see the wikilink there has updated from `[[recipes.italian.desserts.gelato#ingredients]]` to `[[recipes.italian.desserts.gelato#best-ingredients]]`, ensuring links are updated!
<
< If you want to migrate a section beneath a certain header to a different note entirely, [Move Header](https://wiki.dendron.so/notes/srajljj10V2dl19nCSFiC) will do the job and ensure links are updated.
<
< ### Refactor Hierarchy
<
< The [Refactor Hierarchy](https://wiki.dendron.so/notes/srajljj10V2dl19nCSFiC) command is used to change the name of, and all links to, a collection of notes.
<
< > 🌱 **ACTION:** This section will be all about creating and refactoring notes, in order to get an idea for how it all works.
<
< - Create the following notes by placing the cursor inside each, and pressing `Ctrl+Enter`:
<   - `[[recipes.vegetarian.tikka-masala]]`
<   - `[[recipes.vegetarian.aloo-paratha]]`
<   - `[[recipes.vegetarian.dosa]]`
< - Each of these are Indian food! Let's move them into the `indian` hierarchy.
<   - Run the command `Dendron: Refactor Hierarchy`
<   - At the first prompt, enter `recipes.vegetarian` (note hierarchy you are targeting)
<   - At the second prompt, enter `recipes.indian` (renaming the targeted hierarchy)
<
< A refactor preview should be provided, before making the change, to confirm the changes you are about to make.
<
< ![Refactor Hierarchy Preview prompt](https://org-dendron-public-assets.s3.amazonaws.com/images/refactor-hierarchy-preview-prompt.png)
<
< This will change all of the targeted notes:
<
< ```
< recipes.vegetarian.md               -->	recipes.indian.md
< recipes.vegetarian.tikka-masala.md  -->	recipes.indian.tikka-masala.md
< recipes.vegetarian.aloo-paratha.md  -->	recipes.indian.aloo-paratha.md
< recipes.vegetarian.dosa.md          -->	recipes.indian.dosa.md
< ```
<
< Selecting `Proceed` renames all the notes! There's no need to worry about any broken links throughout your notes, as those are updated by `Refactor Hierarchy`, too.
<
< > 💡 **TIP:** We could have chosen to move them all to `[[recipes.indian.vegetarian]]`, too, if wanting to add `vegetarian` to recipe hierarchies. Doing this would make it easy to do a lookup against all vegetarian recipes with `recipes. vegetarian`. Another alternative is to take a look at using [Tags](https://wiki.dendron.so/notes/8bc9b3f1-8508-4d3a-a2de-be9f12ef1821).
<
```

### tutorial.rich-formatting.md
```diff
diff packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.rich-formatting.md packages/plugin-core/assets/dendron-ws/tutorial/alt/tutorial.rich-formatting.md
2c2
< id: lxrp006mal1tfsd7nxmsobe
---
> id: ie5x2bq5yj7uvenylblnhyr
5c5
< updated: 1652123135299
---
> updated: 1652126894307
>
```

### tutorial.conclusion.md
```diff
diff packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.conclusion.md packages/plugin-core/assets/dendron-ws/tutorial/alt/tutorial.conclusion.md
2c2
< id: kyjfnf2rnc6vn71iyn9liz7
---
> id: wmbd5xz40ohjb8rd5b737cq
5c5
< updated: 1652123184798
---
> updated: 1652126817206
>
```

hikchoi:
will create an async on the process of adding alt tutorials / syncing them once this PR gets an okay.

packages/common-all/src/abTests.ts
```diff
@@ -138,4 +157,5 @@ export const CURRENT_AB_TESTS = [
   MEETING_NOTE_TUTORIAL_TEST,
```
jonathanyeung, line 157: as per offline discussion, let's go ahead and remove everything related to `MEETING_NOTE_TUTORIAL_TEST`

packages/common-all/src/abTests.ts
```diff
@@ -138,4 +157,5 @@ export const CURRENT_AB_TESTS = [
...
   GRAPH_THEME_TEST,
   GRAPH_THEME_FEATURE_SHOWCASE_TEST,
+  AB_2022_05_REFACTOR_MENTION,
```
jonathanyeung, line 160: Can we add 'Tutorial' somewhere in this test name?

packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.md
```diff
@@ -0,0 +1,31 @@
...
+updated: 1652122742329
+created: 1608051264282
+nav_order: 1.1
+---
+
+## Pre-requisites
+
+This walkthrough will help you get started with the basics of Dendron. Before you start, make sure you do the following:
+
+- Install [VS Code](https://code.visualstudio.com/download) or a VS Code compatible editor like [VSCodium](https://vscodium.com/)
```
jonathanyeung, line 14: We shouldn't display this should we?

jonathanyeung (changes requested):
Can we also do the following: 
- Create a note to track this experiment as per [[A/B Testing|dendron://private/area.metrics.sop.ab-testing]]

packages/plugin-core/assets/dendron-ws/tutorial/main/tutorial.md
kevinslin: since we are already copying over the tutorial, lets just tailor `tutorial.md` for vscode. can we replace the text with the following?

```
Dendron is an IDE for general knowledge.
Combine the simplicity of markdown with the power of a modern development environment. 

To begin, lets do a quick [[overview of the Dendron UI|tutorial.user-interface]]!
```

packages/dendron-cli/src/commands/devCLICommand.ts
```diff
@@ -225,6 +234,17 @@ export class DevCLICommand extends CLICommand<CommandOpts, CommandOutput> {
           await this.syncAssets(opts);
           return { error: null };
         }
+        case DevCommands.SYNC_TUTORIAL: {
+          if (!opts.treatmentName) {
```
kevinslin, line 238: is there every a case where we don't just want to sync all the tutorials? anything under `tutorial*` should be valid

kevinslin (changes requested):
added some comments about making tutorial sync a single click function vs worrying about treatment names

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -61,36 +54,28 @@ export class TutorialInitializer
 
     const vpath = vault2Path({ vault: opts.vaults[0], wsRoot: opts.wsRoot });
 
-    fs.copySync(path.join(dendronWSTemplate.fsPath, "tutorial"), vpath);
-
-    const ABUserGroup = MEETING_NOTE_TUTORIAL_TEST.getUserGroup(
+    const ABUserGroup = AB_2022_05_REFACTOR_MENTION.getUserGroup(
```
kevinslin, line 57: instead of AB testing specifically for refactor mention, maybe this should just be a long running ab test for tutorial changes? `eg. AB_TUTORIAL_TEST. the groups here can change over time with different values that represent what we're testing

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -61,36 +54,28 @@ export class TutorialInitializer
...
-    rif.replaceInFile(options).catch((err: Error) => {
-      Logger.error({
-        ctx,
-        error: new DendronError({
-          innerError: err,
-          message: "error replacing tutorial placeholder text",
-        }),
-      });
-    });
+    switch (ABUserGroup) {
```
kevinslin, line 61: also - for the initial tutorial sync, can we also include the original tutorial with the meeting notes addition as an alternative? this way, we can compare the current tutorial as a baseline, compared to updated tutorial and updated tutorial without refactoring

kevinslin (changes requested):

packages/plugin-core/src/workspace/tutorialInitializer.ts
hikchoi: Nice idea. That will make updating the tutorial easier each time since we don't have to ever touch `tutorialInitializer`. We just modify the test group.

@jonathanyeung thoughts?

packages/plugin-core/src/workspace/tutorialInitializer.ts
hikchoi: As in, have three groups? e.g.) `"mention" | "noMention" | "original"`

hikchoi: In this case I think we should also include the meeting note paragraph in the other two tutorials.

packages/plugin-core/src/workspace/tutorialInitializer.ts
jonathanyeung: I think in this situation it _should_ work. 

For other situations, I don't think this is good practice though, as modifying the number of groups or their weights will cause people to change groups over time, which will muddy any findings.  Given that users most likely only go through the tutorial once, it shouldn't be much of an issue here.

packages/plugin-core/src/workspace/tutorialInitializer.ts
jonathanyeung: Agree with all comments. 

So then we can understand
- Difference between 'old' (current) and 'new' (wiki.dendron.so version)
- Difference between 'refactor mention' and 'no-refactor-mention'

One thing to keep in mind is that if we have 3 groups, we should expect 1.5x the amount of time to see the same amount of data, and 1.5x the amount of time to see results with the same statistical significance as 2 groups.

hikchoi: > One thing to keep in mind is that if we have 3 groups, we should expect 1.5x the amount of time to see the same amount of data, and 1.5x the amount of time to see results with the same statistical significance as 2 groups.

Noted. I know @kevinslin is planning to set up a short/long tutorial AB test to come up next. Are you fine with this test running for the next 2 releases?

packages/plugin-core/src/workspace/tutorialInitializer.ts
hikchoi: I think one way we can tackle this (if we want to take care of the case where user goes through tutorial multiple times during their use of Dendron) is to track what group the user was initially assigned to in metadata store.

```
lastTutorialTestGroup = metadata.tutorialTestGroup
if AB_TUTORIAL_TEST[lastTutorialTestGroup] exists:
  show tutorial for lastTutorialTestGroup regardless of test group change
else
  setMeta(newlyAssignedTestGroup) # track it for the next round
  handleCurrentTestGroup
```

This will at least preserve what we show the user between AB tests if we keep some groups around from last time as baseline.

hikchoi: but like @jonathanyeung said, user will most likely only go through the tutorial once, so I'll just keep this noted and not implemented for the time being.

packages/plugin-core/src/workspace/tutorialInitializer.ts
kevinslin: yep

hikchoi:
Updates

- Changed test to `AB_TUTORIAL_TEST`
  - This will a long-running task to be re-used for subsequent AB testing on tutorials
- Next round of tutorial AB testing will look at three groups: `main`, `no-refactor`, and `original`
  - `main`: identical to `dendron.tutorial.*`, except for the landing note which is tailored to be read within VSCode, and also includes the meeting note mention
  - `no-refactor`: identical to `main`, except it doesn't mention refactoring
  - `original`: the _old_ tutorial, including the meeting note mention
- Removed `MEETING_NOTE_TUTORIAL_TEST` related code
- `tutorialInitializer` will now accept an arbitrary number of treated tutorials based on the test group defined for the A/B test, doesn't need modification on each new test.
- `sync:tutorial` now doesn't ask for a specific tutorial treatment and will grab everything from `tutorial.*` hierarchy from `dendron-site`
- `sync:tutorial` will take everything under `tutorial.{treatmentName}.*` and sync it to `assets/dendron-ws/tutorial/treatments/{treatmentName}`.

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -16,7 +15,6 @@ import {
 } from "@dendronhq/engine-server";
 import fs from "fs-extra";
 import path from "path";
```
kevinslin, line 17: can we also replace `rif` from our dependencies now? (I don't think we're using it anywhere)

I believe you can just run `yarn remove {pkg}` inside the workspace

kevinslin:
comment about removing old dep. otherwise looks good

packages/dendron-cli/src/commands/devCLICommand.ts
```diff
@@ -412,6 +417,65 @@ export class DevCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+
+    const commonDirPath = path.join(tutorialDirPath, "common");
+
+    // wipe everything in /assets/dendron-ws/tutorial/treatments
+    const treatmentsDirPath = path.join(tutorialDirPath, "treatments");
+
+    fs.removeSync(treatmentsDirPath);
+    fs.ensureDirSync(treatmentsDirPath);
+
+    // grab everything from `tutorial.*` hierarchy
```
jonathanyeung, line 444: nit: maybe expand on this comment to say exclude 'tutorial.md' because the websites' version says to download vs code, etc.

packages/plugin-core/assets/dendron-ws/tutorial/treatments/main/tutorial.taking-notes.md
```diff
@@ -0,0 +1,143 @@
...
+3. Press **Enter**. The note is opened.
+
+> 💡 **INFO:** Lookup uses _fuzzy search_, which means you can type out partial queries and still see the results.Entering multiple keywords delimited by space will lookup matching notes regardless of order of the > keywords.
+>
+> - Example: `vege rec` or `rec vege` will match the `recipes.vegetarian` note.
+
+> 💡 **TIP:** Read [Finding Notes](https://wiki.dendron.so/notes/84a0366a-eab5-4862-9bef-360f92a196dc) to understand the different ways of querying the lookup.
+> Most queries can be satisfied using [lookup](https://wiki.dendron.so/notes/84a0366a-eab5-4862-9bef-360f92a196dc), but sometimes you want to do a full text search across the **contents** of your notes. This can be done with `Search: Find in Files` (`Ctrl+Shift+F` / `Cmd+Shift+F`), which leverages VS Code [built-in search](https://code.visualstudio.com/docs/editor/codebasics#_search-across-files) capabilities.
+
+Dendron also has a few built-in note types. For example, if you find yourself taking meeting notes often, you can use the `Dendron: Create Meeting Note` command to create a note with a pre-built template for meetings. Try it out!
```
jonathanyeung, line 139: for parity (and experimental consistency), can we also add a level 3 header for the meeting notes section? 

```markdown
### Meeting Notes

Dendron also has a few built-in note types...
```

jonathanyeung (approved):
Left another comment on the testing group thread, but we can defer making a decision on that later.

packages/plugin-core/src/workspace/tutorialInitializer.ts
jonathanyeung: @hikchoi @kevinslin On second thought, it's better to never change the groups.  Which AB group a user is in is added to User Props, so if a user changes groups, we won't be able to see their historical data properly because their user props have changed (at least not easily in Amplitude).

We can defer making a decision until the next time around, so I have no problems to go ahead and merge as-is.

packages/dendron-cli/src/commands/devCLICommand.ts
hikchoi: that's not related to this piece of code.

what you are describing would be already treated when we are updating the `tutorial` hierarchy in `dendron-site`.

We exclude `tutorial.md` here because the actual tutorials are in `tutorial.{treatmentName}.*`.

`tutorial.md` is just there to add a warning message that this is not the hierarchy that is being published to the wiki. We don't want to sync this note because it's not actually part of the tutorial asset.

packages/plugin-core/assets/dendron-ws/tutorial/treatments/main/tutorial.taking-notes.md
hikchoi: oops. missed that from the constant.

this already got merged so I raised another PR to sync this: https://github.com/dendronhq/dendron/pull/2919
PR for `dendron-site` already has it included.

PR #2890, opened 2022-05-06
chore: make full graph view as default

Harshita-mindfire:
This PR aims to:
- Make Full Graph view as default view when you first open the graph.
- Hides Graph Theme for Schema graph view
- fixes the following bug with display of [Note Graph Message](https://github.com/dendronhq/dendron/blob/chore/make-full-graph-view-as-default/packages/dendron-plugin-views/src/components/graph.tsx#L406:L406) overlaps with full note graph in background

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [~] 1-2 Edge cases tested

### Extended
NA

## Docs

### Extended

### DIFF
--- a/packages/dendron-plugin-views/src/hooks/useSyncGraphWithIDE.tsx
+++ b/packages/dendron-plugin-views/src/hooks/useSyncGraphWithIDE.tsx
@@ -47,10 +47,6 @@ const useSyncGraphWithIDE = ({ graph, ide, engine, config }: Props) => {
       ) {
         graphActiveNode.select();
         graph.center(graphActiveNode);
-        graph.zoom({
-          level: 1.5, // the zoom level
-          renderedPosition: graphActiveNode.position(),
-        });
 
         setLastSelectedID(graphActiveNode.id());
       }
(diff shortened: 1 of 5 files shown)


jonathanyeung (approved):
Tested the change, lgtm - as part of the same release that this change goes out with, can we also make the change to add the local vs full telemetry markers? (can be separate PR)

packages/dendron-plugin-views/src/hooks/useSyncGraphWithIDE.tsx
```diff
@@ -47,10 +47,6 @@ const useSyncGraphWithIDE = ({ graph, ide, engine, config }: Props) => {
       ) {
         graphActiveNode.select();
         graph.center(graphActiveNode);
-        graph.zoom({
```
SeriousBug, line 50: This is how the graph looks like for me when I open it:



Would it make sense to add back the initial zoom here? I think it might make more sense if it was at least zoomed in enough to display the current note.

SeriousBug (approved):
Works fine for me, left one comment about the default zoom level.

packages/dendron-plugin-views/src/hooks/useSyncGraphWithIDE.tsx
Harshita-mindfire: The only reason behind removing it was, that it gets way too zoomed in. It gives an impression that the graph isn't loaded but when we zoom out we see all the nodes. Will find an optimum zoom value for this an update.

kevinslin (approved):
nitpick: this should be classified as a `enhance`, not a `chore` because it is user visible

PR #2893, opened 2022-05-09
chore: fixing messageType prop in tip of day telem

jonathanyeung:
## chore: fixing messageType prop in tip of day telem

Fixing the `messageType` property of the `FeatureShowcaseDisplayed` event for the 'Tip of the Day' view. It's currently reporting blank.

Verified the behavior under the debugger.  Updated Airtable record.


hikchoi (approved):
LGTM

PR #2898, opened 2022-05-10
chore: adding telemetry to backlinks panel

jonathanyeung:
## chore: adding telemetry to backlinks panel

Adding telemetry to Backlinks Panel.  

EDIT: Airtable Updated.

### DIFF
--- a/packages/plugin-core/src/constants.ts
+++ b/packages/plugin-core/src/constants.ts
@@ -498,6 +498,10 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     },
     when: DendronContext.PLUGIN_ACTIVE,
   },
+  GOTO_BACKLINK: {
+    key: "dendron.gotoBacklink",
+    title: `${CMD_PREFIX} Go To Backlink`,
+  },
   // --- Workspace
   ADD_AND_COMMIT: {
     key: "dendron.addAndCommit",
--- a/packages/plugin-core/src/workspace.ts
+++ b/packages/plugin-core/src/workspace.ts
@@ -8,6 +8,7 @@ import {
   ERROR_STATUS,
   getStage,
   ResponseUtil,
+  VSCodeEvents,
   WorkspaceSettings,
   WorkspaceType,
 } from "@dendronhq/common-all";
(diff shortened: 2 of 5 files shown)


packages/plugin-core/src/workspace.ts
```diff
@@ -591,12 +592,29 @@ export class DendronExtension implements IDendronExtension {
...
       }
     );
+
+    backlinkTreeView.onDidExpandElement(() => {
+      AnalyticsUtils.track(VSCodeEvents.BacklinksPanelUsed, {
+        type: "ExpandElement",
+      });
+    });
+
+    backlinkTreeView.onDidChangeVisibility(() => {
```
kevinslin, line 602: does this fire on both visible and hide events?

packages/plugin-core/src/workspace.ts
```diff
@@ -591,12 +592,29 @@ export class DendronExtension implements IDendronExtension {
...
+      });
+    });
+
     this.backlinksDataProvider = backlinksTreeDataProvider;
     this.context.subscriptions.push(backlinksTreeDataProvider);
 
     vscode.commands.registerCommand(
       DENDRON_COMMANDS.BACKLINK_SORT_BY_LAST_UPDATED.key,
       sentryReportingCallback(() => {
+        AnalyticsUtils.track(VSCodeEvents.BacklinksPanelUsed, {
```
kevinslin, line 614: can we also report on what the value was set to?

packages/plugin-core/src/constants.ts
```diff
@@ -498,6 +498,10 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     },
     when: DendronContext.PLUGIN_ACTIVE,
   },
+  GOTO_BACKLINK: {
```
kevinslin, line 501: can we hide this from command palette?

kevinslin (changes requested):

packages/plugin-core/src/workspace.ts
jonathanyeung: updated event name + added a state flag.

packages/plugin-core/src/constants.ts
jonathanyeung: good catch, forgot how the commands manifest worked haha

SeriousBug (approved):

PR #2901, opened 2022-05-10
feat: local graph view in the Dendron Side Panel

Harshita-mindfire:
This PR aims to add a new view in the Dendron side panel: `Graph Panel`

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended
NA
### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
NA

## Docs

### Extended
 ## Analytics
Will update the airtable when PR is merged. 
- Added a GraphEvent  `GraphPanelUsed` and its properties `type`: string, `state`:string

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -28,7 +28,6 @@ export enum VSCodeEvents {
   FeatureShowcaseResponded = "FeatureShowcaseResponded",
   HelpAndFeedbackItemClicked = "HelpAndFeedbackItemClicked",
   BacklinksPanelUsed = "BacklinksPanelUsed",
-  GraphPanelClicked = "GraphPanelClicked",
 }
 
 export enum CLIEvents {
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -28,6 +28,7 @@ export enum VSCodeEvents {
   FeatureShowcaseResponded = "FeatureShowcaseResponded",
   HelpAndFeedbackItemClicked = "HelpAndFeedbackItemClicked",
   BacklinksPanelUsed = "BacklinksPanelUsed",
+  GraphPanelClicked = "GraphPanelClicked",
 }
 
 export enum CLIEvents {
@@ -156,6 +157,7 @@ export enum AppNames {
 export enum GraphEvents {
   GraphThemeChanged = "GraphThemeChanged",
   GraphViewUsed = "GraphViewUsed",
+  GraphPanelUsed = "GraphPanelUsed",
 }
 
 export const DendronEvents = {
--- a/packages/dendron-plugin-views/src/components/DendronSideGraphPanel.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronSideGraphPanel.tsx
@@ -1,0 +1,16 @@
+import { ideHooks, ideSlice } from "@dendronhq/common-frontend";
+import { DendronComponent } from "../types";
+import DendronGraphPanel from "./DendronGraphPanel";
+
+/**
+ * Wrapper component around DendronGraphPanel
+ * @param props
+ * @returns DendronGraphPanel component with ide.isSidePanel set to true
+ */
+const DendronSideGraphPanel: DendronComponent = (props) => {
+  const ideDispatch = ideHooks.useIDEAppDispatch();
+  // set side panel to true
+  ideDispatch(ideSlice.actions.setIsSidePanel(true));
+  return <DendronGraphPanel {...props} />;
+};
+export default DendronSideGraphPanel;
--- a/packages/dendron-plugin-views/src/components/DendronSideGraphPanel.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronSideGraphPanel.tsx
@@ -1,4 +1,3 @@
-import { ideHooks, ideSlice } from "@dendronhq/common-frontend";
 import { DendronComponent } from "../types";
 import DendronGraphPanel from "./DendronGraphPanel";
 
@@ -8,9 +7,7 @@ import DendronGraphPanel from "./DendronGraphPanel";
  * @returns DendronGraphPanel component with ide.isSidePanel set to true
  */
 const DendronSideGraphPanel: DendronComponent = (props) => {
-  const ideDispatch = ideHooks.useIDEAppDispatch();
-  // set side panel to true
-  ideDispatch(ideSlice.actions.setIsSidePanel(true));
+  props.isSidePanel = true;
   return <DendronGraphPanel {...props} />;
 };
 export default DendronSideGraphPanel;
--- a/packages/plugin-core/src/showcase/GraphPanelTip.ts
+++ b/packages/plugin-core/src/showcase/GraphPanelTip.ts
@@ -15,7 +15,7 @@ export class GraphPanelTip implements IFeatureShowcaseMessage {
   }
 
   getDisplayMessage(_displayLocation: DisplayLocation): string {
-    return "Dendron has a new Graph Panel. Check it out!";
+    return "We're experimenting with a note graph panel in the new Dendron sidebar. Check it out!";
   }
 
   onConfirm() {
--- a/packages/plugin-core/src/showcase/GraphPanelTip.ts
+++ b/packages/plugin-core/src/showcase/GraphPanelTip.ts
@@ -1,0 +1,32 @@
+import { ShowcaseEntry } from "@dendronhq/engine-server";
+import * as vscode from "vscode";
+import {
+  DisplayLocation,
+  IFeatureShowcaseMessage,
+} from "./IFeatureShowcaseMessage";
+import { DendronTreeViewKey } from "@dendronhq/common-all";
+
+export class GraphPanelTip implements IFeatureShowcaseMessage {
+  shouldShow(_displayLocation: DisplayLocation): boolean {
+    return true;
+  }
+  get showcaseEntry(): ShowcaseEntry {
+    return ShowcaseEntry.GraphPanel;
+  }
+
+  getDisplayMessage(_displayLocation: DisplayLocation): string {
+    return "Dendron has a new Graph Panel. Check it out!";
+  }
+
+  onConfirm() {
+    vscode.commands.executeCommand(`${DendronTreeViewKey.GRAPH_PANEL}.focus`);
+  }
+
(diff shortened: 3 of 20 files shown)


kevinslin (commented):
excited about this. might actually use the graph with this :)

jonathanyeung:
Tested it out - looks very cool. 

I'm hitting an issue though - it's likely a pre-existing issue.  When I navigate a few links using the graph view, particularly in a circle (say A->B->C->B->A), sometimes the graph view ends up turning blank, and then I see some output in the webview console. Can you repro? 

```log
index.bundle.js:314916 [ERROR] 58:39.209 getLocalNoteGraphElements - Couldn't find vault of note config.ts, aborting link creation
ConsoleLogger._log @ index.bundle.js:314916
2index.bundle.js:314916 [WARN] 58:39.209 getLocalNoteGraphElements - Failed to link note vault/dendron.ref.links to vault/void. Most likely, this note does not exist.
ConsoleLogger._log @ index.bundle.js:314916
index.bundle.js:314916 [WARN] 58:39.209 getLocalNoteGraphElements - Failed to link note vault/dendron.ref.links to vault/void.*. Most likely, this note does not exist.
ConsoleLogger._log @ index.bundle.js:314916
index.bundle.js:314919 [INFO] 58:39.209 DendronNoteGraphView Object
index.bundle.js:314919 [INFO] 58:39.209 useGraphElements Object
index.bundle.js:314919 [INFO] 58:39.209 useGraphElements Object
index.bundle.js:314916 [INFO] 58:39.211 Graph - Rendering graph...
index.bundle.js:132389 You have set a custom wheel sensitivity.  This will make your app zoom unnaturally when using mainstream mice.  You should change this value from the default only if you can guarantee that all your users will use the same hardware and OS configuration as your current machine.
warn @ index.bundle.js:132389
index.bundle.js:269343 Error: Can not create edge `73eb67ea-0291-45e7-8f2f-193fd6f00643_Zq8Kz00hOPxCo4HdS5rdt` with nonexistant target `Zq8Kz00hOPxCo4HdS5rdt`
    at error (index.bundle.js:132371:9)
    at Collection.elesfn$u.restore (index.bundle.js:144566:11)
    at new Collection (index.bundle.js:144235:10)
    at Core.add (index.bundle.js:144974:18)
    at index.bundle.js:47636:25
    at Array.forEach (<anonymous>)
    at applyDisplayConfig (index.bundle.js:47634:237)
    at applyConfig (index.bundle.js:47645:646)
    at index.bundle.js:47645:955
    at fk (index.bundle.js:270846:19)
```

Harshita-mindfire:
> I'm hitting an issue though - it's likely a pre-existing issue with graph view. When I navigate a few links using the graph view, particularly in a circle (say A->B->C->B->A), sometimes the graph view ends up turning blank, and then I see some output in the webview console. Can you repro?
> 
Hey, yes was able to repro. I have a fix for this already.

kevinslin (changes requested):
1. where you able to get this to work on org-workspace? it wasn't loading for me.
1. noticed that we can toggle between local and full graph in the webview. is that on purpose? because the webview is titled "local graph". to give us more flexibility, can we just call this the `graphView` for the sidepanel?
1. i might have changed my settings but by default, it seemed to load the full graph instead of the local graph
1. can you also add some analytics around the panel?
- when its visible
- when its been clicked on

see https://github.com/dendronhq/dendron/blob/feat/local-graph-view/packages/plugin-core/src/features/TipOfTheDayWebview.ts#L63:L63

Harshita-mindfire:
> 1. where you able to get this to work on org-workspace? it wasn't loading for me.

Yes, does load for me
image

> 2. noticed that we can toggle between local and full graph in the webview. is that on purpose? because the webview is titled "local graph". to give us more flexibility, can we just call this the `graphView` for the sidepanel?

This shouldn't be the case. The side panel should only display local graph and there should not be any filters as well as toggle option.

hikchoi:
@kevinslin Just did a local build myself while pairing with @Harshita-mindfire. I can't repro what you are experiencing there. Might be an incomplete / old build.

packages/common-all/src/analytics.ts
```diff
@@ -154,6 +154,7 @@ export enum AppNames {
 
 export enum GraphEvents {
   GraphThemeChanged = "GraphThemeChanged",
+  LocalGraphSidePanel = "LocalGraphSidePanel",
```
jonathanyeung, line 157: Can we follow the noun + verb convention?  Given that you've named it `DendronGraphPanel.tsx`, let's keep that terminology - so how about `GraphPanelUsed` for this event name? 

We can start calling this GraphPanel (side) vs GraphView (full page webview)

packages/plugin-core/src/views/LocalGraphView.ts
```diff
@@ -0,0 +1,178 @@
...
+    Logger.info({ ctx, data: msg });
+    switch (msg.type) {
+      case GraphViewMessageEnum.onSelect: {
+        const note = this._ext.getEngine().notes[msg.data.id];
+        await new GotoNoteCommand(this._ext).execute({
+          qs: note.fname,
+          vault: note.vault,
+        });
+        AnalyticsUtils.track(GraphEvents.LocalGraphSidePanel, {
+          message: GraphViewMessageEnum.onSelect,
```
jonathanyeung, line 65: for consistency, let's try to use the same property names across different instances of this `GraphEvents.LocalGraphSidePanel` event type.  So maybe something like 

```typescript
AnalyticsUtils.track(GraphEvents.GraphPanelUsed, {
          type: "NodeClicked",
}
```

packages/plugin-core/src/views/LocalGraphView.ts
```diff
@@ -0,0 +1,178 @@
...
+              isSidePanel: true,
+            },
+            source: "vscode",
+          });
+        }
+        break;
+      }
+
+      case GraphViewMessageEnum.onLocalGraphVisible: {
+        AnalyticsUtils.track(GraphEvents.LocalGraphSidePanel, {
```
jonathanyeung, line 108: Rather than track 'visible' here, let's track it on the VSCode panel's visibility event.  `vscode.WebviewView.onDidChangeVisibility`

jonathanyeung (changes requested):

packages/common-all/src/analytics.ts
Harshita-mindfire: updated event and changed `Local Graph View` to `Graph Panel`

kevinslin:
> @kevinslin Just did a local build myself while pairing with @Harshita-mindfire. I can't repro what you are experiencing there. Might be an incomplete / old build.

it looks like this line that is causing the issue: https://github.com/dendronhq/dendron/blob/feat/local-graph-view/packages/plugin-core/src/views/GraphPanel.ts#L100:L100

if i don't have a custom theme or a graph theme set, `isSidePanel` is never set to `true`. instead of passing in `isSidePanel`, since we know this view is always going to be a side panel, why not create a `DendronSideGraphPanel` view that is just a wrapper around `DendronGraphPanel` with `isSidePanel` set to true?

packages/plugin-core/src/views/GraphPanel.ts
```diff
@@ -0,0 +1,177 @@
...
+      key: DendronTreeViewKey.GRAPH_PANEL,
+      webviewView,
+    });
+    webviewView.webview.onDidReceiveMessage(
+      this.onDidReceiveMessageHandler,
+      this
+    );
+
+    webviewView.onDidChangeVisibility(() => {
+      AnalyticsUtils.track(VSCodeEvents.GraphPanelClicked, {
```
jonathanyeung, line 56: To reduce the number of types of events, can we use the `GraphPanelUsed` event here as well, set the type property to `VisibilityChanged` and have a state property?

Similar to what I did for Backlinks Panel: 

```typescript
    backlinkTreeView.onDidChangeVisibility((e) => {
      AnalyticsUtils.track(VSCodeEvents.BacklinksPanelUsed, {
        type: "VisibilityChanged",
        state: e.visible ? "Visible" : "Collapsed",
      });
    });
```

jonathanyeung (changes requested):
One last nit on telem event naming, otherwise lgtm

SeriousBug:
Looks good to me. I have some nitpicks, but it's better if you spin these off into a followup task so this is not blocked.

- The zoom level in the graph is not maintained between note switches. So if I zoom into the graph, then switch to another note, the zoom resets.
- The zoom level was too far by default for me, which meant note names were not visible. It looks like the zoom is tied to the size of the panel, could we cap it so at minimum it's zoomed in enough that text appears?
  


- Resizing the panel doesn't seem to resize or re-zoom the view until I switch to another note and back. It would be nice if it automatically updated as the panel is resized.

packages/dendron-plugin-views/src/components/DendronSideGraphPanel.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Wrapper component around DendronGraphPanel
+ * @param props
+ * @returns DendronGraphPanel component with ide.isSidePanel set to true
+ */
+const DendronSideGraphPanel: DendronComponent = (props) => {
+  const ideDispatch = ideHooks.useIDEAppDispatch();
+  // set side panel to true
+  ideDispatch(ideSlice.actions.setIsSidePanel(true));
```
kevinslin, line 13: instead of making this a state property, can we just have this as part of the `props` for `DendronGraphPanel`? this way we avoid an unnecessary dispatch

packages/plugin-core/src/showcase/GraphPanelTip.ts
```diff
@@ -0,0 +1,32 @@
...
+export class GraphPanelTip implements IFeatureShowcaseMessage {
+  shouldShow(_displayLocation: DisplayLocation): boolean {
+    return true;
+  }
+  get showcaseEntry(): ShowcaseEntry {
+    return ShowcaseEntry.GraphPanel;
+  }
+
+  getDisplayMessage(_displayLocation: DisplayLocation): string {
+    return "Dendron has a new Graph Panel. Check it out!";
```
jonathanyeung, line 18: nit: Can we amend this message to maybe also help with side bar discoverability? 

"We're experimenting with a note graph panel in the new Dendron sidebar.  Check it out!"

jonathanyeung (approved):

PR #2903, opened 2022-05-11
enhance(workspace): Speed up workspace initialization

tma66:
**Background**
While timestamping the initialisation process, I found out that the bulk of the time was spent in `addOrUpdateParents`. This was due to the linear search component of using `_.find` with a noteprops array. By changing the signature to use a combination of `NotePropsDict` and `NoteFNamesDict`, I was able to make drastic improvements to init speed.

Before (org-workspace on local Macbook): ~50 sec
After: ~4 sec

**Changes**
1. Use `NotePropsDict` and `NoteFNamesDict` in place of `NoteProps[]` within noteParser.ts

**Future work**
Continue to scrub code for more usages of `_.find` with a noteprops array or `getNoteByFnameV5`

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-frontend/src/features/engine/slice.ts
+++ b/packages/common-frontend/src/features/engine/slice.ts
@@ -15,8 +15,6 @@ import { createLogger } from "../../utils";
 // @ts-ignore
 import internal from "@reduxjs/toolkit/node_modules/immer/dist/internal";
 
-let NOTE_FNAMES_DICT: NoteFNamesDict | undefined;
-
 /**
  * Equivalent to engine.init
  */
@@ -164,13 +162,12 @@ export const engineSlice = createSlice({
   reducers: {
     setFromInit: (state, action: PayloadAction<DEngineInitPayload>) => {
       const { notes, wsRoot, schemas, vaults, config } = action.payload;
-      NOTE_FNAMES_DICT = new NoteFNamesDict(_.values(notes));
       state.notes = notes;
       state.wsRoot = wsRoot;
       state.schemas = schemas;
       state.vaults = vaults;
       state.config = config;
-      state.noteFName = NOTE_FNAMES_DICT;
+      state.noteFName = new NoteFNamesDict(_.values(notes));
     },
     setConfig: (state, action: PayloadAction<ConfigGetPayload>) => {
       state.config = action.payload;
@@ -192,6 +189,7 @@ export const engineSlice = createSlice({
       state.schemas = {};
       state.notesRendered = {};
       state.error = null;
+      state.noteFName = new NoteFNamesDict([]);
     },
     setRenderNote: (
       state,
@@ -207,16 +205,19 @@ export const engineSlice = createSlice({
       }
       // this is a new node
       if (!state.notes[note.id]) {
-        const notesByFnameDict = _.isUndefined(NOTE_FNAMES_DICT)
-          ? new NoteFNamesDict(_.values(state.notes))
-          : NOTE_FNAMES_DICT;
-        NoteUtils.addOrUpdateParents({
+        // TODO: Support passing in fname dictionary
+        const stubs = NoteUtils.addOrUpdateParents({
           note,
           notesDict: state.notes,
-          notesByFnameDict,
           createStubs: true,
           wsRoot: state.wsRoot!,
         });
+        state.noteFName.add(note);
+        stubs
+          .filter((noteChangeEntry) => noteChangeEntry.status === "create")
+          .forEach((noteChangeEntry) =>
+            state.noteFName.add(noteChangeEntry.note)
+          );
       }
       state.notes[note.id] = note;
     },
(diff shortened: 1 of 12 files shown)


packages/engine-server/src/cache/dendronFileSystemCache.ts
```diff
@@ -31,7 +31,11 @@ export abstract class DendronFileSystemCache<T extends FileSystemCache, V>
    */
   private readFromFileSystem(): T {
     if (fs.existsSync(this._cachePath)) {
-      return fs.readJSONSync(this._cachePath) as T;
+      try {
+        return fs.readJSONSync(this._cachePath) as T;
+      } catch (_err: any) {
```
jonathanyeung, line 36: What kind of errors did you see here? Can we log it?

packages/common-frontend/src/features/engine/slice.ts
```diff
@@ -161,12 +164,13 @@ export const engineSlice = createSlice({
...
     setFromInit: (state, action: PayloadAction<DEngineInitPayload>) => {
       const { notes, wsRoot, schemas, vaults, config } = action.payload;
+      NOTE_FNAMES_DICT = new NoteFNamesDict(_.values(notes));
       state.notes = notes;
       state.wsRoot = wsRoot;
       state.schemas = schemas;
       state.vaults = vaults;
       state.config = config;
-      state.noteFName = new NoteFNamesDict(_.values(notes));
+      state.noteFName = NOTE_FNAMES_DICT;
```
jonathanyeung, line 173: I think we'll run into that similar issue I encountered w/ the GraphView optimization where some code complains about redux state not being serializable.  We should create a more primitive type for this dict like `[key: string]: string;` and use that type here (and in the graph view) instead.

Consequently, we'll also need to amend the arguments of `addOrUpdateParents`

jonathanyeung (changes requested):
50s -> 3s!  🚀 

Good catch on `_.find`, I had previously only been looking for `NoteUtils.getNoteByFnameV5()` for these kinds of optimizations.

packages/common-frontend/src/features/engine/slice.ts
kevinslin: yeah, this is why we added NOTE_FNAMES_DICT in the module instead of inside of redux. @tma66 do we still need this in state?

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -430,36 +416,39 @@ export class FileStorage implements DStore {
...
           numCacheUpdates: _.size(cacheUpdates),
         });
 
-        return notes;
+        return notesById;
       })
     );
-    const allNotes = _.flatten(out);
-    if (allNotes.length === 0) {
+    this.notes = Object.assign({}, ...out);
```
kevinslin, line 449: you can also just use `{...out}` as a more concise verison of this

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -468,9 +457,9 @@ export class FileStorage implements DStore {
       );
     }
 
-    this._addBacklinks({ notesWithLinks, allNotes });
+    this._addBacklinks({ notesWithLinks, allNotes: _.values(this.notes) });
```
kevinslin, line 460: can we cache this value (aka instantiate separately) since we deconstruct this 2x (first time in line 450)?

kevinslin (changes requested):
nice job! just a few comments

packages/engine-server/src/drivers/file/storev2.ts
tma66: Good point. fixed

packages/engine-server/src/drivers/file/storev2.ts
tma66: Hmm got type error here. FYI `out` is `NotePropsDict[]` here that I'm trying to merge into one `NotePropsDict`

packages/engine-server/src/cache/dendronFileSystemCache.ts
tma66: In case file gets corrupted or is an invalid json file. Sure I'll log it

packages/common-frontend/src/features/engine/slice.ts
tma66: Looks like its being used in graph view. https://github.com/dendronhq/dendron/pull/2903. I'd have to expose  `NOTE_FNAMES_DICT` if we wanted to remove from state. @jonathanyeung What do you think here?

jonathanyeung: As discussed offline, since this is a pre-existing issue, let's defer that to the next iteration.

jonathanyeung (approved):

SeriousBug (approved):
The speed improvement is amazing, great work! The code looks good and I didn't see anything wrong in manual testing either.

PR #2904, opened 2022-05-11
feat(navigate): Backlink Panel with Hover

jonathanyeung:
### Breaking Change: Minimum supported VSCode version bumped from `1.58.0` to `1.62.0`
_[1.62.0](https://code.visualstudio.com/updates/v1_62) was released in October 2021. This upgrade is necessary to have the hover controls support HTML._

---

Enhances the usability of the backlinks panel:
- Reduces the tree view from having 3 levels to 2 levels. Candidates and Links are no longer in separate folders, but are listed together. An icon to represent candidate vs link is inlined with each link to differentiate between the two. I did this to reduce nesting, especially since I suspect almost no one has candidate links enabled (telemetry on this incoming).
- For the filename treeview items, I moved some info around for the descriptions. Each note name entry now contains # of links, # of candidates, and updated time. There used to be a redundant fname value as the description for filename items, I've removed that.
- Adds hover to filename items. This will render a note header, followed by a list of all references in the note in separate sections, with variable amounts of context. The rendered style is the same as the per-reference items (bullet below).
- Adds hover to references. This will render the markdown with +/- 20 lines of context. It also highlights the reference line and adds some contextual line/EOF information to make it easier to get a sense of where you're at in the document.
- Sort Order is now a sticky setting. I changed the default to sort by last updated and also changed from a single button to using the ellipsis `...` control to be consistent with VS Code tree views. 

Some feedback from the team on the initial revision:

@kevinslin
- [x] make the sort order sticky when toggled
- [x] tree view, change icon or show a clear status of what the sort order is when toggled
     - Made it work the same way as vscode by moving it into the '...' ellipsis control. 
- [x] when clicking on a reference, focus text editor and add selection to the link that is doing the referencing
    - It looks like this was already working.
    
@hikchoi 
- hover isn't scrollable for long notes. In this case I don't think we should show the entire note.
    - somewhat mitigated this by changing what we render for the note level tree view items.  Not perfect, but we can iterate on it. 
- in the hover for each backlink, I notice that we are bolding the line. Can we make this more pronounced? do we have the capability to highlight it?
    - Changed to use a highlight. I tried to color the highlighter to observe the user theme, but to no avail - don't think it's possible in the current API.  Stuck with highlighter yellow.
- If we can provide multiple hovers, might be nice if we can have another hover result that tells the user what the link type is (candidate or backlink). Otherwise, we could always make it pretty with some Markdown formatting and stick it on the top or something. A nicely formatted view of the frontmatter of the hovered note item?
    - Made some enhancements that includes both of these data for note-level tree items. 
    
@Harshita-mindfire 
- The links in the "Hover Preview on Note" are clickable and do not open the correct file in the editor.
    - Fixed in a separate PR https://github.com/dendronhq/dendron/pull/2949  
- I see that scroll is not possible for preview of longer notes but it does look incomplete this way. Maybe if we add ellipses... to show that the note preview is incomplete.
    - I don't think we get any feedback from the API on when we're getting truncated.  I changed the style to no longer render the full note. 

@personjerry 
- Does it stay usable on smaller screen sizes?
    - The API here is pretty simple so it doesn't give us those capabilities like typical frontend frameworks, but I added some mitigations to reduce content length.

### Updated Screenshots:
    
image

image

image

Docs: https://github.com/dendronhq/dendron-site/pull/514
Airtable Telem Updated

**Note**: If you want to test this locally, I suggest you modify the following vscode setting. Otherwise, it'll feel sluggish - VSCode by default adds a 1.5 second delay between putting your mouse over the item and showing the tooltip in treeviews.  The hover preview in the editor is set to 300 ms, so I matched it and it feels pretty natural:

```json
"workbench.hover.delay": 300
```

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs
Updated Docs PR Link: https://github.com/dendronhq/dendron-site/pull/514
(Prev Commit was backed out: https://github.com/dendronhq/dendron-site/commit/53ad96ef94c20bd1cbe2224509e9ebb4e83af2b6)

### Extended


jonathanyeung:
> for the preview:
> 
> * depends on if you want to go through the effort but if backlinks panel supports it, consider using HTML view. see (you'll need to sync docs)
>   
>   * [[Create Note from Arbitrary Text|dendron://dendron.docs/pkg.plugin-core.dev.cook.create-note-from-arbitrary-text]]
>   * [[Render Note|dendron://dendron.docs/pkg.dendron-engine.api.engine.render-note]]

Unfortunately, it doesn't look like the hover API supports this - it only permits a `MarkdownString`, not a full blown iFrame.  There are a few other things we can do to make the view look better though.

> some other requests not related with this change:
> * make the sort order sticky when toggled
> * tree view, change icon or show a clear status of what the sort order is when toggled
> * when clicking on a reference, focus text editor and add selection to the link that is doing the referencing

👍🏼

personjerry:
Does it stay usable on smaller screen sizes?

packages/engine-server/src/markdown/remark/backlinksHoverProcessor.ts
```diff
@@ -0,0 +1,232 @@
...
+ * 2. Changes the backlink node away from a wikilink/noteref to prevent the
+ *    backlink text from being altered
+ * 3. Adds contextual " --- line # ---" information
+ * 4. Removes all elements that lie beyond the contextual lines limit of the
+ *    backlink
+ * @param this
+ * @param _opts
+ * @returns
+ */
+export function BacklinkHoverProcessor(
```
kevinslin, line 38: can we stick with the existing style? so lowerCamelCase, and without using `Processor` suffix?

packages/engine-test-utils/src/__tests__/engine-server/markdown/backlinksPanelHover.spec.ts
```diff
@@ -0,0 +1,690 @@
...
+  backlinkHoverOpts,
+  match,
+}: {
+  testName: string;
+  noteBody: string;
+  backlinkHoverOpts: BacklinkOpts;
+  match: string[];
+}) {
+  return createProcTests({
+    name: testName,
```
kevinslin, line 29: ah, this is deprecated. can you mark this as such and use `createProcCompileTests` instead? 

see sample usage here > https://github.com/dendronhq/dendron/blob/enhance/backlinks-panel-hover/packages/engine-test-utils/src/__tests__/engine-server/markdown/v5/hierarchies.spec.ts#L7:L7

packages/plugin-core/src/features/BacklinksTreeDataProvider.ts
```diff
@@ -54,7 +68,11 @@ export default class BacklinksTreeDataProvider
     isLinkCandidateEnabled: boolean | undefined
   ) {
     this._isLinkCandidateEnabled = isLinkCandidateEnabled;
-    this.updateSortOrder(BacklinkSortOrder.PathNames);
+
+    // Set default sort order to use last updated
+    this.SortOrder =
```
kevinslin, line 73: nitpick. js properties generally aren't capitalized

kevinslin (commented):
excited for this!  had a few nitpicks as well as some additional comments below

- hover feels a little long still. can we shorten it?

- can you add docs? should go in [[Backlinks|dendron://dendron.dendron-site/dendron.topic.sidebar.backlinks]]
    - would list out functionality under `### Features`, see [[Features|dendron://dendron.dendron-site/dendron.topic.templates#features]] as an example
    - add screenshots + loom videos of hover actions, see [[Screenshots and Videos|dendron://dendron.handbook/area.docs.sop.screenshots-and-videos]]

packages/engine-test-utils/src/__tests__/engine-server/markdown/backlinksPanelHover.spec.ts
jonathanyeung: Sounds good.  Next time, can we more proactively tag old methods with `@deprecated`?

jonathanyeung:
> * hover feels a little long still. can we shorten it?

Reduced all context line counts by about half.

> * can you add docs? should go in [[Backlinks|dendron://dendron.dendron-site/dendron.topic.sidebar.backlinks]]

https://github.com/dendronhq/dendron-site/commit/53ad96ef94c20bd1cbe2224509e9ebb4e83af2b6

kevinslin (approved):
left some comments on docs, otherwise lgtm

PR #2915, opened 2022-05-12
feat(workspace): Add a command to migrate existing regular vaults into self contained vaults

SeriousBug:
This PR adds a new command which can migrate existing vaults into self contained vaults. The command only supports regular vaults, no seed or workspace vaults, which can be supported on a later iteration. The prompt will only display vaults that can be migrated.

Docs: https://github.com/dendronhq/dendron-site/pull/498

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- [x] If you are adding a new language feature (graphically visible in VS Code/preview/publishing), an example is included in the [test workspace](https://wiki.dendron.so/notes/dtMsF12SF2SUhLN10sYe2.html)

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended


packages/plugin-core/package.json
```diff
@@ -41,21 +41,6 @@
         "configuration": "./language-configuration.json"
       }
     ],
-    "viewsWelcome": [
```
kevinslin, line 44: it looks like the new views weren't added to the config/constants. i know this isn't part of this feature but could you add them to the `gen:config` script?

packages/plugin-core/src/commands/MigrateSelfContainedVault.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+    return undefined;
+  }
+
+  constructor(ext: IDendronExtension) {
+    super();
+    this.extension = ext;
+  }
+
+  async gatherInputs(opts?: CommandOpts): Promise<CommandOpts | undefined> {
```
kevinslin, line 39: since this is a destructive operation, can we warn folks to backup all their notes before doing this?

packages/engine-server/src/workspace/service.ts
```diff
@@ -467,6 +468,53 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+    }
+    const newVault: SelfContainedVault = { ...vault, selfContained: true };
+
+    const vaultFolder = path.join(this.wsRoot, vault.fsPath);
+    const tmpFolder = `${vault.fsPath}-migrate-${genUUID()}`;
+    // Rename the vault temporarily, create a new folder with the same name,
+    // then move the vault into the `notes` folder inside the new folder. This
+    // is the fastest and most reliable way to move all notes and assets into
+    // the new folder, since we don't have to move each file individually.
+    await fs.move(vaultFolder, tmpFolder);
```
kevinslin, line 488: does this need to be prefixed? (eg. `wsRoot/tmpFolder` or `tmpdir()/tmpfolder`?) getting this error https://www.loom.com/i/028126cbf1e942baad2387fd4e45973d

packages/engine-server/src/workspace/service.ts
```diff
@@ -467,6 +468,53 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+    await this.createSelfContainedVault({
+      addToCodeWorkspace: false,
+      addToConfig: false,
+      vault: newVault,
+    });
+    // Except if there's `.git` inside the vault, that has to stay at the root of the vault
+    if (await fs.pathExists(path.join(vaultFolder, FOLDERS.NOTES, ".git"))) {
+      await fs.move(
+        path.join(vaultFolder, FOLDERS.NOTES, ".git"),
+        path.join(vaultFolder, ".git")
```
kevinslin, line 510: could there be a .git in the desitnation that gets overridden if we do this?

kevinslin (changes requested):
some comments on safeguards and questions

packages/engine-server/src/workspace/service.ts
SeriousBug: No because that's the new folder we just created, but I'll add some checks to make the whole code more resilient in case something does go wrong.

packages/engine-server/src/workspace/service.ts
SeriousBug: Yup, thanks!

kevinslin:
looks like a build error

packages/plugin-core/package.json
```diff
@@ -64,18 +64,12 @@
...
-        {
-          "id": "dendron.tip-of-the-day",
-          "name": "Tip of the Day",
-          "when": "dendron:pluginActive",
-          "type": "webview"
         },
         {
-          "id": "dendron.backlinks",
-          "name": "Backlinks",
+          "id": "dendron.calendar-view",
```
kevinslin, line 69: heads up that this should be in the `dendron-view`, not in the `explorer`. currently, `gen:config` doesn't account for this so it will need to be updated

kevinslin (changes requested):

packages/plugin-core/package.json
SeriousBug: `gen:config` is generating them correctly now after merging master. I had to correct some more things but I think it's good now.

kevinslin:
- there's an issue with how we do the migration. a regular workspace has the following setup
```
- Dendron
    - dendron.yml
    - vault
        - note1.md
```
- when we migrate the vault, we end up with the following configuration
```
- Dendron
    - dendron.yml
    - vault
        - dendron.yml
        - dendron.code-workspace
        - notes
            - note1.md
```

this makes sense if the user has multiple vaults but is confusing in the common case (user has a single vault to their workspace). for the sake of iteration, lets launch it but also highlight it in the docs. can we also have a followup to automatically take care of the common case of a workspace with a single vault?

kevinslin:
this looks good. can you merge and follow the new merge process in [[Async Update to Pr Merge Process|dendron://private/user.kevin.journal.2022.05.20.m.async-update-to-pr-merge-process]]?

also, this command introduces a new analytics point (the MigrateSelfContainedVaults command). can you add this to upcoming metrics in airtable?

PR #2916, opened 2022-05-12
chore: added telemetry for toggle graph view

Harshita-mindfire:
This PR aims to add some telemetry around the toggle of graph view and the default graph theme saved by user

### Basics

### Extended
- General
- Operations
- Architecture

### Basics
TODO: update Airtable

### Extended

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended
NA

## Docs
NA

### Extended
NA

## Telemetry
Added a property `defaultTheme` for an existing graph event `GraphThemeChanged`
Added a GraphEvent `GraphViewUsed`.
Added a property `state: string` and `type: string` for event `GraphViewUsed`.

### DIFF
--- a/packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
+++ b/packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
@@ -173,6 +173,13 @@ export class NoteGraphPanelFactory {
             });
             break;
           }
+
+          case GraphViewMessageEnum.toggleGraphView: {
+            AnalyticsUtils.track(DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key, {
+              type: msg.data.graphType,
+            });
+            break;
+          }
           default:
             break;
         }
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
```diff
@@ -173,6 +173,13 @@ export class NoteGraphPanelFactory {
             });
             break;
           }
+
+          case GraphViewMessageEnum.toggleGraphView: {
+            AnalyticsUtils.track(DENDRON_COMMANDS.SHOW_NOTE_GRAPH.key, {
```
jonathanyeung, line 178: Instead of conflating with the command telemetry, can we use the same 'GraphViewUsed' ("view
 here, not "panel") convention as with the other review?
 
 type: "GraphTypeChanged"
 state: "Local" / "Full"

jonathanyeung (changes requested):

jonathanyeung (approved):

PR #2918, opened 2022-05-12
chore: upgrade jest to 28.0

kevinslin:
chore: upgrade jest to 28.0

Changes:
- upgrade jest from v26 to v28
- consolidate jest dependencies as a workspace dependency (vs dependency of each individual package)
- update jest scripts to run from workspace 
- fix issues with tests being both async and calling `done`
- remove slow and un-used test
- remove un-used dependencies from children

Out of scope:
- to timebox, did not touch jest dependencies from the following packages:
    - nextjs-template
    - dendron-plugin-views

DX Changes:
- changes where the watch task and debug task are being run from (`root` of workspace instead of `engine-test-utils`) (will update in [[dev]] when changes have been mergged)

Context:
- this started off as a task to experiment with [mdx2](https://mdxjs.com/blog/v2/) support in dendron which would make it possible to extend publishing (we currently embed raw html for published pages which make modifying the outputted HTML extremely difficult)
- in order to support mdx2, we need to upgrade `unified` to `+v13` as they did a rewrite of their parsing engine (supposed to be much faster)
    - we had a backlog item to do this as our existing version was several major versions out of date and prevented us from updating any of our plugins
- the reason we did not do this originally was because `> 13` `unified` were `esm` only modules and did not play well with `jest` 
- `jest` and specifically `ts-jest` has better support for `esm` at the latest version, hence pivotting to upgrade jest before tackling upgrading unified


SeriousBug:
VSCode still doesn't support ESM, so just a heads up we still can't upgrade unified until they do.

I'm also getting the following error when I try to run the tests (using the `Test (bootstrap)` task). Is there a different way to run the tests now?

```
FAIL src/__tests__/site.spec.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    SyntaxError: /home/kaan/Projects/Dendron/dendron/packages/engine-test-utils/src/__tests__/site.spec.ts: Unexpected token (30:33)

      28 | import { ENGINE_HOOKS, ENGINE_HOOKS_MULTI } from "../presets";
      29 |
    > 30 | const basicSetup = (preSetupHook?: SetupHookFunction) => ({
         |                                  ^
      31 |   createEngine: createEngineFromEngine,
      32 |   expect,
      33 |   preSetupHook: async (opts: WorkspaceOpts) => {

      at instantiate (../../node_modules/jest-cli/node_modules/@babel/parser/src/parse-error/credentials.js:61:22)
```

jonathanyeung (changes requested):

kevinslin:
> VSCode still doesn't support ESM, so just a heads up we still can't upgrade unified until they do.
> 
> I'm also getting the following error when I try to run the tests (using the `Test (bootstrap)` task). Is there a different way to run the tests now? I tried cleaning the entire workspace and starting from scratch (`git clean -dfx; yarn bootstrap:bootstrap; yarn bootstrap:build`) but the error continues.
> 
> ```
> FAIL src/__tests__/site.spec.ts
>   ● Test suite failed to run
> 
>     Jest encountered an unexpected token
> 
>     Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
> 
>     Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.
> 
>     By default "node_modules" folder is ignored by transformers.
> 
>     Here's what you can do:
>      • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
>      • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
>      • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
>      • If you need a custom transformation specify a "transform" option in your config.
>      • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.
> 
>     You'll find more details and examples of these config options in the docs:
>     https://jestjs.io/docs/configuration
>     For information about custom transformations, see:
>     https://jestjs.io/docs/code-transformation
> 
>     Details:
> 
>     SyntaxError: /home/kaan/Projects/Dendron/dendron/packages/engine-test-utils/src/__tests__/site.spec.ts: Unexpected token (30:33)
> 
>       28 | import { ENGINE_HOOKS, ENGINE_HOOKS_MULTI } from "../presets";
>       29 |
>     > 30 | const basicSetup = (preSetupHook?: SetupHookFunction) => ({
>          |                                  ^
>       31 |   createEngine: createEngineFromEngine,
>       32 |   expect,
>       33 |   preSetupHook: async (opts: WorkspaceOpts) => {
> 
>       at instantiate (../../node_modules/jest-cli/node_modules/@babel/parser/src/parse-error/credentials.js:61:22)
> ```

yep, this has been changed. thanks for the catch. updated instruction is 

```
cd $MONOREPO_ROOT
yarn ci:test:cli
```

jonathanyeung (approved):

PR #2920, opened 2022-05-12
chore: end the upgrade toast wording AB test

SeriousBug:
This PR ends the AB test for the upgrade toast wording. After several weeks of data, it appears that "see what's new" is a clear winner.

The chart below is the cumulative number of unique users who clicked the button after an upgrade. Over several weeks "see what's new" maintained a clear lead over the other options, so this PR sets that as the default and ends the experiment.

Users who looked at the upgrade changelog

### DIFF
--- a/packages/common-all/src/abTests.ts
+++ b/packages/common-all/src/abTests.ts
@@ -2,15 +2,6 @@
 import { ABTest } from "./abTesting";
 import { GraphThemeEnum } from "./types/typesv2";
 
-export enum UpgradeToastWordingTestGroups {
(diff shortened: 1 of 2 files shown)


jonathanyeung (approved):
Nice! @SeriousBug - when you get a chance, would you mind backfilling a dendron note / airtable entry to record this experiment as per [[Code - Setup an AB Test|dendron://private/area.metrics.sop.ab-testing#code---setup-an-ab-test]]

Doesn't need to be that detailed.  Thanks!

kevinslin (approved):

PR #2929, opened 2022-05-17
chore(internal): Refactor notepropsfnamedict to primitive typing

tma66:
**Background**
According to their official docs, Redux highly discourages use of non-serializable types (this includes classes and maps/sets). To bring parity between the redux state and storev2 state, we should use the same primitive javascript object typing.

In addition, `getNotesByFnameFromEngine` currently returns a reference to notes in the engine state. This means that any changes to the notes will affect the engine state. To prevent this side effect, we will return a copy instead.

**Major Changes**
1. Change type of `storev2.noteFnames` from NoteFNamesDict class object to NotePropsByFnameDict Object
2. Create `NoteFullDictUtils` class with static methods to work with both `NotePropsByFnameDict` and `NotePropsDict`. This will be used to add/update/delete from both dictionaries at same time
3. `getNotesByFnameFromEngine` will return a DEEP COPY instead of reference of notes from the engine. 
4. Update `DStore` api to include `getNote` and `findNotes`. These will be called by the engine to access notes from the store

**Minor Changes**
1. Move usages of `getNoteOrThrow` to `getNoteByFnameFromEngine` + throw themselves. getNoteOrThrow takes linear time to search
2. When DoctorService regenerates id, old note should be deleted
3. Minor bug fixes in store where we didn't keep track of all the NoteChangedEntry
4. Remove hydration step from `updateNote`. Notes should be hydrated on the client/engine side. This can lead to bugs where it removes parents/children from newNote if oldNote doesn't have.

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/notePropDict.ts
+++ b/packages/common-all/src/notePropDict.ts
@@ -1,0 +1,134 @@
+import _ from "lodash";
+import {
+  NotePropsDict,
+  NotePropsByFnameDict,
+  DVault,
+  NoteProps,
+  NotePropsFullDict,
+} from "./types";
+import { cleanName, isNotUndefined } from "./utils";
+import { VaultUtils } from "./vault";
+
+/**
+ * Utilities for working with NotePropsFullDict. The reason NotePropsFullDict is not a class is due to needing
+ * to work with primitive objects with redux
+ */
+export class NoteFullDictUtils {
+  /**
+   * Find notes by fname. If vault is provided, filter results so that only notes with matching vault is returned
+   * Return empty array if no match is found
+   *
+   * @param fname
+   * @param notesDict
+   * @param vault If provided, use to filter results
+   * @returns Copy of NoteProps array
+   */
+  static findNotesByFname(
+    fname: string,
+    notesDict: NotePropsFullDict,
+    vault?: DVault
+  ): NoteProps[] {
+    const { notesById, notesByFname } = notesDict;
+    const cleanedFname = cleanName(fname);
+    const ids: string[] | undefined = notesByFname[cleanedFname];
+    if (!ids) {
+      return [];
+    }
+    let notes = ids.map((id) => notesById[id]).filter(isNotUndefined);
+    if (vault) {
+      notes = notes.filter((note) => VaultUtils.isEqualV2(note.vault, vault));
+    }
+    return _.cloneDeep(notes);
+  }
+
+  /**
+   * Remove note from both notesById and notesByFname
+   *
+   * @param note note to delete
+   * @param notesDict
+   */
+  static deleteNote(note: NoteProps, notesDict: NotePropsFullDict) {
+    const { notesById, notesByFname } = notesDict;
+    delete notesById[note.id];
+    NoteFnameDictUtils.deleteNote(note, notesByFname);
+  }
+
+  /**
+   * Add note to notesById and notesByFname.
+   * If note id already exists, check to see if it corresponds to same note by fname.
+   * If fname match, then we only need to update notesById. If fname doesn't match, remove old id from notesByFname first before updating both.
+   *
+   * Otherwise, if note id doesn't exist, add to both dictionaries
+   *
+   * @param note to add
+   * @returns
+   */
+  static addNote(note: NoteProps, notesDict: NotePropsFullDict) {
+    const { notesById, notesByFname } = notesDict;
+    const maybeNote = notesById[note.id];
+    if (maybeNote) {
+      if (cleanName(maybeNote.fname) === cleanName(note.fname)) {
+        notesById[note.id] = note;
+        return;
+      } else {
+        // Remove old fname from fname dict
+        NoteFnameDictUtils.deleteNote(maybeNote, notesByFname);
+      }
+    }
+    notesById[note.id] = note;
+    NoteFnameDictUtils.addNote(note, notesByFname);
+  }
+}
+
+/**
+ * Utilities for working with NotePropsByFnameDict.
+ */
+export class NoteFnameDictUtils {
+  /**
+   * Create a map of {key -> value} where key = noteFname and value is list of ids corresponding to that fname from NotePropsDict
+   *
+   * @param notesById NotePropsDict used to populate map
+   * @returns
+   */
+  static create(notesById: NotePropsDict): NotePropsByFnameDict {
+    const notesByFname = {};
+    _.values(notesById).forEach((note) =>
+      NoteFnameDictUtils.addNote(note, notesByFname)
+    );
+    return notesByFname;
+  }
+
+  /**
+   * Add note to notesByFname dictionary. If note fname exists, add note id to existing list of ids
+   *
+   * @param note to add
+   * @param notesByFname dictionary to modify
+   */
+  static addNote(note: NoteProps, notesByFname: NotePropsByFnameDict) {
+    const fname = cleanName(note.fname);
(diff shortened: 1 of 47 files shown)


packages/common-all/src/types/typesv2.ts
```diff
@@ -169,10 +170,25 @@ export type DNodePropsDict = {
...
 };
 
+/**
+ * Map of noteFname -> list of noteIds. Since fname is not unique across vaults, there can be multiple ids with the same fname
+ */
+export type NotePropsByFnameDict = {
+  [key: string]: string[];
+};
+
+export type NotePropsFullDict = {
```
kevinslin, line 187: name is a little obtuse, especially since this doesn't return `NoteProps` by a dictionary instead. rename to `NoteDicts` instead?

packages/common-all/src/notePropDict.ts
```diff
@@ -0,0 +1,134 @@
...
+  NotePropsFullDict,
+} from "./types";
+import { cleanName, isNotUndefined } from "./utils";
+import { VaultUtils } from "./vault";
+
+/**
+ * Utilities for working with NotePropsFullDict. The reason NotePropsFullDict is not a class is due to needing
+ * to work with primitive objects with redux
+ */
+export class NoteFullDictUtils {
```
kevinslin, line 16: rename to NoteDictsUtils

packages/common-all/src/notePropDict.ts
```diff
@@ -0,0 +1,134 @@
...
+   * Return empty array if no match is found
+   *
+   * @param fname
+   * @param notesDict
+   * @param vault If provided, use to filter results
+   * @returns Copy of NoteProps array
+   */
+  static findNotesByFname(
+    fname: string,
+    notesDict: NotePropsFullDict,
```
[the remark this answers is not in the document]
kevinslin: also apply to subsequent methods

packages/common-all/src/notePropDict.ts
```diff
@@ -0,0 +1,134 @@
...
+    return _.cloneDeep(notes);
+  }
+
+  /**
+   * Remove note from both notesById and notesByFname
+   *
+   * @param note note to delete
+   * @param notesDict
+   */
+  static deleteNote(note: NoteProps, notesDict: NotePropsFullDict) {
```
kevinslin, line 50: should we return an error if the key isn't present?

packages/common-all/src/notePropDict.ts
```diff
@@ -0,0 +1,134 @@
...
+ * Utilities for working with NotePropsByFnameDict.
+ */
+export class NoteFnameDictUtils {
+  /**
+   * Create a map of {key -> value} where key = noteFname and value is list of ids corresponding to that fname from NotePropsDict
+   *
+   * @param notesById NotePropsDict used to populate map
+   * @returns
+   */
+  static create(notesById: NotePropsDict): NotePropsByFnameDict {
```
kevinslin, line 93: this might be a big refactor but for consistency, can we rename this to `NotePropsByIdDict`?

packages/engine-test-utils/src/presets/engine-server/rename.ts
```diff
@@ -1599,7 +1599,7 @@ const NOTES = {
           expected: "foo1",
         },
         {
-          actual: changedEntries && changedEntries.length === 5,
+          actual: changedEntries && changedEntries.length === 6,
```
kevinslin, line 1602: curiosu - why did this change?

kevinslin (changes requested):
nice job! hopefully this should drive down the corner cases and engine being out of sync issues. left comments, mostly around conventions and naming

packages/common-all/src/types/typesv2.ts
SeriousBug: I'd also recommend a docstring that links to the utils class to remind anyone using this that they should be accessing through that class.

packages/engine-server/src/drivers/file/noteParser.ts
```diff
@@ -214,21 +218,22 @@ export class NoteParser extends ParserBase {
                 hash: resp.noteHash,
               });
             }
+
+            // Check for duplicate IDs when adding new note to the map
+            if (notesById[notes[0].id] !== undefined) {
```
SeriousBug, line 223: The duplicate note ID check is now only happening for the first note in `notes`, is there any reason for that?

packages/engine-server/src/doctor/service.ts
```diff
@@ -380,7 +380,8 @@ export class DoctorService implements Disposable {
           note.id = genUUID();
           await engine.writeNote(note, {
             runHooks: false,
-            updateExisting: true,
+            // Old note needs to be removed
```
jonathanyeung, line 383: Is this change related to this PR?

jonathanyeung (changes requested):

jonathanyeung:
Can you also go ahead and add an `@deprecated` tag to `NoteUtils.getNoteOrThrow()`?

packages/common-all/src/types/typesv2.ts
tma66: Fixed

packages/common-all/src/notePropDict.ts
tma66: Looks like its fairly standard to return a boolean instead to denote whether it was deleted or it. I can return that instead. We can let the clients construct the error (eg 404) if needed

packages/common-all/src/notePropDict.ts
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Find notes by fname. If vault is provided, filter results so that only notes with matching vault is returned
+   * Return empty array if no match is found
+   *
+   * @param fname
+   * @param notesDict
+   * @param vault If provided, use to filter results
+   * @returns Copy of NoteProps array
+   */
+  static findNotesByFname(
```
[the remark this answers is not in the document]
tma66: sure

packages/engine-test-utils/src/presets/engine-server/rename.ts
tma66: Previously we were missing the changeEntry where root.children got updated from "[foo]" to "[]" before foo got deleted.

In the diff, I added a couple places where we forgot to propagate all the changedEntries.

packages/engine-server/src/drivers/file/noteParser.ts
tma66: `resp.propsList` now returns the note we are trying to create + all the parents that were updated. I wrote this with the idea that only the first note should matter for duplicate check but now that I think about it, its possible that stubs could also have duplicate ids. Let me rework this.

packages/engine-server/src/doctor/service.ts
tma66: Needed to make this change to make one of the tests to pass which expected note ids to change.
In addition, before this change, the old note would still exist in the engine.

SeriousBug (approved):

jonathanyeung (approved):

kevinslin (approved):

PR #2930, opened 2022-05-17
fix: fix corrupt .dendron.ws version to unblock activation

jonathanyeung:
## fix: fix corrupt .dendron.ws version to unblock activation

Some users have a corrupt value for 'version' in their `.dendron.ws` file if they ran the CLI.  This would cause their activation to fail.

Related: https://github.com/dendronhq/dendron/pull/2871

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -680,6 +680,8 @@ export async function _activate(
       const wsService = new WorkspaceService({ wsRoot });
       let previousWorkspaceVersionFromWSService = wsService.getMeta().version;
       if (
+        // Fix a temporary issue where CLI was writing an invalid version number to .dendron.ws:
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/_extension.ts
```diff
@@ -680,6 +680,8 @@ export async function _activate(
       const wsService = new WorkspaceService({ wsRoot });
       let previousWorkspaceVersionFromWSService = wsService.getMeta().version;
       if (
+        // Fix a temporary issue where CLI was writing an invalid version number to .dendron.ws:
+        !semver.valid(previousWorkspaceVersionFromWSService) ||
```
kevinslin, line 684: there is a chance that previousWorkspaceVersionFromState is undefined. can we do a separate if block and just write the previous version to be equal to `currentVersion`?

PR #2931, opened 2022-05-17
feat: display task note status when linking to task notes in publishing and in preview

SeriousBug:
Displays the task note status when linking to task notes, both in the preview and in publishing.

Docs: https://github.com/dendronhq/dendron-site/pull/505

### DIFF
--- a/packages/engine-server/src/markdown/remark/dendronPub.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPub.ts
@@ -520,14 +520,22 @@ function linkExtras({
       taskConfig,
     });
     if (status) {
-      before.push({
+      const checkbox: { [key: string]: any } = {
         type: "element",
-        tagName: "span",
+        tagName: "input",
         properties: {
+          type: "checkbox",
+          disabled: true,
           className: "task-before task-status",
         },
-        children: [span("task-status-text", status)],
-      });
+      };
+      if (TaskNoteUtils.isTaskComplete({ note, taskConfig })) {
+        checkbox.properties.checked = true;
+      } else if (status.trim().length > 0) {
+        checkbox.children = [span("task-status-text", `(${status})`)];
+      }
+
+      before.push(checkbox);
     }
     const priority = TaskNoteUtils.getPrioritySymbol({
       note,
--- a/packages/engine-server/src/markdown/remark/dendronPub.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPub.ts
@@ -2,6 +2,7 @@ import {
   ConfigUtils,
   DendronError,
   ERROR_SEVERITY,
+  IntermediateDendronConfig,
   isNotUndefined,
   isWebUri,
   NoteProps,
@@ -9,6 +10,7 @@ import {
   ProcFlavor,
   StatusCodes,
   TAGS_HIERARCHY,
+  TaskNoteUtils,
 } from "@dendronhq/common-all";
 import _ from "lodash";
 import type { Image, Root } from "mdast";
@@ -322,6 +324,9 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
         // for rehype
         //_node.value = newValue;
         //_node.value = alias;
+
+        const { before, after } = linkExtras({ note, config });
+
         _node.data = {
           vaultName: data.vaultName,
           alias,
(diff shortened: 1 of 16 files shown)


packages/engine-server/src/markdown/remark/dendronPub.ts
```diff
@@ -491,5 +498,66 @@ function plugin(this: Unified.Processor, opts?: PluginOpts): Transformer {
...
+  const after = [];
+
+  // For task notes, add the status, priority, due, and owner info
+  if (note && TaskNoteUtils.isTaskNote(note)) {
+    const taskConfig = ConfigUtils.getTask(config);
+    const status = TaskNoteUtils.getStatusSymbolRaw({
+      note,
+      taskConfig,
+    });
+    if (status) {
```
kevinslin, line 522: since we already have css that styles checkboxes, instead of adding additional styles, can we convert them to the checkbox instead? 

when i add the following in the preview `- [ ] foo`, I get the following HTML:

```html
<ul class="contains-task-list">
    <li class="task-list-item">
    <input type="checkbox" disabled=""> foo bar</li>
</ul>
```

we can add the same html for the case when a task has no status or the status is marked as done. some caveats here:

- it appears that in gfm, checkboxes are only valid inside a bulleted list
    ```md
    [ ] this is not a task and will be rendered plainly
    - [ ] this is a task and will be rendered as a checkbox
    ```
    - if we wanted to follow this spec, we would need to inspect opening and closing `ul` tags (there's some extra complications here because users can intermix reegular bullet points with task bullet points)
- an 80/20 approach here could be to generate the following 
    ```html
    <div>
    <input type="checkbox" disabled=""> foo bar</li>
    </div>
    ```
    - this will cause the chexbox to force a line break and interacts well with lists

given this, I would propose the following:
- add the `div` based `checkbox` for tasks
- if a task is done, add a checked checkbox 
    - for now, consider a task checked if the status is set to `done` or `x` 
        - as a followup, for task configuration, we should be able to map `status` symbol as a `completion` symbol
- if a task has a status that is not `done`, add the status in brackets next to it `<input type="checkbox" disabled=""> [status] foo bar</li>`
- if a task has no status, add a empty disabled checkbox

SeriousBug: I went ahead and added some configuration for people to be able to define what a complete task is, and switched to using a checkbox instead.

- A status that is an empty string (or whitespace) will be displayed with an unchecked box
- A status that matches a state in the `taskCompleteStatus` will be displayed with a checked box
- Any other status will be displayed with an unchecked box, followed by the status in parenthesis.

I think this is a pretty clean look. The only caveat is that it looks like you can't really style disabled checkboxes, the workaround I've seen suggested is replicating a checkbox with HTML&CSS if you want disabled checkboxes with custom styles. That means the checkboxes have a dark color.

kevinslin:
looks good. a few notices:
- this will be a followup task but if you change the task status and go back to the note that the task was in, the preview won't change (this is because nothing about the note has changed)
    - we'll need to update the preview cache logic (if any of the notes referenced by the note has changed, update the preview)
- why do tasks have underline?
- tasks right now are wrapped in `span` instead of `div`. this means a task note by itself won't force a newline
    - the following
    ```md
    - foo
    [[task.some-task-note]]
    ```
    - results in html that looks like the following
    ```
    - foo [ ] some task note
    ```
    - should be okay and if anything, make tasks notes a bit more versatile than gfm checkboxes. would call out in documentation
- can you add docs?
- also a followup task - we should track in startup, number of task notes. don't think we currently calculate this. also in engine events when task notes get created

SeriousBug:
@kevinslin Tasks have underlines because they are links (including the check boxes). I could rework it so the checkbox and status parts are not inside the link, should I do that?

I added docs (linked in PR body), and I'm creating followup tasks for the preview updates and telemetry. We can see the usage of "Create Task Note" already for how often people are creating task notes though, it looks like a little over 2% of users create task notes, creating about 60 task notes a week: https://analytics.amplitude.com/dendron/chart/m02fanh?source=redirect%3A%20chart%20saved

jonathanyeung:
I think we should also add an option to allow users to disable this rendering.  

`TaskNoteUtils.isTaskNote` defines a task note as having any one of the TaskNoteProps fields in the frontmatter.  It's quite possible that users may have included such fields on their own without using `Create Task Note` or without meaning the note to be a "task" (for example, the 'status' field can be used in many different scenarios).

SeriousBug:
Good catch on the config, added one so people can disable it. I'm also updating the docs for this option.

kevinslin:
test failure - looks like the initial config check is failing

kevinslin (approved):

kevinslin:
looks good, feel free to merge

jonathanyeung (approved):

PR #2936, opened 2022-05-17
fix: block anchor after table crashes preview

SeriousBug:
A block anchor located after a table would cause the preview to crash (and likely publishing too, although I did not test if it does). This fix solves the issue.

We also decided to standardize on supporting block anchors after tables and not next to them. This works around a bug with the table parser, but also is a best practice since formatters and markdown table extensions otherwise misinterpret block anchors.

Co-authored-by: Kevin Lin <kevin@dendron.so>


kevinslin (approved):
looks good. can you also submit doc changes to block anchors calling out that it can't be used in a table as well as update examples of it? otherwise, feel free to merge and add to ready to ship

SeriousBug:
>  can you also submit doc changes

Already did so onto `dev`, I replaced "bad" examples and added a warning. Thanks for the reminder!

Issue #2943, opened 2022-05-19
enableHierarchyDisplay problem


wupingju:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [x] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

I’ve tried to set `enableHierarchyDisplay` to false but my site still shows the Children section.

## To Reproduce

1. My codes are here: https://github.com/wupingju/pinchlime-dendron/blob/master/dendron.yml
2. This is my site: https://wiki.pinchlime.com/ 

## Expected behavior

If I set `enableHierarchyDisplay` to false, my site shouldn't show the Children section.

## Screenshots

https://media.discordapp.net/attachments/735365126227493004/976479018360639559/CleanShot-2022-05-18-21-38-08.png

## Desktop (please complete the following information)

- OS: macOS
- Browser (if with published websites; include version): Sidekick / Safari
- Versions (all that apply to bug):
  - Dendron Extension: Latest
  - VS Code version: Latest
  - Dendron CLI: 

## Dendron Log file

(I'll attach this file when I go home)

Harshita-mindfire:
Thank @wupingju for reporting the bug. We'll let you know when this gets fixed 🙏

Harshita-mindfire:
This should be fixed with Dendron v0.97. Feel free to comment here if that is not the case.

Harshita-mindfire:
Hey @PJ Wu#4806 What is the dendron-cli version you have? 
Also can you try uncommenting this and see if that works https://github.com/wupingju/pinchlime-dendron/blob/master/dendron-publish-site.sh#L14?

wupingju:
Hi @Harshita-mindfire 

I've tried to delete
# yarn add @dendronhq/dendron-cli@latest

but still the same~

I don't know how to check my dendron-cli version (don't remember if I've installed this), could you please tell me how to check?

wupingju:
These are deploy logs , don't know if this will help.

[...]

Issue #2824, opened 2022-04-26
Publishing: Change the naming of published 'Children' pages to something that is intuitive to non-dendronite


nickolay-kondratyev:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [ ] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup - Dendron's Lookup Command
- [ ] Views - Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [x] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

I showed a published Dendron website to one of my friends, and the first things she said "`Children` what do you mean by children?" (Referring that she thought it was literal 'Children' meaning in her mind instead of Child Notes) 

## Describe the solution you'd like ##

Rename "Children" heading for child notes to something that is intuitive to non Dendron initiate :)

Something like "Sub pages" comes to mind. 

## Describe alternatives you've considered ##

Alternative modify the output of publishing.

SR--:
I agree, causes confusion here too. It is especially confusing if the note content deals with children. :)

Harshita-mindfire:
Hey @nickolay-kondratyev , This is configurable with the `hierarchyDisplayTitle` config under publishing namespace.
Click [here](https://wiki.dendron.so/notes/fJtzFwKpMU9hrYBBvwTcY/) for more info.

PR #2946, opened 2022-05-19
fix: publishing config to respect `enableHierarchyDisplay`

Harshita-mindfire:
Related issue: https://github.com/dendronhq/dendron/issues/2943
Also resolves these enhancements: https://github.com/dendronhq/dendron/issues/581 https://github.com/dendronhq/dendron/issues/2824

Co-authored by : kevin

### Basics

### Extended
- General
- Operations
- Architecture

### Basics
NA

### Extended
NA
### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
NA

### Extended


packages/engine-server/src/markdown/remark/hierarchies.ts
```diff
@@ -65,15 +65,22 @@ function footnoteDef2html(definition: FootnoteDefinition) {
...
+  let hierarchyDisplay = config?.hierarchyDisplay;
+
+  if (MDUtilsV5.shouldApplyPublishingRules(proc)) {
+    hierarchyDisplay =
+      ConfigUtils.getHierarchyDisplayConfigForPublishing(
+        config
+      ).hierarchyDisplay;
+    hierarchyDisplayTitle =
+      ConfigUtils.getHierarchyDisplayConfigForPublishing(config)
+        .hierarchyDisplayTitle || "Children";
```
SeriousBug, line 81: This is a little confusing, does `config` we get from `getProcData` meant to override what's set in the configuration, or does the publishing config override whatever is passed in? Why do we have the `|| "Children"` fallback twice? Making this clearer might be worth it even if it's a bit more verbose.

SeriousBug (commented):
This is supposed to disable the display of `Children: ...` and `Backlinks: ...` at the end of published notes right? I tried it out in the test-workspace by exporting it and using the nextjs-template, but it doesn't seem to be disabling it for me. Maybe I'm misunderstanding it or doing something wrong, can you confirm with manual testing that this works in publishing?

Harshita-mindfire:
> This is supposed to disable the display of `Children: ...` and `Backlinks: ...` at the end of published notes right? I tried it out in the test-workspace by exporting it and using the nextjs-template, but it doesn't seem to be disabling it for me. Maybe I'm misunderstanding it or doing something wrong, can you confirm with manual testing that this works in publishing?

The `enableHierarchyDisplay` is responsible for displaying `Children` on the published page. This config does not affect the `Backlinks`. Updated the test-workspace/dendron.yml with `enableHierarchyDispplay: false` and `hierarchyDisplayTitle: Sub Notes` setting. If you do `dendron publish init` and then `dendron publish dev`, the `Children` should not be visible when `enableHierarchyDisplay` is set to false.

packages/engine-server/src/markdown/remark/hierarchies.ts
Harshita-mindfire: the config we get from `getProcData` could be v4 or v5, for publishing hierarchies, we get the hierarchy display configs from utility `getHierarchyDisplayConfigForPublishing` that gives us the values from appropriate version. 
Updated the Children fallback twice mistake.

kevinslin (approved):

kevinslin:
approved, can you merge according to [[Async Update to Pr Merge Process|dendron://private/user.kevin.journal.2022.05.20.m.async-update-to-pr-merge-process]]

PR #2949, opened 2022-05-20
enhance(navigate): make wikilinks clickable in hover

jonathanyeung:
## enhance(navigate): make wikilinks clickable in hover

Makes any wikilinks in hover clickable, which will navigate to the note via `GoToNote`

Coauthored by @SeriousBug.

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/engine-server/src/markdown/remark/dendronPreview.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPreview.ts
@@ -1,13 +1,13 @@
 import { vault2Path } from "@dendronhq/common-server";
 import _ from "lodash";
 import { Image } from "mdast";
-import { DendronASTTypes } from "../types";
 import path from "path";
 import Unified, { Transformer } from "unified";
 import { Node } from "unist";
 import visit from "unist-util-visit";
 import { VFile } from "vfile";
-import { RemarkUtils } from ".";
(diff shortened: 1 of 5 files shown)


packages/engine-server/src/markdown/remark/dendronPreview.ts
```diff
@@ -30,6 +30,32 @@ export function makeImageUrlFullPath({
   node.url = fpath;
 }
 
+/**
+ * Transforms any wiklinks into a vscode command URI for gotoNote.
+ */
+function commandUriForWikilink({
```
SeriousBug, line 36: nitpick: the name sounds like it returns the command URI, when it actually sets it

SeriousBug (approved):
Looks good to me, and works well in manual testing. Just left one nitpick.

Harshita-mindfire (approved):
Lgtm, works well in manual testing.

packages/engine-server/src/markdown/remark/dendronPreview.ts
jonathanyeung: amended to `modifyWikilinkValueToCommandUri()`

PR #2952, opened 2022-05-21
feat(notes): Auto generate template/schema for daily journal

tma66:
**Background**
When a user runs Daily Journal for the first time, then auto-generate a template and a schema for them that will apply the template to the journal note. This should make it easier for users to already have a working schema template syntax that they can use to build additional schemas for.

More info: [[Summary|dendron://private/task.schema.2022.01.04.include-pre-built-journal-template#summary]]

**Changes**
1. Check if user is running Daily Journal for first time. If yes, auto generate template/schema
2. Run A/B test (for first time users, half will get auto generated files, other half will get current behavior (nothing new)
3. Generate UUID for auto generated templates. @jonathanyeung Is there a reason why the template id was hardcoded before?

**As the schema is hard coded, this currently does not support the `dateFormat` journalConfig.


packages/plugin-core/assets/dendron-ws/templates/dendron.templates.daily.md
```diff
@@ -0,0 +1,43 @@
+_Edit the [[dendron.templates.daily]] to change this template generated for Dendron Daily Journal._
```
jonathanyeung, line 1: nit: "Edit the [[dendron.templates.daily]] **note** to change"

packages/plugin-core/src/commands/CreateDailyJournal.ts
```diff
@@ -1,15 +1,43 @@
...
+import { ExtensionProvider } from "../ExtensionProvider";
 import { JournalNote } from "../traits/journal";
 import {
   CommandOpts,
   CreateNoteWithTraitCommand,
 } from "./CreateNoteWithTraitCommand";
+import { VSCodeUtils } from "../vsCodeUtils";
+import { MetadataService } from "@dendronhq/engine-server";
+
+export type ExecuteData = {
```
jonathanyeung, line 31: nit: can you rename this to something like `CreateDailyJournalData` or something specific to this command type?

packages/plugin-core/src/commands/CreateDailyJournal.ts
```diff
@@ -21,22 +49,197 @@ export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
...
+    }
+
+    const assetUri = VSCodeUtils.getAssetUri(this._extension.context);
+    const dendronWSTemplate = VSCodeUtils.joinPath(assetUri, "dendron-ws");
+
+    const src = path.join(dendronWSTemplate.fsPath, "templates", fileName);
+    const body = (await fs.readFile(src)).toString();
+
+    // Ensure that engine state is aware of the template before returning so
+    // that the template can be found when creating the meeting note. TODO: This
```
jonathanyeung, line 226: nit: copy paste error on 'meeting'.

packages/plugin-core/src/commands/CreateDailyJournal.ts
```diff
@@ -21,22 +49,197 @@ export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
...
-      const vault = dailyVault
-        ? VaultUtils.getVaultByName({
-            vaults: this._extension.getEngine().vaults,
-            vname: dailyVault,
-          })
-        : undefined;
-
-      return super.execute({ ...opts, vaultOverride: vault });
+    const metaData = MetadataService.instance().getMeta();
+    // Only create default schema/template if running Daily Journal for first time
```
jonathanyeung, line 67: We're going to run into an issue here with existing users - we don't know if they've created Daily Journals in the past.  I'm thinking to make the experience less jarring for existing users, that we don't do this auto template creation for anyone who is an existing user, irrespective of whether they've run create daily journal in the past (as we just don't know). 

I think we can do this by adding an additional check here, something like `if MetadataService.firstInstallTime < May 31st, 2022` THEN don't auto-create the template.  

Ideally, we have a prop in metadata svc like `DendronVersionOnFirstInstall` so that we can do a version based check instead of a time based check.

jonathanyeung (changes requested):

packages/plugin-core/src/commands/CreateDailyJournal.ts
tma66: Ah good point. I can do the time based check while we come up with something better later

packages/plugin-core/assets/dendron-ws/templates/dendron.templates.daily.md
```diff
@@ -0,0 +1,43 @@
+_Edit the [[dendron.templates.daily]] note to change this template generated for Dendron Daily Journal._
+
+<!--
+Based on the journaling method created by Intelligent Change:
```
kevinslin, line 4: Lets link it back to our own blog. Replace with 

> Based on the 5 minute journal method. See more details [here](https://blog.dendron.so/notes/P1DL2uXHpKUCa7hLiFbFA/)

packages/common-all/src/schema.ts
```diff
@@ -58,6 +60,9 @@ export class SchemaCreationUtils {
               pattern: currPattern.pattern,
             };
           }
+          if (currPattern.desc) {
+            curr["desc"] = currPattern.desc;
```
kevinslin, line 64: what was this for?

packages/plugin-core/src/commands/CreateDailyJournal.ts
```diff
@@ -21,22 +50,202 @@ export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
...
-
-      return super.execute({ ...opts, vaultOverride: vault });
+    const metaData = MetadataService.instance().getMeta();
+    // Only create default schema/template if running Daily Journal for first time after 5/31/22
+    if (_.isUndefined(metaData.firstDailyJournalTime)) {
+      isFirstTime = true;
+      MetadataService.instance().setFirstDailyJournalTime();
+      if (
+        !_.isUndefined(metaData.firstInstall) &&
+        metaData.firstInstall > Time.DateTime.fromISO("2022-05-31").toSeconds()
```
kevinslin, line 74: NOTE: the user could already have a daily journal schema. this only checks if they have a `dendron.daily` schema. instead of checking for the file, can we check if they have a schema that starts with `daily` pattern?

packages/plugin-core/src/commands/CreateDailyJournal.ts
```diff
@@ -21,22 +50,202 @@ export class CreateDailyJournalCommand extends CreateNoteWithTraitCommand {
...
+
+    const src = path.join(dendronWSTemplate.fsPath, "templates", fileName);
+    const body = (await fs.readFile(src)).toString();
+
+    // Ensure that engine state is aware of the template before returning so
+    // that the template can be found when creating the daily journal note.
+    const templateNoteProps = NoteUtils.create({
+      fname,
+      vault,
+      id: genUUID(),
```
kevinslin, line 238: this should happen automatically if you don't supply an id

kevinslin (changes requested):

packages/common-all/src/schema.ts
tma66: To create the desc fields in the auto-generated schema. Since we can't do comments easily, I'm using the desc fields to describe each line in the schema

packages/plugin-core/src/commands/CreateDailyJournal.ts
tma66: Good point. Fixed to check for schema id instead

packages/plugin-core/assets/dendron-ws/templates/dendron.templates.daily.md
tma66: I feel linking to [[dendron.templates.daily]] is useful as it doesn't require them to read the tutorial to make any modifications. I added your line to the blog right underneath for those that are curious

kevinslin (approved):

PR #2962, opened 2022-05-24
feat(publish): ability to exclude children in dendron side nav

kevinslin:
feat(publish): ability to exclude children in dendron side nav

For notes with a lot of children, the side nav can become overcrowded. Introduce an option to disable children on the nav. Also make this the default when `has_collection` is `true`

- before: https://loom.com/i/710825588ced436487dc7687d5cf30f8
- after: https://www.loom.com/i/5b0330bb90b347838158b99ac3f3a5ef

- diff size:
    - small: https://www.loom.com/i/0fe71bb4be94402a90d77f999ea2fd06
    - large: https://www.loom.com/i/caf5d9e47fa446e1bb4d864de5f437a9
    - reg: https://www.loom.com/i/5b0330bb90b347838158b99ac3f3a5ef

Docs: https://github.com/dendronhq/dendron-site/pull/509

***

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [~] 1-2 Edge cases tested

### Extended

- CSS
    - [x] lg: screen ≥ 992px
    - [x] safari
    - [x] firefox
    - [x] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/types/index.ts
+++ b/packages/common-all/src/types/index.ts
@@ -73,7 +73,7 @@ export type DendronSiteFM = {
   /**
    * Should exclude children from nav
    */
-  nav_children_exclude?: boolean;
+  nav_exclude_children?: boolean;
   permalink?: string;
   /**
    * If collection, don't show in nav
--- a/packages/common-all/src/types/index.ts
+++ b/packages/common-all/src/types/index.ts
@@ -70,6 +70,10 @@ export type DendronSiteFM = {
   canonicalUrl?: string;
   nav_order?: number;
   nav_exclude?: boolean;
+  /**
+   * Should exclude children from nav
+   */
+  nav_children_exclude?: boolean;
   permalink?: string;
   /**
    * If collection, don't show in nav
--- a/packages/common-all/src/util/treeUtil.ts
+++ b/packages/common-all/src/util/treeUtil.ts
@@ -101,7 +101,7 @@ export class TreeUtils {
       vaultName: VaultUtils.getName(note.vault),
       navExclude: fm.nav_exclude || false,
       children:
-        fm.nav_children_exclude || fm.has_collection
+        fm.nav_exclude_children || fm.has_collection
           ? []
           : this.sortNotesAtLevel({
               noteIds: note.children,
--- a/packages/common-all/src/util/treeUtil.ts
+++ b/packages/common-all/src/util/treeUtil.ts
@@ -1,5 +1,6 @@
 import _ from "lodash";
 import { TAGS_HIERARCHY, TAGS_HIERARCHY_BASE } from "../constants";
+import { PublishUtils } from "../publish";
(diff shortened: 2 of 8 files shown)


packages/common-all/src/types/index.ts
```diff
@@ -70,6 +70,10 @@ export type DendronSiteFM = {
   canonicalUrl?: string;
   nav_order?: number;
   nav_exclude?: boolean;
+  /**
+   * Should exclude children from nav
+   */
+  nav_children_exclude?: boolean;
```
jonathanyeung, line 76: should we name this as `nav_exclude_children` instead?

jonathanyeung (changes requested):

packages/common-all/src/util/treeUtil.ts
```diff
@@ -90,26 +91,30 @@ export class TreeUtils {
...
-        .map((noteId) =>
-          TreeUtils.note2Tree({
-            noteId,
-            noteDict,
-          })
-        )
-        .filter(isNotUndefined),
+      navExclude: fm.nav_exclude || false,
+      children:
+        fm.nav_children_exclude || fm.has_collection
```
SeriousBug, line 104: nitpick: This part is getting pretty unreadable, could we refactor it to be cleaner?

packages/engine-test-utils/src/__tests__/common-all/util/treeUtil.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+        const treeData = TreeUtils.generateTreeData(engine.notes, [domain]);
+        expect(treeData.roots).toMatchSnapshot();
+        expect(treeData.roots[0].children).toEqual([]);
+      },
+      {
+        expect,
+        preSetupHook: ENGINE_HOOKS.setupBasic,
+      }
+    );
+  });
```
SeriousBug, line 55: Could you also add a case for when `has_collection` is enabled, but `nav_children_exclude` is set to false?

packages/common-all/src/types/index.ts
SeriousBug: Also how about `nav_include_children`, where the default is true? If I remember it right we were trying to go with positive names for configs (e.g. enableX instead of disableX), it might make sense to follow that here too.

SeriousBug (changes requested):
A few minor changes I suggested. This is a great idea by the way!

packages/common-all/src/types/index.ts
kevinslin: we switched to `enable*` in the `dendron.yml`. for this case, since the overwhelming case is that children should be included and because `nav_exclude` is also still a property, would follow the "stick to existing conventions" part for this prop

jonathanyeung (approved):

PR #2968, opened 2022-05-24
chore: check vault path usages for self contained vaults

SeriousBug:
Fixed some usages of `vault2Path` and `VaultUtils.getRelPath` to be self contained vault aware.

Also performed some minor cleanup of the codebase as I hit warnings/deprecations.

### Basics

- [ ] circular dependency check: make sure your code is not introducing new circular dependencies in plugin-core.  See [Avoiding Circular Dependencies](https://wiki.dendron.so/notes/pMS27sHxbWeKMoPRrWEzs.html).

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [ ] [Confirm manual testing](https://wiki.dendron.so/notes/99q7A73uGmCwu2KvSHZro.html) 
- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/common-all/src/vault.ts
+++ b/packages/common-all/src/vault.ts
@@ -66,6 +66,11 @@ export class VaultUtils {
     return vault.fsPath;
   }
 
+  static getRelVaultRootPath(vault: DVault) {
+    if (VaultUtils.isSelfContained(vault)) return vault.fsPath;
+    return VaultUtils.getRelPath(vault);
+  }
+
   static getVaultByName({
     vaults,
     vname,
--- a/packages/common-all/src/vault.ts
+++ b/packages/common-all/src/vault.ts
@@ -47,9 +47,16 @@ export class VaultUtils {
   }
 
   /**
-   * Path of vault relative to workspace root
+   * Path for the location of notes in this vault, relative to the workspace
+   * root.
+   *
+   * While for old vaults this is the same as
+   * {@link VaultUtils.getRelVaultRootPath}, for self contained vaults the notes
+   * are located inside the vault in the `notes` subdirectory.
+   *
    * @param vault
-   * @returns
+   * @returns path for location of the notes in the vault, relative to the
+   * workspace root
    */
   static getRelPath(vault: DVault) {
     if (VaultUtils.isSelfContained(vault)) {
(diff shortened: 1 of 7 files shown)


packages/common-all/src/vault.ts
```diff
@@ -66,6 +66,11 @@ export class VaultUtils {
     return vault.fsPath;
   }
 
+  static getRelVaultRootPath(vault: DVault) {
+    if (VaultUtils.isSelfContained(vault)) return vault.fsPath;
+    return VaultUtils.getRelPath(vault);
```
jonathanyeung, line 71: any reason not to put this logic into `VaultUtils.getRelPath()` itself? It's maybe a little bit confusing on knowing when to use `VaultUtils.getRelVaultRootPath()` vs `VaultUtils.getRelPath()`

jonathanyeung (changes requested):

packages/common-all/src/vault.ts
SeriousBug: `getRelPath` gets the location of the notes, while `getRelVaultRootPath` gets the root of the vault. This is only different for self contained vaults, where `getRelPath() == getRelVaultRootPath() + "/notes"`.

When first implementing self contained vaults, I added a feature into `getRelPath` to get the path of the notes folder, since most of the code uses it to access the note files and schema files. But there are a few parts that need the root of the vault, so I'm adding this.

One option would be to rename `getRelPath` to `getRelNotesPath`, and `getRelVaultRootPath` can be the new `getRelPath`. I avoided doing that since it would be a very large diff, but I'm happy to do so.

jonathanyeung: Maybe 80/20 is to just add some of this info as docstrings to these two util methods.

kevinslin (approved):
can you add an async and add an entry to [[Cook|dendron://dendron.docs/dev.process.code.cook]] about best practices when trying to get the vault path?

packages/common-all/src/vault.ts
SeriousBug: Good idea, I'll add the docstrings and merge.

PR #2970, opened 2022-05-24
enhance(internal): Create GET and FINDER for engine notes

tma66:


packages/plugin-core/src/commands/CopyNoteURL.ts
```diff
@@ -55,7 +55,7 @@ export class CopyNoteURLCommand extends BasicCommand<
     }
     const fname = path.basename(maybeTextEditor.document.uri.fsPath, ".md");
     const engine = getDWorkspace().engine;
-    const note = _.find(engine.notes, { fname });
+    const note = (await engine.findNotes({ fname }))[0];
```
jonathanyeung, line 58: This is a pre-existing issue, so feel free to defer.  But we should be using `engine.getNote(id)` in this situation. (applies to the other places you've removed `engine.notes` references too)

jonathanyeung (approved):

packages/plugin-core/src/commands/CopyNoteURL.ts
tma66: Looking at this, it seems `maybeNote` is already the note we want here? @kevinslin Let me know if this line is safe to remove (see latest commit). It failed one of the tests but it seems the test was incorrect?

@jonathanyeung As for `fileWatcher`, do you know how to get the `id` in `onDidDelete`?

jonathanyeung: O, let me amend my comment a bit for the fileWatcher case- maybe we can do something better than using [0] index, i.e. passing in a vault as well to `findNotes()`.  Something in VaultUtils should do the trick in getting the vault? Maybe VaultUtils.getVaultByDirPath(fsPath)

tma66: gotcha, makes sense

jonathanyeung: updates lgtm

PR #2973, opened 2022-05-25
chore(internal): add tutorial note count information

jonathanyeung:
## chore(internal): add tutorial note count information

- A quick fix for identifying notes that are tutorial notes, and adjusting our numNotes, numWikilinks, numBacklinks, and numNoteRefs values accordingly. 

@kevinslin and @hikchoi - please add to the hardcoded list of tutorial ID's as we have more experimental tutorial versions.

Tested for startup time - org-workspace still only spending ~15 ms on this part, so we're ok.

---

### Basics

Airtable Updated

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -265,7 +265,7 @@ class ExtensionUtils {
 
     // Note IDs that are part of our tutorial(s). We want to exclude these from
     // our 'numNotes' telemetry.
-    const tutorialIDs = [
+    const tutorialIds = new Set<string>([
       "ks3b4u7gnd6yiw68qu6ba4m",
       "mycf53kz1r7swcttcobwbdl",
       "kzry3qcy2y4ey1jcf1llajg",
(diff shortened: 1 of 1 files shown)


hikchoi:
@jonathanyeung 
do we plan to keep this list moving forward? I'm in the process of spec-ing out how we would support arbitrary-length tutorials. Not sure how well this pattern is going to fit into that.

jonathanyeung:
> @jonathanyeung do we plan to keep this list moving forward? I'm in the process of spec-ing out how we would support arbitrary-length tutorials. Not sure how well this pattern is going to fit into that.

What I've done here is obviously not a very scalable solution - it's a quick fix to help us to better analysis on first n-week activity.  We can sync up after you're done with the spec on a better way to handle this going forward.

packages/plugin-core/src/_extension.ts
```diff
@@ -254,13 +254,36 @@ class ExtensionUtils {
...
     let numNoteRefs = 0;
     let numWikilinks = 0;
     let numBacklinks = 0;
     let numLinkCandidates = 0;
     let numFrontmatterTags = 0;
+    let numTutorialNotes = 0;
+
+    // Note IDs that are part of our tutorial(s). We want to exclude these from
+    // our 'numNotes' telemetry.
+    const tutorialIDs = [
```
tma66, line 268: Not biggie since size is small right now, but I'd use a set here if we want to further optimize

kevinslin:
@hikchoi do we have an sop on adding to the tutorial? if so, can we include this as part of that sop?

kevinslin (approved):
when we initialize a workspace, we have workspace specific metadata. can we track `numNotesOnFirstStartup`?

Issue #2311, opened 2022-01-30
"Insert note" to insert frontmatter from template


abunavas:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [ ] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup - Dendron's Lookup Command
- [ ] Views - Dendron Preview, Tree View, Side Panels in the UI
- [x] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [ ] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

Not a problem, just enhancement. 

## Describe the solution you'd like ##

When inserting note, it will be useful to insert the frontmatter also

## Describe alternatives you've considered ##

Obviously, I can use schemas, but (for now) it's too confusing for me. :-(

kevinslin:
Adding this to our backlog as a good first task for newcomers

aleksey-rowan:
The community has selected this issue as this week's [CROP](https://wiki.dendron.so/notes/tayowNP4tMvcqH7AraoQ4/) winner. 🎉

bpathakota:
Looks like this is complete, can we mark this as closed? @kevinslin @aleksey-rowan

aleksey-rowan:
@bpathakota, yeap, this is done. I can confirm that frontmatter is being inserted as well. Note that the `insert note` command got replaced by the [`Apply Template`](https://wiki.dendron.so/notes/ftohqknticu6bw4cfmzskq6/) command.

@abunavas, could you please check if the new command is working the way you expect?

PR #2982, opened 2022-05-25
feat(edit): introduce apply template command

kevinslin:
feat(edit): introduce template apply command

This command will use a selected note as a template and apply it to the current note. 
This is the successor of the `InsertNoteCommand` which will be deprecated after this change and removed in a subsequent release.
This addresses the crop issue #2311

New Configuration:
- [templateHierarchy](https://github.com/dendronhq/dendron-site/blob/a65e9a663d35727e6c2d0cd7be3ac6d84d63239f/vault/dendron.topic.templates.config.templateHierarchy.md)

New Commands:
- [Template Apply](https://github.com/dendronhq/dendron-site/blob/a65e9a663d35727e6c2d0cd7be3ac6d84d63239f/vault/dendron.topic.templates.commands.template-apply.md)

Deprecated Commands:
- InsertNoteCommand

New Metrics:
- EventProperty: source
  - Event: TemplateApplied
  - values: CommandName
  - desc: what caused the template to be applied
- Event: "DeprecationNoticeShow"
  - source: CommandName
  - desc: what caused the deprecation notice to be shown
- Event: "DepreactionNoticeAccept"
  - source: CommandName
  - desc: what caused the deprecation notice to be accepted

Test Updates:
- fix issue with stubbing for initial install status since we no longer rely on `StateService` for global version
- add convenience methods for tests:
  ```sh
  yarn test:cli
  yarn test:cli:update-snapshots
  ```

Docs: https://github.com/dendronhq/dendron-site/pull/512

Next:
- post pr, will create an async about test updates


packages/plugin-core/src/commands/NoteLookupCommand.ts
```diff
@@ -557,7 +555,9 @@ export class NoteLookupCommand
           targetNote: nodeNew,
           engine,
         });
-        AnalyticsUtils.track(EngagementEvents.TemplateApplied);
+        AnalyticsUtils.track(EngagementEvents.TemplateApplied, {
+          source: this.key,
```
jonathanyeung, line 559: I think it'll be clearer for us when looking at telem to have this property called 'SchemaTemplate' instead of 'NoteLookup'?

packages/plugin-core/src/utils/LookupUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+  NoteProps,
+  RespV3,
+  SchemaTemplate,
+  VaultUtils,
+} from "@dendronhq/common-all";
+import _ from "lodash";
+import { WSUtilsV2 } from "../WSUtilsV2";
+import { parseRef } from "./md";
+
+export class LookupUtils {
```
jonathanyeung, line 14: can we place this file somewhere in `src/components/lookup` ? We have a bunch of other lookup utils there already; there's also already a file called `LookupUtils.test.ts` that's unrelated to this file.

jonathanyeung (changes requested):
I'm not sure I understand the PR description: 

> Instead of inserting the note, applies the note frontmatter instead

It does both right? 

In a later iteration, it might be good to have an option on whether to include FM or not during apply template.  I've had an issue in the past where I had the `publish:false` FM property set because I didn't want my templates to get published, but then this caused my notes that I had applied the template to also no longer get published, which took me a little while to debug.

packages/plugin-core/package.json
```diff
@@ -258,6 +258,10 @@
         "command": "dendron.createTask",
         "title": "Dendron: Create Task Note"
       },
+      {
+        "command": "dendron.templateApply",
```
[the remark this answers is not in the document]
kevinslin: will write an async about this. we have some commands that do verb+object (go to note) and others that do object+verb (vault add). i would like to standardize to object+verb as it works better when we have lots of commands and is more *hierarchical*

packages/plugin-core/src/commands/InsertNoteCommand.ts
```diff
@@ -79,6 +84,24 @@ export class InsertNoteCommand extends BasicCommand<
     const pos = editor.selection.active;
     const selection = new Selection(pos, pos);
     await editor.insertSnippet(snippet, selection);
+    AnalyticsUtils.track(ExtensionEvents.DeprecationNoticeShow, {
+      source: DENDRON_COMMANDS.INSERT_NOTE.key,
+    });
+    window
+      .showWarningMessage(
+        "Heads up that InsertNote is being deprecated and will be replaced with the 'Template Apply' command",
```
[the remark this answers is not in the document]
kevinslin: same comment as above

packages/plugin-core/src/commands/NoteLookupCommand.ts
kevinslin: a schema template can be applied via `NoteLookup` or `GotoNote`. essentially, a template will always be a "schema template" unless it is called via `TemplateApply`

jonathanyeung: But do we actually care if it's `NoteLookup` vs `GoToNote`? Seems like we should only care if it's automatic application (schema template) vs manual application (apply template command)

Anyway - I'm fine with it, let's just add notes in Airtable about how to interpret what a value with 'NoteLookup' means.

jonathanyeung (approved):

PR #2984, opened 2022-05-26
feat(views): Preview setting for light, dark, or custom themes

SeriousBug:
This PR adds the ability to set the preview to a light or dark theme independent of the IDE theme, as well as a custom theme option for custom themes.

The custom theme support reads the same CSS file as the publishing right now, which I think is advantageous as it means the preview will look identical to published notes. We can add support to set independent custom themes for preview and publishing in a later iteration if desired.

### DIFF
--- a/packages/common-all/data/dendron-yml.validator.json
+++ b/packages/common-all/data/dendron-yml.validator.json
@@ -937,6 +937,13 @@
           },
           "description": "Maps each status to a symbol, word, or sentence. This will be displayed for the task."
         },
+        "taskCompleteStatus": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "Sets which statuses mark the task as completed."
+        },
         "prioritySymbols": {
           "type": "object",
           "additionalProperties": {
@@ -969,6 +976,7 @@
         "name",
         "prioritySymbols",
         "statusSymbols",
+        "taskCompleteStatus",
         "todoIntegration"
       ],
       "description": "Namespace for configuring scratch note behavior"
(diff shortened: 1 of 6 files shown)


packages/common-all/data/dendron-yml.validator.json
```diff
@@ -937,6 +937,13 @@
           },
           "description": "Maps each status to a symbol, word, or sentence. This will be displayed for the task."
         },
+        "taskCompleteStatus": {
```
jonathanyeung, line 940: The task related stuff is just backfill from the previous PR right?

jonathanyeung (changes requested):
Nice! Looking forward to this. 

From prev. discussion (https://github.com/dendronhq/dendron/pull/2887#issuecomment-1122885713), let's also add a FeatureToaster/TipOfDay to highlight this feature. 

Also, it might be cool to later pre-build a css that can match the user's current theme using CSS Variables (https://code.visualstudio.com/api/extension-guides/webview#theming-webview-content)

SeriousBug:
@jonathanyeung Well, turns out it was pretty easy to add support for that too! It now follows your current theme, and actually changes dynamically as you change your theme color. See here:

[Dendron Preview Color Theme Change - Watch Video](https://www.loom.com/share/053db1930a814d6b93a7818c15b18967)

jonathanyeung (approved):
Theme matched preview looks awesome!

Will this cause any issues when publishing with default settings? As in - the CSS variables won't get replaced in published sites.

SeriousBug:
None of this touches published sites, I only made changes to the preview. Published sites should be unaffected.

The only catch is that both preview and publishing use the same CSS file for the custom themes, which I think is good enough for now. If there's demand for setting separate custom themes later we can tackle that.

PR #2985, opened 2022-05-26
chore: removing commands circular dependency

jonathanyeung:
## chore: removing commands circular dependency

Remove the circular dependency related to commands.  This change also removes a deprecated Tree View reference which was no longer being used.

While looking at initialization logic, I realized that getting rid of the command circular dependency was quite trivial now - I think a lot of our previous work around preview clean up and other refactoring work have made it easier.  

Only 1 circular dependency left!


kevinslin (approved):
awesome!

PR #2991, opened 2022-05-26
chore(workspace): detect duplicate note id on lifecycle events

hikchoi:
# chore(workspace): detect duplicate note id on lifecycle events

This PR:
- Adds a workspace utility method that given a text document, detects duplicate note ids existing in the engine
- detects duplicate note ids on:
  - `onDidSaveTextDocument`
  - `onDidChangeActiveTextEditor`
  - `onDidOpenTextDocument`
- Adds a `source` field to the payload when tracking `WorkspaceEvent.DuplicateNoteFound` to differentiate when the detection happened.

## Caveats
- This detection is based on the engine state on init. This means you can have two notes with the same id at one point, but the engine only has one.
  - Perhaps a dupe was found during init, was warned, but ignored.
  - Perhaps there wasn't a dupe, and a note was created with vanilla VSCode create, and is currently not recognized by the engine.
  - If there are dupes on init, depending on which was parsed first, you will have different notes in the engine.
- Because of this, detection only works if the lifecycle events originated / resulted in displaying the note _not_ currently recognized by the engine.
  - i.e.) given `foo.md` with id `foo`, if a user created `foo-dupe.md` and copy pasted everything from `foo.md`, `foo-dupe` is not in the engine, and only when you save / open / change to `foo-dupe.md` will the detection trigger. Saving / opening / changing to `foo.md` will not trigger the detection.
- *This PR only adds the detection part; This will show the same warning message that we already get during init.*
  - I will follow up with a doctor command for automatic fix.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/WSUtilsV2.ts
+++ b/packages/plugin-core/src/WSUtilsV2.ts
@@ -4,22 +4,19 @@ import path from "path";
 import {
   DendronError,
   DNoteAnchorBasic,
-  DuplicateNoteError,
   DVault,
   NoteProps,
   NoteUtils,
   RespV3,
   VaultUtils,
-  WorkspaceEvents,
 } from "@dendronhq/common-all";
 import _ from "lodash";
 import { IWSUtilsV2 } from "./WSUtilsV2Interface";
 import { Logger } from "./logger";
-import { MessageSeverity, VSCodeUtils } from "./vsCodeUtils";
+import { VSCodeUtils } from "./vsCodeUtils";
 import { ExtensionProvider } from "./ExtensionProvider";
-import { file2Note, isInsidePath, vault2Path } from "@dendronhq/common-server";
+import { isInsidePath, vault2Path } from "@dendronhq/common-server";
 import { AnchorUtils, WorkspaceUtils } from "@dendronhq/engine-server";
-import { AnalyticsUtils } from "./utils/analytics";
 
 let WS_UTILS: IWSUtilsV2 | undefined;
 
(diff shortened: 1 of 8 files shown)


kevinslin:
looks like plugin test is failing

packages/plugin-core/src/WSUtilsV2.ts
```diff
@@ -268,4 +271,69 @@ export class WSUtilsV2 implements IWSUtilsV2 {
     const fnameWithExtension = `${fname}.md`;
     return this.openFileInEditorUsingFullFname(vault, fnameWithExtension);
   }
+
```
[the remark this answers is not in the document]
kevinslin: would add it to [[../packages/engine-server/src/doctor/service.ts]]

packages/plugin-core/src/commands/ReloadIndex.ts
```diff
@@ -241,7 +241,9 @@ export class ReloadIndexCommand extends BasicCommand<
         errors.forEach((error) => {
           if (DuplicateNoteError.isDuplicateNoteError(error) && error.code) {
             VSCodeUtils.showMessage(MessageSeverity.WARN, error.message, {});
-            AnalyticsUtils.track(WorkspaceEvents.DuplicateNoteFound);
+            AnalyticsUtils.track(WorkspaceEvents.DuplicateNoteFound, {
+              source: "ReloadIndex",
```
kevinslin, line 245: for commands, can we just use `this.key`?

packages/plugin-core/src/WSUtilsV2.ts
```diff
@@ -268,4 +271,78 @@ export class WSUtilsV2 implements IWSUtilsV2 {
...
+        fsPath,
+      });
+    } catch (error: any) {
+      // document doesn't exist in workspace.
+      return;
+    }
+
+    // we do this because the note in document would _not_ be in our store
+    // if it is a duplicate note.
+    const currentNote = file2Note(fsPath, vault);
```
jonathanyeung, line 292: Looking at the code - this may throw if we're not in a Dendron note at all?  Say someone opens a random non-Dendron Markdown file with no frontmatter, or a plain text file.  Are we still good here?

jonathanyeung (changes requested):
For a different iteration - maybe it'd be good if we don't auto-collapse frontmatter if we detect that the note has a duplicate ID.  I'm wondering if there's a discoverability issue on the frontmatter since we auto collapse it.

packages/plugin-core/src/commands/ReloadIndex.ts
SeriousBug: We already use `DuplicateNoteFound` event when we find them during the initialization, so I think it makes sense to use the same event here.

packages/plugin-core/src/WSUtilsV2.ts
```diff
@@ -268,4 +271,78 @@ export class WSUtilsV2 implements IWSUtilsV2 {
...
+
+    // find the potentially-duplicate note that's currently in our store.
+    const noteById = await engine.getNote(currentNote.id);
+
+    let hasDuplicate = false;
+
+    if (noteById !== undefined) {
+      if (currentNote.id === noteById.id) {
+        // id of note in store and from document is the same. we _might_ have hit a duplicate.
+        if (VaultUtils.isEqualV2(currentNote.vault, noteById.vault)) {
```
SeriousBug, line 302: nitpick: Could simplify this as

```ts
const hasDuplicate = !VaultUtils.isEqualV2(currentNote.vault, noteById.vault) || currentNote.fname !== noteById.fname;
```

SeriousBug (commented):
Looks good apart from comments others already raised

packages/plugin-core/src/WSUtilsV2.ts
hikchoi: I did start with one statement but unfurled it here for readability. I think it's clearer this way.

packages/plugin-core/src/commands/ReloadIndex.ts
hikchoi: > for commands, can we just use this.key?

sure.

> We already use DuplicateNoteFound event when we find them during the initialization, so I think it makes sense to use the same event here.

Are you replying to @kevinslin? Otherwise I'm a bit confused what you are saying. This is the one that's being called during initialization.

packages/plugin-core/src/WSUtilsV2.ts
hikchoi: `file2Note` will result in a `NoteProps` with the frontmatter newly populated, in which case the logic moving forward will detect that there is no note that has the same id in store, which short circuit the rest of the checks.

So we are still good here. Unless you think specifically checking for that case is preferable.

packages/plugin-core/src/WSUtilsV2.ts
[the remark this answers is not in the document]
hikchoi: Can you elaborate? This PR doesn't add a doctor action. It just detects duplicates.
Even when we do add a doctor action to fix the duplicate id, the prompt will have the button that refers to the doctor action to fix it, not the other way around.

Unless you are suggesting that detection itself should be a doctor action, which is not what this PR is doing.

packages/plugin-core/src/WSUtilsV2.ts
hikchoi: Anyways, I'll add a test for the case you are describing.

packages/plugin-core/src/commands/ReloadIndex.ts
SeriousBug: I replied without reading the code, please disregard. Sorry :sweat_smile:

packages/plugin-core/src/WSUtilsV2.ts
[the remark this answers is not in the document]
kevinslin: so I think of `doctorservice` as a service that is able to fix various things that are broken in dendron. any function that might be thought of a "fix a broken workspace" issue can be associated with the doctor service, even if it is not a doctor command (though might be a good backlog item to add fixing duplicate ids as a backlog item)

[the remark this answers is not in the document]
hikchoi: Got it. I'll come up with a sensible location to hold these utils.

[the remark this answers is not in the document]
hikchoi: The utils are only used for the plugin (we hook into lifecycle events), so I opted to not add them to the doctor service, but rather create a `DoctorUtils` that will only be used in the plugin.

jonathanyeung (approved):

kevinslin (approved):

PR #2992, opened 2022-05-26
chore(view): migrate seed browser to plugin views

Harshita-mindfire:
This PR aims to migrate Seed Browser to plugin views from `dendron-next-server`. Came across it while deprecating next server.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA

## Docs
NA

### Extended
NA

### DIFF
--- a/packages/dendron-plugin-views/src/components/SeedBrowser.tsx
+++ b/packages/dendron-plugin-views/src/components/SeedBrowser.tsx
@@ -1,0 +1,97 @@
+import { SeedConfig, SEED_REGISTRY } from "@dendronhq/common-all";
+import { Avatar, Card, Layout, List, PageHeader } from "antd";
+import _ from "lodash";
+import seedStyles from "../styles/scss/seeds.module.scss";
+import { DendronComponent } from "../types";
+import { AddToWorkspaceButton, GoToSiteButton } from "./Seeds";
+
+const SeedBrowser: DendronComponent = (props) => {
+  const { ide, workspace } = props;
+  const { browser } = workspace;
+  const seedDataToRender = _.values(SEED_REGISTRY).map((data) => {
+    if (data) {
+      let payload: SeedConfig & { seedInWorkspace: boolean } = {
+        ...data,
+        seedInWorkspace: false,
+      };
+      if (_.includes(ide.seedsInWorkspace, data?.id)) {
+        payload = {
+          ...payload,
+          seedInWorkspace: true,
+        };
+      }
+      return payload;
+    }
+    return undefined;
+  });
+
+  const { Meta } = Card;
+  const { Content } = Layout;
+
+  return (
+    <>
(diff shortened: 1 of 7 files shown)


packages/dendron-plugin-views/src/components/SeedBrowser.tsx
```diff
@@ -0,0 +1,97 @@
...
+                  xxl: 4,
+                }}
+                dataSource={seedDataToRender}
+                renderItem={(item) => (
+                  <List.Item>
+                    <Card
+                      className={seedStyles.card}
+                      hoverable
+                      actions={[
+                        <GoToSiteButton
```
kevinslin, line 59: both action buttons currently have a `.` next to the icon > https://www.loom.com/i/bc2fef4657ba4625935c13054a73d737

kevinslin (changes requested):
there's a flicker when loading the seed browser for the first time. see https://www.loom.com/share/fa2b41e1f5a949a4ae776919b07bf8b4

can we not load the images until the page is ready? (low priority since the seed browser will need some re-work now that we have self contained vaults). if you can do it in <30min, feel free, otherwise ignore it

packages/dendron-plugin-views/src/components/Seeds.tsx
```diff
@@ -0,0 +1,93 @@
...
+      } else {
+        window.open(url);
+      }
+    }
+  }
+
+  if (url) {
+    return (
+      <Tooltip placement="top" title="Go to Site">
+        <GlobalOutlined key="website" onClick={onClick} />
```
SeriousBug, line 48: Not the point of the PR, but if it's not too much work could we also switch these to put the text next to the icon, instead of having it be a tooltip? So:

```jsx
<Text><GlobalOutlined ... /> Go to Site </Text>
```

Same for the other buttons. That's more accessible for screen readers, and more intuitive for everyone I think. But again not the point of this PR so feel free to ignore this.

Harshita-mindfire: This is pretty easy to do but I think the reason behind keeping the tooltip is to keep the design clean since some of the messages are long `Site Unavailable `Add to workspace` and `Already in Workspace`

Harshita-mindfire:
> can we not load the images until the page is ready? (low priority since the seed browser will need some re-work now that we have self contained vaults). if you can do it in <30min, feel free, otherwise ignore it

gave it a quick try by enclosing it in useEffect and adding `document.readyState` as dependency, didn't work out. Will try again in a follow-up PR

jonathanyeung:
> gave it a quick try

Let's defer it - usage of this view is very low; let's just try to get across the line of deprecating next-server

jonathanyeung (approved):

PR #3006, opened 2022-05-29
fix: Images with encoded URI are not rendered in the Preview

tenheadedlion:
For example: `` is not rendered in the Preview, while this is perhaps not caused by pasting images directly into the Dendron Paste Image extension, some users might use other editors to edit the same note, those editors may store the image from clipboard with a space-separated name and return an unexpectedly encoded URI.

### DIFF
--- a/packages/engine-server/src/markdown/remark/dendronPreview.ts
+++ b/packages/engine-server/src/markdown/remark/dendronPreview.ts
@@ -26,7 +26,7 @@ export function makeImageUrlFullPath({
   }
   // assume that the path is relative to vault
   const { wsRoot, vault } = MDUtilsV5.getProcData(proc);
-  const fpath = path.join(vault2Path({ wsRoot, vault }), node.url);
+  const fpath = path.join(vault2Path({ wsRoot, vault }), decodeURI(node.url));
   node.url = fpath;
 }
 
--- a/packages/plugin-core/src/test/suite-integ/PreviewPanel.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/PreviewPanel.test.ts
@@ -116,6 +116,34 @@ suite("GIVEN PreviewPanel", function () {
         });
       });
 
+      describe("AND image URI is encoded", () => {
+        test("THEN URL is correctly rewritten", async () => {
+          const { vaults } = ExtensionProvider.getDWorkspace();
+          const note = await makeTestNote({
+            previewPanel,
+            body: "![](assets/Pasted%20image%20%CE%B1.png)",
+          });
+          expect(
+            await AssertUtils.assertInString({
+              body: note.body,
+              match: [
+                "https://file",
+                "vscode",
+                path.posix.join(
+                  VaultUtils.getRelPath(vaults[0]),
+                  "assets",
+                  // `makeTestNote()` will invoke `rewriteImageUrls()`
+                  //  in which `makeImageUrlFullPath()` will expectedly decode "Pasted%20image%20%CE%B1.png"
+                  //    to "Pasted image α.png",
+                  //  then `panel.webview.asWebviewUri` encodes it back to "Pasted%20image%20%CE%B1.png".
+                  "Pasted%20image%20%CE%B1.png"
+                ),
+              ],
+            })
+          ).toBeTruthy();
+        });
+      });
+
       describe("AND image is an absolute path", () => {
         test("THEN URL is correctly rewritten", async () => {
           const { wsRoot } = ExtensionProvider.getDWorkspace();


kevinslin:
thanks for the PR!
would you mind adding a test for this as well?

relevant test code here > https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/PreviewPanel.test.ts#L135:L135
instructions for running tests here > [[Testing|dendron://dendron.docs/dev.process.qa.test]]

tenheadedlion:
> thanks for the PR! would you mind adding a test for this as well?
> 
> relevant test code here > https://github.com/dendronhq/dendron/blob/master/packages/plugin-core/src/test/suite-integ/PreviewPanel.test.ts#L135:L135 instructions for running tests here > [Testing](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro)

I think adding a test as follows to PreviewPanel.test.ts is trivial, it just creates a markdown file with an encoded image URI in it.

```
      describe("AND image URI is encoded", () => {
        test("THEN URL is correctly rewritten", async () => {
          const { vaults } = ExtensionProvider.getDWorkspace();
          const note = await makeTestNote({
            previewPanel,
            body: "![](assets/Pasted%20image%20%CE%B1.png)",
          });
          expect(
            await AssertUtils.assertInString({
              body: note.body,
              match: [
                "https://file",
                "vscode",
                path.posix.join(
                  VaultUtils.getRelPath(vaults[0]),
                  "assets",
                  //"Pasted%20image%20%CE%B1.png"
                  encodeURI("Pasted image α.png")
                ),
              ],
            })
          ).toBeTruthy();
        });
      });
```
which results in:

```
./tmp-1185486-yfjRlyRxCU2l/vault/preview-test-image.md:![](assets/Pasted%20image%20%CE%B1.png)
```
 It doesn't create a real image in the test vault so there is no way to make sure the URI points to that image. And the function `makeImageUrlFullPath()` is never hit during debugging, the codebase has no test for `makeImageUrlFullPath()` and it's kind of difficult to write one, it seems we can only verify it with our eyes to see if the Preview renders the image.

However, I think the commit is fine and necessary for 3 reasons:

1. [`decodeURI()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) is safe to use in any scenario, it does no harm if the URI doesn't have encoded components, the result is the same as before the commit. But if the URI does have an encoded component, `makeImageUrlFullPath()` will translate it to the same URI as the actual image URI in the disk, for example, `%CE%B1.png` must be decoded to `α.png` otherwise the filesystem will be looking for `%CE%B1.png`.
2. it is unlikely for a user or software to create a `%CE%B1.png` or `Pasted%20image.png` in the disk. If there is a need to name an image with a space-separated name, the user would happily type a space instead of `%20`, and `α` instead of `%CE%B1` because the latter is impossible to type without looking it upon on a table of something, and that requires the input of `α`, which the user would just live with it and need not bother with a `%CE%B1`. And although it's unwise to name a file with spaces there are editors out there doing just that, for example, Obsidian when it pastes images. So it's pretty safe to ignore the possibility of somebody really typing `%CE%B1`, and we should take care of the major use cases.
3. Dendron does not recognize `` as an image, and maybe it should? Typora can do that. But it doesn't matter, because `decodeURI("Pasted image 20220529172111.png")` returns `"Pasted image 20220529172111.png"`, that is to say, even if Dendron is going to support such syntax, the commit is still valid.

kevinslin:
>  And the function makeImageUrlFullPath() is never hit during debugging,

`makeImageUrlFullPath` is hit on `makeTestNote` every time as long as you are not using a web url (eg. http://...). 

instead of verifying whether an image is on disk, can we just check that an image with an encoded ur is properly decoded when its rewritten? `makeImageUrlFullPath` gets called

tenheadedlion:
> makeImageUrlFullPath is hit on makeTestNote every time as long as you are not using a web URL 

Ah yes, now I understand how it works. The test involves a lot of changes because in the `rewriteImageUrls()` function, `panel.webview.asWebviewUri()` is called right after  `makeImageUrlFullPath()`, the URI that is decoded in `makeImageUrlFullPath()` is again encoded by `panel.webview.asWebviewUri()`, if we want to test the decoded URI we have to break into the code and add some flags to control the workflow, so that it may return a decoded URI.

Is the commit appropriate? Looking forward to your advice.

tenheadedlion:
without `decodeURI()`, an image URI will [go through](https://github.com/dendronhq/dendron/blob/50ebb87b6111840ed0e95614fe88ea721d980b0f/packages/plugin-core/src/components/views/PreviewPanel.ts#L294) these changes:

```
before makeImageUrlFullPath(): assets/Pasted%20image%20%CE%B1.png
after makeImageUrlFullPath(): /tmp/tmp-1266476-qixR7dQSq3NV/vault/assets/Pasted%20image%20%CE%B1.png
after panel.webview.asWebviewUri(): https://file%2B.vscode-resource.vscode-cdn.net/tmp/tmp-1266476-qixR7dQSq3NV/vault/assets/Pasted%2520image%2520%25CE%25B1.png
```

kevinslin:
in that case, instead of adding a conditional in `rewriteImageUrls` just for the test, can we just test that the output is `Pasted%20image%20%CE%B1.png`? (we'll know decode worked because output isn't `Pasted%2520image%2520%25CE%25B`). you can add a comment that links back to this PR or just a sentence explaining the logic

kevinslin:
nice job! thanks for the pr. will be released in next weeks release :)

PR #3009, opened 2022-05-31
chore(workspace): support initializing variable length tutorials

hikchoi:
# chore: support initializing variable length tutorials

This PR:
- Proposes a new convention for tutorials for consolidating tutorial view events into one
- Add additional information about tutorial view event for easier analysis
  - `fname`, `currentStep`, `totalSteps`
- syncs tutorial that has additional frontmatter as per new convention
- fixes an issue where some tutorial view events were omitted

## New convention
- Tutorials now should have frontmatter `currentStep` and `totalSteps`, indexed by zero. (entry point is always `currentStep: 0`)

## omitted tutorial view events
- We have a sizable number of users who have never reported some tutorial events.
- This seems to be rooted from a race between registering the callback to `onDidChangeVisibleTextEditors`, and actually opening the note
  - This is mitigated by making sure we register the callback before we proceed with opening tutorial notes for the user.

## Deprecated events
- `Tutorial_X_Show` events

## Introduced events
- `TutorialNoteViewed`
  - `testGroup`
  - `fname`
  - `currentStep`
  - `totalSteps`

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -47,7 +47,15 @@ export enum TutorialEvents {
   Tutorial_3_Show = "Tutorial_3_Show",
   Tutorial_4_Show = "Tutorial_4_Show",
   Tutorial_5_Show = "Tutorial_5_Show",
+  TutorialNoteViewed = "TutorialNoteViewed",
 }
+
+export type TutorialNoteViewedPayload = {
+  testGroup: string;
+  fname: string;
+  currentStep: number;
+  totalSteps: number;
+};
 
 export enum KeybindingConflictDetectedSource {
   activation = "activation",
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -41,17 +41,11 @@ export enum CLIEvents {
 export enum TutorialEvents {
   WelcomeShow = "WelcomeShow",
   ClickStart = "Getting_Started_Clicked",
-  Tutorial_0_Show = "Tutorial_0_Show",
-  Tutorial_1_Show = "Tutorial_1_Show",
-  Tutorial_2_Show = "Tutorial_2_Show",
-  Tutorial_3_Show = "Tutorial_3_Show",
-  Tutorial_4_Show = "Tutorial_4_Show",
-  Tutorial_5_Show = "Tutorial_5_Show",
   TutorialNoteViewed = "TutorialNoteViewed",
 }
 
 export type TutorialNoteViewedPayload = {
-  testGroup: string;
+  tutorialType: string;
   fname: string;
   currentStep: number;
   totalSteps: number;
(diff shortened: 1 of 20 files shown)


personjerry:
I've also seen a bunch of users who go straight to tutorial 2 without ever hitting tutorial 0. Will this fix this issue as well?

packages/plugin-core/src/workspace/tutorialInitializer.ts
```diff
@@ -67,9 +68,51 @@ export class TutorialInitializer
...
     const ctx = "TutorialInitializer.onWorkspaceOpen";
 
+    // Register a special analytics handler for the tutorial:
+    // This needs to be registered before we open any tutorial note.
+    // Otherwise some events may be lost and not reported properly.
+    const disposable = vscode.window.onDidChangeActiveTextEditor((e) => {
+      const document = e?.document;
+
+      if (document !== undefined) {
+        const payload = this.getAnalyticsPayloadFromDocument({
```
SeriousBug, line 103: I think this throws if the opened document is not in a vault. Should add a `try...catch` block around it.

SeriousBug (commented):
Looks good, just had one comment about a possible point of errors.

kevinslin (commented):
for the tutorial frontmatter, can we make sure they are also present in dendron-site? otherwise, syncing them would remove them, no?

packages/common-all/src/analytics.ts
```diff
@@ -47,8 +47,16 @@ export enum TutorialEvents {
   Tutorial_3_Show = "Tutorial_3_Show",
   Tutorial_4_Show = "Tutorial_4_Show",
   Tutorial_5_Show = "Tutorial_5_Show",
```
jonathanyeung, line 49: Let's remove these events since we're no longer using them.

packages/common-all/src/analytics.ts
```diff
@@ -47,8 +47,16 @@ export enum TutorialEvents {
...
+  TutorialNoteViewed = "TutorialNoteViewed",
 }
 
+export type TutorialNoteViewedPayload = {
+  testGroup: string;
```
jonathanyeung, line 54: subtle distinction, but rather than 'testGroup', I think we should be tracking something called 'tutorialType'

jonathanyeung (changes requested):

packages/plugin-core/src/workspace/tutorialInitializer.ts
hikchoi: Seems like the extension host is swallowing the error that is thrown when trying to grab the vault of the file, but I wrapped it around a try catch just to make it safe.

hikchoi:
> for the tutorial frontmatter, can we make sure they are also present in dendron-site? otherwise, syncing them would remove them, no?

@kevinslin those are here: https://github.com/dendronhq/dendron-site/pull/519

kevinslin (approved):

PR #3014, opened 2022-06-01
feat(sync): Obsidian Import Flow

jonathanyeung:
## feat(sync): Obsidian Import Flow

MVP flow for Obsidian Import, specifically targeting onboarding users who are coming from Obsidian. This utilizes markdown import pod under the hood, but gets rid of the need to fill out a `config.yml` file.  All a user needs to do is use a file picker control to pick the folder. 

This will show up in a feature toaster to users under the following conditions:
1. user marked 'Obsidian' in prior tools in the survey
2. user opened a tutorial workspace
3. Either the user reached the last page of the tutorial, or 10 minutes has elapsed, whichever comes first.

This change doesn't improve or fix existing issues in the Markdown Import Pod.  That will be deferred to a subsequent PR. 

TODO: 
1. Need to resolve the tutorialInitializer logic against Mark's change here: https://github.com/dendronhq/dendron/pull/3009

---

### Basics

### Extended
- General
  - [N/A] check if this change adversely impact performance
- Operations
  - [N/A] when shipping this change, will it just work or will it introduce additional operational overhead due to complicated interface or known bugs?
- Architecture

### Basics
- [N/A] if you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs
https://github.com/dendronhq/dendron-site/pull/522

### DIFF
--- a/packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
+++ b/packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
@@ -1,0 +1,70 @@
+import {
+  getAllImportPods,
+  MarkdownImportPod,
+  podClassEntryToPodItemV4,
+} from "@dendronhq/pods-core";
+import vscode from "vscode";
+import { PickerUtilsV2 } from "../../components/lookup/utils";
+import { DENDRON_COMMANDS } from "../../constants";
+import { PodQuickPickItemV4 } from "../../utils/pods";
+import { CommandOpts, ImportPodCommand } from "../ImportPod";
+
+/**
+ * Convenience command that uses the same flow as {@link ImportPodCommand} for
+ * Markdown Pod but simplifies the steps by not requiring the user to fill out a
+ * config.yml file.
+ */
+export class ImportObsidianCommand extends ImportPodCommand {
+  key = DENDRON_COMMANDS.IMPORT_OBSIDIAN_POD.key;
+
+  constructor(_name?: string) {
+    super(_name);
+    this.pods = getAllImportPods();
+  }
+
+  /**
+   * Hardcoded to use markdown pod, as Obsidian is a markdown import.
+   * @returns
+   */
+  override async gatherInputs() {
+    const markdownPod = podClassEntryToPodItemV4(MarkdownImportPod);
+
+    const podChoice = {
+      label: markdownPod.id,
+      ...markdownPod,
+    } as PodQuickPickItemV4;
+
+    return { podChoice };
+  }
+
+  /**
+   * Use a file picker control instead of a pod config YAML file to get the
+   * Obsidian vault location. Also, just default to the current vault.
+   * @returns
+   */
+  override async enrichInputs(): Promise<CommandOpts | undefined> {
+    const podChoice = podClassEntryToPodItemV4(MarkdownImportPod);
+
+    const uri = await vscode.window.showOpenDialog({
+      title: "Obsidian vault location to import",
+      canSelectFiles: false,
+      canSelectFolders: true,
+      canSelectMany: false,
+    });
+
+    if (!uri || uri.length === 0) {
+      return;
+    }
+
+    const src = uri[0].fsPath;
+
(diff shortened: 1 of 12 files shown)


packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+import { DENDRON_COMMANDS } from "../../constants";
+import { PodQuickPickItemV4 } from "../../utils/pods";
+import { CommandOpts, ImportPodCommand } from "../ImportPod";
+
+/**
+ * Convenience command that uses the same flow as {@link ImportPodCommand} for
+ * Markdown Pod but simplifies the steps by not requiring the user to fill out a
+ * config.yml file.
+ */
+export class ImportObsidianCommand extends ImportPodCommand {
```
kevinslin, line 17: this might be a markdown import pod issue but right now, we don't show a progress bar when importing which makes it hard to tell if import is still going or something has crashed. can we show a window.progress notification during the import?

packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+  key = DENDRON_COMMANDS.IMPORT_OBSIDIAN_POD.key;
+
+  constructor(_name?: string) {
```
kevinslin, line 20: can we track how many notes were imported?

packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
kevinslin, line 17: can we also have this just be a dedicated command?  import pod -> obsidian

kevinslin (changes requested):
a few nitpicks, otherwise looks good.

as a future step, we should have this be triggered by the last page of the tutorial via a command uri under next steps (eg. `Import your notes from obsidian`)

packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
jonathanyeung: Er - you mean 'dedicated pod' instead of being a dedicated command?

If that's what you meant - I think that might require more tweaking of how the import pod flow works, which is what I was trying to avoid in this 80/20 iteration.

packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
```diff
@@ -0,0 +1,70 @@
...
+      return;
+    }
+
+    const src = uri[0].fsPath;
+
+    const vault = PickerUtilsV2.getVaultForOpenEditor();
+
+    const config = {
+      src,
+      vaultName: vault.name,
```
Harshita-mindfire, line 65: Can we update this to use vault util method to get name(`VaultUtils.getName`)? For vaults that has name property as undefined (only have `fsPath`), this would result in ajv validation error: must have required property 'vaultName'

packages/plugin-core/src/commands/pods/ImportObsidianCommand.ts
jonathanyeung: It looks like we already do this in Import Pod:

https://github.com/dendronhq/dendron/blob/08e8320dc88ef52c1c49e76ace843f8a50642225/packages/plugin-core/src/commands/ImportPod.ts#L129

https://github.com/dendronhq/dendron/blob/08e8320dc88ef52c1c49e76ace843f8a50642225/packages/plugin-core/src/commands/ImportPod.ts#L172

kevinslin: got it. lets go ahead with this then

kevinslin (approved):

PR #3023, opened 2022-06-01
fix: bullet points missing in new theme-matching preview

SeriousBug:
Fixes an issue where bullet points in lists would be missing when using the new theme-matching preview.


jonathanyeung (approved):
Tested locally, looks good

PR #3026, opened 2022-06-01
chore(internal): Clean up store api

tma66:
**Background**
As described in https://docs.google.com/document/d/133blpKIJuWL2nBar3rgQdnQMEOuqs2gCgf9fqluyNUI/edit, we will be moving to a more consistent RESTful interface between the store and engine and clients. This PR and upcoming PRs will focus on cleaning up and simplifying the current store before we do any new implementations

**Changes**
1. Rename `bulkAddNotes` to `bulkWriteNotes`. This will be the first of our bulk operations that will perform the same business logic as `writeNote`, just on multiple notes. It will also return a `DendronCompositeError` instead of a single error
2. Add `getAllNotes` method on store + engine. This will replace our usages of `engine.notes` moving forward. EDIT: Removed in favor of migrating usages of `engine.notes` first
3. Deprecate usages of `engine.notes`. We should avoid direct access to the notes
4. Fix bugs on engine side where notefnames may not be properly updated

**Todo**
1. Rename API to match store apis. I will focus on cleaning up the backend apis before touching anything client facing


jonathanyeung (commented):
Took a quick pass, generally looks good; I'll try to take a deeper look soon but if others approve feel free to merge.

packages/engine-test-utils/src/presets/engine-server/index.ts
```diff
@@ -33,7 +33,7 @@ export const ENGINE_SERVER = {
   ENGINE_RENAME_PRESETS,
   ENGINE_GET_NOTE_BLOCKS_PRESETS,
   ENGINE_QUERY_PRESETS,
-  ENGINE_BULK_ADD_NOTES_PRESETS,
+  ENGINE_BULK_ADD_NOTES_PRESETS: ENGINE_BULK_WRITE_NOTES_PRESETS,
```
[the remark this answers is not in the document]
tma66: oops, fixed

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -199,6 +200,13 @@ export class FileStorage implements DStore {
     }
   }
 
+  /**
+   * See {@link DStore.getAllNotes}
+   */
+  async getAllNotes(): Promise<NotePropsByIdDict> {
```
kevinslin, line 206: this seems much more expensive then return `this.notes`. 
since this is used a lot - can we get some benchmarks of how this performs, latency wise, in org-workspace?

packages/engine-server/src/enginev2.ts
```diff
@@ -235,6 +235,10 @@ export class DendronEngineV2 implements DEngine {
     return DendronEngineV2._instance;
   }
 
+  /**
+   * @deprecated see {@link DendronEngineV2.getAllNotes}
+   * For accessing a specific note by id, see {@link DendronEngineV2.getNote}
+   */
   get notes(): NotePropsByIdDict {
```
kevinslin, line 242: do we need to do this if we are going to be updating the engine interface anyways to `engine.notes.[method]`? might represent unnecessary churn

kevinslin (changes requested):

packages/engine-server/src/enginev2.ts
tma66: I want to deprecate this first to avoid usages of `this.engine.notes` before we get a clear idea of the engine interface. `engine.notes.method` might not be valid anymore since the business logic lies in the engine vs notes.

Either way, I think its worth preventing others from using the wrong method now.

packages/engine-server/src/drivers/file/storev2.ts
tma66: I added this to unblock future clients that require the entire engine notes and to avoid usages of `engine.notes`. Most should be using `getNote`.

I'll add some benchmarks for this to see its impact

tma66: For 24695 notes in org-workspace, `getAllNotes` takes ~634ms

jonathanyeung: 634 ms may be non-trivial, depending on the type of operation.  Do we have any active references to this new `getAllNotes` method yet?

tma66: None now, but after talking to @kevinslin offline, I'm going to remove this for now. Ideally most clients shouldn't need to get all notes so I think we can do a sweep first and provide methods on the engine/store side.

Looking at some current usages, its cases of getting all notes and doing own custom filtering so those can be refactored first

kevinslin (approved):

PR #3033, opened 2022-06-02
feat(structure): Set Task Status and Complete Task commands

SeriousBug:
Adds the commands "Set Task Status" and "Complete Task".

Set Task Status will prompt the user for the status to use based on the task symbol mappings, while "Complete Task" uses the first `taskCompleteStatus` without prompting.

For both commands, if the user has a link to a note selected, they will set the status for that note. If they don't have any links selected but are currently in a task note, then it will set the status for that note (only if the current note is a task note, to avoid setting the status accidentally in non-task notes).

If the user has a link selected but the link is ambiguous (multiple notes across vaults with the same name, and the link isn't x-vault), then it will also prompt for which note the user wanted.

Docs: https://github.com/dendronhq/dendron-site/pull/525

### Basics

### Extended

- General
- Operations
- Architecture

### Basics

### Extended

## 

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended

- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

### Basics

## 

### Basics

## 

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/TaskComplete.ts
+++ b/packages/plugin-core/src/commands/TaskComplete.ts
@@ -1,25 +1,20 @@
-import { ConfigUtils, NoteProps } from "@dendronhq/common-all";
+import { ConfigUtils } from "@dendronhq/common-all";
 import { DENDRON_COMMANDS } from "../constants";
 import { BasicCommand } from "./base";
 import { VSCodeUtils, MessageSeverity } from "../vsCodeUtils";
 import { IDendronExtension } from "../dendronExtensionInterface";
 import { TaskStatusCommand } from "./TaskStatus";
 
-type CommandInput = {
-  setStatus?: string;
-};
+type CommandOpts = {};
 
-type CommandOpts = Required<CommandInput> & {
-  note: NoteProps;
-};
+type CommandOutput = {} | undefined;
 
-type CommandOutput = {};
-
-export class CreateTaskCommand extends BasicCommand<
+export class TaskCompleteCommand extends BasicCommand<
   CommandOpts,
   CommandOutput
 > {
-  key = DENDRON_COMMANDS.TASK_CREATE.key;
+  key = DENDRON_COMMANDS.TASK_COMPLETE.key;
+  public static requireActiveWorkspace: boolean = true;
   private _ext: IDendronExtension;
 
   constructor(extension: IDendronExtension) {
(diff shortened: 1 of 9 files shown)


packages/plugin-core/src/commands/TaskComplete.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  async execute(_opts: CommandOpts) {
+    const complete: string | undefined = ConfigUtils.getTask(
+      this._ext.getDWorkspace().config
+    ).taskCompleteStatus[0];
+
+    if (complete === undefined) {
+      await VSCodeUtils.showMessage(
+        MessageSeverity.ERROR,
+        "You have no task statuses marked as complete. Please add something to 'taskCompleteStatus' in your configuration file.",
```
kevinslin, line 33: can we show a link to the relevant docs here and open dendron.yml?

kevinslin (commented):
nice! just a nitpick about better ergonomics on the error case, otherwise looks good!

bpathakota (commented):
Since this relates to task notes, would include 'structure' as the scope tag to the PR title

packages/plugin-core/src/commands/TaskComplete.ts
SeriousBug: It actually looks like this never gets triggered, because the `ConfigUtils` fills in the configuration if it is empty or missing. I did add the link and config command for future proofing in case that changes in the future.

kevinslin (approved):

packages/plugin-core/src/test/suite-integ/TaskStatus.test.ts
```diff
@@ -0,0 +1,250 @@
...
+    after(() => {
+      showQuickPick.restore();
+    });
+
+    test("THEN prompts for the status", () => {
+      expect(showQuickPick.calledOnce).toBeTruthy();
+    });
+    test("THEN updates the task status", async () => {
+      const { engine } = ExtensionProvider.getDWorkspace();
+      const task = NoteUtils.getNoteByFnameFromEngine({
```
tma66, line 54: nit: Use `engine.findNotes` instead

SeriousBug: Thanks, switched to it.

On a side note, could we add a "find 1 note" function for the engine too? It's just a little nicer in ergonomics, because otherwise you end up with `(await engine.find(...))[0]` when you are sure that there's only a single note.

tma66: Sure, I can add a method on engine to make it a little simpler, Something like findFirstNote that returns a NoteProps or undefined.

PR #3034, opened 2022-06-02
chore: `.skip` and `.only` for `describe*WS` tests & ensure `props` can't be set for overridden properties when creating notes

SeriousBug:
This PR makes 2 small enhancements to testing utilities:

- `describeSingleWS` and `describeMultiWS` functions now support `.only` and `.skip`, using the same format as `describe`. This means you can now do `describeSingleWS.only(...` when working with these tests.
- When creating a testing note, setting some properties for the note under `props` is incorrect because they will get overridden by the more general properties of the function. It looks like there was an attempt to enforce this with types, but an error meant that it wasn't getting applied. This PR also adds `custom` to the "incorrect" props because it also gets overridden.

I'll write an async once this is released for the `.only` and `.skip` now being available.

### DIFF
--- a/packages/common-test-utils/src/noteUtils.ts
+++ b/packages/common-test-utils/src/noteUtils.ts
@@ -26,7 +26,7 @@ export type CreateNoteOptsV4 = {
   wsRoot: string;
   fname: string;
   body?: string;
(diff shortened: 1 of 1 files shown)


kevinslin (approved):
nice!

PR #3036, opened 2022-06-03
enhance/publish upgrade

kevinslin:
- add templateVersion config variable
- dendron publish init improvements

### DIFF
--- a/packages/dendron-cli/src/commands/publishCLICommand.ts
+++ b/packages/dendron-cli/src/commands/publishCLICommand.ts
@@ -4,29 +4,30 @@ import {
   DendronError,
   DendronSiteConfig,
   error2PlainObject,
+  ErrorFactory,
   getStage,
   Stage,
 } from "@dendronhq/common-all";
-import { SiteUtils } from "@dendronhq/engine-server";
+import { GitUtils } from "@dendronhq/common-server";
+import { DConfig, SiteUtils, VersionProvider } from "@dendronhq/engine-server";
 import {
+  BuildOverrides,
   NextjsExportConfig,
   NextjsExportPod,
   NextjsExportPodUtils,
-  BuildOverrides,
   PublishTarget,
 } from "@dendronhq/pods-core";
+import fs from "fs-extra";
 import _ from "lodash";
+import ora from "ora";
 import path from "path";
+import prompts from "prompts";
 import yargs from "yargs";
 import { CLIUtils, SpinnerUtils } from "../utils/cli";
 import { CLICommand } from "./base";
 import { ExportPodCLICommand } from "./exportPod";
 import { PodSource } from "./pod";
 import { SetupEngineCLIOpts } from "./utils";
-import prompts from "prompts";
-import fs from "fs-extra";
-import ora from "ora";
-import { GitUtils } from "@dendronhq/common-server";
 
 type CommandCLIOpts = {
   cmd: PublishCommands;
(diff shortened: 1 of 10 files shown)


packages/dendron-cli/src/commands/publishCLICommand.ts
```diff
@@ -414,14 +408,39 @@ export class PublishCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
-  async _initialize(opts: { nextPath: string; spinner: ora.Ora }) {
-    const { spinner } = opts;
+  /**
+   * Clone template
+   * If templateVersion is set in dendron.yml, clone at version
+   * Otherwise, get same version as engine
+   * @param opts
+   * @return: string, version of template
+   */
+  async _initialize(opts: {
```
jonathanyeung, line 418: since this only seems to clone a template, and its return value is a templateVersion, should we just rename this method 'cloneTemplate'?

packages/dendron-cli/src/commands/publishCLICommand.ts
```diff
@@ -414,14 +408,39 @@ export class PublishCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+  }) {
+    const { spinner, wsRoot } = opts;
     SpinnerUtils.renderAndContinue({
       spinner,
       text: "Initializing NextJS template.",
     });
+    const config = DConfig.readConfigAndApplyLocalOverrideSync(opts.wsRoot);
     await this._cloneTemplate(opts);
+    const templateVersion =
+      NextjsExportPodUtils.templateVersion(config) ||
```
jonathanyeung, line 431: What happens if config version exceeds engine Version?  Shouldn't we do `Math.min(configVersion?, engineVersion)` ?

packages/pods-core/src/builtin/NextjsExportPod.ts
```diff
@@ -161,6 +171,16 @@ export class NextjsExportPodUtils {
     return { error: null };
   }
 
+  static async switchToBranch({
```
jonathanyeung, line 174: It's a bit unintuitive that this method is called `switchToBranch` but the arguments are a path and a version.  

How come this method is never invoked anywhere in plugin-core? Does this functionality only apply to CLI flow for publishing right now?

packages/pods-core/src/builtin/NextjsExportPod.ts
```diff
@@ -117,6 +117,16 @@ export type NextjsExportConfig = ExportPodConfig & NextjsExportPodCustomOpts;
 type NextjsExportPlantOpts = ExportPodPlantOpts<NextjsExportConfig>;
 
 export class NextjsExportPodUtils {
+  static git(baseDir: string) {
```
jonathanyeung, line 120: why do we need to expose this?

packages/engine-test-utils/src/__tests__/dendron-cli/commands/publishCLICommand.spec.ts
```diff
@@ -47,19 +49,124 @@ describe("WHEN run `dendron publish init`", () => {
...
+          expect(resp.error.message).toEqual("template exists");
+        },
+        {
+          expect,
+        }
+      );
+    });
+  });
+
+  // test("THEN succeed", async () => {
```
jonathanyeung, line 157: still needed?

jonathanyeung (changes requested):

packages/dendron-cli/src/commands/publishCLICommand.ts
kevinslin: we don't want to do a min in this case because `configVersion` is source of truth. if the engineVersion is lower, we need to ask the user to update the engine because they've published using a higher version. i've done a writeup of this here, will also share as an async > https://docs.google.com/document/d/1dZ3X039TqQyUMaRYx9KJSFVpJMEHgtjRbqdAdFDGWX8/edit

packages/dendron-cli/src/commands/publishCLICommand.ts
kevinslin: it does a few things outside of cloning so would keep the current name, especially since we already have a `this._cloneTemplate` method

- read config to get templateVersion
- clone template
- switch to correct version of template based on `templateVersion`
- install depdenencies
- write `templateVersion` to `dendron.yml` if not set

packages/pods-core/src/builtin/NextjsExportPod.ts
kevinslin: see the gdco. we are currently not using this process but want to adopt it to have versioned template updates

packages/pods-core/src/builtin/NextjsExportPod.ts
kevinslin: we don't. fixed!

packages/engine-test-utils/src/__tests__/dendron-cli/commands/publishCLICommand.spec.ts
kevinslin: nope

jonathanyeung (approved):

bpathakota (approved):

PR #3038, opened 2022-06-03
chore(workspace): sync quickstart-v1 tutorial

hikchoi:
> NOTE: First pass of the quickstart tutorial can be found here: dendronhq/dendron-site#527
> I took the liberty of changing around some wording and adding some useful links compared to the initial draft in gdoc.
> Please leave any feedbacks for the tutorial content in the PR in `dendron-site`. Once the copy is finalized I will sync the changes to this PR.

# chore(workspace): sync quickstart-v1 tutorial

This PR:
- ends tutorial a/b test for refactor mention.
- adds new a/b test for short-form quickstart tutorial.
- syncs newly treated `quickstart-v1` tutorial.
- enables command uris for note preview
- allows providing custom preview link handler for `PreviewPanelFactory`.
- implements custom `TutorialPreviewLinkHandler` to hook into `onLinkClicked` event for tracking purposes.
  - adds new event `TutorialEvent.TutorialPreviewLinkClicked` event and properties.
- adds new `LinkType.COMMAND` in `PreviewLinkHandler`.

## New event
- `TutorialEvent.TutorialPreviewLinkClicked`
  - properties: `linkType`, `href`
### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

## Docs

### Extended


packages/common-all/src/abTests.ts
```diff
@@ -65,49 +65,30 @@ export const GRAPH_THEME_FEATURE_SHOWCASE_TEST = new ABTest(
...
-  "main" = "main", // shows main tutorial
-  "no-refactor" = "no-refactor", // shows main tutorial, minus mentions to refactoring
-  "original" = "original", // shows old (origianl) tutorial
+export enum QuickstartTutorialTestGroups {
+  "main" = "main",
+  "quickstart-v1" = "quickstart-v1",
 }
 
 /**
- * Long running A/B test for differently treated tutorials.
```
hikchoi, line 89: fyi: I'm removing these since we decided (@jonathanyeung pointed out) that we shouldn't keep a long-running A/B test group that changes over time.

packages/plugin-core/assets/dendron-ws/tutorial/treatments/quickstart-v1/tutorial.md
```diff
@@ -0,0 +1,60 @@
+---
```
hikchoi, line 1: I did a first pass copy of the quickstart. This is synced from https://github.com/dendronhq/dendron-site/pull/527

If you have any feedback, please leave it in that PR.

personjerry:
What's the easiest way that I can try it out?

To be clear, the "tutorial workspace" in this context just becomes the one page?

packages/common-all/src/util/regex.ts
```diff
@@ -15,3 +15,9 @@ export function isWebUri(uri: string): boolean {
   if (scheme === "http" || scheme === "https") return true;
   return false;
 }
+
+export function isCommandUri(uri: string): boolean {
```
jonathanyeung, line 19: can we add a quick docstring for this util method? 

Also - this might be better if it has 'VSCode' somewhere in the function name (it's VSCode Command Uri right)?

packages/plugin-core/src/components/views/PreviewViewFactory.ts
```diff
@@ -15,13 +15,19 @@ export class PreviewPanelFactory {
...
+    extension: IDendronExtension,
+    linkHander?: IPreviewLinkHandler
+  ): PreviewProxy {
     // Simple singleton implementation, since we only want one preview panel at
     // any given time.
-    if (!PreviewPanelFactory._preview) {
+
+    // if preview panel doesn't exist yet, create a new one.
+    // if a custom link handler is provided, use that to create a new preview panel.
+    if (!PreviewPanelFactory._preview || linkHander) {
```
jonathanyeung, line 27: I don't think it's too correct to have a singleton implemented this way.  Can we remove the `|| linkHandler` part?  I think it should still function fine since the `PreviewLinkHandler` implementation only acts on that one preview ID.

packages/plugin-core/src/components/views/TutorialPreviewLinkHandler.ts
```diff
@@ -0,0 +1,39 @@
+import { TutorialEvents } from "@dendronhq/common-all";
+import { AnalyticsUtils } from "../../utils/analytics";
+import { LinkType, PreviewLinkHandler } from "./PreviewLinkHandler";
+
+export class TutorialPreviewLinkHandler extends PreviewLinkHandler {
+  /**
+   *  set of tutorial note ids that we will allow tracking of link clicked events.
+   */
+  private _allowedIds = new Set<string>(["c1bs7wsjfbhb0zipaywqfbg"]);
```
jonathanyeung, line 9: While I'm still relying on the hacky impl for workspace note telemetry, can we add this ID to `_extension.ts`? https://github.com/dendronhq/dendron/blob/55630a47455e960b01a7499de108d16a7ef832f1/packages/plugin-core/src/_extension.ts#L269

jonathanyeung (changes requested):

packages/common-all/src/util/regex.ts
kevinslin: instead of `vscode`, would use `code` (since it can also refer to `vscodium`)

hikchoi: > instead of vscode, would use code (since it can also refer to vscodium)

I think we can stick to `vscode` since every other places we just refer it to vscode. (i.e. `VSCodeUtils`)

I propose that we keep it consistent and later discuss if we want to change the naming scheme, and then bulk refactor everything that refers to it as `vscode`

packages/plugin-core/src/components/views/PreviewViewFactory.ts
hikchoi: merging `TutorialPreviewLinkHandler` and `PreviewLinkHandler` logic for now as per sync discussion

packages/plugin-core/src/components/views/PreviewViewFactory.ts
```diff
@@ -18,6 +18,9 @@ export class PreviewPanelFactory {
   public static create(extension: IDendronExtension): PreviewProxy {
     // Simple singleton implementation, since we only want one preview panel at
     // any given time.
+
+    // if preview panel doesn't exist yet, create a new one.
+    // if a custom link handler is provided, use that to create a new preview panel.
```
jonathanyeung, line 23: nit: comment is stale now.

jonathanyeung (approved):

packages/plugin-core/src/components/views/PreviewViewFactory.ts
hikchoi: fixed.

PR #3042, opened 2022-06-04
chore: add `create_test_vault` dev method, fix issue with `createWorkspace`, other backend changes

kevinslin:
chore: add `create_test_vault` dev method, fix issue with `createWorkspace`, other backend changes

- introduce `dendron dev create_test_vault` which can create vaults with pre-configured distribution of notes/vaults (useful for diagnosing performance issues)
- fix issue with `createWorkspace` cmd - when passing in multiple vaults to a self contained workspace, only the first vault is actually added to the `dendron.yml`
  - this led to a larger refactoring of the `createWorkspace` to make a distinction between the `wsVault` (for sc vault, this is the vault in `notes` folder) and `additionalVaults`
- also modified `WorkspaceInitializer` interface to account for `wsVault` and `additionalVaults`
- cleaned up some tests

### DIFF
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -78,7 +78,16 @@ export type PathExistBehavior = "delete" | "abort" | "continue";
 
 export type WorkspaceServiceCreateOpts = {
   wsRoot: string;
-  vaults?: DVault[];
+  /**
+   * Does workspace come with a vault?
+   * - for self contained vault, this is the `notes` folder
+   * - for non-self contained vault, this is whatever the user passes in
+   */
+  wsVault?: DVault;
+  /**
+   * Additional vaults to create
+   */
+  additionalVaults?: DVault[];
   /**
    * create dendron.code-workspace file
    */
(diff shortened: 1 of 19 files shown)


packages/engine-server/src/workspace/service.ts
```diff
@@ -1005,7 +1017,13 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
   }
 
   static async createStandardWorkspace(opts: WorkspaceServiceCreateOpts) {
-    const { wsRoot, vaults } = opts;
+    const { wsRoot, wsVault, additionalVaults } = _.defaults(opts, {
+      additionalVaults: [],
+    });
+    // for a standard workspace, there is no difference btw a wsVault and any other vault
+    const vaults: DVault[] = [wsVault, ...additionalVaults].filter(
+      (v): v is DVault => !_.isUndefined(v)
```
SeriousBug, line 1025: nitpick: `isNotUndefined` helper would also typecheck here

SeriousBug (approved):

PR #3043, opened 2022-06-04
enhance/better completion

kevinslin:
- refactor lookup utilities
- refactoring
- add lookupv2 method


jonathanyeung:
Why `dendron note find` and not `dendron lookup note` to be consistent?

packages/engine-server/src/utils/engine.ts
```diff
@@ -0,0 +1,142 @@
...
+} from "@dendronhq/common-all";
+import fs from "fs-extra";
+import _ from "lodash";
+import path from "path";
+
+export function openPortFile({ fpath }: { fpath: string }): number {
+  return _.toInteger(_.trim(fs.readFileSync(fpath, { encoding: "utf8" })));
+}
+
+export class EngineUtils {
```
jonathanyeung, line 20: can we rename this file to EngineUtils.ts? I know it's redundant given the path it's in, but it helps me tremendously when viewing code when the file name matches the class / functionality. Otherwise, when I see the file name I'll think this is some `DEngine` implementation.

packages/dendron-cli/src/commands/notes.ts
```diff
@@ -32,28 +36,110 @@ type CommandOpts = CommandCLIOpts & SetupEngineResp & CommandCommonProps;
...
+  stringOutput: string;
+  notesOutput: NoteProps[];
+};
+
 export enum NoteCommands {
+  /**
+   * Like lookup, but only look for notes
+   * Returns a list of notes
+   */
+  FIND = "find",
```
jonathanyeung, line 52: I'm not a fan of introducing the terminology of 'find'.  It's inconsistent with 'lookup'.  This is unintuitive for users coming from the plugin.  I'd rather have `lookup` take additional parameters to specify note vs schema lookup.

packages/plugin-core/src/features/completionProvider.ts
```diff
@@ -105,7 +105,7 @@ export const provideCompletionItems = sentryReportingCallback(
     }
 
     const line = document.lineAt(position).text;
-    Logger.debug({ ctx, position, msg: "enter" });
```
jonathanyeung, line 108: Did you mean to leave this change in?  This is pretty noisy right?

jonathanyeung (changes requested):

packages/dendron-cli/src/commands/notes.ts
SeriousBug: If we don't mind a breaking change, one option would be to name this lookup, and name the old lookup `goto` or something similar. That's more similar to what the old lookup CLI command does.

packages/dendron-cli/src/commands/notes.ts
```diff
@@ -96,53 +182,55 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+      output: NoteCLIOutput.JSON,
+    });
 
     try {
       switch (cmd) {
+        case NoteCommands.FIND: {
+          const { query } = checkQuery(opts);
+          const notes = await NoteLookupUtils.lookup({ qsRaw: query, engine });
+          const resp = await formatNotes({
+            output: NoteCLIOutput.JSON,
```
SeriousBug, line 195: Doesn't respect the selected output format

SeriousBug (changes requested):
Looks good, just one bug and asking for additional tests.

packages/engine-server/src/utils/engine.ts
kevinslin: hmm, fine renaming this but do we want to have a broader conversation around this? this pattern is used frequently across the codebase. we should adopt one of the following:
- always use `utils/{x}`
- always use `utils/{x}Utils`
- always use `utils/{x}` unless it conflicts with an existing file in which case we apply `utils/{x}Utils`

Also - while we're on it, we should also have a conversation around casing since we use it inconsistently. when to uppercase vs lowercase. I think everything is fine as long as we're consistent. here is airbnb's style guide on naming as an example: https://github.com/airbnb/javascript#naming--filename-matches-export

packages/dendron-cli/src/commands/notes.ts
kevinslin: i've switched this to `lookup` and renamed old lookup to `lookup_legacy`. we actually want to deprecate/remove `getNoteByPath` and add seperate `note get` and `note create` methods. there's less than 7 people currently using this method. will update docs for deprecation notice

packages/plugin-core/src/features/completionProvider.ts
kevinslin: yeah - its helpful to debug. completion provider only gets triggered on triggers so not that noisy. but will switch back to debug

jonathanyeung (approved):
Looks good, but I think the docs also needs updates to reflect the new commands.

SeriousBug (approved):

PR #3047, opened 2022-06-06
chore: automate the schema updates

SeriousBug:
This PR automates the updates to the JSON schema files by doing the updates during the release process. The schema updates should get merged back in along with the PR after the release. The updates are also done for the nightly version, which won't get merged but will keep the schema up-to-date in the nightly versions as well.


kevinslin (approved):

kevinslin:
nice! great example of `Don't Rely on Good Intentions` :)

PR #3048, opened 2022-06-06
add dendron-viz pkg

kevinslin:


SeriousBug (approved):

SeriousBug:
Actually, do you also need to modify the `dendron-cli`s `package.json` so the CLI can access this package?

kevinslin:
> Actually, do you also need to modify the `dendron-cli`s `package.json` so the CLI can access this package?

yeah, for that, i'll leave it to you/hayata since this pr only introduces the dendron-viz package

PR #3049, opened 2022-06-06
chore: use self contained vaults for all tests

SeriousBug:
This PR sets self contained vaults to be the default in all plugin tests, which better reflects our current default.

Most tests function correctly after the change, others have been fixed, and for a couple tests that I though was non-critical I opted to skip them for now so we can update the default sooner than later.


kevinslin (commented):
looks good - can we add a comment next to the skipped tests to add rationale (eg. skipping to iterate on using sc vaults as default. will come back and fix later)

SeriousBug:
@kevinslin Thanks, added the comments. I'll also create a task and add it as a next item to self contained vaults project.

kevinslin (approved):

PR #3052, opened 2022-06-07
feat(views): Recent Workspaces Panel

jonathanyeung:
## feat(views): Recent Workspaces Panel

Adds a panel displaying recently opened Dendron workspaces.  This panel will show the last 5 opened Dendron workspaces.  If a user has never opened a workspace before, then they'll get a message to get started (see below). Clicking on an item will open the workspace. 

This only shows up if not in a currently active Dendron workspace. This is to prevent clutter in the side bar.  If a user wants to manage multiple projects, then I think we should point them to the Project Manager extension, which is a fully featured version of the same concept.

I also had to make some adjustments in LaunchTutorial code + some telemetry tweaks.

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics
Added to 'Upcoming'

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs


packages/plugin-core/src/constants.ts
```diff
@@ -872,15 +897,32 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
...
+    key: "dendron.launchTutorialWorkspace",
+    title: `${CMD_PREFIX} Launch Tutorial Workspace`,
+    when: "false",
+  },
   OPEN_BACKUP: {
     key: "dendron.openBackup",
     title: `${CMD_PREFIX} Open Backup`,
     when: DendronContext.PLUGIN_ACTIVE,
   },
+  INSTRUMENTED_WRAPPER_COMMAND: {
```
kevinslin, line 921: can you add a comment to what this means?

packages/plugin-core/src/commands/LaunchTutorialCommand.ts
```diff
@@ -10,6 +10,10 @@ type CommandInput = {};
 
 type CommandOutput = void;
 
+/**
+ * This command is a bit of a misnomer - it actually launches the welcome
+ * webview page
+ */
 export class LaunchTutorialCommand extends BasicCommand<
```
kevinslin, line 17: can we rename it in that case?

packages/plugin-core/src/commands/InstrumentedWrapperCommand.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/**
+ * This command is a simple wrapper around commands, but this wrapper also fires
+ * a telemetry event when it is invoked. This is intended to be used as a
+ * wrapper around built-in VSCode commands that need to be invoked via a command
+ * URI, such as within webviews or in TreeView items, but where we still want a
+ * telemetry event to get reported.
+ */
+export class InstrumentedWrapperCommand extends BasicCommand<
```
kevinslin, line 30: can we do an async and add it to the instrumentation cookbook after the pr has been merged?

packages/plugin-core/src/features/RecentWorkspacesTreeview.ts
```diff
@@ -0,0 +1,75 @@
...
+      command: {
+        title: "Open Workspace",
+        command: "vscode.openFolder",
+        arguments: [Uri.file(element.fsPath)],
+      },
+      event: VSCodeEvents.RecentWorkspacesPanelUsed,
+    };
+
+    return {
+      label: element.fsPath,
```
kevinslin, line 37: in terms of ergonomics, the whole path to the note overflows the sidebar which is not that great: https://www.loom.com/i/b8d556dc80584b0da3caf2b2b8073c1f
can we list vaults by their names instead (and tooltip to the full path)?

packages/engine-server/src/metadata/service.ts
```diff
@@ -233,6 +237,10 @@ export class MetadataService {
     return this.getMeta().priorTools;
   }
 
+  get RecentWorkspaces(): string[] | undefined {
```
kevinslin, line 240: we have a mix of `get {Prop}` and `get{Prop}` calls here (and everywhere). can we have an async to standardize? ()

kevinslin (changes requested):
looking forward to this feature. some callouts:
- we are calling this the `recent workspaces panel` - moving forward with self contained vaults, are we still using that terminology even if everything is a vault? should this just be recent vaults?
- there is a difference btw vaults that exist independently and vaults that exist as dependencies to the current vault. we should think about the UX for this even if we're not implementing it now. currently, the only way to switch between vaults consistently is to look up the `root` note of a particular vault. does it make sense to show `dependent vaults` here?
- in terms of clutter on the sidebar, can we not just minimize the `recent` panel if in an workspace vs hiding it completely?

packages/plugin-core/src/features/RecentWorkspacesTreeview.ts
jonathanyeung: I considered this but thought it'd be more informative to a new user to list the file path as opposed to the vault name - the main intention here is to make sure new users can get back into Dendron if they had closed it out. Having this be a convenience to opening vaults/workspaces for experience users is a secondary goal.

packages/plugin-core/src/commands/InstrumentedWrapperCommand.ts
jonathanyeung: sure

packages/plugin-core/src/constants.ts
jonathanyeung: The comment is in `InstrumentedWrapperCommand.ts`

packages/plugin-core/src/commands/LaunchTutorialCommand.ts
jonathanyeung: sure - at first I left it b/c I thought it would interfere with our existing charts, but it looks like we don't look at this.

jonathanyeung:
> we are calling this the recent workspaces panel - moving forward with self contained vaults, are we still using that terminology even if everything is a vault? should this just be recent vaults?

> there is a difference btw vaults that exist independently and vaults that exist as dependencies to the current vault. we should think about the UX for this even if we're not implementing it now. currently, the only way to switch between vaults consistently is to look up the root note of a particular vault. does it make sense to show dependent vaults here?

I think maybe we're slightly misaligned on the spec for this particular feature. The main intention of _this_ change is to make sure new users can get back into Dendron if they had closed it out. Having this be a convenience to opening vaults/workspaces for experienced users is a secondary goal.

With this in mind, the list of things that appear in this panel are either workspace files or workspace folders that Dendron recognizes.  So for accuracy, I think this should be called 'Recent Workspaces' and not 'Recent Vaults'.  This list represents the VS Code concept of workspace/folder, not Dendron Vaults.

> in terms of clutter on the sidebar, can we not just minimize the recent panel if in an workspace vs hiding it completely?

For the reason listed above, I think it actually makes more sense to not be displayed when you're in a Dendron workspace.

kevinslin (approved):

PR #3057, opened 2022-06-07
feat(cli): add a cli command that generates a packed-circles visualization of workspace

hayata-suenaga:
# Visualize CLI command

*[[Add a CLI command that will generate the visualization graph|task.2022.06.03.add-a-cli-command-that-will-generate-the-visualization-graph]]*

CLI command to generate a svg file per vault that contains packed circles visualization.


packages/dendron-viz/package.json
```diff
@@ -25,14 +25,34 @@
...
-    "buildCI": "yarn compile",
-    "compile": "tsc -p tsconfig.build.json "
+    "build": "yarn copyNonTSFiles && yarn compile",
+    "buildCI": "yarn copyNonTSFiles && yarn compile",
+    "copyNonTSFiles": "mkdir -p lib && cp ./src/loadModule.js lib/",
+    "compile": "tsc -p tsconfig.build.json",
+    "watch": "yarn copyNonTSFiles && yarn compile --watch"
   },
   "dependencies": {
+    "@actions/artifact": "^0.5.2",
```
kevinslin, line 35: you probably don't need the `@actions/*` dependency. the github action will be in a separate package when we get to it

packages/dendron-viz/src/dendron.tsx
```diff
@@ -0,0 +1,89 @@
...
+  console.log(JSON.stringify(resp, null, 2));
+}
+
+export type InputArgs = {
+  wsRoot: string;
+  out?: string;
+};
+
+export function generateSVG(args: InputArgs) {
+  main(args);
```
kevinslin, line 88: this should return so that the calling function can `await` the return value

packages/dendron-viz/src/process-dir.js
```diff
@@ -0,0 +1,68 @@
+import fs from "fs";
+import * as nodePath from 'path';
+import { shouldExcludePath } from './should-exclude-path';
```
kevinslin, line 3: why is this a `js` file?

hayata-suenaga: removed `process-dir.js` as it is not used

packages/dendron-viz/package.json
```diff
@@ -25,14 +25,34 @@
...
+    "react-dom": "^17.0.2",
+    "yargs": "^17.4.1"
   },
   "devDependencies": {
-    "rimraf": "^2.6.2"
+    "rimraf": "^2.6.2",
+    "@types/jest": "^27.0.1",
+    "@types/micromatch": "^4.0.2",
+    "husky": "^7.0.0",
+    "jest": "^27.0.6",
```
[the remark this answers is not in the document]
hayata-suenaga: also removed `husky`

packages/dendron-viz/package.json
hayata-suenaga: Removed the use of @actions/core from `dendron.tsx`

SeriousBug:
@hayata-suenaga Just a heads up that you'll want to resolve the merge conflicts. Github should have a "Resolve conflicts" button at the bottom of this page, or you can follow the merge instructions on [this tutorial](https://kalkus.dev/2020/05/07/how-to-update-feature-branch-from-master-in-git/).

packages/dendron-viz/src/dendron.tsx
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env node
 
 import { VaultUtils } from "@dendronhq/common-all";
-import { DendronEngineV2 } from "@dendronhq/engine-server";
 import fs from "fs";
 import React from "react";
 import ReactDOMServer from "react-dom/server";
 import { processDir } from "./process-dendron-notes";
 import { createTree } from "./Tree";
+import { VisualizeCLICommandOpts } from "@dendronhq/dendron-cli";
```
SeriousBug, line 9: You're introducing a circular dependency here: `dendron-viz` imports `dendron-cli` and `dendron-cli` imports `dendron-viz`.

You should write a new type in `dendron-viz` that just lists what `generateSVG` needs. That will be more flexible since you'll probably need to reuse this code in the plugin too.

hayata-suenaga: I'll fix it.

packages/dendron-cli/src/commands/visualizeCLICommand.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type CommandOpts = InputArgs & SetupEngineResp & CommandCommonProps;
+
+export { CommandOpts as VisualizeCLICommandOpts };
+
+export class VisualizeCLICommand extends CLICommand {
+  constructor() {
+    super({
+      name: "visualize",
+      desc: "generates a fingerprint visualization of Dendron workspace",
```
SeriousBug, line 14: It looks like "packed circles" is the right term for this type of visualization, can we use that name for the description: https://datavizproject.com/data-type/packed-circle-chart/

packages/dendron-viz/src/language-colors.json
```diff
@@ -0,0 +1,1053 @@
+{
```
SeriousBug, line 1: Do we need this file? It's pretty big and I don't think it makes sense for us since we're not visualizing different programming languages. We can look at how to do colors later.

If it is hard to remove, it's fine to keep it for now.

packages/dendron-viz/src/should-exclude-path.ts
```diff
@@ -0,0 +1,29 @@
+import { isMatch } from "micromatch";
```
SeriousBug, line 1: For this and other files, could we follow `camelCase` for the file name instead of `kebab-case`? I think that's the standard in the code base.

SeriousBug (changes requested):
Looks good to me, just a few small comments. Good work!

jest.config.js
```diff
@@ -7,6 +7,10 @@ const commonConfig = {
   snapshotSerializers: ["jest-serializer-path"],
   testEnvironment: "node",
   testPathIgnorePatterns: ["utils.ts"],
+  transformIgnorePatterns: [
+    // These are ESM modules that need to be transpiled before Jest can run them
```
kevinslin, line 11: do you mean the opposite? these are esm modules that *should not be transpiled*

kevinslin (commented):
nice job! just left one small comment

jest.config.js
hayata-suenaga: @kevinslin Because `jest` transforms test codes to `commonjs`, the transpiled test code was having trouble in importing esm module libraries. That's why we decided to let `jest` transform esm module libraries to `commonjs` libraries so that they can be imported using `require`.

Because `jest` by default ignores all `node_modules`, we reset the pattern to ignore everything except `esm module` libraries that need to be transpiled.

SeriousBug (approved):

hayata-suenaga:
[PR for `dendron-site` doc]( https://github.com/dendronhq/dendron-site/pull/559)

SeriousBug:
@kevinslin @jonathanyeung Could you do a review of this as well when you have the time so we can get it merged?

hayata-suenaga:
For documentation, I mistakingly pushed my commit to `dev` branch directly. The link to my note is here:
[https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.topic.visualize.cli.md](https://github.com/dendronhq/dendron-site/blob/dev/vault/dendron.topic.visualize.cli.md)

yarn.lock
```diff
@@ -2420,6 +2420,28 @@
     lodash "^4.17.21"
     node-fetch "^2.6.1"
 
+"@dendronhq/common-all@^1.00.0":
```
kevinslin, line 2423: how was dendronhq added to dependencies? if you add using `lerna bootstrap`, this should not be here. can you reset the yarn.lock and try running lerna bootstrap instead?

kevinslin:
> 

other timeouts are okay, they're known and we're working on fixing them. is 5e4 right? its on the high side. can we try 3e4 first?

hayata-suenaga:
> > 
> 
> other timeouts are okay, they're known and we're working on fixing them. is 5e4 right? its on the high side. can we try 3e4 first?

@kevinslin I decreased the timeout amount from 5e4 to 3e4. Timeout issues are not happening. However, in windows environments, the build for `dendron-viz` seems to be failing. Would this be a problem?

[https://github.com/dendronhq/dendron/runs/7013826518?check_suite_focus=true#step:9:42]([](https://github.com/dendronhq/dendron/runs/7013826518?check_suite_focus=true#step:9:42))

kevinslin:
looks like test are failing

bpathakota (approved):
commit title looks good!

PR #3059, opened 2022-06-08
chore: make lookup control appear faster

jonathanyeung:
## chore: make lookup control appear faster

There's a lag between hitting Ctrl+L and the lookup quick pick showing up when you have a lot of notes.  This causes a lot of stray keystrokes to get added in notes.  This change attempts to fix the initial lag.

---

### Basics

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested


SeriousBug (approved):
I tried very hard to beat the lookup opening, but it opens too fast now :D Looks good!

PR #3060, opened 2022-06-08
enhance(views): add tooltip on hover for graph nodes

Harshita-mindfire:
This PR is one of the follow-up items from Graph BML learnings. It aims to display the note's fname as a tooltip on graph node hover.
This is an 80-20 solution for better visualization of dated hierarchies.

**BREAKING CHANGE**
Added 2 new dependencies to `dendron-plugin-views`
- tippy.js : 6.3.7
- cytoscape-popper: 2.0.0
***
image

### Basics

### Extended
- General
- Operations
- Architecture

### Basics
NA

### Extended
NA
### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA

## Docs
NA

### Extended


jonathanyeung (approved):
cool! lgtm - I'm down to try this out and iterate.

SeriousBug (commented):
Left one comment on code quality.

I do have a question, would it be possible to use [AntD's popover](https://ant.design/components/popover/) to do this without adding new dependencies? I'm just not sure how much benefit we are getting out of the new deps, we can keep them if they are useful.

Harshita-mindfire:
> I do have a question, would it be possible to use [AntD's popover](https://ant.design/components/popover/) to do this without adding new dependencies? I'm just not sure how much benefit we are getting out of the new deps, we can keep them if they are useful.

`Tippy.js` is a recommended dependency to integrate with `cytoscape-popper`. Integrating with AntD can be tricky since we need to provide the reference of node to the tooltip. cytoscape-popper also expects the return value to be a DOM element.

Issue #3039, opened 2022-06-03
Unable to prevent Dendron from treating @ as a user tag


bugfixd:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [X] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [X] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

I sometimes want to prevent Dendron from treating an "at sign" followed by a word (e.g. `@hello`) as a user tag.

I've tried escaping the @ sign, e.g. `\@hello`, but Dendron still treats it as a user tag.

My current workaround is to enclose the whole thing in backticks, e.g. \`@hello\`.

Note: This is not exactly a bug — maybe just an oversight? Escaping works with hashtags (e.g. `\#not-a-tag`) and links (e.g. `\[not a link]`), and so imho it should also work with @ signs. 😉

## Expected behavior

I would like to be able to escape an @ sign to prevent Dendron from interpreting it as a user tag.

## Desktop (please complete the following information)

- OS: Windows 10 Pro
- Browser (if with published websites; include version): Chrome
- Versions (all that apply to bug):
  - Dendron Extension: 0.97.0
  - VS Code version: 1.67.2

SeriousBug:
Thanks for the report, I'll take a look!

I didn't know hashtags supported escaping actually :sweat_smile: but it should be easy enough to add to user tags too.

bugfixd:
@SeriousBug ,

I've noticed a related issue:

When a word has an @ sign in the *middle* of it, Dendron treats it as a user tag — although it correctly handles words that look like email addresses:

![image](https://user-images.githubusercontent.com/48696524/172389716-6c49ea07-9689-40ef-8423-9a457b809f9d.png)

SeriousBug:
Thanks, I can recreate the issue. The same actually seems to happen with hashtags, `test#ing` will cause the `#ing` part to be recognized as a tag.

I'm pretty sure this used to work, I think our parser changed how it works so it's not functioning anymore. In any case, I'm working on a fix for this. I'll also add support for escaping.

SeriousBug:
[...]

PR #3064, opened 2022-06-08
fix(edit): email addresses and hash symbols inside words are parsed as tags

SeriousBug:
For example `foo@bar` or `foo#bar` would be parsed as user or hash tags. This PR fixed this issue by requiring tags to be preceded by a space, or to start at the beginning of a line.

#3039


bpathakota:
@SeriousBug: Would either tie this to #scope.structure (since it relates to tags) or #scope.edit (since it impacts the editing experience). Let me know if you have any other thoughts on this!

kevinslin (approved):
looks good, can you update the commit message as per bindhu's comment?

SeriousBug:
Thanks Bindhu! Updated and merging.

PR #3072, opened 2022-06-10
feat(views): depth filter for local graph

Harshita-mindfire:
In the Graph Filter menu, an additional filter for "depth" has been added. The depth value varies from 1(default) through 3. This is to improve the usability of the local graph view. This value is also saved so that the graph opens with the same depth the next time.
Use the graph panel buttons to update the graph depth. The depth is saved back to meta.json file when panel is Collapsed

As the graph becomes dense with incoming and outgoing links, it begins to look messy. This may be resolved by adding more filters to show/hide incoming/outgoing links.

Level 1 (default) displays: active note, its parent, children and links
Level 2 displays active note, parent, grandparent, children, grandchildren, active note's links, active note's links' links

Docs link: https://github.com/dendronhq/dendron-site/pull/539

### Basics

### Extended
- General
- Operations
- Architecture

### Basics
Added a property `graphDepth` for an existing event `GraphViewUsed`

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -176,6 +176,7 @@ export enum GraphEvents {
   GraphThemeChanged = "GraphThemeChanged",
   GraphViewUsed = "GraphViewUsed",
   GraphPanelUsed = "GraphPanelUsed",
+  GraphDepthChanged = "GraphDepthChanged",
 }
 
 export enum TreeViewEvents {
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -176,7 +176,6 @@ export enum GraphEvents {
   GraphThemeChanged = "GraphThemeChanged",
   GraphViewUsed = "GraphViewUsed",
   GraphPanelUsed = "GraphPanelUsed",
-  GraphDepthChanged = "GraphDepthChanged",
 }
 
 export enum TreeViewEvents {
--- a/packages/dendron-plugin-views/src/components/DendronGraphPanel.tsx
+++ b/packages/dendron-plugin-views/src/components/DendronGraphPanel.tsx
@@ -74,6 +74,20 @@ const DendronGraphPanel: DendronComponent = (props) => {
   }, [ide.graphTheme]);
 
   useEffect(() => {
+    if (ide.graphDepth) {
+      logger.log("updating graph depth in config ", ide.graphDepth);
+      setConfig((c) => ({
+        ...c,
+        "filter.depth": {
+          ...c["filter.depth"],
+          value: ide.graphDepth!,
+        },
+      }));
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [ide.graphDepth]);
+
+  useEffect(() => {
     if (isSidePanel) {
       setConfig((c) => ({
         ...c,
--- a/packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
+++ b/packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
@@ -42,6 +42,7 @@ type QueueData = {
    */
   distance: number;
   classes: string;
+  isParent?: boolean;
 };
 
 function computeGraphElements({
@@ -102,9 +103,35 @@ function computeGraphElements({
     });
 
     if (data.distance < maxDistance) {
-      const children = note.children;
-
       const noteVaultClass = getVaultClass(note.vault);
+
+      const parentNote =
+        (data.distance === 0 || data.isParent) && note.parent
+          ? notes[note.parent]
+          : undefined;
+      if (parentNote) {
+        nodesQueue.enqueue({
+          note: parentNote,
+          distance: data.distance + 1,
+          isParent: true,
+          classes: `${DEFAULT_NODE_CLASSES} parent ${getVaultClass(
+            parentNote.vault
+          )}`,
+        });
+        edges.hierarchy.push({
+          data: {
+            group: "edges",
+            id: `${parentNote.id}_${note.id}`,
+            source: parentNote.id,
+            target: note.id,
+            fname: parentNote.fname,
+            stub: _.isUndefined(parentNote.stub) ? false : parentNote.stub,
+          },
+          classes: `${DEFAULT_EDGE_CLASSES} hierarchy ${noteVaultClass}`,
+        });
+      }
+
+      const children = note.children;
 
       // Setup the edges for children now:
       children.forEach((child: string) => {
--- a/packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
+++ b/packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
@@ -67,8 +67,6 @@ function computeGraphElements({
   const nodes: GraphNodes = [];
   const activeNote = notes[noteActive.id];
 
-  // TODO: Add parents and grandparents
-
   // Add descendents (children, grandchildren, etc) depending on the specified
   // distance parameter:
   const nodesQueue = new FIFOQueue<QueueData>();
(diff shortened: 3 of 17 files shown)


packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
```diff
@@ -34,166 +35,246 @@ const DEFAULT_CLASSES = ""; //graph-view
...
+   */
+  distance: number;
+  classes: string;
+  /**
+   * is node an ancestor of the active note
+   */
+  isParent?: boolean;
+};
+
+function computeGraphElements({
```
jonathanyeung, line 51: Gave the graph view a test, and I don't quite recall what we had discussed is the functionally right thing to do, but I _think_ the view will be more useful if we treat hierarchies and links as different types of nodes - meaning that if we set depth = 2, then we'd include the following nodes: 

- parent, grandparent
- children, grandchildren
- active note's links
- active note's links' links

In this proposal, things we will **not** include are:
- siblings, cousins, aunts, uncles, nieces, nephews, etc. If a user wants to view siblings, then they can navigate to the parent note instead (and the sibling will show up as the parent's children)
- parent's links, grandparent's links, childrens' links, etc.
- a link's parent/child. 

What do you think about this proposal?

jonathanyeung (changes requested):
One issue I noticed is that in full graph view, if I try to manually type a number into the 'depth' text box, it'll disappear.  It works if I click the up and down arrows though.

packages/common-all/src/analytics.ts
```diff
@@ -170,6 +170,7 @@ export enum GraphEvents {
   GraphThemeChanged = "GraphThemeChanged",
   GraphViewUsed = "GraphViewUsed",
   GraphPanelUsed = "GraphPanelUsed",
+  GraphDepthChanged = "GraphDepthChanged",
```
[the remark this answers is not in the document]
Harshita-mindfire: updating it as a part of Graph View Used, since we update filter from Graph View

Harshita-mindfire:
> One issue I noticed is that in full graph view, if I try to manually type a number into the 'depth' text box, it'll disappear. It works if I click the up and down arrows though.

Thanks for catching it!

packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
Harshita-mindfire: updated.

packages/dendron-plugin-views/src/components/DendronGraphPanel.tsx
```diff
@@ -73,6 +73,20 @@ const DendronGraphPanel: DendronComponent = (props) => {
...
   }, [ide.graphTheme]);
 
+  useEffect(() => {
+    if (ide.graphDepth) {
+      logger.log("updating graph depth in config ", ide.graphDepth);
+      setConfig((c) => ({
+        ...c,
+        "filter.depth": {
+          ...c["filter.depth"],
+          value: ide.graphDepth!,
```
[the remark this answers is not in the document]
jonathanyeung: ah, misread the code then - please ignore.

packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
```diff
@@ -34,166 +35,333 @@ const DEFAULT_CLASSES = ""; //graph-view
...
+            source: parentNote.id,
+            target: note.id,
+            fname: parentNote.fname,
+            stub: _.isUndefined(parentNote.stub) ? false : parentNote.stub,
+          },
+          classes: `${DEFAULT_EDGE_CLASSES} hierarchy ${noteVaultClass}`,
+        });
+      }
+      // do not include children of parent/grandparent
+      const children = data.isParent ? [] : note.children;
```
jonathanyeung, line 134: nit: calling this children is a bit misleading.  Maybe better would be to do something like 

```typescript
// do not include children of parent/grandparent
if(!data.isParent) {
children.forEach(...);
}
```

packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
```diff
@@ -34,166 +35,333 @@ const DEFAULT_CLASSES = ""; //graph-view
...
+              id: `${note.id}_${childNote.id}`,
+              source: note.id,
+              target: childNote.id,
+              fname: note.fname,
+              stub: false,
+            },
+            classes: `${DEFAULT_EDGE_CLASSES} hierarchy ${noteVaultClass}`,
+          });
+
+          nodesQueue.enqueue({
```
jonathanyeung, line 152: In this case, won't we iterate over the active note multiple times, since active note === active note's child's parent?

packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
```diff
@@ -34,166 +35,333 @@ const DEFAULT_CLASSES = ""; //graph-view
...
-        fname: activeNote.fname,
-        color: getNoteColor({ fname: activeNote.fname, notes }),
-        stub: _.isUndefined(activeNote.stub) ? false : activeNote.stub,
-        localRoot: true,
-      },
-      classes: `${DEFAULT_NODE_CLASSES} ${getVaultClass(activeNote.vault)}`,
-      selected: true,
-    },
-  ];
+  // Breadth First Search for Descendants
```
jonathanyeung, line 80: 'for Descendants' is no longer accurate.

jonathanyeung (changes requested):
Just a few minor nits. 

A few usability things I observed, which we can address later
- reloading window causes the filter setting to not get saved, as it likely won't get to save in time.  I think we should just persist it every time the user toggles the setting.
- Adjusting the filter level in the view doesn't cause the panel to update the filter level until you reload the view.  This one may be tricky to implement, so we can defer it.

Harshita-mindfire:
TODO: grey out graph depth update icons when depth is at its max/min

packages/dendron-plugin-views/src/hooks/useGraphElements.tsx
Harshita-mindfire: I don't think so, this is because we add a property `isParent: true` for active note's parent and only enqueue parent if it is either active note's parent or its ancestor. Parent of child notes are not added to the queue.

packages/plugin-core/src/views/GraphPanel.ts
```diff
@@ -25,15 +26,46 @@ export class GraphPanel implements vscode.WebviewViewProvider {
...
   constructor(extension: IDendronExtension) {
     this._ext = extension;
     this._ext.context.subscriptions.push(
       vscode.window.onDidChangeActiveTextEditor(this.onOpenTextDocument, this)
     );
   }
 
-  public postMessage(msg: DMessage) {
-    this._view?.webview.postMessage(msg);
+  get graphDepth(): number | undefined {
```
jonathanyeung, line 38: nit: add 'private' in front of the get and set.

packages/plugin-core/src/workspace.ts
```diff
@@ -687,6 +682,29 @@ export class DendronExtension implements IDendronExtension {
     return backlinkTreeView;
   }
 
+  private setupGraphPanel() {
+    const graphPanel = new GraphPanel(this);
+
+    vscode.commands.registerCommand(
+      DENDRON_COMMANDS.GRAPH_PANEL_INCREASE_DEPTH.key,
+      sentryReportingCallback(() => {
+        // post message to DendronGraphPanel
```
jonathanyeung, line 691: nit: these comments are slightly out of date now

jonathanyeung (approved):

bpathakota:
@Harshita-mindfire - What is the main benefit of this feature we would want to highlight for the changelog? the saved value or is there more specific detail about improved usability? Would update the commit title to emphasize on the benefit of adding the depth filter.

PR #3081, opened 2022-06-13
enhance(retrieve): support xvault templates for gotonote command

tma66:
**Background**
There are multiple places where a template can be applied to a note (NoteLookup command and getNoteByPath). This leads to differing behaviours where some paths support xvault templates and others don't. We want to consolidate the template creation logic to one place (`template.ts`) and remove the template logic from `getNoteByPath`.

By updating GoTo Note command to use the same template logic used by NoteLookup command, we can also support xvault templates for this command.

See [[Support Multiple Templates in Gotonote|dendron://private/task.2022.05.24.support-multiple-templates-in-gotonote]]

**Changes**
1. Extract template logic from `NoteLookup` command to `template.ts`. This new method will find and apply a template tied to a given note (if applicable). If there are multiple template options, use callback to pick correct note
2. Remove `getNoteByPath` from engine and related apis
3. Update previous users of `getNoteByPath` to call updated template logic (GoToNote command, note CLI)
4. Update doctor action `CREATE_MISSING_LINKED_NOTES` to call `engineWrite` instead of `engineGetNoteByPath`

**Todo**
Update docs on removal of `getNoteByPath`


packages/dendron-cli/src/commands/notes.ts
```diff
@@ -205,31 +206,73 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
-              } as NoteCommandData,
-            };
+            // Until we support user prompt, pick template note for them if there are multiple matches in order of
+            // 1. Template note that lies in same vault as note to lookup
+            // 2. First note in list
+            await TemplateUtils.findAndApplyTemplate({
+              note,
+              engine,
+              pickNote: async (choices: NoteProps[]) => {
+                const sameVaultNote = choices.filter((ent) => {
```
SeriousBug, line 228: nitpick: Could use `choices.find` since we're trying to find only one matching

packages/engine-server/src/enginev2.ts
```diff
@@ -409,11 +408,13 @@ export class DendronEngineV2 implements DEngine {
     }
   }
 
+  /**
+   * @deprecated - See {@link DendronEngineV2.findNotes}
```
SeriousBug, line 412: It looks like there are very few, if any, usages of this. Could we drop this entirely if it's not too hard?

SeriousBug (commented):
Left a few comments, otherwise looks good.

packages/engine-server/src/enginev2.ts
kevinslin: +1 on this. it should only be used by goto note and the cli. are there any other usages @tma66 ?

kevinslin (commented):
looks good - nothing to add on top of comments from @SeriousBug

packages/common-all/src/util/stringUtil.ts
```diff
@@ -1,8 +1,24 @@
...
+import { CONSTANTS } from "../constants";
 
 /**
  * Returns levenshtein distance between the two strings, the higher the number
  * the further apart the strings are. 0 signals that the strings are equal. */
 export function levenshteinDistance(s1: string, s2: string): number {
   return levenshtein.get(s1, s2);
 }
+
+export function parseDendronURI(linkString: string) {
```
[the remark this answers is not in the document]
tma66: I actually moved that from LinkUtils to here. Since `LinkUtils` is part of `engine-server`, I felt it was easier to just move this function than the entire class

packages/common-server/src/template.ts
```diff
@@ -156,22 +160,69 @@ export class TemplateUtils {
...
+          return {
+            error: new DendronError({
+              message: `No vault found for ${vaultName}`,
+            }),
+          };
+        }
+      }
+
+      const maybeNotes = await engine.findNotes({ fname, vault: maybeVault });
+      const maybeTemplateNote = await pickNote(maybeNotes);
```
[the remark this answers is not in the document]
tma66: `pickNote` should be generic enough to handle list of notes of any size. I realize the doc is a little confusing so I'm going to fix that

packages/engine-server/src/enginev2.ts
tma66: I can remove it. I wasn't sure if we wanted to change the engine api as part of this change but I can go ahead and do that now

packages/dendron-cli/src/commands/notes.ts
tma66: good point, fixed

SeriousBug (approved):

PR #3082, opened 2022-06-13
fix(workspace): try to patch `EPERM` issues for windows

SeriousBug:
This PR tries to patch the `EPERM` issues some users reported on Windows. I can't reproduce the bug, but I patched all places that are likely to cause issues.

Looking at Sentry, it looks like the issue happens inside `_activate`, but that's as much as Sentry gives me because the real stack trace seems to get lost.

Also removed the deprecated `RespRequired` type, and updated the `info()` engine interface which was the only place where it was used.

### DIFF
--- a/packages/api-server/src/modules/notes/index.ts
+++ b/packages/api-server/src/modules/notes/index.ts
@@ -93,6 +93,14 @@ export class NoteController {
     getLogger().info({ ctx, msg: "enter" });
     try {
       const version = NodeJSUtils.getVersionFromPkg();
+      if (!version) {
+        return {
+          data: undefined,
+          error: DendronError.createPlainError({
+            message: "Unable to read the Dendron version",
+          }),
+        };
+      }
       return {
         data: {
           version,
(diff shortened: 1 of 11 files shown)


kevinslin (commented):
might be unrelated with this change but see this error consistently in the windows test - can you investigate? https://github.com/dendronhq/dendron/runs/6886427256?check_suite_focus=true

jonathanyeung (approved):
Thanks for the cleanup. 

Was there more context reported by the user as to what they were doing when they got the EPERM error? 

There are a ton of different EPERM errors in Sentry 😅

SeriousBug:
@kevinslin Sure. Do you have any context on those tests by the way, especially the one titled `THEN do right thing`, I'm not sure what it's supposed to be testing for.

kevinslin:
> @kevinslin Sure. Do you have any context on those tests by the way, especially the one titled THEN do right thing, I'm not sure what it's supposed to be testing for.

yeah, those tests were me converting legacy style tests to our new test harness. the description sucks because I was trying to time box the conversions. "do the right thing" translates into add the proper vaults and config updates (aka whatever the expect clauses are asserting)

PR #3087, opened 2022-06-14
enhance(workspace): make vault add command warn for transitive deps

SeriousBug:
This PR adds a check when a self contained vault is being added to the workspace. If the self contained vault has transitive dependencies, the user will be warned with a prompt which they'll need to acknowledge before we continue. I opted to have this blocking prompt because the user otherwise won't have enough time to see the prompt before the window reloads.

Docs: https://github.com/dendronhq/dendron-site/pull/545

### DIFF
--- a/packages/common-all/src/analytics.ts
+++ b/packages/common-all/src/analytics.ts
@@ -149,6 +149,7 @@ export enum ContextualUIEvents {
 export enum WorkspaceEvents {
   AutoFix = "AutoFix",
   DuplicateNoteFound = "DuplicateNoteFound",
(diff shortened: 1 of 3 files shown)


kevinslin (commented):
can we also write a guide on what to do in this situation in `dendron.topic.vaults.cook.transitive-vaults` and link it here?

the guide would go into adding the vaults into the dendron.yml of the current vault

SeriousBug:
@kevinslin Good idea. Added a PR for the docs, and added a button to the prompt to open the docs page.

kevinslin (approved):

PR #3089, opened 2022-06-15
fix(views): double click issue on help and feedback panel

jonathanyeung:
## fix(views): double click issue on help and feedback panel

There's an issue with the Help and Feedback panel where if you click an item multiple times, it'll only launch the link once.  This may make it feel like the button is broken. This change addresses that issue.

I also added an additional link to our FAQ page.

### DIFF
--- a/packages/plugin-core/src/features/HelpFeedbackTreeview.ts
+++ b/packages/plugin-core/src/features/HelpFeedbackTreeview.ts
@@ -1,4 +1,10 @@
 import {
+  assertUnreachable,
+  DendronTreeViewKey,
+  VSCodeEvents,
+} from "@dendronhq/common-all";
+import * as vscode from "vscode";
+import {
   Event,
   ProviderResult,
   ThemeIcon,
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/features/HelpFeedbackTreeview.ts
```diff
@@ -6,17 +12,12 @@ import {
...
-  DendronTreeViewKey,
-  VSCodeEvents,
-} from "@dendronhq/common-all";
-import { AnalyticsUtils } from "../utils/analytics";
+import { InstrumentedWrapperCommand } from "../commands/InstrumentedWrapperCommand";
 
 enum MenuItem {
   getStarted = "Get Started",
   readDocs = "Read Documentation",
+  seeFaqs = "See FAQ's",
```
kevinslin, line 20: nitpick: can we change to `See FAQ` (usually not expressed with the s)

kevinslin (approved):
looks good, just a nitpick about title

PR #3094, opened 2022-06-15
enhance(edit): add capture helper for hb templates

kevinslin:
enhance(edit): add capture helper for hb templates

This helper lets users capture parts of a text. Docs: https://github.com/dendronhq/dendron-site/pull/544

### DIFF
--- a/packages/common-server/src/template.ts
+++ b/packages/common-server/src/template.ts
@@ -121,6 +121,14 @@ class TemplateHelpers {
       const day = date.getDay();
       return day;
     },
+
+    capture: (text: string, pattern: string): string => {
+      const out = text.match(new RegExp(pattern, "i"));
+      if (out) {
+        return out[0];
+      }
+      return `ERROR: no match found for ${pattern} in ${text}`;
+    },
   };
 }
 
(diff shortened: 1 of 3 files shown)


packages/common-server/src/template.ts
```diff
@@ -121,6 +121,14 @@ class TemplateHelpers {
       const day = date.getDay();
       return day;
     },
+
+    capture: (text: string, pattern: string): string => {
```
jonathanyeung, line 125: Let's also modify `genTrackPayload` in this file to get data for this helper method.

packages/engine-test-utils/src/__tests__/common-all/template.spec.ts
```diff
@@ -129,6 +129,41 @@ describe(`WHEN running applyTemplate tests`, () => {
       });
     });
 
+    describe("AND WHEN using capture helper", () => {
```
jonathanyeung, line 132: very nitpick: `AND` usually combos with the THEN (the verification clause, as in doing multiple verifications) as opposed to the WHEN.

jonathanyeung (changes requested):

jonathanyeung:
lgtm, but looks like there's a cli test error.

kevinslin:
> lgtm, but looks like there's a cli test error.

fixed

Issue #3040, opened 2022-06-03
OS dependent path for vaults


rioka:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When additional vaults are created in Windows, paths in `dendron.code-workspace` and `dendron.yml` are stored using `\` as path separator; this causes dendron not to fetch them if you later try use the same repository under linux.

```json
    "folders": [
        {
            "path": "dependencies\\localhost\\work\\notes",
            "name": "work"
        },
```

```yml
workspace:
    vaults:
        -
            fsPath: dependencies\localhost\work
            selfContained: true
            name: work
        -
```

Apparently, dendron already handle this correctly for other settings (e.g. `"pasteImage.path": "${currentFileDir}/assets/images"`), so I tried to do the same, and that (ie manually changing `\` to `/`) solved the issue.

## To Reproduce

- Create a workspace in Windows
- Add another (local) vault
- Remove the line from `.gitignore` so that it will be in the same repository
- Commit to remote repository
- Restore in Linux

## Expected behavior

Should be able to use all vaults in Linux, too

## Desktop (please complete the following information)

- OS: Windows, Linux
- Versions (all that apply to bug):
  - Dendron Extension: 0.97.0
  - VS Code version: 1.67.2
  - Dendron CLI:

SeriousBug:
Thanks for the report. Agreed that Dendron should either normalize the paths when writing them, or when reading them.

SeriousBug:
Hi! Starting with 0.101, Dendron will always use UNIX-style paths for new vaults. For existing vaults, you'll need to manually update them. Please let us know if this issue continues to happen.

PR #3096, opened 2022-06-15
enhance(workspace): Vaults always use UNIX style separators in config files & Adding an existing remote vault avoids creating workspace files

SeriousBug:
This PR does 2 things:

1. Newly added vaults will always use UNIX style paths, as opposed to platform specific paths. This is required for cross compatibility for vaults created in Windows. Dendron will continue to correctly read these paths on Windows. #3040 
2. When adding an existing vault as a remote vault to a workspace, Dendron will no longer try to create `.gitignore` and `dendron.yml` files inside the added vault.


kevinslin (changes requested):
looks like this broke a CLI test for windows - can you fix?

SeriousBug:
@kevinslin Fixed now, thanks!

kevinslin (approved):

PR #3098, opened 2022-06-16
enhance(structure): improved note traits

jonathanyeung:
## enhance(structure): improved note traits

Various improvements to Note Traits:
- Note Traits now supports application of a template
- 'Adding' support for lodash, luxon access in the trait definition JS.  It turns out that it was already accessible via Node JS' `module.require`, but just completely undiscoverable.  I've added require statements to the traits template.
- Revamping the traits template to include more thorough documentation, including rudimentary Intellisense support via JSDoc strings
- Allows traits to 'hot' reload, so you no longer need to reload window to test if your code is working
- Add a watch-like compile capability - when a user saves their JS file, the code does a test run of each function during the traits registration process.  If any of the functions have errors, we'll provide a helpful error toast pointing directly at the error source.
- Fixes some usability issues with the note traits related commands, such as when you try to create a note with a trait but have no traits defined.

Docs: https://github.com/dendronhq/dendron-site/pull/560

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs

https://github.com/dendronhq/dendron-site/pull/560

### DIFF
--- a/packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
+++ b/packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
@@ -4,6 +4,8 @@ import {
   NoteTrait,
   OnCreateContext,
   cleanName,
+  EngagementEvents,
+  NoteUtils,
 } from "@dendronhq/common-all";
 import { HistoryEvent } from "@dendronhq/engine-server";
 import path from "path";
@@ -17,6 +19,8 @@ import { GotoNoteCommand } from "./GotoNote";
 import { ExtensionProvider } from "../ExtensionProvider";
 import { VaultSelectionModeConfigUtils } from "../components/lookup/vaultSelectionModeConfigUtils";
 import { NoteLookupProviderUtils } from "../components/lookup/NoteLookupProviderUtils";
+import { TemplateUtils } from "@dendronhq/common-server";
+import { AnalyticsUtils } from "../utils/analytics";
 
 export type CommandOpts = {
   fname: string;
@@ -104,19 +108,6 @@ export class CreateNoteWithTraitCommand extends BaseCommand<
     let title;
     let body;
 
-    if (this.checkWorkspaceTrustAndWarn()) {
-      if (this.trait.OnCreate?.setTitle) {
-        const context = await this.getCreateContext();
-        context.currentNoteName = fname;
-
-        title = this.trait.OnCreate.setTitle(context);
-      }
-
-      if (this.trait.OnCreate?.setBody) {
-        body = await this.trait.OnCreate.setBody();
-      }
-    }
-
     // TODO: GoToNoteCommand() needs to have its arg behavior fixed, and then
     // this vault logic can be deferred there.
     let vault = opts.vaultOverride;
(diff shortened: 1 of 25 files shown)


packages/plugin-core/src/commands/RegisterNoteTraitCommand.ts
```diff
@@ -14,33 +14,88 @@ type CommandOutput = {} | undefined;
...
+/**
+ * lodash module is available. See https://lodash.com/docs for documentation.
+ */
+const _ = module.require("lodash");
+
+/**
+ * luxon is available for Date functions. See
+ * https://moment.github.io/luxon/api-docs/index.html for documentation
+ */
+const luxon = module.require("luxon");
```
kevinslin, line 54: what is the difference btw `module.require` and `require`?

packages/plugin-core/src/workspace/workspaceActivater.ts
```diff
@@ -60,11 +60,11 @@ export class WorkspaceActivator {
...
 
-    // --- Setup Traits
     ext.workspaceImpl = workspace;
-    // Only set up note traits after workspaceImpl has been set, so that the
-    // wsRoot path is known for locating the note trait definition location.
-    ext.setupTraits();
+
+    // HACK: Only set up note traits after workspaceImpl has been set, so that
+    // the wsRoot path is known for locating the note trait definition location.
+    ext.traitRegistrar.initialize();
```
kevinslin, line 67: do we need to check if workspace trust is active and not initialize if thats not the case?

packages/plugin-core/src/services/NoteTraitManager.ts
```diff
@@ -0,0 +1,263 @@
...
+      ? path.join(this._wsRoot, CONSTANTS.DENDRON_USER_NOTE_TRAITS_BASE)
+      : undefined;
+
+    if (!userTraitsPath) {
+      return;
+    }
+
+    const pattern = new vscode.RelativePattern(userTraitsPath, "*.js");
+
+    this._watcher = vscode.workspace.createFileSystemWatcher(
```
kevinslin, line 216: will this work for native workspaces or only code workspaces?

kevinslin (commented):
nice job with this. only minor comments. when you do the docs for this, can you use the [[Standard Documentation|dendron://dendron.handbook/leaflet.journal.2022.05.10.standard-documentation]]?

do you feel good to merge it in for next week?

packages/plugin-core/src/services/NoteTraitManager.ts
jonathanyeung: Native workspaces will include the traits path in the folder set, so there's no problem there.  There is however an issue with self-contained vaults, which seems to exclude all paths that aren't 'notes' in the generated code workspace file.  So pods, traits, etc. don't appear - I'm not sure if that was intentional or not, but I can look into it in the next iteration.

kevinslin (commented):
looking good. some small comments

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
```diff
@@ -139,6 +142,68 @@ export class CreateNoteWithTraitCommand extends BaseCommand<
...
+        const dummy = NoteUtils.createForFake({
+          contents: "",
+          fname: "trait-tmp",
+          id: "trait-tmp",
+          vault: vault!,
+        });
+
+        if (notes && notes.length > 0) {
+          // Only apply schema if note is found
+          TemplateUtils.applyTemplate({
```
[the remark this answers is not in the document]
jonathanyeung: fixed, good catch

kevinslin (approved):

PR #3104, opened 2022-06-17
chore(internal): add proxy metrics for refactoring commands

hikchoi:
# chore(internal): add proxy metrics for refactoring commands

This PR:
- Adds proxy metrics for various refactoring commands for easier analytics.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/StatisticsUtils.ts
+++ b/packages/common-all/src/StatisticsUtils.ts
@@ -7,18 +7,29 @@
  */
 
 import _ from "lodash";
-import { DendronError } from ".";
+import { DNodeUtils } from ".";
+import { NoteProps } from "../lib";
+
+export type NonEmptyArray<T> = [T, ...T[]];
 
 export class StatisticsUtils {
+  static isNonEmptyArray<T>(arr: T[]): arr is NonEmptyArray<T> {
+    return arr.length > 0;
+  }
+
+  static getBasicStats(arr: NonEmptyArray<number>) {
+    return {
+      mean: _.mean(arr),
+      median: StatisticsUtils.median(arr),
+      stddev: StatisticsUtils.stddev(arr),
+      max: _.max(arr),
+    };
+  }
+
   /**
    * Get standard deviation from array of numbers.
    */
-  static stddev(arr: number[]) {
-    if (arr.length === 0) {
-      throw new DendronError({
-        message: "Cannot calculate standard deviation of empty array.",
-      });
-    }
+  static stddev(arr: NonEmptyArray<number>): number {
     const population = arr.length;
     const mean = _.mean(arr);
     const deviations = arr.map((value) => value - mean);
@@ -30,15 +41,93 @@ export class StatisticsUtils {
   /**
    * Get median value from array of numbers.
    */
-  static median(arr: number[]) {
-    if (arr.length === 0) {
-      throw new DendronError({
-        message: "Cannot calculate median of empty array.",
-      });
-    }
+  static median(arr: NonEmptyArray<number>): number {
     const population = arr.length;
     const mid = Math.floor(population / 2);
     const sorted = [...arr].sort((a, b) => a - b);
     return population % 2 ? sorted[mid] : (sorted[mid] + sorted[mid - 1]) / 2;
+  }
+
+  /**
+   * Convenience command to grab a collection of statistics from a set of notes.
+   *
+   * Given an array of notes, aggregate and get basic statistics of the following:
+   *
+   * number of children,
+   * number of links,
+   * number of characters in body,
+   * note depth
+   *
+   * This is used to grab statistics about quantifiable properties from a set of notes
+   * for analytics purposes.
+   *
+   * @param notes Notes to get basic stats from
+   * @returns an object holding all basic stats
+   */
+  static getBasicStatsFromNotes(notes: NoteProps[]) {
+    // accumulate values from all notes.
+    const numChildrenAcc = notes.map((note) => note.children.length);
+    const numLinksAcc = notes.map((note) => note.links.length);
+    const numCharsAcc = notes.map((note) => note.body.length);
+    const noteDepthAcc = notes.map((note) => DNodeUtils.getDepth(note));
+
+    // make sure they are nonempty arrays.
+    if (
+      !StatisticsUtils.isNonEmptyArray(numChildrenAcc) ||
+      !StatisticsUtils.isNonEmptyArray(numLinksAcc) ||
+      !StatisticsUtils.isNonEmptyArray(numCharsAcc) ||
+      !StatisticsUtils.isNonEmptyArray(noteDepthAcc)
+    ) {
+      // this should never be the case, but we need to do this
+      // because typescript can't infer that.
+      return;
+    }
+
+    const {
+      mean: numChildren,
+      median: medianNumChildren,
+      stddev: stddevNumChildren,
+      max: maxNumChildren,
+    } = StatisticsUtils.getBasicStats(numChildrenAcc);
+
+    const {
+      mean: numLinks,
+      median: medianNumLinks,
+      stddev: stddevNumLinks,
+      max: maxNumLinks,
+    } = StatisticsUtils.getBasicStats(numLinksAcc);
+
+    const {
+      mean: numChars,
+      median: medianNumChars,
+      stddev: stddevNumChars,
+      max: maxNumChars,
+    } = StatisticsUtils.getBasicStats(numCharsAcc);
+
+    const {
+      mean: noteDepth,
+      median: medianNoteDepth,
+      stddev: stddevNoteDepth,
+      max: maxNoteDepth,
+    } = StatisticsUtils.getBasicStats(noteDepthAcc);
+
+    return {
+      numChildren,
+      numLinks,
+      numChars,
+      noteDepth,
+      maxNumChildren,
+      medianNumChildren,
+      stddevNumChildren,
+      maxNumLinks,
+      medianNumLinks,
+      stddevNumLinks,
+      maxNumChars,
+      medianNumChars,
+      stddevNumChars,
+      maxNoteDepth,
+      medianNoteDepth,
+      stddevNoteDepth,
+    };
   }
 }
--- a/packages/common-all/src/StatisticsUtils.ts
+++ b/packages/common-all/src/StatisticsUtils.ts
@@ -7,8 +7,7 @@
  */
 
 import _ from "lodash";
-import { DNodeUtils } from ".";
-import { NoteProps } from "../lib";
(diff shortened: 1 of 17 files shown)


packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -597,12 +602,72 @@ export class MoveHeaderCommand extends BasicCommand<
...
+    AnalyticsUtils.track(EngagementEvents.RefactoringCommandUsed, {
+      command: this.key,
+      ...noteChangeEntryCounts,
+      traits: origin.traits || [],
+      numVaults: vaults.length,
+      numChildren: origin.children.length,
+      numlinks: origin.links.length,
+      numChars: origin.body.length,
+      noteDepth: DNodeUtils.getDepth(origin),
+      numOriginHeaders,
```
kevinslin, line 652: if we're tracking refactor operations, it will be good to have a type that enumerates what these refactor operations could be and use them when doing analytics across the various refactor commands

packages/plugin-core/src/commands/RefactorHierarchyV2.ts
```diff
@@ -438,9 +444,101 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
...
+      const newPath = operation.newUri.fsPath;
+      const fname = path.basename(newPath, ".md");
+      if (newPath !== undefined) {
+        const vault = VaultUtils.getVaultByFilePath({
+          vaults,
+          wsRoot,
+          fsPath: newPath,
+        });
+        const newNote = (
+          await engine.findNotes({
```
kevinslin, line 483: can we get this from the return statement vs calling find notes?

kevinslin (changes requested):

packages/plugin-core/src/commands/MoveNoteCommand.ts
```diff
@@ -316,10 +322,97 @@ export class MoveNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
...
+    const numLinksAcc: number[] = [];
+    const numCharsAcc: number[] = [];
+    const noteDepthAcc: number[] = [];
+    let traitsAcc: string[] = [];
+    await asyncLoop(moves, async (move) => {
+      const { fname, vaultName: vname } = move.newLoc;
+      if (fname !== undefined && vname !== undefined) {
+        const newLocVault = VaultUtils.getVaultByName({ vaults, vname });
+        const newLocNote = (
+          await engine.findNotes({
```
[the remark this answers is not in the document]
hikchoi: Because the outputted `NoteChangeEntry[]` doesn't hold enough context to track back the actual transactions made (or is non-trivial and unreliable to try to do so). 

There are cases for move note (and all other commands you suggested to use `NoteChangeEntry` to get notes) that will output multiple `create` entries for a single moved note. 

for example, cases where move note results in a creation of a stub in place of the old note.

I could filter for that case specifically, but rather than try to find all edge cases and filtering them out of the list of note change entries, relying on the actual moves passed to `.execute` as a command option would be the surest way to make sure we are only grabbing metrics from the right place.

@jonathanyeung and I briefly discussed the state of the events being emitted in our engine, and that just reading them doesn't give us the most accurate / high level picture of what has happened. We haven't come to any solution but I think some enhancements in this area is needed to just rely on note change entries for now.

As an interim solution, I will instead grab these information from during the execution, and pass it around so that it eventually ends up in the commands' respective `CommandOutput`s

packages/plugin-core/src/commands/RenameNoteCommand.ts
```diff
@@ -0,0 +1,110 @@
...
+    const { moves } = opts;
+
+    const move = moves[0];
+
+    const { fname, vaultName: vname } = move.newLoc;
+    if (fname === undefined || vname === undefined) {
+      return;
+    }
+    const vault = VaultUtils.getVaultByName({ vaults, vname });
+    const note = (await engine.findNotes({ fname, vault }))[0];
```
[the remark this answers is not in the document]
hikchoi: see: https://github.com/dendronhq/dendron/pull/3104/files/db7df7f3bb29dc6ca63f22447d194f96374c7e2f#r901846929

hikchoi:
update:

- cleaned most `engine.findNotes` calls and various issues found along the way.
- took some time to add more stats. refactoring commands that aggregate multiple operations now also have median and standard deviation values of each props.

packages/plugin-core/src/commands/RenameNoteCommand.ts
```diff
@@ -0,0 +1,106 @@
...
+    const extension = ExtensionProvider.getExtension();
+    const engine = extension.getEngine();
+    const { vaults } = engine;
+
+    AnalyticsUtils.track(EngagementEvents.RefactoringCommandUsed, {
+      command: this.key,
+      numVaults: vaults.length,
+      ...noteChangeEntryCounts,
+      ..._.omit(this._moveNoteCommand._proxyMetricPayload, [
+        "maxNumChildren",
```
jonathanyeung, line 70: maybe it makes sense to create a type for this so we can make sure we're spelling the props right across the various usages.

packages/plugin-core/src/commands/RefactorHierarchyV2.ts
```diff
@@ -372,19 +377,54 @@ export class RefactorHierarchyCommandV2 extends BasicCommand<
...
+      numProcessed: capturedNotes.length,
+      numChildren: _.mean(numChildrenAcc),
+      maxNumChildren: _.max(numChildrenAcc),
+      medianNumChildren: StatisticsUtils.median(numChildrenAcc),
+      stddevNumChildren: StatisticsUtils.stddev(numChildrenAcc),
+      numLinks: _.mean(numLinksAcc),
+      maxNumLinks: _.max(numLinksAcc),
+      medianNumLinks: StatisticsUtils.median(numLinksAcc),
+      stddevNumLinks: StatisticsUtils.stddev(numLinksAcc),
+      numChars: _.mean(numCharsAcc),
```
jonathanyeung, line 402: Is this number of characters in a note?  Is this relevant for refactoring?

hikchoi: I want to do a general survey of the context of refactoring. i.e. at what point / state of a note do the users feel the need to refactor?

This is more of an exploratory approach and I'm grabbing as many things as I can from `NoteProps` that I can anonymize and aggregate to a number that I can potentially use to derive the inflection point of refactor usefulness.

So in terms of relevancy, I am adding this to figure that out. My guess is that it will be hard to quantify and we wouldn't have the ability to do so without knowing the content of the note, but I'm shooting for some learning here about a general trend.

packages/plugin-core/src/commands/RenameNoteCommand.ts
hikchoi: makes sense. I'll create a type for the common props that all the proxy metrics use

hikchoi: Good call because I found one typo 😄

packages/plugin-core/src/commands/MoveHeader.ts
kevinslin: not enumerating keys but the properties. eg. `numLinks`

hikchoi: ah okay, I misunderstood. @jonathanyeung gave the same suggestion, and it's pushed now.

packages/common-all/src/StatisticsUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import _ from "lodash";
+import { DendronError } from ".";
+
+export class StatisticsUtils {
+  /**
+   * Get standard deviation from array of numbers.
+   */
+  static stddev(arr: number[]) {
```
kevinslin, line 16: since this will be used in many places and its not obvious an error could be thrown, can we return a `RespV3` so that this is explicit?

packages/common-all/src/StatisticsUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+    const deviations = arr.map((value) => value - mean);
+    const powers = deviations.map((value) => value ** 2);
+    const variance = _.sum(powers) / population;
+    return Math.sqrt(variance);
+  }
+
+  /**
+   * Get median value from array of numbers.
+   */
+  static median(arr: number[]) {
```
kevinslin, line 33: see comment above

packages/plugin-core/src/commands/MoveNoteCommand.ts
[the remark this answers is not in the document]
kevinslin: in this case, I would rather we change the output to include both `NoteChangeEntry` as well as the original input. this is also the pattern for all AWS calls. this way, instead of calling a `findNotes` with the engine, we can grab it from the result.

long term, for `NoteChangeEntry`, does it make sense to separate entries by `sideEffect` vs `command`?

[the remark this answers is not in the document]
hikchoi: that would make sense, and also would be preferable if it is enhanced in a way (either by the way it is, or by how we emit the change entries everywhere) that piecing all change entries together we can derive the pre-operation state. Either way, this needs a proper separate discussion.

packages/common-all/src/StatisticsUtils.ts
hikchoi: went with disallowing empty arrays altogether by narrowing the argument type.

packages/plugin-core/src/commands/MoveNoteCommand.ts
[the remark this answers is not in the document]
kevinslin: hold up - so whats the resolution here? until we kick of an actual async or address it in code, can we leave this open?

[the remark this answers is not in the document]
hikchoi: @kevinslin 
I don't think this issue should be a blocking one.

Just for clarification, after I made changes (https://github.com/dendronhq/dendron/pull/3104#issuecomment-1162012062) I removed all `engine.findNotes` calls except for one case where `moveNote` is not a multi move. It is a single call to find one note.

In every other cases I changed it so that we derive the payload from what we can infer from the initial input and intermediate steps.

bpathakota (approved):
commit title looks good!

kevinslin (approved):

PR #3107, opened 2022-06-18
fix(edit): autocomplete issues with tags and mentions

kevinslin:
fix(edit): autocomplete issues with tags and mentions

Enhancement to autocomplete broke tags and user mentions. 

Also adds some tests and light refactoring of utilities

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/CompletionProvider.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/CompletionProvider.test.ts
@@ -157,6 +157,7 @@ suite("completionProvider", function () {
             engine,
           });
           expect(found.length > 0).toBeTruthy();
+          expect(items?.items![0].insertText).toEqual("bar");
         }
       });
     }
@@ -265,6 +266,7 @@ suite("completionProvider", function () {
             engine,
           });
           expect(found.length > 0).toBeTruthy();
+          expect(items![0].insertText).toEqual("bar");
         }
       });
     }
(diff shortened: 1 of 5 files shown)


jonathanyeung (approved):

packages/plugin-core/src/test/suite-integ/CompletionProvider.test.ts
```diff
@@ -43,7 +43,7 @@ suite("completionProvider", function () {
       test("THEN provide completions", async () => {
         const { engine, vaults } = ExtensionProvider.getDWorkspace();
         // Open a note, add [[]]
-        await WSUtils.openNote(
+        await new WSUtilsV2(ExtensionProvider.getExtension()).openNote(
```
hikchoi, line 46: nitpick: Shouldn't this be: 

```suggestion
        await ExtensionProvider.getExtension().wsUtils().openNote(
```

instead of creating a new `WSUtilsV2` instance?

hikchoi: Same for all remainder of test changes.

hikchoi (changes requested):
Just a small nitpick on tests

packages/plugin-core/src/test/suite-integ/CompletionProvider.test.ts
jonathanyeung: @hikchoi - I think we want to bring this forward into the upcoming release, so @kevinslin - want to take this one up as a later follow up?  I'm going to go ahead and merge / push to v100

PR #3109, opened 2022-06-18
feat(extend): add visualization plugin command

hayata-suenaga:
# Still in draft

*[[task.2022.06.03.add-a-plugin-command-to-open-the-visualization-in-the-plugin]]*


bpathakota:
@hayata-suenaga Understand this is still a draft but, this should be titled as per the current process (details here: [[Submitting Code for Review|dendron://dendron.docs/dev.process.review]]). e.g. this would be would be categorized `feat(extend)` (details here: [[Extend|dendron://dendron.dendron-site/dendron.ref.scopes#extend]]).. For the actual title, it should highlight the benefit of adding in the command.

PR #3110, opened 2022-06-18
enhance: autocomplete Lookup based on cursor position when no text is selected

tenheadedlion:
## Enhancement

Remove the need to select a word so as to create a note, VSCode can recognize a word from the cursor's position, and that recognized word is possibly a link, in the case of a link, the users can create the note directly without further editing the name in the lookup dialog box. 

Selecting a word is less efficient and often distracting, double clicks and shortcuts are less economic than not doing anything at all, Dendron could be smart enough to help reduce the burden.

## Implementation

- Get the word from the cursor's current position, with the pattern of a wiki link `\[\[.*\]\]`
- If no text is selected by the users, then set the value of `quickPicker` to that link, if there is one.
- else if the user intended to use the selected text, Dendron will handle the situation the way it used to do.

## Usecase

I have been using Dendron a lot. When jotting down ideas, at some point I will have to refer to other notes. But some of the notes do not exist at the time, so I have to create them, but I find creating notes is not so convenient, and the hint provided on the lookup dialog box is not so helpful, in both modes:  selectionExtract and selection2link. In the latter mode, I still need to edit the name a lot. If I need to change the preset name a lot, I would prefer to do it in the editor, on the fly, to avoid switching focus to somewhere else. And If I decide to create that note later, I can continue the writing and then move the cursor back to the link of the non-existent note and, without selection, which often fails to select a link if no complicated shortcuts are applied, I can quickly prompt the lookup window and jump directly to that note.

## Effect

https://imgur.com/a/BuOrEFA

### DIFF
--- a/packages/plugin-core/src/components/lookup/LookupControllerV3.ts
+++ b/packages/plugin-core/src/components/lookup/LookupControllerV3.ts
@@ -629,6 +629,27 @@ export class LookupControllerV3 implements ILookupControllerV3 {
     }
   }
 
+  // For an empty selection, check if the cursor is within a wiki link,
+  //  if so, let the link be the default path in the lookup dialog box;
+  // For a non empty selection, just pass it to the `handler`;
+  //
+  // the definition guarantees that quickPick is passed by reference;
+  private onSelectionBtnToggled(
+    quickPick: DendronQuickPickerV2,
+    handler?: (text: string, quickPick: DendronQuickPickerV2) => void
+  ) {
+    let { text } = VSCodeUtils.getSelection();
+    text = text!.trim();
+    if (text === "") {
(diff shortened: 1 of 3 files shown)


SeriousBug:
Thanks for your work! This is an interesting addition, but did you know though that you can use "Dendron: Go to Note", bound to `Ctrl+Enter` by default, to create a note when you have a link selected? I think that already covers this use case for the most part.

tenheadedlion:
> Thanks for your work! This is an interesting addition, but did you know though that you can use "Dendron: Go to Note", bound to `Ctrl+Enter` by default, to create a note when you have a link selected? I think that already covers this use case for the most part.

Oh I did not know, I try `Ctrl+Enter` myself and that is what I want to implement, I will close the PR. Thanks for the information :)

PR #3111, opened 2022-06-19
enhance(publish): Support a custom banner on published page

kevinslin:
https://www.loom.com/i/f573ef84983148d6ad8dafcd339f5b65

This is still an experimental feature - still noodling on how to best incorporate custom code points into Dendron.
Will not be included in release notes yet.  Docs here: task.2022.06.18.banner-on-site.docs

### DIFF
--- a/packages/pods-core/src/builtin/NextjsExportPod.ts
+++ b/packages/pods-core/src/builtin/NextjsExportPod.ts
@@ -353,7 +353,6 @@ export class NextjsExportPod extends ExportPod<NextjsExportConfig> {
     }
 
     // custom components
-    debugger;
     if (PublishUtils.hasCustomSiteBanner(config)) {
       const bannerPath =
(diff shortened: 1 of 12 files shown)


packages/pods-core/src/builtin/NextjsExportPod.ts
```diff
@@ -350,6 +351,21 @@ export class NextjsExportPod extends ExportPod<NextjsExportConfig> {
         fs.copySync(headerPath, path.join(destPublicPath, "header.html"));
       }
     }
+
+    // custom components
+    debugger;
```
jonathanyeung, line 356: lingering debugger statement

jonathanyeung (approved):

packages/pods-core/src/builtin/NextjsExportPod.ts
jonathanyeung: I went ahead and removed it.

Issue #2988, opened 2022-05-26
Error when running dendron publish export (selfcontained)


joshuajeschek:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Publish.  External Site Publish

## Describe the bug

I have a self-contained remote vault (just migrated) and if I run `yarn dendron publish export --target github --yes` I get an error `Cannot read properties of undefined`
After the error occurs, it just hangs and has to be stopped manually to prevent it running indefinitely. 
Look at the end for complete log.

## To Reproduce

I have no idea, I don't know what's wrong.

## Expected behavior

Exporting should work.

## Screenshots

not applicable

## Desktop (please complete the following information)

- OS: Windows and Ubuntu (Desktop and Github Actions)
- Browser (if with published websites; include version): /
- Versions (all that apply to bug):
  - Dendron Extension: /
  - VS Code version: /
  - Dendron CLI: 0.96.0

### The Error
(cleaned up long paths and linebreaks)
```console
yarn dendron publish export --target github --yes
🌱 checking if NextJS template is initialized
🌱 NextJS template is already initialized.
⠋ generating metadata for publishing...
⠧ error {"stack":"TypeError: Cannot read properties of undefined (reading 'dendron.defaultTimestampDecorationFormat')
    at Function.getFM (...\\node_modules\\@dendronhq\\engine-server\\lib\\markdown\\utilsv5.js:123:102)
    at Function._procRemark (...\\node_modules\\@dendronhq\\engine-server\\lib\\markdown\\utilsv5.js:187:53)
    at Function.procRemarkParse (...\\node_modules\\@dendronhq\\engine-server\\lib\\markdown\\utilsv5.js:342:21)
    at Function._procRehype (...\\node_modules\\@dendronhq\\engine-server\\lib\\markdown\\utilsv5.js:284:35)
    at Function.procRehypeFull (...\\node_modules\\@dendronhq\\engine-server\\lib\\markdown\\utilsv5.js:364:27)
    at NextjsExportPod._renderNote (...\\node_modules\\@dendronhq\\pods-core\\lib\\builtin\\NextjsExportPod.js:174:48)
    at NextjsExportPod.renderBodyToHTML (...\\node_modules\\@dendronhq\\pods-core\\lib\\builtin\\NextjsExportPod.js:278:32)
    at ...\\node_modules\\@dendronhq\\pods-core\\lib\\builtin\\NextjsExportPod.js:334:22
[...]

PR #3114, opened 2022-06-20
fix(publish): issue publishing note with ref without a `code-worksapce` file

kevinslin:
Dendron will try to format `created`/`updated` according to `dendron.defaultTimestampDecorationFormat` option from `dendron.code-workspace` when processing note references. This will fail on any workspace that does not have `dendron.code-workspace`. This leads to an error on publish (see https://github.com/dendronhq/dendron/issues/2988) and probably errors in other places

For frontmatter variable substitution, we never meant for `created`/`updated` to actually be formatted. To simplify the method call, remove the requirement to parse `dendron.code-workspace`

### DIFF
--- a/packages/engine-server/src/engineClient.ts
+++ b/packages/engine-server/src/engineClient.ts
@@ -440,7 +440,12 @@ export class DendronEngineClient implements DEngineClient, EngineEventEmitter {
     const resp = await this.api.engineUpdateNote({ ws: this.ws, note, opts });
     const noteClean = resp.data;
     if (_.isUndefined(noteClean)) {
-      throw new DendronError({ message: "error updating note", payload: resp });
+      throw new DendronError({
+        message: `error updating note: ${JSON.stringify(
+          NoteUtils.toNoteLoc(note)
+        )}`,
+        payload: resp,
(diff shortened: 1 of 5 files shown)


hikchoi (approved):
> For frontmatter variable substitution, we never meant for created/updated to actually be formatted. To simplify the method call, remove the requirement to parse dendron.code-workspace

Makes sense to remove this. `getFM` should only be concerned with grabbing the frontmatter property values as is.

I must have added them a while ago because of a request from the community so that they can substitute them in the body in a human readable form.

Any timeline on supporting handlebarsjs in frontmatter substitution?

kevinslin:
> > For frontmatter variable substitution, we never meant for created/updated to actually be formatted. To simplify the method call, remove the requirement to parse dendron.code-workspace
> 
> Makes sense to remove this. `getFM` should only be concerned with grabbing the frontmatter property values as is.
> 
> I must have added them a while ago because of a request from the community so that they can substitute them in the body in a human readable form.
> 
> Any timeline on supporting handlebarsjs in frontmatter substitution?

optimistically sometime in july

PR #3116, opened 2022-06-20
chore(workspace): debounce completion provider trigger to reduce lag when autocompleting wikilinks

hikchoi:
# chore(workspace): debounce completion provider trigger to reduce lag when autocompleting wikilinks
This PR:
- register a debounced version of `provideCompletionItem` to avoid triggering lookup (potentially expensive depending on workspace size) on every keystroke.

## Note
- We have another completion item provider for block completions, but this does not get triggered on every keystroke, and doesn't use lookup so it is generally lighter than `provideCompletionItem`. For this reason this is not debounced.

## Testing
This is not straightforward to write a test so confirming manually that it is successfully debounced:

as-is:
image

to-be:
image

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/features/completionProvider.ts
+++ b/packages/plugin-core/src/features/completionProvider.ts
@@ -185,7 +185,6 @@ export const provideCompletionItems = sentryReportingCallback(
     }
 
     const line = document.lineAt(position).text;
-    console.log({ ctx, line });
     Logger.info({ ctx, position, msg: "enter" });
 
     // get all matches
@@ -341,6 +340,15 @@ export const provideCompletionItems = sentryReportingCallback(
   }
 );
 
+/**
+ * Debounced version of {@link provideCompletionItems}.
+ *
+ * We trigger on both leading and trailing edge of the debounce window because:
+ * 1. without the leading edge we lose focus to the Intellisense
+ * 2. without the trailing edge we may miss some keystrokes from the users at the end.
+ *
+ * related discussion: https://github.com/dendronhq/dendron/pull/3116#discussion_r902075154
+ */
 export const debouncedProvideCompletionItems = _.debounce(
   provideCompletionItems,
   100,
--- a/packages/plugin-core/src/features/completionProvider.ts
+++ b/packages/plugin-core/src/features/completionProvider.ts
@@ -341,6 +341,12 @@ export const provideCompletionItems = sentryReportingCallback(
   }
 );
 
+export const debouncedProvideCompletionItems = _.debounce(
+  provideCompletionItems,
+  100,
+  { trailing: false, leading: true }
+);
+
 export const resolveCompletionItem = sentryReportingCallback(
   async (
     item: CompletionItem,
@@ -613,7 +619,8 @@ export const activate = (context: ExtensionContext) => {
     languages.registerCompletionItemProvider(
       "markdown",
       {
-        provideCompletionItems,
+        // we debounce this provider since we don't want lookup to be triggered on every keystroke.
+        provideCompletionItems: debouncedProvideCompletionItems,
       },
 
       "[", // for wikilinks and references
@@ -626,6 +633,10 @@ export const activate = (context: ExtensionContext) => {
     languages.registerCompletionItemProvider(
       "markdown",
       {
+        /**
+         * we don't have to debounce this since it is not triggered on every keystroke
+         * and is ligher than {@link provideCompletionItems} in general.
+         */
         provideCompletionItems: provideBlockCompletionItems,
       },
       "#",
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/features/completionProvider.ts
```diff
@@ -341,6 +341,12 @@ export const provideCompletionItems = sentryReportingCallback(
   }
 );
 
+export const debouncedProvideCompletionItems = _.debounce(
+  provideCompletionItems,
+  100,
+  { trailing: false, leading: true }
```
jonathanyeung, line 347: do we actually want it on leading?  This'll potentially make us lose the last keystroke - I think trailing is better in this case? (and 100 ms is probably too fast for a user to perceive lag)

hikchoi: You are right, but:

if we remove the leading edge, we lose the trigger focus entirely and any successive keystrokes will not trigger the completion. (imagine mashing [ twice. This triggers the provider twice. if we remove trailing, we lose the completion on the second [ and that's the end.)

we should probably invoke on both edges and play around with the time a bit.

jonathanyeung: > if we remove the leading edge, we lose the trigger focus entirely

I haven't played around much with the VSCode completion provider API, but are you saying we'll hit some VS Code timeout since we didn't provide any data in time and then the provider UI box will no longer show?

hikchoi: typing [[asdfasdf]] very quickly, all on 100ms:

1. leading false, trailing true:
 
image

2. leading true, trailing false:

image

3. leading true, trailing true

image

hikchoi: > I haven't played around much with the VSCode completion provider API, but are you saying we'll hit some VS Code timeout since we didn't provide any data in time and then the provider UI box will no longer show?

@jonathanyeung 

I think what's happening is this:

1. we register a few trigger characters, `[` being one of them.
2. the first `[` will trigger the provider
3. the second `[`, if within the debounce window, will not trigger the provider, but will obviously still be typed in the editor.
4. At this point, Intellisense missed the trigger, VSCode saw it as a plain keystroke, didn't find anything in the suggestions, and disengaged. We lose focus here.

jonathanyeung: I see - thanks for the experimentation.  So it looks like doing leading + trailing should get us where we want? 

One other thing we can consider is using the cancellation token: 

```ts
provideCompletionItems(document: TextDocument, position: Position, token: CancellationToken, context: CompletionContext): ProviderResult<T[] | CompletionList<T>>;
```

jonathanyeung (approved):

packages/plugin-core/src/features/completionProvider.ts
hikchoi: Yes I think for now leading + trailing is sufficient.
I notice that we did at some point use the `resolveCompletionItems` and use cancellation tokens there. Not sure why we stopped using that? I'll follow up with it for the next step.

packages/plugin-core/src/features/completionProvider.ts
```diff
@@ -179,13 +179,13 @@ export const provideCompletionItems = sentryReportingCallback(
...
     const ctx = "provideCompletionItems";
     const startTime = process.hrtime();
-
     // No-op if we're not in a Dendron Workspace
     if (!ExtensionProvider.getExtension().isActive()) {
       return;
     }
 
     const line = document.lineAt(position).text;
+    console.log({ ctx, line });
```
kevinslin, line 188: remove console statement

packages/plugin-core/src/features/completionProvider.ts
```diff
@@ -341,6 +341,12 @@ export const provideCompletionItems = sentryReportingCallback(
   }
 );
 
+export const debouncedProvideCompletionItems = _.debounce(
+  provideCompletionItems,
+  100,
+  { leading: true, trailing: true }
```
kevinslin, line 347: can you link to the gh pr discussion with @jonathanyeung for context

bpathakota:
@hikchoi - Apologies on the late comments. Since this is user facing, this should be an enhancement and would change the scope to edit as it impacts the user's editing experience. e.g. `enhance(edit): reduce lag on autocomplete by adding a debounce`. Would be better to emphasize on the benefit over the change since it will be more intuitive for the changelog. Let me know if you have any questions!

PR #3118, opened 2022-06-20
enhance(workspace): support single-vault workspace vaults for self contained vault migration

SeriousBug:
Self contained vault migration will now allow users to migrate workspace vaults that only contain one vault.

### DIFF
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -496,15 +496,19 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
   }
 
   async migrateVaultToSelfContained({ vault }: { vault: DVault }) {
-    if (vault.seed || vault.workspace) {
+    const backupInfix = "migrate-vault-sc";
+    if (vault.seed) {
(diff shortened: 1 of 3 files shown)


packages/engine-server/src/workspace/service.ts
```diff
@@ -496,15 +496,19 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
   async migrateVaultToSelfContained({ vault }: { vault: DVault }) {
-    if (vault.seed || vault.workspace) {
+    const backupInfix = "migrate-vault-sc";
+    if (vault.seed) {
       // Unsupported vaults are filtered in the commands that use this function,
       // but also adding a sanity check here.
       throw new DendronError({
         message:
-          "Seed vaults and workspace vaults are not yet supported for automated migration.",
+          "Seed vaults vaults are not yet supported for automated migration.",
```
jonathanyeung, line 505: nit: 'vaults vaults'

jonathanyeung (commented):
lgtm @kevinslin can you also review?

PR #3132, opened 2022-06-22
fix(publish): missing "link to anchor" icon by headings and block anchors

SeriousBug:
This PR restores the link icon that was located by headings and at places where block anchors are attached. The link icons are always visible but are colored the same color as the text. Once the related heading or paragraph is hovered over, the icon will change into the link color for the current theme.

Widescreen:

Small screen:


kevinslin:
its a little jarring to have the handle show up in every element. can we make it invisible unless the user hovers over it?

bpathakota (changes requested):
Think it would be better if we followed a standardized approach for fix commits - to indicate what issue is being fixed. e.g. `fix(publish): missing link icon by headings and block anchors`

SeriousBug:
@kevinslin Done, it now only appears on hover, but only on devices that can hover (e.g. desktops and laptops). Other devices like smartphones will always display the icons. While that looks a little jarring in the example image, that's because there are tons of anchors in that example. In real world usage most of the time it will only be next to headings.

kevinslin:
> @kevinslin Done, it now only appears on hover, but only on devices that can hover (e.g. desktops and laptops). Other devices like smartphones will always display the icons. While that looks a little jarring in the example image, that's because there are tons of anchors in that example. In real world usage most of the time it will only be next to headings.

can we default to not show on mobile vs show?

kevinslin:
> @kevinslin Sure, but the anchors would be completely unusable on mobile then. Users won't have any way of seeing them.

got it. we'll take that tradeoff for now until we can make it work more unobtrusively on mobile.

SeriousBug:
@kevinslin Sure, done.

kevinslin (approved):

PR #3133, opened 2022-06-22
fix(publish): compile error with no banner present

kevinslin:
fix(publish): compile error with no banner present

replace require with no-op to appease compiler


SeriousBug (approved):
Solves the problem for me, that's a very smart solution. Thanks!

PR #3146, opened 2022-06-24
chore: refactor startup logic to work without reload

kevinslin:
chore: refactor startup logic to work without reload

- main:
    - move extension startup logic into `WorkspaceActivator`
        - this is divided into 2 phases:
            - init
                ```ts
                /**
                * Initialize workspace. All logic that happens before the engine is initialized happens here
                * - create workspace class
                * - register traits
                * - run migrations if necessary
                */
                ```
            - activate
                ```ts
                /**
                * Initialize engine and activate workspace watchers
                */
                ```
    - update trait logic (trait can be injected at runtime), this simplifies the command registration process (we can register all commands on activation vs conditionally registration some commands only when the ws has activated)
    - nice sideeffect, this removes all circular dependencies
    - LaunchTutorial command will now launch the tutorial without reloading the workspace

- misc
    - fix issue in `describeSingleWS.only` function

- NOTE: since this is quite a large change, I've held back on adding the auto-init into the tutorial workspace. will be trivial to add once this is merged


hikchoi (commented):
some comments.

packages/plugin-core/src/_extension.ts
```diff
@@ -998,67 +730,3 @@ function _setupLanguageFeatures(context: vscode.ExtensionContext) {
 }
 
 // ^qxkkg70u6w0z
```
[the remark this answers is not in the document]
kevinslin: removed

packages/plugin-core/src/utils/StartupUtils.ts
```diff
@@ -11,33 +11,30 @@ import {
...
-import { StateService } from "../services/stateService";
 import { SurveyUtils } from "../survey";
 import { VSCodeUtils } from "../vsCodeUtils";
-import { showWelcome } from "../WelcomeUtils";
 import { AnalyticsUtils } from "./analytics";
 import { ConfigMigrationUtils } from "./ConfigMigration";
-import { ExtensionProvider } from "../ExtensionProvider";
-import { readMD } from "@dendronhq/common-server";
 
 export class StartupUtils {
```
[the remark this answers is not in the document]
hikchoi: Got it.

hikchoi:
fyi: the extension name when accessed during dev looks to be `dendron.@dendronhq/plugin-core`

in case we need to use `vscode.extensions.getExtension()` later.

image

hikchoi:
> NOTE: since this is quite a large change, I've held back on adding the auto-init into the tutorial workspace. will be trivial to add once this is merged

Just so I'm understanding correctly. What is the current expected behavior when we click on the big green button?

In a non-dendron workspace, ran `Dendron: Launch Tutorial` to bring up the welcome page, and then clicked the button.

From the logs I can see that a new `Dendron-X` workspace has been initialized, but nothing happened in the VSCode side. 

Is this part what you are referring to be added once merged?

packages/plugin-core/src/dendronExtensionInterface.ts
```diff
@@ -127,4 +132,5 @@ export interface IDendronExtension {
    * about registered Note Traits
    */
   get traitRegistrar(): NoteTraitService;
+  setEngine(svc: EngineAPIService): void;
```
jonathanyeung, line 135: should be IEngineAPIService (if possible). 

Can we also move this up to be next to `getEngine`?

packages/plugin-core/src/telemetry.ts
```diff
@@ -3,11 +3,17 @@ import { SegmentClient } from "@dendronhq/common-server";
 import { Logger } from "./logger";
 
 /** Creates a SegmentClient for telemetry, if enabled, and listens for vscode telemetry settings to disable it when requested. */
-export function setupSegmentClient(ws: DWorkspaceV2, cachePath?: string) {
+export function setupSegmentClient({
+  ws,
+  cachePath,
+}: {
+  ws?: DWorkspaceV2;
```
jonathanyeung, line 10: I'm not sure I completely understand - when will this not be set here?  Will we have any issues of not respecting user's disableTelemetry choice because of a race condition?

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
```diff
@@ -39,16 +40,38 @@ export class CreateNoteWithTraitCommand extends BaseCommand<
...
-  trait: NoteTrait;
+  private _trait: NoteTrait | undefined;
+  private initTrait: () => NoteTrait;
   protected _extension: IDendronExtension;
 
-  constructor(ext: IDendronExtension, commandId: string, trait: NoteTrait) {
+  constructor(
+    ext: IDendronExtension,
+    commandId: string,
+    trait: NoteTrait | (() => NoteTrait)
```
jonathanyeung, line 50: I'm not a big fan of having this OR'd type - I think what'd be cleaner is to remove this trait prop from the constructor, and then expose a `protected set trait(trait: NoteTrait)` method that each derived class like `MeetingNoteCommand` or `JournalNoteCommand` call as part of their run/execute.

If that's more than you'd like to refactor right now, can you add a TODO comment with this info?

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
```diff
@@ -39,16 +40,38 @@ export class CreateNoteWithTraitCommand extends BaseCommand<
...
+
+    if (_.isFunction(trait)) {
+      this.initTrait = trait;
+    } else {
+      this.initTrait = () => trait;
+    }
     this._extension = ext;
   }
 
+  get trait(): NoteTrait {
```
jonathanyeung, line 63: can we make this private (or maybe protected)

jonathanyeung (changes requested):
For my understanding, the intention behind splitting the 'init' logic and the 'activate' logic (as defined per your PR description) such that we can re-run activate without needing to re-run init?

> nice sideeffect, this removes all circular dependencies

I don't think this is true - there's still a circular dependency here:

```
  [
    'src/ExtensionProvider.ts',
    'src/dendronExtensionInterface.ts',
    'src/fileWatcher.ts'
  ]
```

kevinslin:
> > NOTE: since this is quite a large change, I've held back on adding the auto-init into the tutorial workspace. will be trivial to add once this is merged
> 
> Just so I'm understanding correctly. What is the current expected behavior when we click on the big green button?
> 
> In a non-dendron workspace, ran `Dendron: Launch Tutorial` to bring up the welcome page, and then clicked the button.
> 
> From the logs I can see that a new `Dendron-X` workspace has been initialized, but nothing happened in the VSCode side.
> 
> Is this part what you are referring to be added once merged?

If you click the green button, it should create a native workspace and initialize dendron with this workspace without reloading the window

packages/plugin-core/src/_extension.ts
```diff
@@ -281,9 +179,26 @@ export async function _activate(
     const ws = await DendronExtension.getOrCreate(context, {
       skipSetup: stage === "test",
     });
+    const existingCommands = await vscode.commands.getCommands();
 
     // Setup the commands
     await _setupCommands({ ext: ws, context, requireActiveWorkspace: false });
+    // Order matters. Need to register `Reload Index` command before `reloadWorkspace`
```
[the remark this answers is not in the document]
kevinslin: updated in comments

```
    // Order matters. Need to register `Reload Index` command before activating workspace
    // Workspace activation calls `RELOAD_INDEX` via {@link WSUtils.reloadWorkspace}
```

packages/plugin-core/src/telemetry.ts
kevinslin: it won't be set when the user doesn't activate dendron inside of a dendron workspace. we read the `dendron.yml` to see if they have a workspace specific override for telemetry. otherwise, we check $HOME/.dendron.telemetry file for the status

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
kevinslin: so the reason for passing in a function to the constructor is so we can register all commands at startup (vs waiting for the workspace to be available to register a command). my preference is for `trait` to be changed to `initTraitCb`. have it as an or type for now to avoid breaking tests and other interfaces. added a TODO statement to come back to this

packages/plugin-core/src/workspace/workspaceActivator.ts
```diff
@@ -0,0 +1,661 @@
...
+      };
+    }
+
+    ExtensionUtils.setWorkspaceContextOnActivate(wsService.config);
+    MetadataService.instance().setDendronWorkspaceActivated();
+    Logger.info({ ctx, msg: "fin startClient", durationReloadWorkspace });
+
+    const stage = getStage();
+    if (stage !== "test") {
+      ext.activateWatchers();
```
hikchoi, line 569: This throws an error if the tutorial workspace is initialized in a non-workspace folder.

```
stack trace: Error: no folders set for workspace
	at DendronExtension.activateWatchers (/Users/hikchoi/dev/dendronhq/dendron/packages/plugin-core/out/src/workspace.js:555:19)
	at WorkspaceActivator.activate (/Users/hikchoi/dev/dendronhq/dendron/packages/plugin-core/out/src/workspace/workspaceActivator.js:421:17)
	at processTicksAndRejections (node:internal/process/task_queues:96:5)
	at async SetupWorkspaceCommand.execute (/Users/hikchoi/dev/dendronhq/dendron/packages/plugin-core/out/src/commands/SetupWorkspace.js:227:21)
	at async LaunchTutorialWorkspaceCommand.execute (/Users/hikchoi/dev/dendronhq/dendron/packages/plugin-core/out/src/commands/LaunchTutorialWorkspaceCommand.js:34:9)
	at async LaunchTutorialWorkspaceCommand.run (/Users/hikchoi/dev/dendronhq/dendron/packages/plugin-core/out/src/commands/base.js:69:20)
	at async /Users/hikchoi/dev/dendronhq/dendron/packages/plugin-core/out/src/WelcomeUtils.js:52:21
```

hikchoi (changes requested):
Just pulled the branch in again to test it locally. Dendron initializes as expected to the tutorial workspace. I must have done something wrong before when building.

One comment about an error thrown when the open VSCode instance is not in a VSCode workspace.

Also, I know we kinda don't want to add a folder to the user's code-workspace without asking, but new users may get confused when they can't see anything in the file explorer.

We do want to nudge users to use the tree view instead of the file explorer, but if we do this we might want to highlight the use of tree view more in the tutorial to avoid confusion.

kevinslin:
> Just pulled the branch in again to test it locally. Dendron initializes as expected to the tutorial workspace. I must have done something wrong before when building.
> 
> One comment about an error thrown when the open VSCode instance is not in a VSCode workspace.
> 
> Also, I know we kinda don't want to add a folder to the user's code-workspace without asking, but new users may get confused when they can't see anything in the file explorer.
> 
> We do want to nudge users to use the tree view instead of the file explorer, but if we do this we might want to highlight the use of tree view more in the tutorial to avoid confusion.

yeah - we'll need to iterate on the ux. this is why this change doesn't change the current onboarding flow - it just does the refactoring needed to make it happen. its a big enough change by itself, will have a separate change to integrate this with tutorial

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
jonathanyeung: I understand the reasoning - my suggestion to move the setting of the trait to a setter and out of the constructor means that we also don't need to have the WS at init time, but only when the user runs the command.

hikchoi (changes requested):
Since we are not reloading and we didn't remove the big green button yet, we should change [this](https://github.com/dendronhq/dendron/blob/bcadd487c706379f0384a6f3a8c728d606a2bba8/packages/plugin-core/src/WelcomeUtils.ts#L42) to immediately track instead of saving it for the next session

packages/plugin-core/src/commands/CreateNoteWithTraitCommand.ts
kevinslin: I'm not quite sure how that would be implemented. Regardless, I left a TODO here to change it - can we merge this and then work out details of the note trait command?

PR #3152, opened 2022-06-25
fix(structure): quickpick stuck issue for refactor hierarchy cmd

jonathanyeung:
## fix(structure): quickpick stuck issue for refactor hierarchy cmd

Fixing an old issue with the refactor hierarchy command, where the quickpick would get "stuck" and require a mouse to click either "proceed" or "cancel".  The QuickPick was losing focus because of the webview we show that contains the list of note renames.


kevinslin:
nice, this has been super annoying issue. looks like theres a stray import

kevinslin (approved):

PR #3154, opened 2022-06-27
fix(structure): hot reload in note traits + no template by default

jonathanyeung:
## fix(structure): hot reload in note traits + no template by default

Fixing two issues with the note traits updates:
- Hot reloading wasn't working in the webpacked build because the wrong `webpack-require-hack` file was being modified.
- Commenting out the `applyTemplate` function of the default trait template.


jonathanyeung:
Heads up @Harshita-mindfire and @kevinslin - I went ahead and cherry-picked this into `release/0.101.0`

kevinslin (approved):

PR #3158, opened 2022-06-27
chore: initial work to remove MDUtilsV4

kevinslin:
`MDUtilsV4` was our old way of parsing markdown. we've migrated many months ago - this is the first part to remove dependencies on mdutils v4
- NOTE: This also removes tests for the `containers` plugin since we don't support it in `MDUtilsV5` nor document it. will remove the container dependency when we remove all of `MDUtilsV4`


jonathanyeung (approved):
nice cleanup!

Issue #1438, opened 2021-09-29
Create a keyboard shortcut to close Preview.


apastuszak:
## Please Select if your Request is Either something new or an Enhancement ##

- [ ] Enhancement of an existing Feature.
- [X ] Request of a new feature.

## Please select the area your request applys to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [X] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [ ] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

Provide keyboard shortcuts to allow open and closing of HTML Markdown previews. Right now, I can type CTRL+K V, and the preview window will pop up. But if you then type CTRL+W it will close the note and not the preview window, unless you click on the preview window first.

## Describe the solution you'd like ##

A keyboard shortcut that will always close the preview windows, regardless of where the cursor focus is.

## Describe alternatives you've considered ##

I was trying to find  if I could set another keybinding and did not see one. Dendron: Show Preview  exists, but there is no Dendron: Hide Preview

## Additional context ##

As I edit notes, I would like to pop open the Preview, see how it looks, and then close it and keep working without needing to take my hands off the keyboard.

jonathanyeung:
Thanks @apastuszak for the suggestion - this seems like a useful command to add.  We'll prioritize against our backlog.

aleksey-rowan:
A dedicated Dendron keybinding to always close the preview regardless of what's focused would be great, but there are other ways to do it right now without taking your hands off the keyboard (maybe not as quick as a single shortcut though):

- if you have just a single editor group, you can close the preview by `Ctrl + 2 Ctrl + W`; where `Ctrl + 2` switches to the second editor group with the preview, and when the preview is closed, the focus goes back to the first editor group with your open note;
- if you have just a single editor group, you can add a shortcut to the `@command:workbench.action.closeEditorsInOtherGroups` command to close the preview;
- alternatively, you can do `Ctrl + K Ctrl + Right Arrow Ctrl + W`; where `Ctrl + K Ctrl + Right Arrow` focuses on the editor group to the right of the current one, which should be a preview. This is a bit longer than the first two sequences though;

memeplex:
IMO having toggle show/hide command is the most intuitive and convenient interface.

[...]

PR #3164, opened 2022-06-28
enhance(plugin-core): add toggle preview command

hayata-suenaga:
# Add `Dendron: Toggle Preview` plugin command to toggle note preview panel

task: `[[task.2022.07.04.create-a-keyboard-shortcut-to-close-preview#^a4ivy7ht61jr]]`
issue: #1438 
doc pr: [569](https://github.com/dendronhq/dendron-site/pull/569)

Same functionality as `Dendron: Show Preview` except that running the command when there is already a preview panel open hides the panel.


packages/plugin-core/src/test/suite-integ/TogglePreview.test.ts
```diff
@@ -33,13 +33,14 @@ suite("GIVEN ShowPreview", function () {
...
       test("THEN the current note is opened", async () => {
-        const cmd = new ShowPreviewCommand(
+        const cmd = new TogglePreviewCommand(
           PreviewPanelFactory.create(ExtensionProvider.getExtension())
         );
         const out = await cmd.run();
         expect(out?.note).toBeTruthy();
         expect(out!.note!.id).toEqual(note.id);
         expect(out!.note!.fname).toEqual(note.fname);
+        await cmd.run();
```
kevinslin, line 43: you are running this command twice here

packages/plugin-core/src/constants.ts
```diff
@@ -840,6 +840,17 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     },
     when: "dendron:pluginActive",
   },
+  TOGGLE_PREVIEW: {
+    key: "dendron.togglePreview",
+    title: `${CMD_PREFIX} Toggle Preview`,
+    icon: `$(open-preview)`,
+    // keybindings: {
```
kevinslin, line 847: can we replace the show preview command key bindings with `togglePreview`? this should be the default going forward

packages/plugin-core/src/commands/TogglePreview.ts
```diff
@@ -46,9 +55,16 @@ export class ShowPreviewCommand extends InputArgCommand<
...
+  async execute(opts?: TogglePreviewCommandOpts) {
     let note: NoteProps | undefined;
 
+    // If this class is used as TogglePreview (not as ShowCommand),
+    // hide (dispose) the previwe panel when it's already visible
+    if (!this._isShowCommand && this._panel.isVisible()) {
+      this._panel.hide();
+      return undefined;
+    }
+
```
kevinslin, line 67: if the user is using `ShowPreview`, can we show a deprecation message and track it? see [[Deprecate Command|dendron://dendron.docs/pkg.plugin-core.dev.cook.deprecate-command]]

kevinslin (changes requested):

packages/plugin-core/src/constants.ts
hayata-suenaga: so regardless if the user invokes `ShowPreview` or `TogglePreview`, the logic for `TogglePreview` should be executed?

hayata-suenaga: so regardless if the user invokes `ShowPreview` or `TogglePreview`, the logic for `TogglePreview` should be executed?

currently, I'm passing `isShowCommand` flag to the command to accommodate the old behavior when user invokes `ShowToggle`. If the above is the case, I can just get rid of the flag and always execute toggle logic.

hayata-suenaga: except I only need the flag for showing deprecation message when `ShowToggle` is used

hayata-suenaga: I think I still need to maintain show command's key bindings for the above reason. I need to pass a flag to show deprecation message.

If our intention is to execute toggle logic for both `ShowPreview` and `TogglePreview`, I can change code inside `execute` instead of changing the key bindings.

kevinslin: we'll execute toggle logic for both. the reason to show the deprecation message is if the user set a custom keyboard shortcut for `show toggle` that will stop working once we remove the command

kevinslin (commented):
nice! one last thing - can you update the package.json according to https://github.com/dendronhq/dendron-docs/blob/main/notes/pkg.plugin-core.dev.cook.update-pkg-json.md#L8:L8?

hayata-suenaga:
> nice! one last thing - can you update the package.json according to https://github.com/dendronhq/dendron-docs/blob/main/notes/pkg.plugin-core.dev.cook.update-pkg-json.md#L8:L8?

Running `gen:config` task didn't change any file

packages/plugin-core/src/constants.ts
kevinslin, line 847: can you uncommen the keyboard bindings here and remove the ones for show preview?

PR #3165, opened 2022-06-28
enhance(workspace): seed vaults are easier to migrate

SeriousBug:
Seed vaults previously read the `fsPath` for the seed vault from the built-in seed config. This makes it very difficult to migrate seed vaults: since the seed config is built into Dendron, one would have to time the migration with the release or users won't be able to access the seed vault. Even if the seed config was downloaded online, we'd have to refresh the seed config with every workspace sync or risk users hitting issues during migration.

This PR instead makes the seed commands read the `fsPath` from the workspace config file inside the seed vault itself. This makes it easier to perform migrations, because as soon as users sync their workspace they will also have the updated fsPath, and migration can be done without updating the seed config. The seed config is reserved for metadata that can't be obtained from the vault itself.


bpathakota (approved):
PR title looks good

PR #3174, opened 2022-06-30
feat(workspace): smart note refs

kevinslin:
- introduce smart note references as described in [[smart header references|dendron://dendron.docs/rfc.47-better-note-references#smart-header-references]]
- introduce `enableSmartRef` config option with default set to true
    - NOTE: decided to go with this instead of `enableLegacyRef` - the plan is to manually migrate everyone to the new syntax and remove legacy ref syntax
- fix bug in `extractHeader` (was mutating the block when extracting)
- copy note ref behavior update, if using `enableSmartRef` and a header is selected, don't generate `#{header},1:#*` but `#{header}` instead

Docs: https://github.com/dendronhq/dendron-site/pull/573

### DIFF
--- a/packages/common-all/data/dendron-yml.validator.json
+++ b/packages/common-all/data/dendron-yml.validator.json
@@ -753,6 +753,9 @@
         "enableEditorDecorations": {
           "type": "boolean"
         },
+        "enableSmartRefs": {
+          "type": "boolean"
+        },
         "feedback": {
           "type": "boolean"
         },
--- a/packages/plugin-core/package.json
+++ b/packages/plugin-core/package.json
@@ -226,6 +226,10 @@
         "title": "Dendron: Random Note"
       },
       {
+        "command": "dendron.renameNoteV2a",
+        "title": "Dendron: Rename Note V2a"
+      },
+      {
         "command": "dendron.renameNote",
         "title": "Dendron: Rename Note"
       },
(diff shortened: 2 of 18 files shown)


packages/common-all/data/dendron-yml.validator.json
```diff
@@ -753,6 +753,9 @@
         "enableEditorDecorations": {
           "type": "boolean"
         },
+        "enableSmartRefs": {
```
jonathanyeung, line 756: I'm not as keen on this name as I don't think it represents what the setting is doing.  I don't think one ref is 'smarter' or 'dumber' than the other, it's primarily just a different syntax.  So my preference would be like 'EnableHeaderGroupedRefSyntax' / 'EnableLegacyRefSyntax' (for the inverse setting) or something along those lines.

packages/plugin-core/package.json
```diff
@@ -225,6 +225,10 @@
         "command": "dendron.randomNote",
         "title": "Dendron: Random Note"
       },
+      {
```
jonathanyeung, line 228: unrelated to this review, but I think we don't actually need this command to be registered? Can we remove from const.ts (maybe separate review).

jonathanyeung (approved):
just one comment on the setting name, otherwise lgtm

packages/common-all/data/dendron-yml.validator.json
kevinslin: i can change it but i'm also planning on removing this as an option all together (we'll remove this in two weeks). its only an option to help with migration. do you think its worth the effort to change?

jonathanyeung: nah- that's the best case :) let's leave it then

PR #3183, opened 2022-07-04
chore(workspace): improve initial startup experience

kevinslin:
chore(workspace): improve initial startup experience

- remove v100 launch screen
- fix lapsed user message being shown to new users

### DIFF
--- a/packages/plugin-core/src/utils/StartupPrompts.ts
+++ b/packages/plugin-core/src/utils/StartupPrompts.ts
@@ -36,17 +36,18 @@ export class StartupPrompts {
       );
     }
 
+    const timeFromFirstInstall = CUR_TIME.minus(
+      Duration.fromObject({ seconds: metaData.firstInstall })
+    );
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/utils/StartupPrompts.ts
```diff
@@ -36,17 +36,18 @@ export class StartupPrompts {
       );
     }
 
+    const timeFromFirstInstall = CUR_TIME.minus(
```
jonathanyeung, line 39: What's the functional difference here? Looks the same to me, but maybe my eyes aren't working.

kevinslin: same, just turned into variables to make it a little more readable

jonathanyeung: Ah ok - so the 'fix lapsed user message being shown to new users' part of the change comes from changing the 'break' to a return' statement - got it.

jonathanyeung (approved):

PR #3186, opened 2022-07-05
fix template applying twice

kevinslin:


kevinslin:
> Can you explain what the proposed fix is? I see a bunch of changes about finding a header, but I'm not sure I see how that's relevant to this issue.

My bad - so this PR actually includes two changes. Added description in pr comment: https://github.com/dendronhq/dendron/pull/3186#issue-1293675756

packages/plugin-core/src/commands/MoveHeader.ts
```diff
@@ -132,18 +135,40 @@ export class MoveHeaderCommand extends BasicCommand<
...
+      targetIndex = index;
       return false;
     });
-    if (!targetHeader) {
+    if (!targetHeader || _.isUndefined(targetIndex)) {
       throw this.headerNotSelectedError;
     }
-    return { proc, origin: maybeNote, targetHeader };
+
+    // eslint-disable-next-line no-unused-expressions
```
jonathanyeung, line 158: nit: is this eslint disable necessary? looks like we're using resp.

jonathanyeung (approved):

PR #3200, opened 2022-07-06
enhance(workspace): Create new GET CLI command to use new engine/store

tma66:
**Background**
Continuing from https://github.com/dendronhq/dendron/pull/3125, this PR will begin to slowly integrate the new store system with the dendron CLI by introducing a new command "GET" that will get a note by id. This will also involve creating a new engine to initialise and populate the new stores.

Through this change, we will be able to sanity check the init and get logic of the new store.

**Changes**
1. New GET cli command to get note by id
2. New engine, DendronEngineV3 (name to possibly change after removing old), that will implement initNotes logic and get logic
3. Introduce `findNotesMeta`. With the new stores storing notemetadata and noteprops separately, the new engine will support retrieval of either components.
4. New `NoteParserV2` (will replace current noteparser). Big differences from current is that we will no longer depend on store/engine and schema logic will be done outside this class

**Test**
Cli testing:
`dendron note get --useLocalEngine --query "apples"` -> returns expected note

init testing:
Compared existing init notes vs new and verified same number of notes were created
```
dendron.community has 136 notes
org-incubator has 543 notes
dendron.blog has 29 notes
dendron-docs has 678 notes
handbook has 281 notes
org-dev has 388 notes
vault has 918 notes
dendron-users has 4906 notes
org-private has 18384 notes
```

### DIFF
--- a/packages/common-all/src/noteDictsUtils.ts
+++ b/packages/common-all/src/noteDictsUtils.ts
@@ -160,4 +160,26 @@ export class NoteFnameDictUtils {
     }
     return true;
   }
+
+  /**
+   * Merge two NotePropsByFnameDict into a single NotePropsByFnameDict
+   * If key exists in both, merge values into a single array
+   *
+   * @return new merged NotePropsByFnameDict without modifying existing NotePropsByFnameDicts
+   */
+  static merge(
+    fnameDictOne: NotePropsByFnameDict,
+    fnameDictTwo: NotePropsByFnameDict
+  ) {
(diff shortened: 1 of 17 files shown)


kevinslin (approved):
nice job! some followups: can you write the docs for `get`? (see [[dendron note lookup|dendron://dendron.dendron-site/dendron.topic.notes.cli.lookup]]) as an example

tma66:
> dendron.topic.notes.cli.lookup

Added [[Get|dendron://dendron.dendron-site/dendron.topic.notes.cli.get]]

PR #3226, opened 2022-07-11
Feat/add private note config

hayata-suenaga:
## Add a publish config option on wikilink conversion behavior

### DIFF
--- a/packages/common-all/data/dendron-yml.validator.json
+++ b/packages/common-all/data/dendron-yml.validator.json
@@ -1174,6 +1174,13 @@
         },
         "siteBanner": {
           "type": "string"
+        },
+        "privateNoteBehavior": {
+          "type": "string",
+          "enum": [
+            "privateLink",
+            "titleFallback"
+          ]
         }
       },
       "required": [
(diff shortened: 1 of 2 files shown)


packages/plugin-core/src/utils/pods.ts
```diff
@@ -44,7 +43,8 @@ export const launchGoogleOAuthFlow = async (id?: string) => {
   );
 
   const stringifiedParams = queryString.stringify({
-    client_id: GOOGLE_OAUTH_ID,
+    client_id:
```
kevinslin, line 46: do not commit your google client id

packages/engine-test-utils/src/__tests__/engine-server/markdown/dendronPub.spec.ts
```diff
@@ -202,20 +210,93 @@ describe("GIVEN dendronPub", () => {
...
+      config = genPublishConfigWithPublicPrivateHierarchies();
+    });
+
+    describe("AND WHEN noteref of published note", () => {
+      let resp: VFile;
 
-    describe("AND WHEN noteRef", () => {
-      describe("AND WHEN noteref of published note", () => {
-        let resp: VFile;
+      beforeAll(async () => {
```
kevinslin, line 222: the reason you are seeing test issues is because `beforeAll` runs before `beforeEach`. you'll want to move the config initialization into the beforeAll block

kevinslin (changes requested):

PR #3233, opened 2022-07-13
fix(workspace): workspace sync will maintain proper engine state

jonathanyeung:
## fix(workspace): workspace sync will maintain proper engine state

This change fixes an issue where notes added by workspace sync were causing engine state drift.  The root cause is that the `updateNote` API doesn't return `NoteChangedEvents[]`, and we were using `updateNote` instead of `writeNote` in the file watcher `onCreate` callback.

This change adds a full `NoteChangedEvents[]` in the updateNote response.  I've intentionally not added tests for this, since we will likely modify this API surface as part of @tma66's refactoring work.

---

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

## Docs

### DIFF
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -1062,29 +1062,41 @@ export class FileStorage implements DStore {
 
     const changes: NoteChangeEntry[] = [];
 
-    const noteDicts = {
-      notesById: this.notes,
-      notesByFname: this.noteFnames,
-    };
-    if (opts?.newNode) {
-      const changesToParents = NoteUtils.addOrUpdateParents({
-        note,
-        noteDicts,
-        createStubs: true,
-        wsRoot: this.wsRoot,
-      });
-      changesToParents.forEach((changedEntry) => {
-        NoteDictsUtils.add(changedEntry.note, noteDicts);
-        changes.push(changedEntry);
-      });
+    try {
+      const noteDicts = {
+        notesById: this.notes,
+        notesByFname: this.noteFnames,
+      };
+      if (opts?.newNode) {
+        const changesToParents = NoteUtils.addOrUpdateParents({
+          note,
+          noteDicts,
+          createStubs: true,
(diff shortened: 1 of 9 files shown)


hikchoi (changes requested):
One small comment, but otherwise looks good.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -1060,24 +1060,32 @@ export class FileStorage implements DStore {
...
+      changesToParents.forEach((changedEntry) => {
+        NoteDictsUtils.add(changedEntry.note, noteDicts);
+        changes.push(changedEntry);
+      });
+
+      changes.push({
+        note,
+        status: "create",
+      });
     }
```
tma66, line 1085: We should also push entry for "update" if its not a newNode.

packages/plugin-core/src/services/TextDocumentService.ts
```diff
@@ -172,7 +175,16 @@ export class TextDocumentService implements ITextDocumentService {
       fname,
       vault,
     });
-    return engine.updateNote(props);
+
+    const resp = await engine.updateNote(props);
```
tma66, line 179: It might be simpler to just return props if there is no resp.error.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -1060,24 +1060,32 @@ export class FileStorage implements DStore {
...
     this.logger.debug({ ctx, note: NoteUtils.toLogObj(note) });
     NoteDictsUtils.add(note, noteDicts);
-    return note;
+    return { data: changes, error: null };
```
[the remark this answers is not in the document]
jonathanyeung: Maybe I'm misunderstanding your question, but it's necessary in order to match the updated interface definition.  I've wrapped in a try catch block to propagate an error if it should occur.

kevinslin (approved):

PR #3237, opened 2022-07-13
enhance(workspace): 'Fix It' button in warning toaster for duplicate note id

Harshita-mindfire:
This PR aims to add a Fix It button in duplicate note id warning toaster. Fix It button runs the doctor regenerate note id command underneath the hood. 
This PR also increases the range of no frontmatter error from 0 to line 8.

*** 
Telemetry:
Adds a property `state: "resolved"` for an existing workspace event: `DuplicateNoteFound`

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA

## Docs
NA

### Extended
NA

### DIFF
--- a/packages/plugin-core/src/components/doctor/utils.ts
+++ b/packages/plugin-core/src/components/doctor/utils.ts
@@ -5,7 +5,9 @@ import {
   WorkspaceEvents,
 } from "@dendronhq/common-all";
 import { file2Note } from "@dendronhq/common-server";
+import { DoctorActionsEnum } from "@dendronhq/engine-server";
 import * as vscode from "vscode";
+import { DoctorCommand } from "../../commands/Doctor";
 import { ExtensionProvider } from "../../ExtensionProvider";
 import { Logger } from "../../logger";
 import { AnalyticsUtils } from "../../utils/analytics";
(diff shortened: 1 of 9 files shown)


packages/plugin-core/src/components/doctor/utils.ts
```diff
@@ -88,7 +90,30 @@ export class DoctorUtils {
...
         });
-        VSCodeUtils.showMessage(MessageSeverity.WARN, error.message, {});
+        VSCodeUtils.showMessage(
+          MessageSeverity.WARN,
+          error.message,
+          {},
+          { title: "Fix It" }
+        ).then((resp) => {
+          if (resp && resp.title === "Fix It") {
+            const cmd: vscode.Command = {
```
hikchoi, line 100: the doctor action probably needs to force a reload after it's run (or make sure the result of the action actually emits the correct note change entry once it's done).

Depending on the note's filename, there will be cases where note A is recognized by the engine or note B, but not at the same time (since we don't allow duplicate ids in the engine). Where you run the doctor command would make a difference I think.

If you ran it in a note that wasn't in the engine state, it would ideally be a `create`, but if you ran it in a note that was in the engine, it would be a `delete` followed by a `create`.

These might be some possible reasons why the note changes are not detected in the test you added.

hikchoi (commented):
left a comment regarding engine state

packages/plugin-core/src/components/doctor/utils.ts
Harshita-mindfire: Did some debugging, it looks like the test is completed before the doctor actions gets completed. Thanks for the help!

kevinslin:
> left a comment regarding engine state

@hikchoi are there any more outstanding issues in this pr?

hikchoi:
> > left a comment regarding engine state
> 
> @hikchoi are there any more outstanding issues in this pr?

Nope. my comment was on the possible reason the test weren't working correctly (synced with @Harshita-mindfire separately). I have no other outstanding comments on the PR itself

hikchoi (approved):

jonathanyeung (approved):

PR #3241, opened 2022-07-14
enhance(workspace): Add new FIND notes CLI command

tma66:
**Background**
Continuing to test the new dendron store via the notes cli command, we will also introduce a new command `FIND` that will find notes based on note properties (currently only fname and vault).

**Changes**
1. New `FIND` cli command that uses new engine
2. Migrate `LOOKUP` command to use new engine

TODO: update docs after approval

### DIFF
--- a/packages/common-all/src/store/INoteStore.ts
+++ b/packages/common-all/src/store/INoteStore.ts
@@ -23,6 +23,15 @@ export interface INoteStore<K> {
   get(key: K): Promise<RespV3<NoteProps>>;
 
   /**
+   * Bulk get NoteProps by list of key
+   * If no notes are found, return empty list.
+   *
+   * @param key: keys of NoteProps
+   * @return list of NoteProps
+   */
+  bulkGet(keys: K[]): Promise<RespV3<NoteProps>[]>;
+
+  /**
    * Get NoteProps metadata by key
    * If key is not found, return error.
    *
--- a/packages/engine-server/src/DendronEngineV3.ts
+++ b/packages/engine-server/src/DendronEngineV3.ts
@@ -30,6 +30,7 @@ import {
   IFileStore,
   INoteStore,
   IntermediateDendronConfig,
+  isNotUndefined,
   NoteChangeEntry,
   NoteDicts,
   NoteDictsUtils,
@@ -40,6 +41,7 @@ import {
   NotePropsMeta,
   NoteUtils,
   Optional,
+  QueryNotesOpts,
   RenameNotePayload,
   RenderNotePayload,
   RespV2,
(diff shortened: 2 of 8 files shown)


packages/common-all/src/store/INoteStore.ts
```diff
@@ -22,6 +22,15 @@ export interface INoteStore<K> {
...
   get(key: K): Promise<RespV3<NoteProps>>;
 
+  /**
+   * Bulk get NoteProps by list of key
+   * If no notes are found, return empty list.
+   *
+   * @param key: keys of NoteProps
+   * @return list of NoteProps
+   */
+  bulkGet(keys: K[]): Promise<RespV3<NoteProps>[]>;
```
jonathanyeung, line 32: should this be `<RespV3<NoteProps[]>>` instead of `<RespV3<NoteProps>[]>`?

that is - do we want to report errors on the overall API call or on a per-get basis.  I think the former makes more sense.

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -341,8 +343,46 @@ export class DendronEngineV3 implements DEngine {
...
+  /**
+   * See {@link DEngine.queryNotes}
+   */
+  async queryNotes(
+    opts: QueryNotesOpts
+  ): ReturnType<DEngineClient["queryNotes"]> {
+    const ctx = "Engine:queryNotes";
+    const { qs, vault, onlyDirectChildren, originalQS } = opts;
+
+    // Need to ignore this because the engine stringifies this property, so the types are incorrect.
```
jonathanyeung, line 355: Hm - where is this stringify happening?

jonathanyeung (approved):
just some minor comments

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -341,8 +343,46 @@ export class DendronEngineV3 implements DEngine {
...
+        onlyDirectChildren,
+        originalQS,
+      })
+      .map((ent) => ent.id);
+
+    if (noteIds.length === 0) {
+      return { error: null, data: [] };
+    }
+
+    // TODO: Support createifNew? Doesn't seem to be actively used
```
kevinslin, line 372: should not be used here, we should refactor so that lookup method will call `engine.create` if `queryNotes` returns empty

kevinslin (approved):

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -341,8 +343,46 @@ export class DendronEngineV3 implements DEngine {
...
   }
 
-  async queryNotes(): ReturnType<DEngineClient["queryNotes"]> {
-    throw Error("queryNotes not implemented");
+  /**
+   * See {@link DEngine.queryNotes}
+   */
+  async queryNotes(
+    opts: QueryNotesOpts
+  ): ReturnType<DEngineClient["queryNotes"]> {
```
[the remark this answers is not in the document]
tma66: Oh that was copied from v2. fixed now

packages/engine-server/src/DendronEngineV3.ts
tma66: Also copied over lol. I'll leave this for now and do some cleanup as follow through

PR #3244, opened 2022-07-14
fix(publish): current menu item on the sidebar wont collapse for a published site

Harshita-mindfire:
This PR aims the fix the issue with the current item on the sidebar not collapsing. 

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA
## Docs
NA

### Extended
NA

### DIFF
--- a/packages/nextjs-template/components/DendronTreeMenu.tsx
+++ b/packages/nextjs-template/components/DendronTreeMenu.tsx
@@ -45,19 +45,18 @@ export default function DendronTreeMenu(
     if (!noteActiveId || !tree) {
       return undefined;
     }
-
     logger.info({
       state: "useEffect:preCalculateTree",
     });
 
     // all parents should be in expanded position
-    const activeNoteIds = TreeUtils.getAllParents({
+    const newActiveNoteIds = TreeUtils.getAllParents({
       child2parent: tree.child2parent,
       noteId: noteActiveId,
     });
 
-    setActiveNoteIds(activeNoteIds);
-  }, [props.noteIndex, dendronRouter.query.id, noteActiveId]);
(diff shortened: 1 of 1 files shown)


kevinslin (changes requested):
looks good, one issue when trying to export is that `DendronTreeMenu` can return undefined which is not a valid react construct. 

```
$ next build && next export
info  - Loaded env from /Users/kevinlin/code/dendron/packages/nextjs-template/.env.production
Failed to compile.

./components/DendronLayout.tsx:49:8
Type error: 'DendronTreeMenu' cannot be used as a JSX component.
  Its return type '"" | Element | null | undefined' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'.

  47 |         </div>
  48 |       )}
> 49 |       <DendronTreeMenu
     |        ^
  50 |         {...props}
  51 |         collapsed={isCollapsed && isResponsive}
  52 |         setCollapsed={setCollapsed}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
kevinlin@Kevins-MacBook-Pro-2 nextjs-template % yarn export
```

packages/nextjs-template/components/DendronTreeMenu.tsx
```diff
@@ -99,15 +98,17 @@ export default function DendronTreeMenu(
   };
 
   return (
```
kevinslin, line 100: this should return `<MenuView .../> || null` if `noteActiveId` isn't defined

kevinslin (approved):

PR #3253, opened 2022-07-15
fix(views): hide sidebar when clicking a non-submenu note on mobile

hkievet:
# About

The SubMenus were closing the sidebar when clicked, but the MenuItems were not.  AntD Menu Items don't have an `onClick` handler built in to them, but their parent `Menu` component does.  So I used that to handle closing the sidebar.

I've decided to create two separate events, even though they do the same thing, but I think that verbosity is good in the long term.

I was very confused by the `changeActiveNote(noteId, { noteIndex: props.noteIndex });` call on line 82.  It made me think that there was some handler somewhere doing something like that for the `MenuItem` click, but it turns out that AFAIK line 82 was extraneous.  Both sub menus and menu items handle their navigation via the `MenuItemTitle` component's `Link` at the bottom of the file.  So I deleted that confusing line.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [x] lg: screen ≥ 992px
    - [x] safari
    - [x] firefox
    - [x] chrome

## Docs

### Extended

### DIFF
--- a/packages/nextjs-template/components/DendronTreeMenu.tsx
+++ b/packages/nextjs-template/components/DendronTreeMenu.tsx
@@ -34,7 +34,6 @@ export default function DendronTreeMenu(
   const tree = ide.tree;
   const logger = createLogger("DendronTreeMenu");
   const dendronRouter = useDendronRouter();
-  const { changeActiveNote } = dendronRouter;
   const [activeNoteIds, setActiveNoteIds] = useState<string[]>([]);
   const noteActiveId = _.isUndefined(dendronRouter.query.id)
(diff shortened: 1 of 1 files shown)


kevinslin:
nice job! we used to use `changeActiveNote`when we weren't explicitly setting the tree menu items as `<Link/>` items so you were right to remove it. you can also remove the `changeActiveNote` de structuring on line 37

kevinslin (approved):

PR #3263, opened 2022-07-18
chore(internal): fix timeout on some tests

hikchoi:
# chore(internal): fix timeout on some tests

This PR:
- Fix timeout issue on some long running tests that commonly fail during the pipeline.
- This should resolve the recent issue with the plugin test that fails for mac in CI.

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -336,6 +336,7 @@ export async function _activate(
         wsRoot: maybeWsRoot,
         engine: resp.data.engine,
         wsService,
+        opts,
       });
       if (respActivate.error) {
         return false;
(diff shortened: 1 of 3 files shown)


hikchoi:
Seems like Github Actions is going through some degraded performance. most likely not related to the fix here. will re-run tomorrow

kevinslin (approved):

hikchoi:
Adding timeouts to tests that fail on each CI run while I'm at it. Will merge once all pipeline are green at least in this PR

packages/plugin-core/src/_extension.ts
```diff
@@ -336,6 +336,7 @@ export async function _activate(
         wsRoot: maybeWsRoot,
         engine: resp.data.engine,
         wsService,
+        opts,
```
hikchoi, line 339: @kevinslin we were not passing in the skip options to workspace activator so all tests were initializing the tree view. This caused the tests that would otherwise run fast to time out.

kevinslin (approved):

PR #3270, opened 2022-07-19
chore(internal): Fix bug where note is not properly deleted via writeNote

tma66:
**Bug**
When `engine.writeNote` is called and the id is different, the intended behavior is to delete the old note with the old id and create a new note with the new id. Currently there is a out-of-sync issue where the engine retains the old note b/c we forgot to filter out the `delete` entries from changeEntries.


hikchoi:
Wow nice catch! I've encountered this a lot after refactor hierarchy; the note graph would render both old and new notes, and also subsequent renames on the new note would fail unless I reload VSCode first.

kevinslin (approved):

Issue #2175, opened 2022-01-14
Inconsistent layout of mobile view of published pages


andrey-jef:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [x] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When reading some specific pages in Dendron wiki with images, i have to manually zoom in to read on mobile. But some other pages on the Dendron wiki can also be read normally on mobile without the need to zoom. 

## To Reproduce

The page that I need to zoom to read on mobile: https://wiki.dendron.so/notes/5U4eAiqshI67VxIL40KWH/

The page that can be read normally without zooming: https://wiki.dendron.so/notes/81da87be-2d4e-47b5-a1d6-c0d647e1ab00/

On mobile, Safari on iOS in my case, go to the 2 url cited previously.
We will see the differences in readability of each page at default zoom level. 

## Expected behavior

The readability of each published page should be consistent and fixed at default zoom level.

## Screenshots

View of page that need to zoom to read
![IMG_0417](https://user-images.githubusercontent.com/81280428/149543610-163f65e3-2f9e-4994-9a5c-ee29c9d1606d.PNG)

View of page that don't need to zoom
![IMG_0418](https://user-images.githubusercontent.com/81280428/149543651-8a425509-bcbf-41bc-bda7-9a8d590efb36.PNG)

## Desktop (please complete the following information)

- OS: iOS 14
- Browser: safari
- Version: 

## Dendron Log file

Please attach the output of `> Dev:Dendron: Open Logs` here

SeriousBug:
Thanks for reporting the issue, I can see the broken layout too.

kevinslin:
this has been fixed in v105.0

andrey-jef:
> this has been fixed in v105.0

The situation is much better, but I doubt that the inconsistent layout bug still persists. Please consider to reopen the issue.

This is a screencap of a perfectly rendered on iOS.
[...]

PR #3272, opened 2022-07-20
fix(publish): renders consitent layout on mobile and non-mobile

namjul:
docs:

- https://github.com/dendronhq/dendron/issues/2175
- [[Inconsistent layout of mobile view|task.2022.07.17.inconsistent-layout-of-mobile-view]]

Whats still missing:

~~Apply `overflow-y: hidden;` to parent element of code blocks. These are rendered with `white-space: pre;` and wrapping them in an container with overflow restriction will prevent the element to incease the page width. This incease in page width resulted in an "zoomed-out" view. For more details checkout comment descriptions.~~ i wondered why this is not a default but prism actually already does [this](https://github.com/dendronhq/dendron/blob/6a4e3caa758ee977d63726872ba44feaab0815b8/packages/common-assets/assets/css/prism/light.css#L56).

~~PS: I flagged the Compenent `packages/common-frontend/src/components/layouts/BasicLayout.tsx`, since I did not found any usages so we could consider deleting that.~~ removed [`6a4e3ca` (#3272)](https://github.com/dendronhq/dendron/pull/3272/commits/6a4e3caa758ee977d63726872ba44feaab0815b8)


kevinslin:
you're right, `BasicLayout` can be removed

kevinslin (approved):

PR #3276, opened 2022-07-20
enhance(retrieve): remove `Show Preview` Command

kevinslin:
enhance(retrieve): remove `Show Preview` Command

**BREAKING**: remove `Show Preview` Command


packages/plugin-core/package.json
```diff
@@ -1301,7 +1292,7 @@
       },
       {
         "command": "dendron.togglePreview",
-        "windows": "windows+ctrl+p",
+        "windows": "ctrl+k v",
```
jonathanyeung, line 1295: How did we come up with this keybinding? It's very different between mac and windows

jonathanyeung (changes requested):

kevins8:
This is the default windows keybinding for vscode preview. the mac one
should actually also be this, didn't want to change it in case people
already had it set.
the reason we can do it for windows is because users can't actually press
windows+p without triggering the menu

PR #3278, opened 2022-07-20
enhance(preview): open note in different editor group then preview

kevinslin:
enhance(preview): open note in different editor group then preview

When opening a note by clicking a link from preview, open it in an different editor group API.
This makes use of the `Tab Groups` API that is only available in vscode `1.67`. 
Dendron currently supports vscode `1.62` which means users using versions prior to `1.67` will not have this feature enabled.

### DIFF
--- a/packages/plugin-core/src/components/lookup/TabUtils.ts
+++ b/packages/plugin-core/src/components/lookup/TabUtils.ts
@@ -2,6 +2,10 @@ import _ from "lodash";
 import semver from "semver";
 import { Event, version as vscodeVersion, ViewColumn, window } from "vscode";
 
+/**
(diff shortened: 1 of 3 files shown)


packages/plugin-core/src/components/lookup/TabUtils.ts
```diff
@@ -0,0 +1,136 @@
+import _ from "lodash";
+import semver from "semver";
+import { Event, version as vscodeVersion, ViewColumn, window } from "vscode";
+
+export interface Tab {
```
jonathanyeung, line 5: can you add comments to each of these interfaces saying that these are copied from VSCode, and that they should be removed after we decide to bump to 1.67 or greater?

jonathanyeung (changes requested):
one comment about code duplication, otherwise good to merge

PR #3284, opened 2022-07-21
enhance(workspace): Add delete/write logic to new engine

tma66:
**Background**
Continuing on from https://github.com/dendronhq/dendron/pull/3200, this PR updates the notes cli `DELETE` command to use the new engine.

**Changes**
1. Migrate delete/write logic from old engine to new engine
2. Add support in `NoteStore` to find notes by child (ie return parent or closest ancestor)
3. Add support in fuse engine to update notes index based on `NoteChangeEntry` entries
4. Update `NoteStore` methods to take into account stub notes. Ie if we are getting/writing a stub, skip filesystem logic
5. Migrate `DELETE` notes cli  command to new engine
6. Update delete logic for both engines. If deleting a note and replacing with a stub, create stub with new id. Previously we were keeping the same id for the stub, but in order to be consistent, all new notes should have new ids.

### DIFF
--- a/packages/common-all/src/store/INoteStore.ts
+++ b/packages/common-all/src/store/INoteStore.ts
@@ -1,5 +1,4 @@
 import {
-  BulkResp,
   DNoteLoc,
   FindNoteOpts,
   NoteProps,
@@ -47,7 +46,7 @@ export interface INoteStore<K> {
    * @param opts: NoteProps find criteria
    * @return List of NoteProps that matches criteria
    */
-  find(opts: FindNoteOpts): Promise<BulkResp<NoteProps[]>>;
+  find(opts: FindNoteOpts): Promise<RespV3<NoteProps[]>>;
 
   /**
    * Find NoteProps metadata by criteria. If no criteria is set, return empty array.
--- a/packages/common-all/src/store/INoteStore.ts
+++ b/packages/common-all/src/store/INoteStore.ts
@@ -93,6 +93,16 @@ export interface INoteStore<K> {
   delete(key: K): Promise<RespV3<string>>;
 
   /**
+   * Delete NoteProps metadata from storage layer for given key.
+   * If key does not exist, do nothing.
+   * Unlike {@link INoteStore.delete}, this will not touch the filesystem
+   *
+   * @param key: key of NoteProps to delete
+   * @return original key
+   */
+  deleteMetadata(key: K): Promise<RespV3<string>>;
+
+  /**
    * Rename location of NoteProps
    * If old location does not exist, return error.
    *
--- a/packages/common-all/src/types/store.ts
+++ b/packages/common-all/src/types/store.ts
@@ -3,10 +3,20 @@ import { DVault } from "./workspace";
 
 // Types used on the store layer
 
+/**
+ * Properties used to find notes by. If multiple properties are provided, then returned notes
+ * must satisfy all constraints.
+ *
+ * For example, if fname = "foo" and vault = "vaultOne", then only notes with fname "foo" in vault "vaultOne" are returned
+ */
 export type FindNoteOpts = {
+  // Find notes by fname
   fname?: string;
   // If vault is provided, filter results so that only notes with matching vault is returned
   vault?: DVault;
+  // Find note by child. This usually refers to parent but could also be closest ancestor if parent doesn't exist. Max one note should be returned
+  // E.g. if childFname = `baz.one.two` and `baz.one` does not exist, then `baz` will be returned
+  child?: NotePropsMeta; // Noteprops metadata of child
 };
 
 export type WriteNoteOpts<K> = {
--- a/packages/common-all/src/types/store.ts
+++ b/packages/common-all/src/types/store.ts
@@ -14,9 +14,6 @@ export type FindNoteOpts = {
   fname?: string;
   // If vault is provided, filter results so that only notes with matching vault is returned
   vault?: DVault;
-  // Find note by child. This usually refers to parent but could also be closest ancestor if parent doesn't exist. Max one note should be returned
-  // E.g. if childFname = `baz.one.two` and `baz.one` does not exist, then `baz` will be returned
-  child?: NotePropsMeta; // Noteprops metadata of child
 };
 
 export type WriteNoteOpts<K> = {
--- a/packages/common-all/src/types/typesv2.ts
+++ b/packages/common-all/src/types/typesv2.ts
@@ -35,10 +35,6 @@ export type EngineDeleteOpts = {
    * If node is deleted and parents are stubs, default behavior is to alsod delete parents
    */
   noDeleteParentStub?: boolean;
-  /**
-   * If the deleted note has children, replace the deleted note with a newly created stub note in place.
-   */
-  replaceWithNewStub?: boolean;
 };
 
 export type NoteLink = {
@@ -327,6 +323,14 @@ export type EngineWriteOptsV2 = {
    * Should any configured hooks be run during the write
    */
   runHooks?: boolean;
+  /**
+   * If true, overwrite existing note with same fname and vault, even if note has a different id
+   */
+  overrideExisting?: boolean;
+  /**
+   * If true, write only to metadata store
+   */
+  metaOnly?: boolean;
 };
 
 export type DEngineInitPayload = {
@@ -456,6 +460,10 @@ export type DCommonMethods = {
   ): Promise<UpdateNoteResp>;
   updateSchema: (schema: SchemaModuleProps) => Promise<void>;
 
+  /**
+   * Write note to metadata store and/or filesystem. This will update existing note or create new if one doesn't exist.
+   * If another note with same fname + vault but different id exists, then return error (otherwise overrideExisting flag is passed)
+   */
   writeNote: (
     note: NoteProps,
     opts?: EngineWriteOptsV2
@@ -516,7 +524,7 @@ export type WorkspaceExtensionSetting = {
 
 // --- KLUDGE END
 
-export type EngineDeleteNoteResp = Required<RespV2<EngineDeleteNotePayload>>;
+export type EngineDeleteNoteResp = RespV2<EngineDeleteNotePayload>;
 export type NoteQueryResp = RespV2<NoteProps[]>;
 export type SchemaQueryResp = Required<RespV2<SchemaModuleProps[]>>;
 export type StoreDeleteNoteResp = EngineDeleteNotePayload;
@@ -582,6 +590,9 @@ export type DEngine = DCommonProps &
      * Find NoteProps metadata by note properties. If no notes metadata match, return empty list
      */
     findNotesMeta: (opts: FindNoteOpts) => Promise<NotePropsMeta[]>;
+    /**
+     * Delete note from metadata store and/or filesystem. If note doesn't exist, return error
+     */
     deleteNote: (
       id: string,
       opts?: EngineDeleteOpts
--- a/packages/dendron-cli/src/commands/notes.ts
+++ b/packages/dendron-cli/src/commands/notes.ts
@@ -193,7 +193,8 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
     if (
       args.cmd === NoteCommands.GET ||
       args.cmd === NoteCommands.LOOKUP ||
-      args.cmd === NoteCommands.FIND
+      args.cmd === NoteCommands.FIND ||
+      args.cmd === NoteCommands.DELETE
     ) {
       args.newEngine = true;
     }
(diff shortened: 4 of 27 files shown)


packages/dendron-cli/src/commands/notes.ts
```diff
@@ -341,11 +342,12 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
         }
         case NoteCommands.DELETE: {
           const { query, vault } = checkQueryAndVault(opts);
-          const note = NoteUtils.getNoteByFnameFromEngine({
-            fname: query,
-            vault,
-            engine,
-          });
+          const note = (
+            await engine.findNotes({
```
kevinslin, line 346: makes me think we should have a `findUnique` method for these cases like [prisma](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#findunique). maybe future iteration

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -267,65 +263,49 @@ export class FileStorage implements DStore {
       await fs.unlink(fpath);
     }
 
-    // if have children, keep this node around as a stub
+    // if have children, create stub note with a new id
     if (!_.isEmpty(noteToDelete.children)) {
```
kevinslin, line 267: do the children need to add the new stub note as their parent id?

packages/common-all/src/types/store.ts
```diff
@@ -3,10 +3,20 @@ import { DVault } from "./workspace";
...
+ * For example, if fname = "foo" and vault = "vaultOne", then only notes with fname "foo" in vault "vaultOne" are returned
+ */
 export type FindNoteOpts = {
+  // Find notes by fname
   fname?: string;
   // If vault is provided, filter results so that only notes with matching vault is returned
   vault?: DVault;
+  // Find note by child. This usually refers to parent but could also be closest ancestor if parent doesn't exist. Max one note should be returned
+  // E.g. if childFname = `baz.one.two` and `baz.one` does not exist, then `baz` will be returned
+  child?: NotePropsMeta; // Noteprops metadata of child
```
kevinslin, line 19: can we rename this to `ancestor` to make this explicit?

packages/engine-server/src/store/NoteStore.ts
```diff
@@ -60,6 +60,15 @@ export class NoteStore implements Disposable, INoteStore<string> {
     if (metadata.error) {
       return { error: metadata.error };
     }
+
+    // If note is a stub, return stub note
+    if (metadata.data.stub) {
+      return {
+        data: _.merge(metadata.data, {
```
kevinslin, line 67: merge will mutate the source object (`metadata.data`). do we want to destructure and return a new object instead?

packages/engine-test-utils/src/__tests__/engine-server/store/noteStore.spec.ts
```diff
@@ -46,17 +48,77 @@ describe("GIVEN NoteStore", () => {
...
+        metadataResp = await noteStore.findMetaData({
+          vault: vaults[0],
+        });
+        expect(metadataResp.data!.length).toEqual(4);
+
+        // Test NoteStore.findMetaData child property
+        // Closest ancestor of foo.ch1 = foo
+        const fooCh1 = await noteStore.get("foo.ch1");
+        metadataResp = await noteStore.findMetaData({
+          child: fooCh1.data,
```
kevinslin, line 73: compiler error, this `data` could be undefined

packages/engine-test-utils/src/__tests__/engine-server/store/noteStore.spec.ts
```diff
@@ -46,17 +48,77 @@ describe("GIVEN NoteStore", () => {
...
+
+        // Closest ancestor of bar.abc.123 = bar
+        const bar123 = await NoteTestUtilsV4.createNote({
+          fname: "bar.abc.123",
+          body: "bar.abc.123",
+          vault: vaults[0],
+          wsRoot,
+        });
+        metadataResp = await noteStore.findMetaData({
+          child: bar123,
```
kevinslin, line 87: see above. also apply for subsequent occurences

kevinslin (changes requested):
nice job. initial pass, left some comments

packages/dendron-cli/src/commands/notes.ts
tma66: `findUnique` or `findFirst` looks tempting. I was concerned about polluting the api with too many methods since we already have findNote and findNoteMeta. Perhaps when we refactor the clients it'll give us a chance to revisit these

packages/engine-server/src/drivers/file/storev2.ts
tma66: Good catch. Fixed and updated test cases

packages/common-all/src/types/store.ts
tma66: So I've intended these to be properties we find notes "by". Ie find notes where fname = "foo" or find notes where child = "baz".

For the child property, it technically should return just the parent note, but I've expanded it to include ancestor since we have multiple use cases for it.

I'd like to keep it child to keep it consistent with the other properties

packages/engine-server/src/store/NoteStore.ts
tma66: good to know, fixed

packages/dendron-cli/src/commands/notes.ts
jonathanyeung: isn't that the same as `getNote`?

packages/common-all/src/store/INoteStore.ts
```diff
@@ -92,6 +92,16 @@ export interface INoteStore<K> {
...
 
+  /**
+   * Delete NoteProps metadata from storage layer for given key.
+   * If key does not exist, do nothing.
+   * Unlike {@link INoteStore.delete}, this will not touch the filesystem
+   *
+   * @param key: key of NoteProps to delete
+   * @return original key
+   */
+  deleteMetadata(key: K): Promise<RespV3<string>>;
```
jonathanyeung, line 103: Why would someone want to do this and not delete the note as well?

packages/common-all/src/types/store.ts
jonathanyeung: Why would someone want to do this? Why not just find the child note explicitly, then get its parent?  

Not sure that this parameter is necessary, but if it is I'd prefer to break it out separately.  It's not clear what it means to "find a note by child".  It'd be better to have a method `getAncestor(id: string)`

packages/common-all/src/types/typesv2.ts
```diff
@@ -456,6 +460,10 @@ export type DCommonMethods = {
   ): Promise<UpdateNoteResp>;
   updateSchema: (schema: SchemaModuleProps) => Promise<void>;
 
+  /**
+   * Write note to metadata store and/or filesystem. This will update existing note or create new if one doesn't exist.
```
jonathanyeung, line 464: super nit: comments run a tad long - there's this neat lightweight extension in vs code that can reflow your comments: https://marketplace.visualstudio.com/items?itemName=stkb.rewrap

packages/engine-server/src/store/NoteMetadataStore.ts
```diff
@@ -118,4 +140,40 @@ export class NoteMetadataStore implements IDataStore<string, NotePropsMeta> {
...
+
+  /**
+   * Recursively search through fname to find next available ancestor note.
+   *
+   * E.g, if fpath = "baz.foo.bar", search for "baz.foo", then "baz", then "root" until first valid note is found
+   * @param fpath of note to find ancestor of
+   * @param vault of ancestor note
+   * @returns closest ancestor note
+   */
+  private async findClosestAncestor(
```
jonathanyeung, line 152: why would we do this and not use parent pointers in the NoteProps?

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -290,12 +296,328 @@ export class DendronEngineV3 implements DEngine {
...
+
+    // Check if another note with same fname and vault exists
+    const resp = await this._noteStore.find({
+      fname: note.fname,
+      vault: note.vault,
+    });
+    const existingNote = resp.data ? resp.data[0] : undefined;
+    // If a note exists with a different id but same fname/vault, then we throw an error unless its a stub or override is set
+    if (existingNote && existingNote.id !== note.id) {
+      // If note is a stub or client wants to override existing note, we need to update parent/children relationships since ids are different
```
jonathanyeung, line 371: are we making it a goal in the new engine to keep link state up to date?

jonathanyeung (changes requested):

packages/engine-server/src/DendronEngineV3.ts
kevinslin: link state needs to be up to date for refactoring operations as well as deletion to work properly. so yes

packages/common-all/src/types/store.ts
kevinslin: this is used internally for cases like delete. see https://github.com/dendronhq/dendron/pull/3284/files#diff-ef9b24e9ac675f8a705cd6fd0c0ce40ab1ba6419dd498e51c66d23ba9bc53de9R376

kevinslin (approved):

packages/common-all/src/types/store.ts
jonathanyeung: But why can't we use parent pointers?

packages/engine-server/src/DendronEngineV3.ts
jonathanyeung: In which case, looks like we're missing some logic here?

packages/dendron-cli/src/commands/notes.ts
tma66: `getNote` currently just supports getting by note id. Are you suggesting we overload it to pass it 2 different sets of parameters `id || (fname + vault)`?

packages/common-all/src/types/store.ts
tma66: Actually the main usage right now is during initNotes or writeNote for when we are adding a new note that doesn't have a parent yet. Ie we want to add `foo.bar.baz` in the engine but only `foo` exists, in which case we need to create a stub for `foo.bar` first.

packages/engine-server/src/store/NoteMetadataStore.ts
tma66: This is for when we are creating a brand new note and we don't know know its parent yet

packages/common-all/src/types/typesv2.ts
tma66: awesome thanks, will take a look

packages/engine-server/src/DendronEngineV3.ts
tma66: If we're referring to backlinks, I'm including that in the next pr as part of the `rename` refactoring since it includes some link modification logic

packages/dendron-cli/src/commands/notes.ts
jonathanyeung: Er, I guess I'm just confused.  If `findUnique` takes fname + vault, what's the difference between that and `findNote()[0]` or something similar. 

Seems like a convenience method, but I don't _think_ we should be trying to do stuff like `findNote()[0]` if possible.

packages/common-all/src/types/store.ts
jonathanyeung: Ah ok - that makes sense.  I still strongly prefer if we either break this out or use different logic to to obtain the parent. 

If `findNotes` becomes the general 'query/SQL-like' API, then it's quickly going to get messy and contradictory.  For example, the natural thing someone else may then add is a 'parent' field - now what do you do if someone sets both arguments? 

As soon as we move beyond name + vault, I think it's gonna be a slippery slope down this path.  name + vault makes sense as parameters since we're saying that it's kind of like a composite primary key for a note.

packages/dendron-cli/src/commands/notes.ts
tma66: The main use case looks to be a convenience method. Secondary use cases could be added verification to ensure that fname + vault is unique and if there are multiple matches, `findUnique` can throw an error

packages/common-all/src/types/store.ts
tma66: Originally I was treating `findNotes` like a query method. So if parent and child was both set, it'd return all notes where child = foo and parent = bar (ie bar.baz.foo).

But since the ancestor search is specific to the engine, I can move it to a separate logic that doesn't use find

jonathanyeung: I think this makes sense as long as we treat the arguments of `findNotes()` as query parameters on the props of the particular note.  But since this particular use case does special ancestral logic, I think it should be named accordingly.

PR #3293, opened 2022-07-25
feat(view): adds lock functionality to `PreviewPanel`

namjul:
# Pull Request Checklist


packages/dendron-plugin-views/src/components/DendronApp.tsx
```diff
@@ -178,7 +179,29 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
     }
   });
 
-  return <Component {...props} />;
+  const isLocked = props.ide.isLocked;
+
+  const handleLock = React.useCallback(() => {
```
hkievet, line 184: Why are you using `useCallback` here?

hkievet: I that that you just copied this from the Preview component, but I think just defining a simple non wrapped function (yes, that gets redefined every render) might be more apt.

https://kentcdodds.com/blog/usememo-and-usecallback

TLDR, this function call is neither expensive nor makes an expensive calculation.

namjul: Thank you for that pointer. I had it previously inside `DendronNotePreview` thinking about "Referential equality" but also there I oversaw its uselessness. Makes me reflect on unconsicous assumptions or automatic responses I find myself inhabiting. [`1834f16` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/1834f167e703eb4b02d1da67432a5c5f3f8a8172)

hkievet: I love that you linked the commit you made the change on.  Wow!  I've never seen that before and it's so smart.

packages/common-frontend/src/features/ide/slice.ts
```diff
@@ -31,6 +31,7 @@ type InitialState = {
   showBacklinks?: boolean;
   showOutwardLinks?: boolean;
   showHierarchy?: boolean;
+  isLocked?: boolean;
```
namjul, line 34: We could also convert the state into local-state inside `DendronApp` as long we only need it there. Maybe premature to put it already inside the global state.

packages/plugin-core/src/constants.ts
```diff
@@ -902,6 +902,12 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     },
     when: "dendron:pluginActive",
   },
+  LOCK_PREVIEW: {
+    key: "dendron.lockPreview",
+    title: `${CMD_PREFIX} Lock Preview`,
+    icon: `$(open-preview)`,
+    when: "dendron:pluginActive",
+  },
```
namjul, line 910: FYI: I did not add shortkeys.

kevinslin:
looks like some of the integration tests are breaking because our mock doesn't have the new methods. could you take a look?

packages/common-frontend/src/features/ide/slice.ts
kevinslin: fine for now. we can simplify later if needed

packages/plugin-core/src/constants.ts
kevinslin: thats fine - we've been avoiding them for new commands

packages/plugin-core/src/constants.ts
```diff
@@ -902,6 +902,12 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     },
     when: "dendron:pluginActive",
   },
+  LOCK_PREVIEW: {
+    key: "dendron.lockPreview",
+    title: `${CMD_PREFIX} Lock Preview`,
+    icon: `$(open-preview)`,
```
kevinslin, line 908: can we get a different icon? see https://code.visualstudio.com/api/references/icons-in-labels

packages/plugin-core/src/commands/LockPreview.ts
```diff
@@ -0,0 +1,40 @@
...
+  constructor(previewPanel: PreviewProxy | undefined) {
+    super();
+    this._panel = previewPanel;
+  }
+
+  async sanityCheck() {
+    if (!this._panel || !this._panel.isVisible()) {
+      return "No preview currently open";
+    }
+    if (this._panel && this._panel.isLocked()) {
```
kevinslin, line 27: can we implement this as toggling behavior? rename to toggle preview lock

packages/dendron-plugin-views/src/components/DendronApp.tsx
```diff
@@ -161,14 +166,45 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
...
+      type: NoteViewMessageEnum.onUnlock,
+      data: {},
+      source: DMessageSource.webClient,
+    });
+  };
+
+  return (
+    <>
+      <Component {...props} />
+      {isLocked && (
```
kevinslin, line 198: to make this discoverable, can we have the lock be transparent if not locked so that users can toggle by clicking on it? its okay to split this out as a separate pr

kevinslin (changes requested):
nice job! left some ux comments

packages/plugin-core/src/constants.ts
namjul: [`75aa659` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/75aa6598d30ca0be5dc4efdb3ebef196d764727a)

packages/plugin-core/src/commands/LockPreview.ts
namjul: [`2ad67d2` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/2ad67d2337df532b7348c34e12eeab5344c91c4f)
[`e827c0f` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/e827c0f0d458712e63645f43a3bbbd17b7cf7fa9)

packages/plugin-core/src/components/views/PreviewPanel.ts
```diff
@@ -137,7 +138,7 @@ export class PreviewPanel implements PreviewProxy, vscode.Disposable {
           this._onTextChanged = undefined;
         }
 
-        this._panel = undefined;
+        this.dispose();
```
namjul, line 141: @kevinslin not super confident about this change. Maybe you can check if this does not break existing behavior.

packages/dendron-plugin-views/src/components/DendronApp.tsx
namjul: will catch up on that here https://github.com/dendronhq/dendron/tree/enhance/toggle-preview-lock-from-inside-preview

packages/plugin-core/src/components/views/PreviewPanel.ts
kevinslin: this shouldn't be necessary since this is invoked after `panel.dispose` is called

namjul: ok so you are pointing to `this._panel = undefined;` being called in  `dispose()` meaning we are confident about that change and can leave it like it is.

kevinslin: if by leaving as is you mean reverting this change (not adding `this.dispose`), than yes. `this.dispose` invokes `this._panel.onDidDispose` 
adding @jonathanyeung in case he has more to add here

packages/plugin-core/src/components/views/PreviewViewFactory.ts
```diff
@@ -30,4 +30,8 @@ export class PreviewPanelFactory {
 
     return PreviewPanelFactory._preview;
   }
+
+  static get preview() {
+    return PreviewPanelFactory._preview
```
jonathanyeung, line 35: Let's not add a static getter for `PreviewProxy`, otherwise it'll be hard to manage coupling around the preview panel later (we had this problem in the past).  Instead, in `_extension.ts` you can create a single object and share it between the implementations of the TOGGLE_PREVIEW_LOCK and TOGGLE_PREVIEW commands

jonathanyeung (changes requested):
Looks cool!

packages/plugin-core/src/components/views/PreviewPanel.ts
namjul: @jonathanyeung the intention with this commit was to go after behavoir 1.

[`d13cef8` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/d13cef8231ee2c9e8bc39ce2a597a31ed38b54dc)

@kevinslin after some [delayering/untangling](https://twitter.com/Malcolm_Ocean/status/1363299005443108864) my confusion was about not seing `dispose` and `onDidDispose` being two different entry points. obvious now :)

packages/plugin-core/src/components/views/PreviewPanel.ts
```diff
@@ -150,13 +151,42 @@ export class PreviewPanel implements PreviewProxy, vscode.Disposable {
   hide(): void {
     this.dispose();
   }
+  lock() {
+    const activeTextEditor = VSCodeUtils.getActiveTextEditor();
+    if (activeTextEditor) {
+      this._lockedEditorFileName = activeTextEditor?.document.fileName;
+      this.sendLockMessage(this._panel!, this.isLocked());
+    } else {
+      vscode.window.showInformationMessage("No active texteditor found");
```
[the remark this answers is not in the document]
namjul: Good point, since we do the sanittyCheck in `packages/plugin-core/src/commands/TogglePreviewLock.ts 24` we should not enter this path.

[`7485b1a` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/7485b1a7f3e8176005b058039d8fc5d60da73866)

packages/plugin-core/src/components/views/PreviewViewFactory.ts
namjul: I was uncomfortable with this change, but did not take time to think about it more. Thank you for pointing this out, thats a good way to solve this.

[`b147ea1` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/b147ea11f4742cd0a18334c7bc32913092d405b1)

packages/plugin-core/src/components/views/PreviewPanel.ts
```diff
@@ -346,6 +387,7 @@ export class PreviewPanel implements PreviewProxy, vscode.Disposable {
           data: {
             note,
             syncChangedNote,
+            isLocked: this.isLocked(),
```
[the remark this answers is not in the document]
namjul: [`64ae6c4` (#3293)](https://github.com/dendronhq/dendron/pull/3293/commits/64ae6c48bd50fbe461b76a7e42a534f65c93f5cf)

agreed. this is not necessary now since we go for behavoir 1 (reseting lock state after dispose). If we conclude behavior 2 at some point we will need to sync this up with the webview panel but then could just call `sendLockMessage` in `packages/plugin-core/src/components/views/PreviewPanel.ts 149`

kevinslin (approved):

jonathanyeung (approved):

PR #3313, opened 2022-07-28
fix: force tree view focus only on tutorial workspace activation

hikchoi:
# fix: force tree view focus only on tutorial workspace activation

This PR:
- Force tree view focus only on tutorial workspace activation.
- Fixes issue with initial tree view not revealing correct node due to timing issue.

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/views/NativeTreeView.ts
+++ b/packages/plugin-core/src/views/NativeTreeView.ts
@@ -10,14 +10,7 @@ import {
 import { WorkspaceUtils } from "@dendronhq/engine-server";
 import _ from "lodash";
 import path from "path";
-import {
-  Disposable,
-  TextEditor,
-  TreeView,
-  window,
-  commands,
-  Uri,
-} from "vscode";
+import { Disposable, TextEditor, TreeView, window } from "vscode";
 import { ExtensionProvider } from "../ExtensionProvider";
 import { VSCodeUtils } from "../vsCodeUtils";
 import { EngineNoteProvider } from "./EngineNoteProvider";
@@ -75,17 +68,31 @@ export class NativeTreeView implements Disposable {
     >;
 
     result.then(() => {
-      this.treeView = window.createTreeView(DendronTreeViewKey.TREE_VIEW, {
+      const treeView = window.createTreeView(DendronTreeViewKey.TREE_VIEW, {
         treeDataProvider,
         showCollapseAll: true,
       });
+
+      treeView.onDidChangeVisibility(async (e) => {
+        // when the tree view is first shown ever during the session,
+        // VSCode fires a tree view `.reveal` internally to initially load in the nodes.
+        // when this is happening, if we trigger a `reveal` again, there will be an error.
+        // onDidChangeVisibility doesn't seem to wait for the reveal to finish, so
+        // we have to waitfor the initial load to finish before triggering `reveal`.
+        // It seems like setting a very short timeout is enough to this cooperate with VSCode.
+        setTimeout(() => {
+          if (e.visible) {
+            this.forceReveal({ treeView });
+          }
+        }, 0);
+      });
+
+      this.treeView = treeView;
 
       this._handler = window.onDidChangeActiveTextEditor(
         this.onOpenTextDocument,
         this
       );
-
-      this.forceReveal({ treeView: this.treeView });
     });
   }
 
@@ -118,15 +125,6 @@ export class NativeTreeView implements Disposable {
       );
       if (activeNote !== undefined) {
         treeView.reveal(activeNote);
-
-        // we want to force reveal the tree view but not steal focus
-        // this yields focus back to the active editor
-        const { wsRoot } = ExtensionProvider.getDWorkspace();
-        const activeNotePath = NoteUtils.getFullPath({
-          note: activeNote,
-          wsRoot,
-        });
-        commands.executeCommand("vscode.open", Uri.parse(activeNotePath));
       }
     }
   }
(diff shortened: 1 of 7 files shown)


jonathanyeung (approved):

packages/plugin-core/src/views/NativeTreeView.ts
```diff
@@ -111,6 +118,15 @@ export class NativeTreeView implements Disposable {
       );
       if (activeNote !== undefined) {
         treeView.reveal(activeNote);
```
kevinslin, line 120: can we just reveal and set `focus` to false? see https://code.visualstudio.com/api/references/vscode-api#TreeView%3CT%3E

hikchoi: the options for `.reveal` pertains to the tree _item_ that is being revealed.

This is irrelevant now (I removed it), as we are restricting the forced focus to only once during the tutorial workspace activation. This was a workaround to give back focus to the already open text editor for just-activated existing workspaces

packages/plugin-core/src/views/NativeTreeView.ts
```diff
@@ -68,17 +68,32 @@ export class NativeTreeView implements Disposable {
...
       });
 
+      treeView.onDidChangeVisibility(async (e) => {
+        // when the tree view is first shown ever during the session,
+        // VSCode fires a tree view `.reveal` internally to initially load in the nodes.
+        // when this is happening, if we trigger a `reveal` again, there will be an error.
+        // onDidChangeVisibility doesn't seem to wait for the reveal to finish, so
+        // we have to waitfor the initial load to finish before triggering `reveal`.
+        // It seems like setting a very short timeout is enough to this cooperate with VSCode.
+        // TODO: Remove timeout once this bug is resolved.
```
hikchoi, line 83: A lot of head-scratching on this one 😄 I'll submit a bug report to VSCode for this and come back to fix this once that's resolved.

packages/plugin-core/src/views/NativeTreeView.ts
hikchoi: This does mean that the focus _will_ be on the tree view for folks who first open a tutorial workspace.
If we want to focus to be on the tutorial note, we'll have to add this back in (ideally not here but during workspace activation.)

packages/plugin-core/src/workspace/workspaceActivator.ts
```diff
@@ -581,8 +584,16 @@ export class WorkspaceActivator {
       return new EngineNoteProvider(engine);
     };
     if (!opts?.skipTreeView) {
```
kevinslin, line 586: why can't we do this in `onWorkspaceOpen`? if we need for state to be initialized, would be better to introduce a new hook (eg. `onWorkspaceActivate` then add initializer specific logic into `workspaceActiavtor`

hikchoi: Are you suggesting we move the tree view init step itself to each respective workspace initializers?

hikchoi: I left the tree view init logic in the activator since the only specific logic here would be the special treatment with the tutorial workspace after it was initialized.

packages/plugin-core/src/workspace/workspaceInitializer.ts
```diff
@@ -12,6 +13,11 @@ export type OnWorkspaceCreationOpts = {
  * Type that can execute custom code as part of workspace creation and opening of a workspace.
  */
 export type WorkspaceInitializer = {
+  /**
+   * name of the workspace initializer
+   */
+  name: string;
```
kevinslin, line 19: can we remove the name property now since we're not using it?

packages/plugin-core/src/views/NativeTreeView.ts
jonathanyeung: But is it possible that the intended behavior is for visibility to be set to true _then_ data loading happens?

hikchoi: that makes sense, but there is no safeguard here for the extension authors that want to trigger `.reveal` on the tree view. nor can I find any reference of a way to wait for that internal call before we go ahead and trigger it ourselves.

hikchoi: If we were to handle only what we are explicitly triggering, we would have a queue that lets us wait for the previous call to resolve (recommended here by the team: https://github.com/microsoft/vscode/issues/142328#issuecomment-1046907141). but the initial call that VSCode makes to populate the tree view on the first `visible: true` we have no info on the status as-is.

kevinslin (approved):

packages/plugin-core/src/views/NativeTreeView.ts
jonathanyeung: Agree with your points - main point I was trying to make is that I'm not sure what kind of 'bug' per se to file against VS Code.  Perhaps an API request to signal when initial rendering is completed? But I'm not sure if that's the design they want.  It's worth asking at the very least

jonathanyeung (approved):

packages/plugin-core/src/views/NativeTreeView.ts
hikchoi: Agreed. that's the plan.

PR #3318, opened 2022-07-28
feat(view): lock tutorial preview

namjul:
docs:

- - [[Tutorial Preview Should Support Locking|dendron://private/task.2022.07.20.tutorial-preview-should-support-locking]]

### DIFF
--- a/packages/common-all/src/types/typesv2.ts
+++ b/packages/common-all/src/types/typesv2.ts
@@ -870,6 +870,10 @@ export type OnDidChangeActiveTextEditorData = {
    * If activeNote is defined, view will set that note as active note. Otherwise default to {@param note}
    */
   activeNote?: NoteProps;
+  /**
+   * if preview is locked
+   */
+  isLocked?: boolean;
 };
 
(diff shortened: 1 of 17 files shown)


namjul:
@jonathanyeung @kevinslin  I am not really able to test this because when launching the tutorial it opens it with the installed dendron extension instead of the local one.

kevinslin:
> @jonathanyeung @kevinslin I am not really able to test this because when launching the tutorial it opens it with the installed dendron extension instead of the local one.

you can use `LaunchTutorial` command to test the preview. note that when you launch the tutorial, you'll need to restart the debug environment

jonathanyeung:
Hmm - I don't think you need to reattach the bugger or anything like that.  It should be able to reattach automatically when a workspace reloads.  So just try the following:

1. Run the `Run: Extension Local` debug task
2. In the Debug instance of VS Code, run the command 'Launch Tutorial', click the green button
3. Workspace will reload, and you should be able to hit your code path.

namjul:
closing since this is being solve in https://github.com/dendronhq/dendron/pull/3352

PR #3321, opened 2022-07-28
chore(workspace): Fix engine sync issue from manual rename

tma66:
**Background**
While working on the engine `rename` refactoring, I've realized that manually renaming a file outside of dendron can easily cause the engine state to drift. This is due to not properly cleaning up the note metadata of the old file

**Changes**
1. Update `renameNote` param to use `metaOnly` instead of `isEventSourceEngine`. Therefore if the change is from vscode, we will delete/write to only the note metadata store
2. Consolidate link/anchor refresh logic into one method under `EngineUtils`
3. Deprecate `getLinks` and `getAnchors` on engine. In the future, we will not need the engine to perform these actions
4. Remove `onDidRenameFiles` from watcher. Since `renameNote` handles writing the note metadata, there is no need to write the same thing again.

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -37,7 +37,6 @@ import {
   NotePropsMeta,
 } from "./types";
 import {
-  ConfigUtils,
   DefaultMap,
   getSlugger,
   isNotUndefined,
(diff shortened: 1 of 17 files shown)


kevinslin (approved):
nice job!

jonathanyeung (approved):

Harshita-mindfire:
Works great, found an inconsistency in note's title in fname:

tma66:
> Works great, found an inconsistency in note's title in frontmatter: 

Great find! Turns out I needed to add back in the `onDidRenameFiles` logic.

PR #3334, opened 2022-08-02
chore(workspace): Migrate rename notes cli command to new engine

tma66:
**Background**
This PR moves the last and final notes cli command (`rename`) to the new engine. After this, I will focus on feature parity between the old and new engine.

**Changes**
1. `RENAME` notes command will use new engine
2. Expanded NoteStore finder to accept `linkedToNote` property. This will find all notes whose `note.links` contains the passed in noteprop
3. Migrated most of `rename` logic from storev2 to the new engine. Left business logic mostly intact
4. Deprecated several methods on engine side (`updateNote`, `getLinks`, `getAnchors`

**Todo**
Backlinks need to be properly updated as part of `writeNote`. Current engine doesn't do this properly so this will be added after feature parity work is done


packages/common-all/src/types/store.ts
```diff
@@ -14,6 +14,9 @@ export type FindNoteOpts = {
   fname?: string;
   // If vault is provided, filter results so that only notes with matching vault is returned
   vault?: DVault;
+  // Find notes with links referencing to passed in note props value.
+  // E.g. if linkedToNote = "{ fname: foo.123, ... }", return all notes pointing to this note
+  linkedToNote?: NotePropsMeta;
```
jonathanyeung, line 19: I guess a similar discussion to last time, but I'm not sure how I feel about these params - it doesn't seem to be too compatible with the fname/vault parameters above. What happens if someone sets both fname and linkedToNote? 

Should this be called backlinksToNote? I think there's also a typo in the description, "referencing _to_ passed in"?

jonathanyeung (changes requested):

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -647,6 +673,173 @@ export class DendronEngineV3 implements DEngine {
     };
   }
 
+  /**
+   * See {@link DEngine.renameNote}
+   */
+  async renameNote(opts: RenameNoteOpts): Promise<RespV2<RenameNotePayload>> {
+    const ctx = "DEngine:renameNote";
```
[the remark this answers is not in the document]
tma66: Talked to Kevin briefly on this before and we'd like this to be atomic as well. Since the current behavior is not atomic, I won't modify it yet, but I'll leave a todo for us to pick up after migration is complete

packages/common-all/src/types/store.ts
tma66: Long term for this method, a note has to fulfill all requirements to be returned (like a sql statement where the conditions are AND together).

On brighter news, I just realized that this is the same thing as a note's backlinks (`NoteUtils.getNotesWithLinkTo` can also be replaced eventually). Currently it won't work b/c backlinks is broken, but for the new engine, I'll use the backlinks prop so we won't need `linkedToNote`

jonathanyeung (approved):

PR #3340, opened 2022-08-03
enhance(views): improve vscode theme integration

namjul:
This PR aims to improve the calendar view by integrating with vscode theme.
Here we make use of vscode theme editor background token and set calendar-view's background to it.

task: [[Better Compatibility with Vscode Theme|dendron://private/task.2022.07.28.better-compatibility-with-vscode-theme]]

### DIFF
--- a/packages/common-assets/styles/scss/main.scss
+++ b/packages/common-assets/styles/scss/main.scss
@@ -20,6 +20,10 @@ $text-color: #111 !default;
 // Portal end
 @import "portal.scss";
 
+.ant-picker-calendar, .ant-picker-calendar .ant-picker-panel {
+  background-color: var(--vscode-editor-background) !important,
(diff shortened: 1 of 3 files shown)


packages/common-assets/styles/scss/main.scss
```diff
@@ -20,6 +20,10 @@ $text-color: #111 !default;
 // Portal end
 @import "portal.scss";
 
+.ant-picker-calendar, .ant-picker-calendar .ant-picker-panel {
+  background-color: var(--vscode-editor-background) !important,
+}
```
namjul, line 25: There is not really a nice way to adjust antd's calendar styles dynamically. So we need to circle around this and apply it this way.

kevinslin (approved):

PR #3343, opened 2022-08-04
feat(workspace): code workspaces support

jonathanyeung:
## feat(workspace): code workspaces support

This change provides an MVP for having a Lookup-supported Dendron as a web extension in virtual filesystems (like vscode.dev, github.dev)

@Harshita-mindfire helped author `VSCodeFileStore.ts`

### MVP Feature Set
- Only lookup is supported. All other commands and views have been disabled
- Lookup is Limited in functionality:
    - No modifier support
    - No schemas - including icon hints and application of templates
    - No Tab Autocomplete
    - Does not observe additional user settings around lookup (such as fuzzy level)
- Lookup supports the following:
    - Querying of Notes
    - Creation of Notes
    - Default of 'smart' vault selection setting.

#### Webpack / Browser Limitations

1. The Web Extension runs off a different entry point of `extension.ts` from the regular Local extension. All code for the web extension is located under `plugin-core/src/web` and there's a separate webpack configuration to bundle a separate .js file that gets executed for the Web Extension version of Dendron.
1. This means that all code under `plugin-core/src/web` have the following restrictions: they can only import from `common-all` or browser supported Node packages, notably excluding `fs`/`fs-extra`.
1. In order to support virtual workspaces, WSRoot and DVault.fsPath need to be converted from simple strings to URI's.

#### Engine

1. A simplified engine was made, `DendronEngineV3Web` which implements a temporary interface `IReducedEngineApiService`. This is a subset of `DEngine`. A follow up is to consolidate the interfaces into common-all and reuse fields.
1. The engine currently runs in-memory in the same process as `Extension Host`. A later optimization is to run it in a web worker for better performance. So far, testing against test-workspace seems snappy enough for a good user experience. 

#### TSyringe / Initialization Sequence 

1. The initialization sequence has been simplified (albeit, only one command is supported). I've adopted the DI framework TSyringe to help with object instantiation. A TSyringe `container` is first configured for the web version of the extension, and the container is then used to instantiate the command(s). 
1. Instead of passing around a monolithic `IDendronExtension`/`DendronWorkspace` objects with static getters, only the configuration and components required by a class are injected into the constructor. This will help us reduce coupling and improve testability.
1. Currently, the two things extracted from `dendron.yml` are wsRoot and the vault configuration.

#### Lookup Logic

Lookup Logic has been simplified, but the code structure is still WIP -  particularly, `LookupQuickpickFactory` still needs simplication. 
1. `ILookupProvider` has been simplified - it now just returns queried items and has no access to the quickpick.
1. The required logic for the client invoking lookup has also been simplified - a client just needs to call `showLookup()`, which returns a promise on what the user selected. There's no more need for history-service queries. This is modeled after how `vscode.window.showQuickPick()` works.

#### Consolidation to common-all; NoteParserV2

1. I've begun to consolidate some key engine classes into common-all so that they can be shared between local and web versions: `NoteStore`, `NoteMetadataStore` have been moved; a consolidation between `DendronEngineV3` and `DendronEngineV3Web` is next (out of scope of this PR).
1. NoteParserV2 had to be forked unfortunately to get this to work.  We can consider refactoring it later, but the way it is tied to `fs` makes it a bit difficult. Furthermore,  `dendron.cache` is not currently supported in the web version so that logic had to be removed.

### Notes for Reviewers

- There's _a lot_ of duplicated code to maneuver around imports to get this POC off the ground. We can (and should) dedupe later.
- I intentionally left some commented code there so that it's easier to see what functionality is missing that we need to add later to get things to parity between web vs local.
- There's test debt with the code under `web/` - it's a little bit trickier to test this code since test code cannot access `fs` either when you're testing against a web target. This makes a lot of our test-utils unavailable. Furthermore, we have to use the web version of mocha, which was giving me some difficulties that I haven't had time to diagnose yet.


packages/common-server/src/filesv2.ts
```diff
@@ -27,7 +27,11 @@ import path from "path";
 import tmp, { DirResult, dirSync } from "tmp";
 import { resolvePath } from "./files";
 import { SchemaParserV2 } from "./parser";
-import textextensionslist from "textextensions";
+
+// TODO: There's some issue with this extension when in browser mode, so for
```
jonathanyeung, line 31: Disregard changes to this file - this will be reverted.

packages/common-all/src/drivers/string2Note.ts
```diff
@@ -0,0 +1,50 @@
...
+import { genHash } from "../utils";
+
+/**
+ * NOTE: Temporarily duplicated from common-server/filesv2.ts to get Dendron Web
+ * Extension working
+ * @param calculateHash - when set, add `contentHash` property to the note
+ *  Default: false
+ * @returns
+ */
+export function string2Note({
```
kevinslin, line 15: can we add a comment in this function in `common-server` to port over any changes we make to that function?

packages/common-all/src/store/NoteStore.ts
```diff
@@ -1,58 +1,48 @@
...
+  ) {
+    this._fileStore = fileStore;
+    this._metadataStore = dataStore;
+    this._wsRoot = wsRoot;
   }
 
-  dispose() {
-    this._loggerDispose();
-  }
+  dispose() {}
```
kevinslin, line 40: why was disposable removed?

packages/common-all/src/store/NoteStore.ts
```diff
@@ -64,7 +54,11 @@ export class NoteStore implements Disposable, INoteStore<string> {
         data: { ...metadata.data, body: "" },
       };
     }
-    const uri = NoteUtils.getURI({ note: metadata.data, wsRoot: this._wsRoot });
+    const uri = Utils.joinPath(
+      this._wsRoot,
+      VaultUtils.getRelPath(metadata.data.vault),
```
kevinslin, line 59: `getURI` would check if the vault path is an absolute path vs a relative path. looks like this doesn't do that? also, should we extract this into its own function

packages/common-all/src/types/typesv2.ts
```diff
@@ -163,6 +163,16 @@ export type NoteQuickInput = NoteProps & {
   alwaysShow?: boolean;
 };
 
+/**
+ * A reduced version of NoteQuickInput that only keeps the props necessary for
+ * lookup quick pick items
+ */
+export type NoteQuickInputV2 = Pick<NoteProps, "fname" | "vault" | "schema"> & {
```
kevinslin, line 170: can we label these types as `WebExt` version so that its easy to pick out what is for web extension and what is not?

packages/common-all/src/types/workspace.ts
```diff
@@ -26,6 +26,10 @@ export enum DVaultSync {
   SYNC = "sync",
 }
 
+export type DVaultUriVariant = DVault & {
```
kevinslin, line 29: will this be the default moving forward?

kevinslin:
nice work. some high level comments.

### 1. Folder structure
- currently, the code structure is `src/web/{web suff}` and `src/(!web)*` non-web stuff.  as we migrate the original code base, can we adopt a similar organization as the vscode code base where similar files are co-located?
        - eg. `src/commands/(web|common|node)`
- this also goes for tests - as we continue, can we merge the web and nodejs tests as well as co-locate them via the above split?

### 2. Tsringe
- is the plan to adopt dependency injection for plugin-core? if so, could you add a note in the dev docs about dependency injection and how to get started as well as debug code under tsringe? this can be after we've shipped this but just calling it out for now

packages/common-all/src/drivers/string2Note.ts
jonathanyeung: I just deleted the common-server one and updated imports.  Code between this and common-server one were identical.

packages/common-all/src/store/NoteStore.ts
jonathanyeung: Current logger implementation in common-server requires fs, so the logger had to be removed from `NoteStore` for now to put it in common-all.  It can be added back later once we have an fs-agnostic logging interface.

packages/common-all/src/store/NoteStore.ts
jonathanyeung: In what situations do vaults have absolute paths specified versus relative paths?

packages/common-all/src/types/typesv2.ts
jonathanyeung: Hm - for this even tho this is the version for web extension now, I think it should eventually be the version for everything though.

packages/common-all/src/types/workspace.ts
jonathanyeung: I think this can probably be removed actually.  Everything should be able to be formulated off a URI variant of wsRoot + the relative path on a `DVault`.  

Unless sometimes `DVaults` have absolute paths? In which case it'll be problematic for the web extension.

jonathanyeung:
> 1. Folder structure
> * currently, the code structure is `src/web/{web suff}` and `src/(!web)*` non-web stuff.  as we migrate the original code base, can we adopt a similar organization as the vscode code base where similar files are co-located?
>   - eg. `src/commands/(web|common|node)`
> * this also goes for tests - as we continue, can we merge the web and nodejs tests as well as co-locate them via the above split?

I'd prefer to keep it as `src/(web|common|node)/*` rather than `src/commands/(web|common|node)`. I think this will make it easier to keep track of which code files allow what kind of imports and to better understand what exactly is going to get bundled into the web extension version vs the local extension version.  Furthermore, eventually 95% of all the code files should be under `common`, with very few files actually in `web` or `node`.  For tests though, I think it may make more sense to consolidate all tests to be under `src/test`.  

> 2. TSyringe

Yes, that's the plan.  I'll work on adding dev docs this week on this + the info in this PR description.

packages/common-all/src/types/typesv2.ts
jonathanyeung: Offline follow up: let's do a validation on dendron.yml on startup to make sure they're relative.

jonathanyeung:
> some ux pointers:
> 
> * remove toggle preview context button from the ui
> * remove dendron tree view container views until we start supporting them
> * remove dendron commands that are not supported from the web extension

@kevinslin - I figured out what's going on here.  I've made an adjustment and now this _should_ be fine - the reason these are showing up is b/c of the debugging environment - it's running the web ext version of Dendron but in a regular Desktop VS Code extension host.  So the environment variables like 'shellExecutionSupported` will still report true, even though for real end users running this in the browser this will be false (and no real end user will be running web Dendron in local VS Code, they'll just be running local Dendron). It should just be an artifact of the debugging environment.

Environment variables specified here: https://code.visualstudio.com/api/extension-guides/web-extensions#migrate-extension-with-code

packages/plugin-core/src/web/commands/NoteLookupCmd.ts
```diff
@@ -0,0 +1,78 @@
...
+          //         "Select which template to apply or press [ESC] to not apply a template",
+          //       nonStubOnly: true,
+          //     });
+          //   },
+          // });
+          // note = _.merge(newNote, overrides || {});
+          const res = await this.engine.writeNote(newNote);
+
+          if (res.error) {
+            vscode.window.showErrorMessage("Failed to write note to engine!");
```
tma66, line 62: Should we show the error message as well?

packages/plugin-core/src/web/commands/lookup/LookupQuickpickFactory.ts
```diff
@@ -0,0 +1,269 @@
...
+    const qp = this.create({
+      title: "Lookup Note",
+      buttons: [],
+      provider: opts.provider,
+      initialValue,
+    });
+
+    const outerPromise = new Promise<LookupAcceptPayload | undefined>(
+      (outerResolve) => {
+        const foo = new Promise<LookupAcceptPayload | undefined>((resolve) => {
```
tma66, line 60: I'd use a more meaningful variable name than foo lol

packages/plugin-core/src/web/engine/IReducedEngineApiService.ts
```diff
@@ -0,0 +1,49 @@
...
+  RenameNotePayload,
+  RespV2,
+} from "@dendronhq/common-all";
+
+// Subset of IEngineAPIService
+export interface IReducedEngineAPIService {
+  /**
+   * Get NoteProps by id. If note doesn't exist, return undefined
+   */
+  getNote: (id: string) => Promise<NoteProps | undefined>;
```
tma66, line 22: Just heads up that I changed the signature from https://github.com/dendronhq/dendron/pull/3359. I feel it'll be useful to propagate any errors to the clients

packages/plugin-core/src/web/commands/lookup/LookupQuickpickFactory.ts
jonathanyeung: good catch haha - there's a lot of cleanup I was planning on doing to this file.  I want to do it as part of implementing a Schema lookup controller so that I can better understand what needs to be abstracted.

PR #3345, opened 2022-08-04
enhance(sync): google docs export pod prompts for destination folder

Harshita-mindfire:
This PR aims to 
- render note refs better, v1 version: embed them without HTML 
- add wiki link to original note to front of the gdoc
- skips converting wikilinks, hashtags, usertags, backlinks
-  prompts the user to select the folder where doc should be exported

***
Docs PR: https://github.com/dendronhq/dendron-site/pull/606
Loom video: https://www.loom.com/share/75d1f0396cc041519c7060931b6dd080

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

## Docs


packages/pods-core/src/v2/podConfig/GoogleDocsConfig.ts
```diff
@@ -71,7 +73,7 @@ export function createRunnableGoogleDocsV2PodConfigSchema(): JSONSchemaType<Runn
       exportScope: {
         type: "string",
       },
-      description: {
```
jonathanyeung, line 74: nit: can we leave description? We haven't been using this, but it let's someone add some text that will show up in the quick pick: 

image

packages/plugin-core/src/commands/pods/GoogleDocsExportPodCommand.ts
```diff
@@ -170,6 +202,77 @@ export class GoogleDocsExportPodCommand extends BaseExportPodCommand<
...
+    );
+    return folderIdsHashMap;
+  }
+
+  /**
+   * prompts to select the folder docs are exported to
+   * @param folderIdsHashMap
+   */
+  async promtForParentFolderId(folderIdsHashMap: string[]) {
+    const pickItems = folderIdsHashMap.map((folder) => {
```
jonathanyeung, line 259: can we skip this prompt if there are <= 1 folders?

packages/plugin-core/src/commands/pods/GoogleDocsExportPodCommand.ts
```diff
@@ -170,6 +202,77 @@ export class GoogleDocsExportPodCommand extends BaseExportPodCommand<
     }
   }
 
+  async candidateForParentFolders(accessToken: string) {
+    try {
+      const result = await this.getAllFoldersInDrive(accessToken);
+      return result;
+    } catch (err: any) {
+      throw new DendronError({ message: stringifyError(err) });
```
jonathanyeung, line 210: can we still default to their root folder if something goes wrong?

packages/pods-core/src/v2/podConfig/GoogleDocsConfig.ts
```diff
@@ -71,7 +73,7 @@ export function createRunnableGoogleDocsV2PodConfigSchema(): JSONSchemaType<Runn
...
+      parentFolderId: {
```
jonathanyeung, line 76: maybe `parentFolder` instead of `parentFolderId`?  The value here is human readable right? Like "notes", not some GUID?

jonathanyeung (changes requested):

packages/pods-core/src/v2/podConfig/GoogleDocsConfig.ts
Harshita-mindfire: `description` is still there in every export pod config, it is just removed from `Runnable` config of google docs export(not a required property to run gdoc export)

packages/pods-core/src/v2/podConfig/GoogleDocsConfig.ts
Harshita-mindfire: The value is GUID. The only readable value is `root` which means at the root of drive.

jonathanyeung: Ah ok - in that case let's leave it as `parentFolderId`

packages/pods-core/src/v2/podConfig/GoogleDocsConfig.ts
jonathanyeung: Ah, makes sense. Misread the code

jonathanyeung (approved):

packages/pods-core/src/v2/pods/export/GoogleDocsExportPodV2.ts
```diff
@@ -178,7 +180,7 @@ export class GoogleDocsExportPodV2
           wsRoot: this._wsRoot,
         });
         // wrap data in html tags
-        data = `<html>${data}</html>`;
+        data = `<html><div>[[${input.fname}]]</div>${data}</html>`;
```
kevinslin, line 183: can we update the text to be the following (in block quotes)?

> This note was exported from `[[{FNAME}]]` with [Dendron](https://dendron.so/)

packages/plugin-core/src/commands/pods/GoogleDocsExportPodCommand.ts
```diff
@@ -118,6 +125,31 @@ export class GoogleDocsExportPodCommand extends BaseExportPodCommand<
       return;
     }
 
+    let parentFolderId = opts?.parentFolderId;
```
kevinslin, line 128: can we store this in the note as well? so if the gdoc was already exported to a folder, we continue exporting it to the same folder

packages/engine-server/src/markdown/utilsv5.ts
```diff
@@ -295,7 +302,14 @@ export class MDUtilsV5 {
 
           // NOTE: order matters. this needs to appear before `dendronPub`
           if (data.dest === DendronASTDest.HTML) {
-            proc = proc.use(hierarchies).use(backlinks);
+            proc = proc.use(hierarchies);
```
kevinslin, line 305: we should also skip children links in this case

kevinslin (changes requested):
nice job. just a few nitpicks

packages/plugin-core/src/commands/pods/GoogleDocsExportPodCommand.ts
Harshita-mindfire: when we export a note again, it always [overwrites](https://github.com/dendronhq/dendron/blob/756395f8a50ab8f38b405a9f61a98dc968e9a2f3/packages/pods-core/src/v2/pods/export/GoogleDocsExportPodV2.ts#L276) the existing document.

kevinslin: when i say store in the note, i meant the frontmatter of the note in dendorn. like we do with `uri`

kevinslin (approved):

PR #3346, opened 2022-08-04
fix(publish): restores inconsistent layout on mobile

namjul:
## Background

This has been adressed in https://github.com/dendronhq/dendron/pull/3272.  It came up again https://github.com/dendronhq/dendron/issues/2175#issuecomment-1204441903.

[[Inconsistent Layout of Mobile View|dendron://private/task.2022.07.17.inconsistent-layout-of-mobile-view]]

## Solution

- [`dc4f42c` (#3346)](https://github.com/dendronhq/dendron/pull/3346/commits/dc4f42cd3c75a7d4484e1e28c18737036f6f2015) makes sure the layout does not can expand beyond its intentional space. Any element that expands beyond it should now become salient. e.g. [here](https://user-images.githubusercontent.com/328836/182880334-28300ee9-d740-4c5f-a533-2c8d009f6e22.png) it becomes clear that the table is causing a wider expanding whereby the content stays.
  -  a solution would be to not let tables [overflow](https://getbootstrap.com/docs/5.2/content/tables/#responsive-tables)
  - @kevinslin can you point to a good place to wrap tables, I am guess somewhere in our markdown parsing logic.


kevinslin:
> 

Before proceeding, I would recommend watching the walkthrough I did with hunter about how we process markdown here > [[Markdown|dendron://dendron.docs/topic.markdown]]

Since the md processing is rather cumbersome, I've did an initial implementation here: https://github.com/dendronhq/dendron/tree/eg/table-modification
An example of adding a new classname is in the PR [here](https://github.com/dendronhq/dendron/blob/eg/table-modification/packages/engine-server/src/markdown/utilsv5.ts#L659:L659)
Docs for further edits here: https://github.com/syntax-tree/mdast-util-to-hast/tree/8.0.0 (NOTE: we're using version 8 of the library, not the latest)

As an aside, I can see how this would fix it for tables but do you know why https://wiki.dendron.so/notes/9MZBqhrijEM4QpZRa5t08/ doesn't show up right on mobile?

namjul:
> As an aside, I can see how this would fix it for tables but do you know why [wiki.dendron.so/notes/9MZBqhrijEM4QpZRa5t08](https://wiki.dendron.so/notes/9MZBqhrijEM4QpZRa5t08/) doesn't show up right on mobile?

So far I detected 2 ways the layout can get broken.

- `<table>` that are wider than the viewport
  - this can be solved by the above mentioned solution
- `<code>` that contain words with a length wider than the viewport.
  - we add a css class that targets all `<code>` tags and apply `overflow-wrap: anywhere`.

Sidenote: codeblocks don't break the layout because they are wrapped with `<pre>` and get the `overflow: auto` applied meaning they will render with a horzontal scrollbar.

namjul:
This  PR still needs to incorporate the insights from https://discord.com/channels/717965437182410783/802578902429597726/1012742270161801307

kevinslin:
nice job! can we add some playwright tests based off of https://github.com/dendronhq/nextjs-template/blob/main/tests/example.spec.ts#L3:L3 for the mobile changes?

namjul:
> nice job! can we add some playwright tests based off of [https://github.com/dendronhq/nextjs-template/blob/main/tests/example.spec.ts#L3:L3](https://github.com/dendronhq/nextjs-template/blob/main/tests/example.spec.ts?rgh-link-date=2022-08-29T00%3A14%3A48Z#L3:L3) for the mobile changes?

What kind of test do you have in mind? Do you refer with "mobile changes" just the code from this PR?

kevinslin:
> > nice job! can we add some playwright tests based off of [https://github.com/dendronhq/nextjs-template/blob/main/tests/example.spec.ts#L3:L3](https://github.com/dendronhq/nextjs-template/blob/main/tests/example.spec.ts?rgh-link-date=2022-08-29T00%3A14%3A48Z#L3:L3) for the mobile changes?
> 
> What kind of test do you have in mind? Do you refer with "mobile changes" just the code from this PR?

yep. don't worry about existing items but lets start adding tests for new changes so we avoid going one step forward and two steps back

kevinslin:
looks like we have a bunch of test failures here?

kevinslin (approved):

Issue #906, opened 2021-07-01
Merge Notes command


SR--:
## Please Select if your Request is Either something new or an Enhancement ##

- [ ] Enhancement of an existing Feature.
- [x] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible) ##

- [ ] Onboard. Initial vault setup and import/export pods 
- [x] Create. Note creation, lookup, snippets and templates
- [ ] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Is your feature request related to a problem? Please describe ##

I have instances where I want to merge notes. Perhaps one was a quick temporary effort and I want to combine it with a better developed one. Maybe I decided two notions belong together. If I paste the content into one of the two, and delete one of the notes, the links to it will be broken.

## Describe the solution you'd like ##

Implement a command Merge Notes that would:

1. take note 1 and paste its content at the end of note 2
2. delete note 1
3. refactor links to point to note 2

## Describe alternatives you've considered ##

One can do this via search/replace but this is slower

teauxfu:
This could be useful, but I think the next-best workaround would be to use note [references](https://wiki.dendron.so/notes/f1af56bb-db27-47ae-8406-61a98de6c78c.html)
Keeping links healthy should be a bit easier as the notes are more atomic this way, I think

SR--:
Sure, possible in many cases. I am just looking at many notes taken in different situations that are all basically the same or are split opportunistically rather than logically. I want to merge them and then split them again as needed. Splitting is easy, merging is messy.

kpathakota:
@SR-- interesting request, haven't heard of this before but I think this would make interesting new functionality.

SR--:
Thanks for considering. Not a high-priority issues perhaps but is ideologically compatible with the dendron manifesto that promises to make organisational structures fungible.

aleksey-rowan:
I would love to see this as well!

Another use case would be when you need to export a series of notes as a single file. Say you are writing a book and each chapter is a separate note and when you are done, you want to import it into some other tool to format and make a `.pdf` or an `.epub` out of it. It would be easier if you could merge the notes and export/copy-paste just a single file.

[...]

PR #3349, opened 2022-08-05
feat(refactor): merge note command

hikchoi:
# feat(refactor): merge note command
This PR:
- adds a new command `Dendron: Merge Note`
- enhances Lookup provider so that it can take a list of `preAcceptValidators` when it is created by the lookup provider factory.
  - this will allow lookup-derivative commands (such as this one) to pass in list of a predicate functions that would be executed when the user hits enter in the quickpick.
  - if any of the `preAcceptValidator` fails, lookup will not be accepted until the user selects a valid item from the quickpick
  - In merge note command, this is used to prevent users from merging the active note to itself.

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/MergeNoteCommand.ts
+++ b/packages/plugin-core/src/commands/MergeNoteCommand.ts
@@ -1,5 +1,5 @@
 import {
-  asyncLoopOneAtATime,
+  asyncLoop,
   extractNoteChangeEntryCounts,
   NoteChangeEntry,
   NoteProps,
@@ -24,6 +24,7 @@ import { BasicCommand, SanityCheckResults } from "./base";
 import * as vscode from "vscode";
 import _ from "lodash";
 import { ProxyMetricUtils } from "../utils/ProxyMetricUtils";
+import { VSCodeUtils } from "../vsCodeUtils";
 
 type CommandInput = {
   source?: string;
@@ -203,7 +204,7 @@ export class MergeNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
 
     // append to end
     const destBody = destNote.body;
-    const newBody = `${destBody}\n\n${appendPayload}`;
+    const newBody = `${destBody}\n---\n\n# ${sourceNote.title}\n\n${appendPayload}`;
     destNote.body = newBody;
     const writeResp = await this.extension.getEngine().writeNote(destNote);
     if (!writeResp.error) {
@@ -233,7 +234,6 @@ export class MergeNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
       const linksToUpdate = noteToUpdate.links.filter(
         (link) => link.value === sourceNote.fname
       );
-      const noteToUpdateDeepCopy = _.cloneDeep(noteToUpdate);
       const modifiedNote = await _.reduce<
         typeof linksToUpdate[0],
         Promise<NoteProps>
@@ -271,7 +271,7 @@ export class MergeNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
           acc.body = newBody;
           return acc;
         },
-        Promise.resolve(noteToUpdateDeepCopy)
+        Promise.resolve(noteToUpdate)
       );
 
       noteToUpdate.body = modifiedNote.body;
@@ -327,7 +327,7 @@ export class MergeNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
     // find all links that need to be updated from end to front.
     // then update them.
     let noteChangeEntries: NoteChangeEntry[] = [];
-    await asyncLoopOneAtATime(noteIDsToUpdate, async (id) => {
+    await asyncLoop(noteIDsToUpdate, async (id) => {
       try {
         const changed = await this.updateLinkInNote({
           sourceNote,
@@ -399,6 +399,9 @@ export class MergeNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
       ...updateBacklinksChanges,
       ...deleteSourceChanges,
     ];
+
+    // close the source note
+    await VSCodeUtils.closeCurrentFileEditor();
 
     // open the destination note
     await this.extension.wsUtils.openNote(destNote);
--- a/packages/plugin-core/src/commands/MergeNoteCommand.ts
+++ b/packages/plugin-core/src/commands/MergeNoteCommand.ts
@@ -86,6 +86,177 @@ export class MergeNoteCommand extends BasicCommand<CommandOpts, CommandOutput> {
     });
   }
 
+  /**
+   * Given a source note and destination note,
+   * append the entire body of source note to the destination note.
+   * @param source Source note
+   * @param dest Dest note
+   */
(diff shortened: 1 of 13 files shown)


hikchoi:
pending:
- [ ] clean up command logic
- [ ] add proxy metrics
- [ ] add tests
- [ ] confirm engine state is sound after the command
- [ ] add documents

hikchoi:
related to: https://github.com/dendronhq/dendron/issues/906

hikchoi:
new proxy metrics props added to airtable ✅

hikchoi:
Will create async on the new `preAcceptValidator` added to lookup provider when this PR is good to go.

hikchoi:
doc updates at https://github.com/dendronhq/dendron-site/pull/609

jonathanyeung (approved):
Nice! 

Maybe as a later enhancement - should we run this command automatically when a user does 'Move Note' and selects a destination that already exists? (Maybe first prompting them w/ "Would you like to merge these two notes?")

hikchoi:
> Nice!
> 
> Maybe as a later enhancement - should we run this command automatically when a user does 'Move Note' and selects a destination that already exists? (Maybe first prompting them w/ "Would you like to merge these two notes?")

Good question. I did think about the UX / implementation of this as Move note and Merge note are pretty similar ideas on a surface level.

I think first we need to be careful about intent. When users try to use move note, and select an existing note for the destination, do they really want to do that? what are they expecting the end result to be? Or stepping back once more, does this actually happen frequently, and do we know if that action is intentional (they want / expect it to work) or is it a mistake?

Slightly tangential as I know you aren't suggesting we lump the implementation together with Move note, but I do want to break down the move note command and simplify it so that it does one thing. Currently it's taking care of rename note as a special case of move note, and as a result there are extra moving parts in it that could potentially break and make everything that depend on it not work if that happens.

So I prefer move note's capability to be simple and have it not serve as an actual entry / trigger point for another note-level refactoring capability. Though I do see the benefits of what you're suggesting.

My ideal way of doing this would be disallow and notify things that shouldn't happen in one command, and think about a way to give the users sufficient and clear signifiers that _if_ their intention was different, that some other command can do it for them. So for example, instead of how move note just short-circuits when they pick an existing note in lookup, enhance it so that it uses a `preAcceptValidator` that disallows it in the first place, and show a toast that serve as a signifier that _maybe you want to do this instead?_, but not chain it there directly so that it becomes a trigger point for the Merge Note command.

packages/plugin-core/src/commands/MergeNoteCommand.ts
```diff
@@ -0,0 +1,453 @@
...
+          .map((backlink) => backlink.from.id)
+          .filter((ent): ent is string => ent !== undefined)
+      )
+    );
+
+    // for each note that needs to be updated,
+    // find all links that need to be updated from end to front.
+    // then update them.
+    let noteChangeEntries: NoteChangeEntry[] = [];
+    await asyncLoopOneAtATime(noteIDsToUpdate, async (id) => {
```
kevinslin, line 330: does this need to be done sequentially vs in parallel?

packages/plugin-core/src/commands/MergeNoteCommand.ts
```diff
@@ -0,0 +1,453 @@
...
+    destNote: NoteProps;
+  }) {
+    const ctx = `${this.key}:updateLinkInNote`;
+    const { id, sourceNote, destNote } = opts;
+    const noteToUpdate = await this.extension.getEngine().getNote(id);
+    if (noteToUpdate !== undefined) {
+      const linksToUpdate = noteToUpdate.links.filter(
+        (link) => link.value === sourceNote.fname
+      );
+      const noteToUpdateDeepCopy = _.cloneDeep(noteToUpdate);
```
kevinslin, line 236: in this case since we are intentionally mutating the note body, do we need to clone?

kevinslin: also `getNote` shoudl already clone the note

packages/plugin-core/src/commands/MergeNoteCommand.ts
```diff
@@ -0,0 +1,453 @@
...
+    const deleteSourceChanges = await this.deleteSource({ sourceNote });
+
+    const noteChangeEntries = [
+      ...appendNoteChanges,
+      ...updateBacklinksChanges,
+      ...deleteSourceChanges,
+    ];
+
+    // open the destination note
+    await this.extension.wsUtils.openNote(destNote);
```
kevinslin, line 404: do we need to close the source note?

packages/plugin-core/src/commands/MergeNoteCommand.ts
```diff
@@ -0,0 +1,453 @@
...
+    sourceNote: NoteProps;
+    destNote: NoteProps;
+  }) {
+    const { sourceNote, destNote } = opts;
+    // grab body from current active note
+    const appendPayload = sourceNote.body;
+
+    // append to end
+    const destBody = destNote.body;
+    const newBody = `${destBody}\n\n${appendPayload}`;
```
kevinslin, line 206: should we have a line here saying the note was merged? if we go with evernote behavior, it would be

```
{body of previous note}
\n
---
# {title of merged note}
\n
{body of new note)
``

kevinslin (changes requested):
left some comments

packages/plugin-core/src/commands/MergeNoteCommand.ts
hikchoi: that's a good idea.

packages/plugin-core/src/commands/MergeNoteCommand.ts
hikchoi: I was being extra to not mutate it before I explicitly call write but I guess this is now taken care of with the recent store / engine improvements 😄

packages/plugin-core/src/commands/MergeNoteCommand.ts
hikchoi: doesn't need to be sequential. good catch!

packages/plugin-core/src/commands/MergeNoteCommand.ts
hikchoi: I initially thought it'd be nice if the deleted note still stuck around in the editor, but it'd be better to close it to be consistent with the behavior of `Dendron: Delete` now that I think about it. Also better to clean up a note in the editor that isn't in the store to prevent the engine going out of sync potentially.

kevinslin:
looks like we have build errors

hikchoi:
> looks like we have build errors

@kevinslin sorry about that. this week's release had a breaking change with the engine (return signature changed for `writeNote`). Rebased / fixed now

kevinslin (approved):

PR #3352, opened 2022-08-05
enhance/tutorial preview lock 2

kevinslin:
chore(workspace): add lock tutorial

Show Dendron Tutorial with the preview locked. Initially starting this off with a 4/1 weight btw the regular tutorial and the new tutorial.


namjul (approved):
Lgtm. I tried to test it locally but was not able to run the tutorial. I forgot what exactly the issue was that we looked into in our last pairing session @kevin but I am assuming we did not fixed that and its still complicated to test the tutorial.

PR #3359, opened 2022-08-08
enhance(workspace): Hook up engine v3 with schema store

tma66:
**Background**
This PR hooks up the engine v3 with the schema stores from https://github.com/dendronhq/dendron/pull/3348. This will support schema related methods on the engine side such as  `initSchema` and CRUD methods related to schemas

**Changes**
1. Implement `initSchema` and add schemas to notes while init
2. Deprecated `engine.schemas` in favor of `engine.getSchema`. Async to follow up next
3. Changed return signature of `getNote` and `getSchema` to return RespV3<...>. This will allow the engine to propagate the error to the clients w/o dropping it. (consistent with rest of engine methods)

**Testing**
1. Updated engine test spec file to include schemas
2. Manually tested against test-workspace and confirmed same number of schemas after init compared with existing engine


jonathanyeung (approved):
Looks good overall - can we wait until https://github.com/dendronhq/dendron/pull/3343/ goes into master before merging this one?

Should we also go ahead and move over SchemaStore stuff into 'common-all'?

kevinslin (approved):

tma66:
> Looks good overall - can we wait until #3343 goes into master before merging this one?
> 
> Should we also go ahead and move over SchemaStore stuff into 'common-all'?

Yeup, I'll move them right now

Issue #3355, opened 2022-08-06
GoToSiblingCommand: You found a bug!


eshoberg:
GoToSiblingCommand - CTRL+Shift+[

## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [X ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug
I'm not really sure, Using VS Code in 'Microsoft Insider' mode and there are so many things going on that I'm not even sure I was working with Dendron. I honestly thought I was doing an auto complete tutorial from with VS Code. But I'm sure Dendron is active.

Doing some kind of walkthrough regarding auto complete features, etc. on an interactive page titled 'Editor Playground' and was told to press CTRL+Shift+[ in order to code fold (picture attached).

## To Reproduce

1. Randomly choose to allow Microsoft to have basically full control of what your computer is doing and downloading. 
2. Decide, because it popped up, that you would start a tutorial that the VS Code seems to have decided you might like. 
3. Press the key combination mentioned above when told to do so on an HTML element in the editor.
4. See error, not fold. 

## Expected behavior
I mean, your guess is as good as mine. All those AI things that the AI Microsoft dude installed are running amuck. I expected the html to do a code fold as I was told it would do. 

## Desktop (please complete the following information)

- OS: Windows 11 Pre-release 
- Browser: Pre-release VS Code
Version: 1.70.0-insider (user setup)
Date: 2022-08-03T05:53:23.899Z
Electron: 18.3.5
Chromium: 100.0.4896.160
Node.js: 16.13.2
V8: 10.0.139.17-electron.0
OS: Windows_NT x64 10.0.22622
  - Dendron CLI: 0.106.0

## Additional context

This is really kind of a pain in the arse reporting bugs like this, why can't Dendron report his own bugs?

## Dendron Log file

Please attach the output of `> Dendron:Dev: Open Logs` here
[dendron.log](https://github.com/dendronhq/dendron/files/9274719/dendron.log)
[...]

PR #3363, opened 2022-08-09
fix (navigate): check if dendron note is active

hayata-suenaga:
This PR fixes the bug caused when `Dendron: Go Next/Previous Sibling` command is run without an active dendron note on the editor.

issue: #3355

### DIFF
--- a/packages/plugin-core/src/test/suite-integ/GoToSibling.test.ts
+++ b/packages/plugin-core/src/test/suite-integ/GoToSibling.test.ts
@@ -3,7 +3,10 @@ import {
   CreateNoteOptsV4,
   NoteTestUtilsV4,
 } from "@dendronhq/common-test-utils";
+import fs from "fs";
 import { describe } from "mocha";
+import path from "path";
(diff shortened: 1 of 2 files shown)


kevinslin (changes requested):
can you also write a test case for this?

packages/plugin-core/src/test/suite-integ/GoToSibling.test.ts
```diff
@@ -182,6 +185,27 @@ suite("GoToSibling", () => {
       });
     });
 
+    describeSingleWS("no active dendron note", {}, () => {
+      test("Warning message should appear", async () => {
+        await VSCodeUtils.closeAllEditors();
+        // Create a file that is not a Dendron note and show it on editor
+        const workspaceRootPath =
```
kevinslin, line 192: would not use `workspaceFolders` since this will write it to the code workspace. instead, get `wsRoot` from `ExtensionProvider.getEngine().wsRoot`

kevinslin (approved):

PR #3368, opened 2022-08-10
fix(publish): deprecate Dendron: Publish Dev command

Harshita-mindfire:
This PR aims to deprecate `Dendron: Publish Dev` command. The user gets a warning toaster to use Dendron CLI instead with a button to publish cook docs.
added it as a `fix` to include this in changelog


kevinslin (commented):
what is the usage of publish dev today? do we have a chart on amplitude on this?

Harshita-mindfire:
> what is the usage of publish dev today? do we have a chart on amplitude on this?

5 unique users in the last 7 days and 24 in the past 30 days: [Chart](https://analytics.amplitude.com/dendron/chart/yhm7i9k?source=dashboard)

jonathanyeung (approved):

kevinslin (approved):

PR #3370, opened 2022-08-10
chore: add exclude include filter for vscode filestore

Harshita-mindfire:
This PR aims to add an include/exclude filter while reading directory in vs filestore.

Dependency update(common-all): minimatch, @types-minimatch updated to 5.1.0 and 3.0.5 resp.


packages/plugin-core/src/web/engine/store/VSCodeFileStore.ts
```diff
@@ -161,3 +159,20 @@ export async function getAllFilesWithTypes(opts: GetAllFilesOpts) {
     };
   }
 }
+
+/**
+ * Implemented this function again here from common-server.
+ *  This uses minimatch of latest version which is independent of path module
+ */
+function globMatch({
```
kevinslin, line 167: can we migrate the `globMatch` from `common-server` to `common-all` instead using the latest `minimatch`?

yarn.lock
```diff
@@ -4925,7 +4925,7 @@
   resolved "https://registry.npmjs.org/@types/mime/-/mime-1.3.2.tgz#93e25bf9ee75fe0fd80b594bc4feb0e862111b5a"
   integrity sha512-YATxVxgRqNH6nHEIsvg6k2Boc1JHI9ZbH5iWFFv/MTkchz3b1ieGDa5T0a9RznNdI0KhVbdbWSN+KWWrQZRxTw==
 
-"@types/minimatch@*", "@types/minimatch@^3.0.3":
+"@types/minimatch@*", "@types/minimatch@^3.0.3", "@types/minimatch@^3.0.5":
```
kevinslin, line 4928: can we standardize on `@types/minimatch@^3.05`?

kevinslin:
> This PR aims to add an include/exclude filter while reading directory in vs filestore.
> 
> Dependency update(common-all): minimatch, @types-minimatch updated to 5.1.0 and 3.0.5 resp.

is there a reason we can't standardize to `5.1.0` across the board?

packages/plugin-core/src/web/engine/store/VSCodeFileStore.ts
Harshita-mindfire: we can. Will migrate `globMatch` from common-server to common-all, this would also require to have minimatch of latest reason. Reason being the latest version does not have dependency on path module.

jonathanyeung (approved):

jonathanyeung:
> > This PR aims to add an include/exclude filter while reading directory in vs filestore.
> > Dependency update(common-all): minimatch, @types-minimatch updated to 5.1.0 and 3.0.5 resp.
> 
> is there a reason we can't standardize to `5.1.0` across the board?

3.05 is the latest version on `@types-minimatch`

kevinslin (approved):

PR #3386, opened 2022-08-12
feat(workspace): tree view in web ext

jonathanyeung:
## feat(workspace): tree view in web ext

Some differences in how the tree view works compared with original:
- The tree view is now lazily retrieved. Instead of retrieving the entire workspace note state at initialization, we only retrieve the parts of the tree that are necessary for rendering what the user sees.
- When engine events happen, we only update the parts of the tree that have changed.  In the past, we would refresh the entire tree.
- Cache `NotePropsMeta` instead of `NoteProps`

This should lead to a faster tree view and a much reduced memory profile when the tree view is in use.

As part of this change, I'm also doing some cleanup of removing `DVaultUriVariant`, as it turns out it wasn't necessary. 

Some outstanding issues:
- tree view won't observe deletions right now; this is pending some porting work on delete functionality in `DendronEngineV3Web`.
- It won't observe user settings on display type based on `MetadataService` - those files are inaccessible in web


packages/plugin-core/src/web/extension.ts
```diff
@@ -1,26 +1,25 @@
...
 import { setupWebExtContainer } from "./injection-providers/setupWebExtContainer";
+import { NativeTreeView } from "./views/treeView/NativeTreeView";
 
 /**
  * This is the entry point for the web extension variant of Dendron
  * @param context
  */
 export async function activate(context: vscode.ExtensionContext) {
   // Use the web extension injection container:
-  try {
```
kevinslin, line 14: did we decide to get rid of this try/catch?

packages/plugin-core/src/web/views/treeView/NativeTreeView.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    if (!doc.fileName.endsWith("md")) {
+      return;
+    }
+
+    const note = await this.wsUtils.getNoteFromDocument(doc);
+
+    if (note && note.length > 0) {
+      await this._provider.prepNodeForReveal(note[0].id);
+      this.treeView!.reveal(note[0].id, { focus: false, expand: 3 });
```
kevinslin, line 123: instead of a `!` cast, can we use a getter for the tree view that throws an error if the tree view is undefined?

packages/plugin-core/src/web/views/treeView/EngineNoteProvider.ts
```diff
@@ -0,0 +1,318 @@
...
+      ? TreeItemCollapsibleState.None
+      : TreeItemCollapsibleState.Collapsed;
+
+    const tn = new TreeNote(this.wsRoot, {
+      note,
+      collapsibleState,
+      labelType: this._treeViewConfig.LabelTypeSetting,
+    });
+    if (note.stub) {
+      tn.iconPath = new ThemeIcon("gist-new"); // TODO: Replace with ICONS.STUB constant
```
kevinslin, line 259: why not just do this?

packages/plugin-core/src/web/extension.ts
jonathanyeung: No - this is a merge error, thanks for the catch.

packages/plugin-core/src/web/views/treeView/EngineNoteProvider.ts
jonathanyeung: I'll defer this for now - cuz constants.ts has a dependency on `engine-server`, so it'll touch quite a lot of files to make the change.

jonathanyeung:
@kevinslin  - heads up I went ahead and merged this b/c I want to rebase some branches I'm working out of.  Your changes should all be addressed, but if there's anything else pending I can handle in a follow up.

Issue #3381, opened 2022-08-11
Schema template does not get populated when `namespace: true` is set


rlh1994:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [x] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When a schema is created and a level has both a template and `namespace: true` is set then the notes created at this and all subsequent levels do not get populated with the template. Removing the `namespace: true`  correctly populates the template

## To Reproduce

1. Create a schema file as follows:
    ```
        version: 1
        imports: []
        schemas:
          - id: test
            children: 
              - testing
            title: test
            parent: root
          - id: testing
            pattern: "*"
            title: "testing"
            # namespace: true
            template:
              id: template.test
              type: note
    ```
3. Create some `template.test` with literally anything in it
4. Create a note called `test.test1` - it should be populated with your template
5. Uncomment the namespace line in the schema and save
6. Create a note called `test.test2` and `test.test2.test3` and see neither are populated with the template.

## Expected behavior

Namespace set to true should apply the template to that and all subsequent levels in the namespace

## Desktop (please complete the following information)

- OS: Windows
- Versions (all that apply to bug):
  - Dendron Extension: 0.107.2
  - VS Code version: 1.68.1
  - Dendron CLI: N/A

hikchoi:
Thank you for the report. I can reproduce this. Taking a look!

hikchoi:
@rlh1994 A fix should go out either with the next release or the release after that.

Note that namespace schema nodes only match with the note with an exact match, and its immediate children. It does not recursively go down to an arbitrary note depth. So not all subsequent levels under the namespace will have the schema (and the template) applied.

See: https://wiki.dendron.so/notes/c5e5adde-5459-409b-b34d-a0d75cbb1052/#namespace
> making a schema node a namespace automatically brings its immediate children as part of the schema, even if they are not explicitly defined in the schema definition. it will also automatically set the pattern to {namespace}.*

This was a bug that went unnoticed for quite a long time 🙂 thanks for catching it!

[...]

PR #3388, opened 2022-08-12
fix(schema): correctly match namespace schema nodes, and correctly apply schema to new note when note existed as stub

hikchoi:
# fix(schema): correctly match namespace schema nodes, and correctly apply schema to new note when note existed as stub

This PR:
- Correctly matches notes with namespace schema nodes
  - namespace node match pattern was trimmed excessively and was not matching the node on the same depth as well as its immediate children
- Correctly apply schema when a note that existed as a stub is accepted as a new item through note lookup
  - When a note is created as a stub (because of a leaf note), it bypasses any schema application.
  - When this note is later accepted as a new note through lookup, this note that can potentially have a schema match does not get created with a schema applied.

### correctly matches namespace
1. create the following schema:
```
    version: 1
    imports: []
    schemas:
      - id: test
        children: 
          - testing
        title: test
        parent: root
      - id: testing
        pattern: "*"
        title: "testing"
        namespace: true
        template:
          id: template.test
          type: note
```
2. Create `template.test` with any note body
3. Create notes `test.test1` and `test.test1.test2` and see that `template.test` was applied to both of them
### correctly applies schema to newly accepted stub notes
1. Create the same schema describe above
2. Create note `test.one.two.three`
    - Note that this has no schema match, but it's parent that is a stub should match the schema above
3. Using lookup, create the note `test.one.two`.
    - This should already be in the quickpick as a stub
5. Since it has a matching schema, it should have the template applied to it.

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -440,6 +440,27 @@ export class NoteUtils {
     return note;
   }
 
+  /**
+   * Given a stub note, update it so that it has a schema applied to it
+   * This is done before the stub note is accepted as a new item
+   * and saved to the store
+   */
+  static updateStubWithSchema(opts: {
+    stubNote: NoteProps;
+    engine: DEngineClient;
+  }): NoteProps {
+    const { stubNote, engine } = opts;
+    const schemaMatch = SchemaUtils.matchPath({
+      notePath: stubNote.fname,
+      schemaModDict: engine.schemas,
+    });
+    if (schemaMatch) {
+      const { schema, schemaModule } = schemaMatch;
+      NoteUtils.addSchema({ note: stubNote, schemaModule, schema });
+    }
+    return stubNote;
+  }
+
   static createRoot(opts: Partial<NoteOpts> & { vault: DVault }): NoteProps {
     return DNodeUtils.create({
       ...opts,
@@ -1671,7 +1692,6 @@ export class SchemaUtils {
     };
 
     const nextNotePath = getChildOfPath(notePath, matched);
-
     const match = SchemaUtils.matchNotePathWithSchemaAtLevel({
       notePath: nextNotePath,
       schemas: schemaCandidates,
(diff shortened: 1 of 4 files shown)


hikchoi:
related: https://github.com/dendronhq/dendron/issues/3381

packages/common-all/src/dnode.ts
```diff
@@ -1762,8 +1782,10 @@ export class SchemaUtils {
     const match = _.find(schemas, (sc) => {
       const pattern = SchemaUtils.getPatternRecursive(sc, schemaModule.schemas);
       if (sc?.data?.namespace && matchNamespace) {
+        // if we are matching a namespace node,
+        // match on the level of the node itself and its immediate children.
         namespace = true;
-        return minimatch(notePathClean, _.trimEnd(pattern, "/*"));
```
hikchoi, line 1766: The issue here was that `_.trimEnd` was removing the entire glob pattern when what we really wanted was to trim only one level of it.

hikchoi:
Sorry about the scope creep with the second fix item. Found it while I was debugging the first issue and had to fix it for completeness 😄

packages/common-all/src/dnode.ts
kevinslin: its not immediately obvious what `slice(0,-2)` means - can you add some comments here?

hikchoi: improved the comments to make clear what's happening here.

hikchoi:
There was a breaking change with the engine method `writeNote` return signature change. rebased / fixed.

PR #3392, opened 2022-08-12
enhance(cli): New write notes CLI command

tma66:
**Background**
Moving forward, we would like the notes CLI to closely mirror engine notes methods. For this pr, we will introduce a `WRITE` cli command that will create a note if fname/vault doesn't exist or update body of existing note.

Options:
`fname`: Required
`vault`: Required if more than one vault exists
`body`: Not required. Defaults to ""

**How to use**
```
dendron note write --useLocalEngine --fname "mytest" --vault vault --body "this is a body"
```

**Other Changes**
1. Renamed options for `DELETE` and `MOVE` to use `fname` instead of `query`

### DIFF
--- a/packages/dendron-cli/src/commands/notes.ts
+++ b/packages/dendron-cli/src/commands/notes.ts
@@ -23,7 +23,7 @@ type CommandCLIOpts = {
   query?: string;
   cmd: NoteCommands;
   output?: NoteCLIOutput;
-  fName?: string;
+  fname?: string;
   destFname?: string;
   destVaultName?: string;
   newEngine?: boolean;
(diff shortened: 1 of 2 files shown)


packages/dendron-cli/src/commands/notes.ts
```diff
@@ -344,27 +364,65 @@ export class NoteCLICommand extends CLICommand<CommandOpts, CommandOutput> {
...
+          if (resp.error) {
+            return {
+              error: ErrorFactory.createInvalidStateError({
+                message: `write failed: ${resp.error.message}`,
+              }),
+              data: undefined,
+            };
+          } else {
+            this.print(`wrote ${note.fname}`);
+            return { data: { payload: note.fname, rawData: resp } };
```
kevinslin, line 391: can we return a status that indicates whether we created a new note or aupdated an existing note?

kevinslin (changes requested):

packages/dendron-cli/src/commands/notes.ts
tma66: sure

jonathanyeung (approved):

kevinslin (approved):

PR #3393, opened 2022-08-12
chore(edit): add aliasMode config to CopyNoteLink cmd

hayata-suenaga:
This PR adds a new configuration option called `aliasMode` that configures if the note alias is included in note link when `Dendron: Copy Note Link` is executed.

task: `[[Support Alias Mode for Copy Note Link|task.2022.08.11.support-alias-mode-for-copy-note-link]]`


packages/plugin-core/src/test/suite-integ/CopyNoteLink.test.ts
```diff
@@ -259,6 +259,29 @@ suite("CopyNoteLink", function () {
     }
   );
 
+  describeSingleWS(
+    "WHEN the alias mode is none, THEN CopyNoteLink should only return a link without a note alias",
+    {
+      postSetupHook: ENGINE_HOOKS.setupBasic,
+      modConfigCb: (config) => {
+        ConfigUtils.setAliasMode(config, "non");
```
kevinslin, line 267: `none`?

kevinslin (changes requested):

packages/common-all/src/constants/configs/commands.ts
```diff
@@ -165,6 +165,10 @@ const COPY_NOTE_LINK: DendronConfigEntryCollection<CopyNoteLinkConfig> = {
     label: "Non-note file",
     desc: "Copy note link settings for non-note files.",
   },
+  aliasMode: {
+    label: "Include a note link",
+    desc: "Whether or not to include alias when copying a ntoe link",
```
jonathanyeung, line 170: nit: type on 'note'

packages/common-all/data/dendron-yml.validator.json
```diff
@@ -647,8 +647,18 @@
             }
           },
           "additionalProperties": false
+        },
+        "aliasMode": {
+          "type": "string",
+          "enum": [
+            "none",
+            "alias"
```
jonathanyeung, line 655: maybe this would be more intuitive if the enum value is called "title" instead of "alias"

packages/common-all/src/constants/configs/commands.ts
```diff
@@ -165,6 +165,10 @@ const COPY_NOTE_LINK: DendronConfigEntryCollection<CopyNoteLinkConfig> = {
     label: "Non-note file",
     desc: "Copy note link settings for non-note files.",
   },
+  aliasMode: {
```
jonathanyeung, line 168: I think we should make this name be specific to the `copyNoteLink` command - how about `copyNoteLinkAliasMode`?

jonathanyeung (changes requested):

packages/common-all/data/dendron-yml.validator.json
kevinslin: the reason we're using `alias` is because we currently use it in `insertNoteTitle`. would recommend changing that in a separate pass (eg. change both together) if we go down that route

hayata-suenaga: @kevinslin do you mean `insertNoteLink`? In that command, we have five options right now, once of which is `title`.

hayata-suenaga: I actually already changed `alias` to `title`. Should I change it back to `alias`?

kevinslin:
looks like the snapshot tests are failing

packages/common-all/data/dendron-yml.validator.json
kevinslin: all good. lets keep it as is

kevinslin (approved):

PR #3401, opened 2022-08-15
enhance: introduce sqlite as a plugabble metadata store

kevinslin:
enhance: introduce sqlite as a plugabble metadata store

30x speed up on basic queries (50k notes, queries complete in ~5ms vs +150ms)

To use, add the following setting in `dendronrc.yml` or `dendron.yml`
```yml
workspace:
  metadataStore: sqlite
```

Also make sure `sqlite` is available on your machine. 

Future work that is needed:
- generate clients on non-mac clients
- support incremental updates (currently, this just initializes sqlite once when the db is empty)


packages/engine-server/src/drivers/SQLiteMetadataStore.ts
```diff
@@ -0,0 +1,171 @@
...
+  return _prisma;
+}
+
+export type NoteIndexLightProps = {
+  fname: string;
+  id: string;
+  foo: string;
+};
+
+export class SQLiteMetadataStore {
```
jonathanyeung, line 27: this should implement `IDataStore<string, NotePropsMeta>` to make sure it's compatible with new engine architecture.

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -486,6 +487,15 @@ export class FileStorage implements DStore {
     let notesWithLinks: NoteProps[] = [];
     let errors: IDendronError<any>[] = [];
     const start = process.hrtime();
+    // instantiate so we can use singleton later
```
jonathanyeung, line 490: I don't think we should be touching `storev2` anymore to instantiate anything. We should instantiate it during construction of `DendronEngineV3`

jonathanyeung (changes requested):

namjul:
Quickly toying around I get `prisma client not initialized` when running `➜ dendron note lookup --query "curious" --vault "wiki" --attach
`.

packages/engine-server/src/drivers/file/storev2.ts
kevinslin: the reason its here is because we want to gain some confidence with it in our existing workspace. unfortunately `DendronEngineV3` hasn't been implemented yet

kevinslin:
> Quickly toying around I get `prisma client not initialized` when running `➜ dendron note lookup --query "curious" --vault "wiki" --attach `.

you'll need to run `yarn buildPrismaClient` inside of `engine-server` to generate the prisma client

namjul:
without buildPrimsaClient i would get `Error: Cannot find module '../generated-prisma-client'`. With having done `yarn buildPrismaClient
` I still get `Error: prisma client not initialized
`

kevinslin:
> without buildPrimsaClient i would get `Error: Cannot find module '../generated-prisma-client'`. With having done `yarn buildPrismaClient ` I still get `Error: prisma client not initialized`

hmm, i created a makefile to make this easier. can you try following the instructions here? [[dendron://dendron.docs/pkg.dendron-engine.dev.cook#updating-the-prisma-schema]]

packages/dendron-cli/src/commands/devCLICommand.ts
```diff
@@ -450,7 +450,7 @@ export class DevCLICommand extends CLICommand<CommandOpts, CommandOutput> {
     await this.syncAssets(opts);
 
     this.print("prep repo...");
-    await BuildUtils.prepPluginPkg();
```
kevinslin, line 453: @jonathanyeung was there a reason why we didn't pass in the extension target directly at this phase?

packages/api-server/package.json
```diff
@@ -54,9 +54,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@dendronhq/common-all": "^0.108.0",
-    "@dendronhq/common-server": "^0.108.0",
-    "@dendronhq/engine-server": "^0.108.0",
+    "@dendronhq/common-all": "^0.108.1",
```
jonathanyeung, line 57: Can we revert these patch updates? Else we'll get merge conflicts with release 109.

packages/engine-server/prisma/migrations/20220819014343_add_davult/migration.sql
```diff
@@ -0,0 +1,24 @@
+-- CreateTable
```
jonathanyeung, line 1: is this path naming intentional? `packages/engine-server/prisma/migrations/20220819014343_add_davult/migration.sql`

jonathanyeung (changes requested):

packages/engine-server/prisma/migrations/20220819014343_add_davult/migration.sql
kevinslin: autogenerated by prisma

packages/dendron-cli/src/commands/devCLICommand.ts
jonathanyeung: Your change should be fine - I think I had just done that to minimize chances of regressing the regular release build when I was first setting up the nightly stuff.

Issue #3290, opened 2022-07-25
Using dendronrc.yaml can cause dnedron to fail to load.


Lodrus:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [X ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

When a vault is specified in the dendronrc.yml file, the Vault Add and Vault Remove commands can create a name conflict error. This prevents Dendron from successfully loading.

## To Reproduce

1. Create a new vault named "Personal Notes" via the Vault Add command
2. Move the vault directory to a new path such as "~/Personal Notes"
3. Remove the settings related to the "Personal Notes" vault from "dendron.yml" and add them to dendronrc.yml instead
4. Reload VSCode via the "Developer: Reload Window" command. Dendron loads normally. The "Personal Notes" vault is available
5. Add a new vault named "Test" via the Vault Add command. The produces an error message: Following vault names have duplicates: Personal Notes
6. Review the "dendron.yml" file and see that it contains settings for the "Personal Notes" and "Test" vaults.
7. Remove the settings for the "Personal Notes" vault from dendron.yml
8. Reload VSCode via the "Developer: Reload Window" command. Dendron loads normally. The "Personal Notes" vault is available
9. Delete the "Test" vault via the Vault Remove command. This produces the same error and merged data in dendron.yml

## Expected behavior

The content of dendronrc.yml should not be merged into dendron.yml. This would also prevent the error message and allow Dendron to function normally.

## Desktop (please complete the following information)

- OS: Windows
- Versions (all that apply to bug):
  - Dendron Extension: v0.104.1
  - VS Code version: 1.69.2

## Additional context

This was discovered while attempting to setup a globally available "Personal Notes" vault that would be accessible in all workspaces in VSCode

jonathanyeung:
Thanks for finding this issue @Lodrus - dendronrc.yaml is still a germ /alpha stage feature, but we can add an item on our backlog to make the vault configuration better w.r.t `dendronrc.yaml`

Lodrus:
[...]

PR #3402, opened 2022-08-16
chore(workspace): Migrate config get/write to DConfig

tma66:
**Background**
We currently have multiple avenues for reading configs
* engine.config
* engine.getConfig
* DConfig.readConfigSync
* ExtensionProvider.getDWorkspace().config

This makes it harder in the future to keep track and introduces out-of-sync config issues.

**Change**
1. Since all of them are currently a variant of DConfig.readConfigSync, we will first migrate all uses of `engine.config` and `engine.getConfig` to `DConfig.readConfigSync`
2. Cleanup unused engine methods (includes moving methods that are utility methods outside the engine)
3. **UPDATE** Move DConfig to `common-server`
4. Explicitly pass in config to remark processor. Since engine.config no longer exists and we cannot use DConfig in `unified`, the only other option is to have the clients pass in the config

**Bug fix**
1. By using DConfig, we read config from file each time. Although this may introduce a tiny latency (1ms), it solves issues where engine.config is out of sync. Currently if a user updates a config file, we don't update engine.config.

**Todo**
Migrate to use `DConfig.readConfigAndApplyLocalOverrideSync` which is the goal long term. Currently there is a bug related https://github.com/dendronhq/dendron/issues/3290 which will be part of the follow up


packages/engine-server/src/markdown/decorations/hashTags.ts
```diff
@@ -46,7 +47,10 @@ export async function decorateTag({
     fname,
     engine,
   });
-  if (colorType === "configured" || !engine.config.noRandomlyColoredTags) {
+  if (
+    colorType === "configured" ||
+    !DConfig.readConfigSync(engine.wsRoot).noRandomlyColoredTags
```
kevinslin, line 52: this will be called a whole bunch of times

packages/engine-server/src/markdown/decorations/decorations.ts
```diff
@@ -81,18 +82,17 @@ export async function runAllDecorators(
       fname: note.fname,
     }
   );
+  const maxNoteLength = ConfigUtils.getWorkspace(
```
kevinslin, line 85: this will be called a whole bunch of times

packages/engine-server/src/config.ts
```diff
@@ -247,7 +247,10 @@ export class DConfig {
   static readConfigSync(wsRoot: string) {
```
kevinslin, line 247: can we cache this in a global object instead of reading it from the fs everytime?

packages/engine-server/src/config.ts
```diff
@@ -247,7 +247,10 @@ export class DConfig {
...
     const configPath = DConfig.configPath(wsRoot);
     // TODO: validate
```
kevinslin, line 249: if we are using this as the source of truth, it should have logic of `readConfigAndApplyLocalOverrideSync`

kevinslin: also, as we discussed offline, there are really multiple configs at play here.

the workspace config, the local config override (if it exists), and the combined config. would recommend returning an object here 

{mergedConfig, wsConfig, localConfig} so that operations that need to update the config can update the `wsConfig`

kevinslin (changes requested):
some notes about the config structure

packages/engine-server/src/config.ts
tma66: Can we do this in a separate pr? I wanted to first move all usages of engine.config to Dconfig first before introducing any new business logic. Otherwise it might be too many moving pieces at once

packages/engine-server/src/config.ts
tma66: We currently don't have a file watcher for config files. This is why if a user manually updates the config, it doesn't get reflected in the engine.config prop.

Once we add watcher support, we can cache, but I'm not sure if thats necessary right now

packages/engine-server/src/markdown/decorations/hashTags.ts
tma66: Got it. I'll pass the config down from the engine so it doesn't need to be called

packages/api-server/src/modules/config.ts
```diff
@@ -18,13 +18,15 @@ export class ConfigController {
     return ConfigController.singleton;
   }
 
-  async get({ ws }: WorkspaceRequest): Promise<RespV2<ConfigGetPayload>> {
+  async get({
```
jonathanyeung, line 21: We should probably try to get rid of this controller altogether at some point (later), I don't see what it's necessary.

packages/common-all/src/api.ts
```diff
@@ -365,15 +364,19 @@ export class DendronAPI extends API {
     });
   }
 
-  configGet(req: WorkspaceRequest): Promise<APIPayload<ConfigGetPayload>> {
+  configGet(
```
jonathanyeung, line 367: maybe I'm thinking about this wrong, but since `DCommonProps` has config removed, can't we get rid of these two methods altogether?

packages/engine-server/src/config.ts
jonathanyeung: I don't think the performance gain is worth the complication right now of a file watcher - fs reading on a single file isn't so bad, we do sync reads into FS all over the place anyway. 

If anything, one optimization we can do is to have an async version `readConfig` where consumers can use the async version if they want to / can.

packages/common-all/src/api.ts
tma66: Looks like the frontend still makes api calls to get the config (`slice.ts`)

packages/engine-server/src/config.ts
tma66: I also feel reading from `fs` is not bad. To compromise, I'll add a global config, but the default behavior is to read from fs. The expensive calls like `getDecorations` will use the cache while the others read from fs.

The global config also doesn't work well with the tests, hence why I'm making default read from fs

packages/plugin-core/src/features/windowDecorations.ts
```diff
@@ -121,9 +122,8 @@ export async function updateDecorations(editor: TextEditor): Promise<{
   try {
     const ctx = "updateDecorations";
     const engine = ExtensionProvider.getEngine();
-    if (
-      ConfigUtils.getWorkspace(engine.config).enableEditorDecorations === false
-    ) {
+    const config = DConfig.readConfigSync(engine.wsRoot);
```
kevinslin, line 125: can this use the cache?

packages/unified/src/utilsv5.ts
```diff
@@ -265,7 +266,7 @@ export class MDUtilsV5 {
             });
           }
           if (!data.config) {
-            data.config = data.engine!.config;
+            data.config = DConfig.readConfigSync(data.engine!.wsRoot);
```
kevinslin, line 269: this gets used all over the place. can it use the cache?

packages/unified/src/utilsv5.ts
```diff
@@ -363,7 +364,7 @@ export class MDUtilsV5 {
           });
         }
         if (!data.config) {
-          data.config = data.engine!.config;
+          data.config = DConfig.readConfigSync(data.engine!.wsRoot);
```
kevinslin, line 367: this gets used all over the place. can it use the cache?

packages/unified/src/utilsv5.ts
```diff
@@ -408,7 +409,7 @@ export class MDUtilsV5 {
       .use(slug);
 
     // apply plugins enabled by config
-    const config = data?.engine?.config as IntermediateDendronConfig;
+    const config = DConfig.readConfigSync(data?.engine?.wsRoot as string);
```
kevinslin, line 412: see above

packages/unified/src/utilsv5.ts
tma66: sure

bootstrap/scripts/buildAll.js
```diff
@@ -13,9 +13,8 @@ const $ = (cmd) => {
 
 console.log("building all...");
 $(`npx lerna run build --scope @dendronhq/common-all`);
-$(
-  `npx lerna run build --parallel --scope "@dendronhq/{unified,common-server}"`
```
jonathanyeung, line 17: Was this a merge conflict? Can we change it back to the --parallel version?

tma66: I needed to build common-server first or it fails. I believe its b/c one of the unified files needs `DConfig` which I moved to common-server

kevinslin (approved):

bootstrap/scripts/buildAll.js
tma66: Synced up with Jon. Unified cannot import from common-server which means we cannot read configs in this package. Therefore I'm passing in the config to the proc methods along with the data (whereas before it used to be part of the engine data).

To ensure clients don't accidentally forget the config, I'm making config a required prop as well

kevins8:
Does config have to be a required prop? Can the default config be in
common-all and we use that if no config is passed in?

tma66:
> Does config have to be a required prop? Can the default config be in common-all and we use that if no config is passed in?

It's safer to be explicit about what is being used; clients should be telling the engine-server what config is being used. Otherwise it is easy to forget and it'll be harder to debug.

For modes that don't require a config (such as `ProcMode.NO_DATA`), we can provide a method that uses the default config for them such as `procRemarkParseNoData`

kevinslin (approved):

PR #3407, opened 2022-08-16
fix(edit): prevent override of already existing frontmatter values when executing Apply Template

hayata-suenaga:


hayata-suenaga:
currently, I'm getting an error when running plugin core tests. The error says that the code is trying to read `body` from `undefined`. But I don't know where this is happening.

when you have time, could you point me to the right direction?

packages/common-server/src/template.ts
```diff
@@ -31,8 +31,16 @@ function copyTemplateProps({
 }: TemplateFunctionProps) {
   const tempNoteProps = _.pick(templateNote, TemplateUtils.TEMPLATE_COPY_PROPS);
   _.forEach(tempNoteProps, (v, k) => {
-    // @ts-ignore
-    targetNote[k] = v;
+    if (k === "custom") {
+      if (targetNote.custom === undefined) targetNote.custom = {};
+      Object.keys(v).forEach((key) => {
```
kevinslin, line 36: this won't work. v can be `undefined` which is leading to your error

packages/common-server/src/template.ts
```diff
@@ -31,8 +31,16 @@ function copyTemplateProps({
 }: TemplateFunctionProps) {
   const tempNoteProps = _.pick(templateNote, TemplateUtils.TEMPLATE_COPY_PROPS);
   _.forEach(tempNoteProps, (v, k) => {
-    // @ts-ignore
-    targetNote[k] = v;
+    if (k === "custom" && v) {
+      if (targetNote.custom === undefined) targetNote.custom = {};
```
kevinslin, line 35: if v is defined, this if statement is useless as it will never be true

hayata-suenaga: @kevinslin I'm checking if target note has any custom properties. A target note might not have any custom properties defined even when the template note has some custom properties.

Is my assumption is misguided?

kevinslin: actually, on second reading, ignore my comment. you're right here

packages/common-server/src/template.ts
```diff
@@ -31,8 +31,16 @@ function copyTemplateProps({
...
+      Object.keys(v).forEach((key) => {
+        // @ts-ignore
+        targetNote[k][key] = targetNote[k][key] || v[key];
```
kevinslin, line 38: instead of using `targetNote[k]` i would use `targetNote['custom']` to make this a little easier to read

kevinslin:
is this still a draft pr?

hayata-suenaga:
> is this still a draft pr?

sorry I did it again. no it's ready for the final review

PR #3411, opened 2022-08-17
Feat/copy note url command for dendron web

Harshita-mindfire:
# Pull Request Checklist


packages/plugin-core/src/web/utils/WSUtils.ts
```diff
@@ -14,19 +22,19 @@ export class WSUtilsWeb {
...
 
-  private getVaultFromDocument(document: vscode.TextDocument) {
+  getVaultFromDocument(document: vscode.TextDocument) {
     const txtPath = document.uri.fsPath;
     const vault = VaultUtils.getVaultByFilePath({
-      wsRoot: this.wsRoot.fsPath,
+      wsRoot: normalizeUnixPath(this.wsRoot.fsPath),
       vaults: this.vaults,
-      fsPath: txtPath,
+      fsPath: normalizeUnixPath(txtPath),
```
Harshita-mindfire, line 30: used normailzed path values here. eventually we compare them normalized anyway in `getVaultByDirPath` and also because I couldn't work around a `Utils` alternative for `path.relative` [here](https://github.com/dendronhq/dendron/blob/master/packages/common-all/src/vault.ts#L234)

packages/plugin-core/src/web/commands/CopyNoteURLCmd.ts
```diff
@@ -0,0 +1,50 @@
+import { ConfigUtils, IntermediateDendronConfig } from "@dendronhq/common-all";
+import _ from "lodash";
+import { env, window, workspace } from "vscode";
+import { injectable, inject } from "tsyringe";
+import { WSUtilsWeb } from "../utils/WSUtils";
+
+@injectable()
+export class CopyNoteURLCmd {
```
jonathanyeung, line 8: Can you try adding a basic test for this?

packages/plugin-core/src/web/extension.ts
```diff
@@ -43,6 +44,17 @@ async function setupCommands(context: vscode.ExtensionContext) {
         await cmd.run();
       })
     );
+
+  if (!existingCommands.includes(CopyNoteURLCmd.key)) {
+    context.subscriptions.push(
+      vscode.commands.registerCommand(
+        CopyNoteURLCmd.key,
+        async (_args: any) => {
+          await container.resolve(CopyNoteURLCmd).run();
```
jonathanyeung, line 53: we should add a test whenever we add a new resolve to make sure that the container is sufficiently configured to be able to instantiate this object.  It should be a two-liner to `setupWebExtContainer.test.ts`: see

```typescript
        const cmd = container.resolve(NoteLookupCmd);

        assert(!_.isUndefined(cmd));
```

jonathanyeung (changes requested):
nice!

packages/plugin-core/src/web/commands/CopyNoteURLCmd.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(
+    @inject("config") private config: IntermediateDendronConfig,
```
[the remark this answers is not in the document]
Harshita-mindfire: okay. 
A query: All the dependencies that we require viz `siteUrl`, `siteUrlIndex`, and `assetsPrefix` are optional. Can we inject undefined values as dependencies?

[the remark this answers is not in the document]
jonathanyeung: I haven't tried it, but it should be possible - if we just specify the type as optional? i.e.
```typescript
@inject("optionalParam") private thisIsOptional: foo?,
```

packages/plugin-core/src/web/utils/site.ts
```diff
@@ -0,0 +1,118 @@
+import {
```
jonathanyeung, line 1: nit: to follow the new convention - can you rename this file to the exported class? so `SiteUtilsWeb.ts`

jonathanyeung (approved):
Great job with the tests! One minor comment about file naming.

Issue #3418, opened 2022-08-18
Case-insensitive alphabetization


ryan-p-randall:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [ ] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup - Dendron's Lookup Command
- [x] Views - Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [x] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

It always surprises me when the alphabetization of notes in the Tree View sorts capital letters before lowercase letters. I would expect the tree view's alphabetization to be case-insensitive.

<img width="295" alt="Dendron's tree view showing Alphabet, Aack, and App" src="https://user-images.githubusercontent.com/850915/185281331-06028037-4201-473f-b013-76e631302cf0.png">

As the screenshot above shows, the current alphabetization logic will sort `ALPHABET` before `Aack` or `App`. 

This becomes a problem when I want notes to be sorted purely by the alphabet, while also having note whose titles are abbreviations or all-caps.  

For instance, if I wanted to have a note about `De La Soul` and one about `DMX` in a published website, right now DMX would be sorted above De La Soul. Similarly, `NYU` would be sorted above `New School` in a set of notes about universities in New York City.

## Describe the solution you'd like ##

I think that the alphabetization logic should entirely ignore capitalization. It's excellent that the tree view will display a note's title as it's written in the YAML header, but I'd prefer that it not affect the sort order.

## Describe alternatives you've considered ##

Alternately, perhaps there could be a setting that allows users to switch between the current behavior and case-insensitive alphabetization.

For a very small set of notes, `nav_order` could be used to change the order how the notes are displayed. However, I think this would be cumbersome to use just to emulate case-insensitive alphabetization.

## Additional context ##

N/A

jonathanyeung:
@ryan-p-randall - so do you prefer DMX or De La Soul? 

Jokes aside - a fix for this should go out in the next release or two: https://github.com/dendronhq/dendron/pull/3420

ryan-p-randall:
[...]

PR #3420, opened 2022-08-18
fix(workspace): case insensitive tree view sorting

jonathanyeung:
## fix(workspace): case insensitive tree view sorting

Make the sort order in tree view case insensitive. This addresses the issue both in current tree view and the upcoming web ext tree view.

related to: https://github.com/dendronhq/dendron/issues/3418

### DIFF
--- a/packages/common-all/src/util/treeUtil.ts
+++ b/packages/common-all/src/util/treeUtil.ts
@@ -170,10 +170,10 @@ export class TreeUtils {
       (noteId) => {
         if (labelType) {
           return labelType === TreeViewItemLabelTypeEnum.filename
-            ? _.last(noteDict[noteId]?.fname.split("."))
-            : noteDict[noteId]?.title;
+            ? _.last(noteDict[noteId]?.fname.split("."))?.toLowerCase()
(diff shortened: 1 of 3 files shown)


hikchoi:
@jonathanyeung 

This reintroduces the problem of underscores not being sorted right, as well as the inconsistencies with how the tree items are labeled vs sorted. (which was addressed here: https://github.com/dendronhq/dendron/pull/2858 ) 

If we want to ignore casing, maybe we should make it a toggle and also label them without casing if toggled on?

kevinslin (approved):

jonathanyeung:
> @jonathanyeung
> 
> This reintroduces the problem of underscores not being sorted right, as well as the inconsistencies with how the tree items are labeled vs sorted. (which was addressed here: #2858 )
> 
> If we want to ignore casing, maybe we should make it a toggle and also label them without casing if toggled on?
> 
> The immediate impact this has is that users who already have an expectation of the tree view being sorted in a certain way would be confused.

As discussed offline, let's make the change as the default and if user's complain, we can introduce a setting.

I still need to fix a test prior to merge.

PR #3424, opened 2022-08-18
fix(workspace): diagnostics regression

hayata-suenaga:


packages/plugin-core/src/commands/DiagnosticsReport.ts
```diff
@@ -32,11 +33,18 @@ export class DiagnosticsReportCommand extends BasicCommand<
       path.dirname(logPath),
       "dendron.server.log"
     );
-    const serverLogFile = fs.readFileSync(serverLogPath, { encoding: "utf8" });
-    const serverLastLines = serverLogFile.slice(-3000);
+
+    const isDevMode = VSCodeUtils.isDevMode();
+    let serverLastLines: string = "";
+    if (isDevMode) {
```
kevinslin, line 39: you want to not do this in dev mode since there won't be server logs here

hayata-suenaga: sorry I forgot to push the latest commit

packages/plugin-core/src/commands/DiagnosticsReport.ts
```diff
@@ -32,11 +33,18 @@ export class DiagnosticsReportCommand extends BasicCommand<
       path.dirname(logPath),
       "dendron.server.log"
     );
-    const serverLogFile = fs.readFileSync(serverLogPath, { encoding: "utf8" });
-    const serverLastLines = serverLogFile.slice(-3000);
+
+    const isDevMode = VSCodeUtils.isDevMode();
+    let serverLastLines: string = "";
+    if (!isDevMode) {
```
kevinslin, line 39: instead of checking for mode, why don't you check if the file exists? if so, then read it

hayata-suenaga:
@kevinslin I modified my code to check the existence of the log file. Some of the diagnostic tests are still failing... I'll try to investigate why this is happening.

kevinslin:
is this supposed to be in draft still?

hayata-suenaga:
> is this supposed to be in draft still?

thought requesting a review changes draft status...

kevinslin (approved):

PR #3441, opened 2022-08-24
chore: dont log traits as custom cmds

kevinslin:
Due to how we log commands, we were creating a new event for each custom
trait. This was blowing up our postgres db as well as inflating our
event count in amplitude
We already log the `dendron.createNoteWithTraits` event when executing
the trait, this change makes it so that we don't log two events for
every trait (the second one being the custom event where the key is
equal to the trait name)


jonathanyeung (approved):
Good catch

PR #3442, opened 2022-08-24
enhance(view): context menu for tree view

Harshita-mindfire:
- This PR aims to add a context menu for tree view. Added Delete Command in tree view context menu.
- It also fixes an issue with Delete Command: If the note is not open in the text editor. it is not deleted when the command is run from the explorer menu.
- It adds a Create Note command in context menu of Tree view, the lookup bar has the initial value of tree item responsible for action.
- adds source: Tree view property for Create Note and Delete Command

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/CreateNoteCommand.ts
+++ b/packages/plugin-core/src/commands/CreateNoteCommand.ts
@@ -8,8 +8,10 @@ import {
 import { AutoCompletableRegistrar } from "../utils/registers/AutoCompletableRegistrar";
 import _ from "lodash";
 import { window } from "vscode";
+import { NoteProps, NoteUtils } from "@dendronhq/common-all";
+import { AnalyticsUtils } from "../utils/analytics";
 
-type CommandOpts = any;
+type CommandOpts = NoteProps & {};
 
 type CommandOutput = {
   lookup: Promise<NoteLookupOutput | undefined>;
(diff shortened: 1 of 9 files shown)


packages/plugin-core/src/commands/DeleteCommand.ts
```diff
@@ -73,6 +68,60 @@ export class DeleteCommand extends BasicCommand<CommandOpts, CommandOutput> {
     return nodePosition?.end.line;
   }
 
+  /**
+   * When Delete Command is ran from Tree view, it gets Noteprops as args
+   */
+  private isNotePropsArgs(opts: CommandOpts) {
```
hkievet, line 74: suggestion:  Maybe make a commandUtils.ts file or something under commands that could share this function between the two commands you have here?

hkievet (commented):
Good job getting this to work!  I'm glad you were able to teach me more about this part of the code base when we paired together!

packages/plugin-core/src/commands/CreateNoteCommand.ts
```diff
@@ -1,28 +1,46 @@
...
+import {
+  CommandOutput as NoteLookupOutput,
+  CommandRunOpts as NoteLookupRunOpts,
+} from "./NoteLookupCommand";
 import { AutoCompletableRegistrar } from "../utils/registers/AutoCompletableRegistrar";
+import _ from "lodash";
+import { window } from "vscode";
 
-type CommandOpts = {};
+type CommandOpts = any;
```
kevinslin, line 12: can we apply the proper types to this?

packages/plugin-core/src/test/suite-integ/NativeTreeView.test.ts
```diff
@@ -117,6 +117,7 @@ suite("NativeTreeView tests", function () {
             genRandomId: true,
           });
         },
+        timeout: 1e6,
```
kevinslin, line 120: is this necessary?

PR #3450, opened 2022-08-25
enhance(editing): new api for onCreate of note traits

hayata-suenaga:
task: `[[Specify Vault for Note Trait|task.2022.08.23.specify-vault-for-note-trait]]`


packages/plugin-core/src/test/suite-integ/components/traits/TestTrait.ts
```diff
@@ -31,5 +31,8 @@ export class TestTrait implements NoteTrait {
     setTemplate: () => {
       return this.template;
     },
+    // setVault: () => {
```
kevinslin, line 34: remove if not used

kevinslin (approved):

PR #3456, opened 2022-08-26
chore(internal): add extended analytics

hikchoi:
# chore(internal): add extended analytics

This PR:
- adds new event properties `extensionsDetail`, `numExtensions`, and `codeFoldersCreated`

# Pull Request Checklist
### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/utils/ExtensionUtils.ts
+++ b/packages/plugin-core/src/utils/ExtensionUtils.ts
@@ -466,27 +466,15 @@ export class ExtensionUtils {
       // something went wrong don't track extension detail
     }
 
+    // NOTE: this will not be accurate in dev mode
     try {
-      // infer install path from non-built-in extension paths
-      const installPaths = allNonBuiltInExtensions.map((ext) => {
-        // vscode-dot-dir-path/extensions/extension-basename
-        // stripping out last two levels
-        return Utils.dirname(Utils.dirname(URI.file(ext.extensionPath))).fsPath;
-      });
-
-      const pathSet = new Set(installPaths);
-      // make sure every element of install paths are identical.
-      // otherwise the inferred path is not correct
-      let installPath: string;
-      if (pathSet.size === 1) {
-        installPath = installPaths[0];
-      } else {
-        // fail-safe case using current extension context
-        // NOTE: this will not be accurate in dev mode
-        installPath = Utils.dirname(
-          Utils.dirname(URI.file(ext.context.extensionPath))
-        ).fsPath;
-      }
+      // infer install path from extension path.
+      // this assumes the user installs all extensions in one place.
+      // that should be the case for almost all cases, but vscode provides a way to
+      // customize install location so this might not be the case in those rare cases.
+      const installPath = Utils.dirname(
+        Utils.dirname(URI.file(ext.context.extensionPath))
+      ).fsPath;
       const fd = fs.openSync(installPath, "r");
       const stat = fs.fstatSync(fd);
       // some file systems don't track birth times.
--- a/packages/plugin-core/src/utils/ExtensionUtils.ts
+++ b/packages/plugin-core/src/utils/ExtensionUtils.ts
@@ -26,6 +26,7 @@ import fs from "fs-extra";
 import _ from "lodash";
 import path from "path";
 import * as vscode from "vscode";
+import { homedir } from "os";
 import { CONFIG, DendronContext } from "../constants";
 import { IDendronExtension } from "../dendronExtensionInterface";
 import { ExtensionProvider } from "../ExtensionProvider";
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -439,6 +442,70 @@ export class ExtensionUtils {
...
+      if (extensionsDetail && extensionsDetail.length > 0) {
+        _.set(trackProps, "extensionsDetail", JSON.stringify(extensionsDetail));
+        _.set(trackProps, "numExtensions", extensionsDetail.length);
+      }
+    } catch (error) {
+      // something went wrong don't track extension detail
+    }
+
+    try {
+      const VSCODE_INSTALLATIONS = [
```
kevinslin, line 466: can we get this from the current vscode installation? 

`codeFoldersCreated` should be a single number that shows when the current initialized version of vscode was created

kevinslin (commented):
one comment about figuring out extension version

packages/plugin-core/src/utils/ExtensionUtils.ts
hikchoi: I've changed it so that we infer the location more intelligently. It will have more coverage (mostly likely will cover portable installs as well) and doesn't rely on trying and failing on known default locations.

packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -439,6 +441,64 @@ export class ExtensionUtils {
...
+      const pathSet = new Set(installPaths);
+      // make sure every element of install paths are identical.
+      // otherwise the inferred path is not correct
+      let installPath: string;
+      if (pathSet.size === 1) {
+        installPath = installPaths[0];
+      } else {
+        // fail-safe case using current extension context
+        // NOTE: this will not be accurate in dev mode
+        installPath = Utils.dirname(
```
kevinslin, line 486: why  an't we just use this from the beginning?

packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -439,6 +441,64 @@ export class ExtensionUtils {
...
+        _.set(trackProps, "extensionsDetail", JSON.stringify(extensionsDetail));
+        _.set(trackProps, "numExtensions", extensionsDetail.length);
+      }
+    } catch (error) {
+      // something went wrong don't track extension detail
+    }
+
+    try {
+      // infer install path from non-built-in extension paths
+      const installPaths = allNonBuiltInExtensions.map((ext) => {
```
kevinslin, line 471: this seems expensive. can we just find the directory based on the directory that the dendron extension is installed form?

kevinslin (commented):
left a comment

packages/plugin-core/src/utils/ExtensionUtils.ts
hikchoi: to avoid false positives in case a custom install path is provided, but after some thought this doesn't really mitigate that at all.

the custom install path case would be very rare so I'll just default to doing this

packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -439,6 +441,52 @@ export class ExtensionUtils {
...
+      });
+    }
+    try {
+      const extensionsDetail = allNonBuiltInExtensions.map((extension) => {
+        const { packageJSON } = extension;
+        const { id, version } = packageJSON;
+        return { id, version };
+      });
+      if (extensionsDetail && extensionsDetail.length > 0) {
+        _.set(trackProps, "extensionsDetail", JSON.stringify(extensionsDetail));
```
jonathanyeung, line 462: Why `JSON.stringify`? I think this property will need to be an array, else it's going to be difficult to analyze the data in amplitude.

packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -439,6 +441,52 @@ export class ExtensionUtils {
...
+        _.set(trackProps, "numExtensions", extensionsDetail.length);
+      }
+    } catch (error) {
+      // something went wrong don't track extension detail
```
kevinslin, line 466: can we log this using sentry?

packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -439,6 +441,52 @@ export class ExtensionUtils {
...
+      const fd = fs.openSync(installPath, "r");
+      const stat = fs.fstatSync(fd);
+      // some file systems don't track birth times.
+      // in this case the value may be ctime (time of inode change), or 0
+      const { birthtimeMs } = stat;
+      if (birthtimeMs) {
+        _.set(trackProps, "codeFolderCreated", birthtimeMs);
+      }
+    } catch (error) {
+      // something went wrong. don't track
```
kevinslin, line 487: same as above

kevinslin (approved):
looks good, just a final nitpick about reporting errors

packages/plugin-core/src/utils/ExtensionUtils.ts
hikchoi: to prevent some unfortunate column number blowout in postgres, but that's not going to be the case here since it's an array, and they will be stringified on Segment's end anyways.

I think Amplitude parses it as json on their end too, so the stringify here is unnecessary / excessive.

I'll change it.

PR #3459, opened 2022-08-27
fix: tree view init problem on web

jonathanyeung:
## fix: tree view init problem on web

There's an issue where the tree view on the web ext won't show any contents when you open up a repository for the first time. It'll show up if you focus on a note, and it seems that it'll remain in a good state on all subsequent times you open that repository.

I'm not sure what's the exact cause of the issue, but this change adds a mitigation: it leaves a message whenever the tree view is empty that tells the user to open a Dendron note.  

This also fixes a tangentially related issue where the tree view will focus on the active note when you toggle its visibility.

Credit to @hkievet for identifying the issue.


kevinslin (approved):

kevinslin (changes requested):
looks like the tests are failing

kevinslin (approved):

PR #3462, opened 2022-08-29
feat(workspace): preview for web extension

jonathanyeung:
## feat(workspace): preview for web extension

Adds a basic preview for the web extension. 

Features supported:
- basic rendering
- clicking on wikilinks for navigtion

Architectural Notes:
- The main difference is how notes get rendered. In the current local plugin, the extension sends NoteProps to the webview whenever an active note has changed. The webview app then uses the engine slice to call the engine API to `renderNote`. This isn't possible in the webview since the engine API doesn't exist. The methodology used by the web ext is to call `renderNote` on the plugin side, and then send the resulting HTML string to the webview to render.  This simplifies the flow quite a bit and makes for a more flexible design. I think we should adopt this flow in the local extension as well.
- I had to do some refactoring of the unified utils as they aren't currently engine v3 compatible, and the web ext requires a v3 engine. In the next iteration of changes, the unified code will be changed to all be v3 compatible, after which some of the duplicated logic can be consolidated again (such as `webUtils.ts`).


kevinslin (changes requested):
since we are not trying to rush this for conference, can we take some time to refactor the preview panel web to re-use the same logic and components we have from preview panel node?

packages/dendron-plugin-views/src/components/DendronApp.tsx
```diff
@@ -96,6 +97,21 @@ function DendronVSCodeApp({ Component }: { Component: DendronComponent }) {
...
+        ideDispatch(
+          ideSlice.actions.setPreviewHTML(updatePreviewMsg.data.html)
+        );
+        const noteToActivate = updatePreviewMsg.data.note;
+        ideDispatch(ideSlice.actions.setNoteActive(noteToActivate));
+        logger.info({
+          ctx,
+          msg: `onUpdatePreviewHTML:post`,
+        });
+        debugger;
```
kevinslin, line 113: remove

packages/dendron-plugin-views/src/components/DendronNotePreview.tsx
```diff
@@ -104,7 +84,15 @@ const DendronNotePreview: DendronComponent = (props) => {
     config,
   });
 
-  const [noteRenderedBody] = useRenderedNoteBody({ ...props, noteProps });
+  // const previewHTML = props.ide.previewHTML;
```
kevinslin, line 87: can we remove?

packages/plugin-core/src/web/engine/INoteRenderer.ts
```diff
@@ -0,0 +1,12 @@
+import {
+  RenderNoteOpts,
+  RespV2,
+  RenderNotePayload,
+} from "@dendronhq/common-all";
+
+/**
+ * Extracted from DEngine
+ */
+export interface INoteRenderer {
```
kevinslin, line 10: is this only applicable to the web extension? if so, should it be IWebNoteRenderer?

packages/unified/src/utilities/getNoteDependencies.ts
```diff
@@ -0,0 +1,82 @@
...
+} from "../types";
+
+/**
+ * For a given AST, find all note dependencies whose data will be needed for
+ * rendering.
+ * @param ast the syntax tree to look for dependencies
+ * @returns an array of fname-vault? combinations that this tree depends on.
+ */
+export function getNoteDependencies(ast: Node<Data>): DNodeCompositeKey[] {
+  const renderDependencies: DNodeCompositeKey[] = [];
```
kevinslin, line 19: do we need to dedup if we push the same keys?

packages/unified/src/utilsWeb.ts
```diff
@@ -0,0 +1,274 @@
...
+    opts?: { flavor?: ProcFlavor }
+  ) {
+    const proc = this._procRehype(
+      { mode: ProcMode.FULL, parseOnly: false, flavor: opts?.flavor },
+      data
+    );
+    return proc.use(rehypeStringify);
+  }
+
+  private static setProcData(proc: Processor, opts: Partial<ProcDataFullV5>) {
```
kevinslin, line 65: these methods look duplicate from utilsv5. can we use the utilsv5 methods where it makes sense?

jonathanyeung:
> can we take some time to refactor the preview panel web to re-use the same logic and components we have from preview panel node?

Which aspects specifically? 

My approach with this refactor is to iterate by first get the web version working, test it out a bit, then migrating the local version to use the web version as a base.

If we consolidate the code now, I'm worried about causing regressions in the current preview.  It'll also make this review a lot bigger.

packages/plugin-core/src/web/engine/INoteRenderer.ts
jonathanyeung: The idea for this is it to not be web-specific, but common.

packages/unified/src/utilsWeb.ts
jonathanyeung: Removed dupes.  I actually hope to delete this `utilsWeb.ts` class as soon as it's possible.

packages/unified/src/utilities/getNoteDependencies.ts
jonathanyeung: yes - I've actually done that in the next set of changes. I'll defer that change until the next PR to reduce conflicts if that's ok.

packages/plugin-core/src/web/engine/INoteRenderer.ts
kevinslin: > > can we take some time to refactor the preview panel web to re-use the same logic and components we have from preview panel node?
> 
> Which aspects specifically?
> 
> My approach with this refactor is to iterate by first get the web version working, test it out a bit, then migrating the local version to use the web version as a base.
> 
> If we consolidate the code now, I'm worried about causing regressions in the current preview. It'll also make this review a lot bigger.

got it, lets wait then

kevinslin (approved):

PR #3472, opened 2022-08-30
chore: remove enableSmartRefs config

Harshita-mindfire:
This PR aims to remove `enableSmartRefs` config from `workspace` scope.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
NA

## Docs

### Extended

### DIFF
--- a/packages/engine-test-utils/src/__tests__/engine-server/markdown/noteRefv2.spec.ts
+++ b/packages/engine-test-utils/src/__tests__/engine-server/markdown/noteRefv2.spec.ts
@@ -329,13 +329,13 @@ describe("noteRefV2", () => {
       );
     });
 
-    describe.skip("AND note has beginning section", () => {
+    describe("AND note has beginning section", () => {
       runTestCases(
         createProcCompileTests({
           name: "THEN parse beginning section",
           fname: "foo",
           setup: async (opts) => {
-            const text = "![[foo#header-1:^end]]";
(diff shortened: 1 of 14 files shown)


packages/engine-test-utils/src/__tests__/engine-server/markdown/noteRefv2.spec.ts
```diff
@@ -333,7 +329,7 @@ describe("noteRefV2", () => {
       );
     });
 
-    describe("AND note has beginning section", () => {
+    describe.skip("AND note has beginning section", () => {
```
Harshita-mindfire, line 332: Hey @kevinslin, need help with this test case. What is the significance of `^end`  in `![[foo#header-1^end]]` and how does it differ from `![[foo#header-1]]` when `enableSmartRefs` is false.

kevinslin: it grabs everything from `#header-1` to the very last line of the doc. it shoudl have the same behavior as `![[foo#header-1]]` when `enableSmartRefs` is false (this is so that users can re-create the original range using smart refs)

kevinslin (approved):

jonathanyeung (approved):

PR #3474, opened 2022-08-31
fix: giscuss integration

kevinslin:
fix: giscuss integration

Two fixes for the giscus integration:

- nextjs-template required a change in common-all that wasn't available, ended up bumping up the patch version of lerna to `0.110.1` and publishing the new common-all packages
- explicityl pass in the term property based on note id (otherwise giscuss would re-use the same path for all pages) - see https://github.com/dendronhq/dendron/pull/3474/files#diff-06608bb6c886eb23d44682f2302083810e4a26aceb141f69d9bd537c1054d46d

### DIFF
--- a/packages/nextjs-template/components/DendronNoteGiscusWidget.tsx
+++ b/packages/nextjs-template/components/DendronNoteGiscusWidget.tsx
@@ -1,7 +1,7 @@
 import {
   ConfigUtils,
   IntermediateDendronConfig,
-  NoteProps,
+  NoteProps
 } from "@dendronhq/common-all";
 import Giscus, { GiscusProps, Repo } from "@giscus/react";
 import _ from "lodash";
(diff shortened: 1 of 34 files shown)


packages/nextjs-template/components/DendronNoteGiscusWidget.tsx
```diff
@@ -32,8 +33,9 @@ export const DendronNoteGiscusWidget = ({
     const cleanGiscusConfig: GiscusProps = {
       ...giscusConfig,
       repo: repoString,
+      term: page,
```
hikchoi, line 36: I thought having `mapping` as pathname would be sufficient. So is the pathname the same at the point of this widget being loaded?

hikchoi: if that is the case we should remove the config `publishing.giscus.term`) as this does nothing at this point (or change this so that if `giscusConfig.term` exists, it uses that, and specify that if `term` is set in configs all giscus widgets will use the same term)

hikchoi (approved):

packages/nextjs-template/components/DendronNoteGiscusWidget.tsx
kevinslin: i thought so too but doesn't look like it works. and yes - can you remove that from the config?

PR #3480, opened 2022-08-31
feat(workspace): copy codespace url command

Harshita-mindfire:
This PR aims to create a new command `Copy Codespace URL` that would copy the github.dev URL to the clipboard.

Docs PR: https://github.com/dendronhq/dendron-site/pull/635

### Basics

### Extended
- General
- Operations
- Architecture

### Basics
- [!] if you are adding analytics related changes, make sure the [Telemetry](https://wiki.dendron.so/notes/84df871b-9442-42fd-b4c3-0024e35b5f3c.html) docs are updated

### Extended
- [!] can we track the performance of this change to know if it is _successful_?

### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA

## Docs

### Extended


packages/plugin-core/src/commands/CopyCodespaceURL.ts
```diff
@@ -0,0 +1,54 @@
...
+    const editor = vscode.window.activeTextEditor;
+
+    if (vscode.workspace.workspaceFolders && editor) {
+      const folder = vscode.workspace.getWorkspaceFolder(editor.document.uri);
+
+      if (!folder) {
+        return;
+      }
+
+      const root = (await GitUtils.getGitRoot(folder.uri.fsPath)) || "";
```
jonathanyeung, line 41: if it's not a git repo, can we have some error message to a user that this command requires a github backed repository to work?

packages/plugin-core/src/constants.ts
```diff
@@ -498,6 +498,11 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     title: `${CMD_PREFIX} Copy To Clipboard`,
     when: "false",
   },
+  COPY_CODESPACE_URL: {
+    key: "dendron.copyCodespaceURL",
+    title: `${CMD_PREFIX} Copy Codespace URL`,
+    when: `${DendronContext.PLUGIN_ACTIVE} && shellExecutionSupported`,
```
jonathanyeung, line 504: it'd be nice if we can add some where clause here to only show this command if the user has a vault / workspace setup that allows it.  Something for later.

jonathanyeung (changes requested):

packages/plugin-core/src/commands/CopyCodespaceURL.ts
Harshita-mindfire: If not a git repository, it is handled by [getGitRoot](https://github.com/dendronhq/dendron/blob/feat/copy-codespace-url-command/packages/common-server/src/git.ts#L323:L323) util method:

jonathanyeung (approved):
lgtm apart from merge conflict that crept in

kevinslin (approved):

Issue #3466, opened 2022-08-29
`Dendron: Migrate to self contained vault` does not migrate the vault correctly


ccamara:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [x] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

Whenever I add a remote, self-contained vault to my workspace I get an error such as the one below + the vault is not recognized by dentron (Lookup won't work + not displayed in dendron's tab).

```
{"fsPath":"cim-shared-vault","remote":{"type":"git","url":"git@<actual_git_repo>.git"},"name":"CIM Shared vault","selfContained":true}
```

The graph panel would show this:

```
{"name":"DendronError","status":"unknown","payload":"No Engine: <path>/Dendron/cim-dendron"}
```

## To Reproduce

1. Add a remote vault from existing dendron workspace
2. Convert the vault to self-contained

## Expected behavior

I would expect dendron to work with that vault, like it does with the other not-self contained.

## Desktop (please complete the following information)

- OS: macOS 12.5
- Browser (if with published websites; include version): [e.g. chrome, safari]
- Versions (all that apply to bug):
  - Dendron Extension: v0.109.1
  - VS Code version: 1.70.2
  - Dendron CLI:

## Dendron Log file

Please attach the [output of `> Dendron:Dev: Open Logs` here](https://cloud.montera34.org/index.php/s/2MNqFwEfZXWZ6Yy)

Harshita-mindfire:
Hey @ccamara , Thanks for reporting the bug.
Can you elaborate on the steps you proceeded? 
I tried to add this remote self contained vault : https://github.com/Harshita-mindfire/test-vault into my workspace and everything works fine. Is there a step I'm missing?

ccamara:
Sure!

I tested something simpler:

1. Open my dendron workspace
2. Create a new local vault: `Dendron: Vault Add\Local`
3. Generate a new vault
4. Create a new note
5. The vault is added properly and works well
[...]

PR #3483, opened 2022-09-02
fix(workspace): migrate self contained vaults command does not migrate the vault correctly 

Harshita-mindfire:
Related issue: https://github.com/dendronhq/dendron/issues/3466

When a regular vault is migrated to a self-contained vault using the `Migrate to Self Contained Vault` command, the folder structure of a self-contained vault is incorrect. The testcases were all passing because of missing await in expect(so expect condition always returns true for `verifyVaultHasMigrated`). Below is the buggy folder structure.
```
workspace
     |_______vaultx
                  |______ notes
                                | _____notes
                                |            |____assets folder
                                |            |____root.md
                                |            |_____root.schema.yml
                                |
                                |_____dendron.yml
                                |_____dendron.code.workspace
```

The below solution updates the wsRoot for vault migration from `wsRoot/vaultx/notes` to `wsRoot/vault`. This would prevent adding `dendron.yml` and `dendron.code-workspace` inside notes folder. 

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
NA

## Docs
NA

### Extended
NA

### DIFF
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -490,16 +490,15 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
       // Also add a gitignore, so files like `.dendron.port` are ignored if the
       // self contained vault is opened on its own
       await WorkspaceService.createGitIgnore(vaultPath);
-    } else {
-      // Update the config and code-workspace for the current workspace
-      if (addToConfig) {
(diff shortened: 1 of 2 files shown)


kevinslin:
nice catch!

PR #3484, opened 2022-09-02
chore(workspace): validate filename that are illegal in SQLite

hikchoi:
# chore(workspace): validate filename that are illegal in SQLite

This PR:
- Extends the file name validation logic to capture illegal SQLite characters.
- Enhance warning / error messages to be more specific about what is invalid.
- Add doctor action `fixInvalidFileNames` on cli and extension
- 

# Pull Request Checklist
### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/engine-server/src/doctor/service.ts
+++ b/packages/engine-server/src/doctor/service.ts
@@ -627,7 +627,6 @@ export class DoctorService implements Disposable {
           changes = await this.fixInvalidFileNames({
             canRename,
             engine,
-            quiet: opts.quiet,
           });
         }
 
@@ -756,9 +755,8 @@ export class DoctorService implements Disposable {
       resp: ValidateFnameResp;
     }[];
     engine: DEngineClient;
-    quiet?: boolean;
   }) {
-    const { canRename, engine, quiet } = opts;
+    const { canRename, engine } = opts;
     let changes: NoteChangeEntry[] = [];
     if (canRename.length > 0) {
       await asyncLoopOneAtATime(canRename, async (item) => {
--- a/packages/engine-server/src/doctor/service.ts
+++ b/packages/engine-server/src/doctor/service.ts
@@ -75,11 +75,15 @@ export type DoctorServiceOpts = {
 export class DoctorService implements Disposable {
   public L: ReturnType<typeof createDisposableLogger>["logger"];
   private loggerDispose: ReturnType<typeof createDisposableLogger>["dispose"];
+  private print: Function;
 
-  constructor() {
+  constructor(opts?: { printFunc?: Function }) {
     const { logger, dispose } = createDisposableLogger("DoctorService");
     this.L = logger;
     this.loggerDispose = dispose;
+    // if given a print function, use that.
+    // otherwise, no-op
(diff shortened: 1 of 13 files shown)


hikchoi:
docs: https://github.com/dendronhq/dendron-site/pull/639

packages/engine-server/src/doctor/service.ts
```diff
@@ -593,6 +597,59 @@ export class DoctorService implements Disposable {
...
+          notes,
+          noteDicts: {
+            notesById,
+            notesByFname,
+          },
+        });
+        resp = stats;
+
+        if (canRename.length > 0 && !opts.quiet) {
+          console.log("Found invalid filename in notes:\n");
```
kevinslin, line 611: ideally, we wouldn't be using `console.log` here but the [print](https://github.com/dendronhq/dendron/blob/master/packages/dendron-cli/src/commands/base.ts#L255:L255) function from the `CLICommand`. can we pass it in as an optional parameter in the constructor? by default, its a no-op.

packages/engine-server/src/doctor/service.ts
```diff
@@ -593,6 +597,59 @@ export class DoctorService implements Disposable {
...
+          canRename.forEach((item) => {
+            const { note, resp, cleanedFname } = item;
+            const { fname, vault } = note;
+            const vaultName = VaultUtils.getName(vault);
+            console.log(
+              `Note "${fname}" in ${vaultName} (reason: ${resp.reason})`
+            );
+            console.log(`  Can be automatically fixed to "${cleanedFname}"`);
```
kevinslin, line 619: extra space

packages/engine-server/src/doctor/service.ts
kevinslin: also apply to other `console.log` statements

packages/plugin-core/src/commands/Doctor.ts
```diff
@@ -334,10 +414,11 @@ export class DoctorCommand extends BasicCommand<CommandOpts, CommandOutput> {
     return engine;
   }
 
-  addAnalyticsPayload(opts: CommandOpts) {
+  addAnalyticsPayload(opts: CommandOpts, out: CommandOutput) {
     return {
       action: opts.action,
       scope: opts.scope,
+      ...out.extra,
```
kevinslin, line 421: we should pick what properties we are adding but not extract everything from an `any` parameter

kevinslin (changes requested):

jonathanyeung (approved):
lgtm apart from existing comments.

packages/engine-server/src/doctor/service.ts
hikchoi: it's an intentional indentation

kevinslin (approved):

PR #3488, opened 2022-09-04
chore(workspace): Clean up deprecated engine methods

tma66:
This PR cleans up deprecated engine methods and migrates clients to use newer methods.
This includes
- `engine.getLinks`. Replaced by `LinkUtils.findLinks`
- `engine.getAnchors`. Replaced by `AnchorUtils.findAnchors`
- `engine.updateNote`. Replaced by `engine.writeNote`
- `engine.updateSchema`. Replaced by `engine.writeSchema`
- - Began replacing few instances of `engine.notes` in plugin-core. Many more to come later

New changes include
- Added `engine.bulkGetNotes` and `engine.bulkGetNotesMeta`

Other janitor work includes:
- Create engine return types (if missing) with *Resp suffix and consolidate in same place. Currently there's a mix of Resp and Payload types. Combining the two for simplicity. Long term I'd like to move these into own file for better readability
- Update `api.ts` to use `RespV3` or `RespWithOptError` instead of `RespV2`
- Remove unused types/functions


packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -43,41 +48,75 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
     return resp.data ? resp.data : [];
   }
 
   /**
    * See {@link DEngine.bulkWriteNotes}
    */
-  async bulkWriteNotes(
-    opts: BulkWriteNotesOpts
-  ): Promise<Required<BulkResp<NoteChangeEntry[]>>> {
+  async bulkWriteNotes(opts: BulkWriteNotesOpts): Promise<BulkWriteNoteResp> {
```
jonathanyeung, line 110: super nit: for consistency, can we change the resp type to `BulkWritNote*s*Resp`

packages/common-all/src/utils/lookup.ts
```diff
@@ -65,6 +65,14 @@ export class NoteLookupUtils {
     return lastDotIndex < 0 ? "" : qs.slice(0, lastDotIndex + 1);
   };
 
+  static fetchRootResultsFromEngine = async (engine: DEngineClient) => {
```
jonathanyeung, line 68: Can we use `ReducedDEngine` instead of `DEngineClient` here and whenever possible? This makes it such that the web engine can also work with the util.

packages/plugin-core/src/web/commands/lookup/NoteLookupProvider.ts
```diff
@@ -45,7 +45,7 @@ export class NoteLookupProvider implements ILookupProvider {
     try {
       // if empty string, show all 1st level results
       if (transformedQuery.queryString === "") {
-        const items = this.fetchRootQuickPickResults({
+        const items = await this.fetchRootQuickPickResults({
```
jonathanyeung, line 48: this await is actually redundant since items is getting returned immediately. This can be condensed to `return this.fetchRootQuickPickResults()`

Separate note: can we replace this impl with `fetchRootResultsFromEngine` that you added in the lookup utils?

jonathanyeung (changes requested):
Looks great, just a few minor comments.

packages/common-all/src/utils/lookup.ts
tma66: Good point. I'll keep that in mind moving forward

jonathanyeung:
Looks like there's a failure in the plugin tests, but otherwise lgtm!

jonathanyeung (approved):

PR #3489, opened 2022-09-05
enhance/dynamic prisma loading

kevinslin:
enhance(workspace): dramatically decrease bundle size 

- reduce bundle size by only loading prisma dependencies when sqlite is enabled
- also change prisma from production to dev dependency (results in smaller `dendron-cli`)

- logic
	- if sqlite is enabled, we check if prisma binaries are in `$HOME/.dendron/`
		- if binaries are not here, we will fetch them from s3
		- subsequent downloads will use the binary from this folder

- other
	- we added `adm-zip` as a single minified js file in order to unzip the downloaded s3 binary. forked source code to generated the bundle [here](https://github.com/dendronhq/adm-zip/commit/433629e50353927f308d237037a04ed59c475d36)

- future work
	- when we update prisma binaries, we need someway of figuring out whether we need to download a new one
		- simple version: 
			- always download latest. use a checksum that is stored in code to check if we have latest binaries
		- if needed, we have have a more detailed versioning scheme
			- eg. encoding version in folder name (eg. `$HOME/.dendron/prisma_{version}`)


packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -491,7 +492,19 @@ export class FileStorage implements DStore {
     // instantiate so we can use singleton later
     if (this.config.workspace.metadataStore === "sqlite") {
       // eslint-disable-next-line no-new
-      new SQLiteMetadataStore({ wsRoot: this.wsRoot, force: true });
+      const store = new SQLiteMetadataStore({
+        wsRoot: this.wsRoot,
+        force: true,
+      });
+
+      // TODO: stop sleeping after max timeout
```
jonathanyeung, line 500: We should do this TODO now, else people will get stuck here.

jonathanyeung (changes requested):

packages/engine-server/src/drivers/file/storev2.ts
kevinslin: updated. not as worried here because this only gets invoked if users are opting in to using sqlite

kevinslin: this change also has some other updates required to make the automated publishing pipeline work

PR #3495, opened 2022-09-07
fix: tree view empty when note title is number

kevinslin:
fix: tree view empty when note title is number

Fix case where badly formatted note title can cause tree view to be blank


packages/plugin-core/src/workspace/workspaceActivator.ts
```diff
@@ -548,12 +548,18 @@ export class WorkspaceActivator {
...
-    });
+    // check if we have errors during init
+    try {
+      // NOTE: tracking is not awaited, don't block on this
+      ExtensionUtils.trackWorkspaceInit({
+        durationReloadWorkspace,
+        activatedSuccess: !!reloadSuccess,
+        ext,
+      });
+    } catch (error) {
```
jonathanyeung, line 559: This won't catch anything since `ExtensionUtils.trackWorkspaceInit` is async - this needs to go in the `catch` clause of the promise.

`ExtensionUtils.trackWorkspaceInit().catch(reason => captureEx(reason));`

packages/common-all/src/util/treeUtil.ts
```diff
@@ -163,6 +163,15 @@ export class TreeUtils {
       });
     }
 
+    const getNoteTitleOrError = (note: NoteProps) => {
+      const { error } = NoteUtils.validate(note);
+      if (!error) {
+        return note.title.toLowerCase();
+      } else {
+        return `BAD NOTE TITLE: ${note.fname}`;
```
jonathanyeung, line 171: Let's return something that will get sorted at the bottom, or undefined. This won't propagate to the user.

jonathanyeung (changes requested):

packages/plugin-core/src/workspace/workspaceActivator.ts
kevinslin: nice catch!

jonathanyeung (approved):

PR #3498, opened 2022-09-08
chore(workspace): First Pass. Clean up usages of engine.notes in plugin-core

tma66:
This is my first pass at migrating `engine.notes` to use the proper engine methods in plugin-core. I've left some of the more complicated ones for future commits.

**Other Changes**
1. Add support in api for `findNotes`
2. Add support in `findNotes` to exclude stubs
3. Replace `NoteProps` with `NotePropsMeta` where applicable

### DIFF
--- a/packages/plugin-core/src/commands/GoToSiblingCommand.ts
+++ b/packages/plugin-core/src/commands/GoToSiblingCommand.ts
@@ -4,8 +4,9 @@ import {
   DWorkspaceV2,
   isNumeric,
   NoteProps,
-  NotePropsByIdDict,
+  NotePropsMeta,
   NoteUtils,
+  ReducedDEngine,
   RespV3,
 } from "@dendronhq/common-all";
 import { DConfig } from "@dendronhq/common-server";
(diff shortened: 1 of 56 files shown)


packages/plugin-core/src/commands/GoToSiblingCommand.ts
```diff
@@ -150,37 +151,57 @@ export class GoToSiblingCommand extends BasicCommand<
...
-    notes: NotePropsByIdDict,
+  private getSiblingsForJournalNote = async (
+    engine: ReducedDEngine,
     currNote: NoteProps
-  ): NoteProps[] => {
-    const monthNote = notes[currNote.parent!];
-    const yearNote = notes[monthNote.parent!];
-    const parentNote = notes[yearNote.parent!];
+  ): Promise<NotePropsMeta[]> => {
+    const monthNote = await engine.getNoteMeta(currNote.parent!);
```
jonathanyeung, line 158: nit: instead of using '!', we can first check if(!currNote.parent) then return []. This also saves 1 call (potentially over the network if it's a remote backend.)

packages/plugin-core/src/commands/RandomNote.ts
```diff
@@ -57,10 +57,10 @@ export class RandomNoteCommand extends BasicCommand<
 
       return isMatch;
     };
+    const notesToPick = await engine.findNotesMeta({ excludeStub: true });
```
jonathanyeung, line 60: can you add a TODO / warning: comment above this that this may be quite expensive if the user has a lot of notes?

I don't think this feature gets much usage so let's defer further changes.

jonathanyeung (changes requested):
just a few small things, otherwise looks good. 

We'll have some conflicts though, depending on the merge order of the other pending refactor changes.

kevinslin (commented):
looks good, no further comments beyond existing ones

jonathanyeung (approved):

PR #3501, opened 2022-09-09
chore: add setVault to note trait template

hikchoi:
# chore: add setVault to note trait template
This PR:
- adds the new `setVault` onCreate method as an example to the note trait template


jonathanyeung:
Hmm, I kinda don't want to expose this until we have a further discussion on this part of the API (I know we already released it though).

hikchoi:
> Hmm, I kinda don't want to expose this until we have a further discussion on this part of the API (I know we already released it though).

Hmm... do we want to omit it from the release notes too then?

kevinslin (approved):

Issue #2200, opened 2022-01-18
Scratch keyboard shortcut conflicts with defaults


chmac:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings

## Describe the bug

The keybinding for creating a scratch note is cmd+shift+s which [is also the default for save-as](https://github.com/codebling/vs-code-default-keybindings/blob/d6873157d1a5076c9d347df28e479ddf77e74b46/macos.keybindings.json#L1012).

This means the keybinding doesn't work on dendron install (at least on macOS).

SeriousBug:
Thanks for adding the report.

@hikchoi Would it be possible to add a migration for keyboard shortcuts where we change the scratch note shortcut if the user:
- hasn't remapped it to another shortcut already
- hasn't remapped the default "save as" shortcut already
If the user has done neither, then this shortcut is basically unusable and I think it would be safe to remap the shortcut then. If that's hard/impossible, we can also just document this issue in the wiki.

hikchoi:
The default keybinding for `Create Scratch Note` has been changed to avoid conflict with `Save as`.

Please see the [changelog](https://wiki.dendron.so/notes/50kdbcwwda3gphjhccb0e5t/#2022-09-13) for more.
Let us know if you have have any pending concerns regarding this issue.

PR #3502, opened 2022-09-09
chore(workspace): Change default keybinding for Create Scratch Note

hikchoi:
# chore(workspace): Change default keybinding for Create Scratch Note
This PR:
- Changes default keybinding for `Create Scratch Note` from `shift+cmd/ctrl+s` to `cmd/ctrl+k s`
- Add a temporary feature toast during startup to notify the change, and a fail-safe backup toast in the command itself
  - This is only shown once ever, and will be tracked with `FeatureShowcaseDisplayed`

# Pull Request Checklist
### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [~] Common cases tested
- [~] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/_extension.ts
+++ b/packages/plugin-core/src/_extension.ts
@@ -71,6 +71,7 @@ import { DendronExtension, getDWorkspace, getExtension } from "./workspace";
 import { TutorialInitializer } from "./workspace/tutorialInitializer";
 import { WorkspaceActivator } from "./workspace/workspaceActivator";
 import { WSUtils } from "./WSUtils";
+import { CreateScratchNoteKeybindingTip } from "./showcase/CreateScratchNoteKeybindingTip";
 
 const MARKDOWN_WORD_PATTERN = new RegExp("([\\w\\.\\#]+)");
 // === Main
@@ -353,7 +354,14 @@ export async function _activate(
         const ONE_MINUTE_IN_MS = 60_000;
         setTimeout(() => {
           const showcase = new FeatureShowcaseToaster();
-          showcase.showToast();
+          // Temporarily show the new toast instead of the rest.
+          // for subsequent sessions this will not be shown as it already has been shown.
+          const hasShown = showcase.showSpecificToast(
+            new CreateScratchNoteKeybindingTip()
+          );
+          if (!hasShown) {
+            showcase.showToast();
+          }
         }, ONE_MINUTE_IN_MS);
       }
     } else {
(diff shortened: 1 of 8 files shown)


hikchoi:
docs: https://github.com/dendronhq/dendron-site/pull/642

hikchoi:
related: https://github.com/dendronhq/dendron/issues/2200

jonathanyeung (approved):

packages/plugin-core/src/showcase/CreateScratchNoteKeybindingTip.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    return true;
+  }
+
+  get showcaseEntry(): ShowcaseEntry {
+    return ShowcaseEntry.CreateScratchNoteKeybindingTip;
+  }
+
+  getDisplayMessage(_displayLocation: DisplayLocation): string {
+    return `Keyboard shortcut for "Dendron: Create Scratch Note" has changed. If you wish to keep the original, please follow the instructions.`;
```
kevinslin, line 28: nitpick: would change the wording to

> If you wish to keep the original, please follow the instructions online

packages/plugin-core/src/_extension.ts
```diff
@@ -353,7 +354,14 @@ export async function _activate(
         const ONE_MINUTE_IN_MS = 60_000;
         setTimeout(() => {
           const showcase = new FeatureShowcaseToaster();
-          showcase.showToast();
+          // Temporarily show the new toast instead of the rest.
+          // for subsequent sessions this will not be shown as it already has been shown.
+          const hasShown = showcase.showSpecificToast(
```
kevinslin, line 359: can we add to the show criteria that we only show this to users who did not just install vscode?

so as part of this change, add a `firstInstallVersion` as part of the install? if `firstInstallVersion` is not undefined or less than `0.112.0`, then show this message

kevinslin (changes requested):

packages/plugin-core/src/_extension.ts
hikchoi: > if firstInstallVersion is not undefined or less than 0.112.0, then show this message

Do you mean if `firstInstallVersion` is _undefined_ or less than 0.112.0? unless we go and backfill them based on the `firstInstall` time one by one, all existing users will have them undefined.

kevinslin: the former

kevinslin (approved):

PR #3505, opened 2022-09-10
chore: first pass at adding auto-generated docs to dendron code base

kevinslin:
chore: first pass at adding auto-generated docs to dendron code base

Usage instructions here: [[dendron://dendron.docs/dev.process.docs.typedoc]]


hikchoi (commented):
Didn't try pushing it to S3 yet, but I managed to pull / build successfully.

I assume all the warnings about unresolved links during build is going to be fixed next phase?

This is great! I can imagine having this run in CI and adopting some practice to have some process for documenting methods and splitting up methods so that they make sense in the docs as well.

packages/common-all/src/types/typesv2.ts
```diff
@@ -106,7 +106,10 @@ export type DNoteLink<TData extends DNoteLinkData = DNoteLinkData> = {
   data: TData;
 };
 
-export type DNoteLinkRaw<TData = any> = Omit<DNoteLink<TData>, "from"> & {
+export type DNoteLinkRaw<TData extends DNoteLinkData = any> = Omit<
```
jonathanyeung, line 109: nit: revert from this particular change list?

jonathanyeung (approved):

PR #3512, opened 2022-09-12
fix(workspace): initial value incorrect in lookup for webext(win)

Harshita-mindfire:
This PR aims to fix the lookbar bar initial value for codespaces(windows)
Bug:
image


kevinslin (approved):

jonathanyeung (approved):
Good catch - thanks for fixing.

PR #3519, opened 2022-09-14
chore(internal): track config diff during startup

hikchoi:
# chore(internal): track config diff during startup
This PR:
- Adds new properties `dendronConfigChanged` and and `changedDendronConfigs`, which tracks configs that deviated from the default configuration defined

# Pull Request Checklist
### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/utils/index.ts
+++ b/packages/common-all/src/utils/index.ts
@@ -1300,9 +1300,10 @@ export class ConfigUtils {
    */
   static flattenConfigObject(opts: { obj: Object; omitPaths?: string[] }) {
     const { obj, omitPaths } = opts;
+    const objDeepCopy = _.cloneDeep(obj);
     if (omitPaths && omitPaths.length > 0) {
       omitPaths.forEach((path) => {
-        _.unset(obj, path);
+        _.unset(objDeepCopy, path);
       });
     }
 
@@ -1334,7 +1335,7 @@ export class ConfigUtils {
         }
       });
     };
-    flattenToPathValuePairs({ obj });
+    flattenToPathValuePairs({ obj: objDeepCopy });
     return accumulator;
   }
 
--- a/packages/common-all/src/utils/index.ts
+++ b/packages/common-all/src/utils/index.ts
@@ -1286,6 +1286,93 @@ export class ConfigUtils {
   static getConfigDescription = (conf: string) => {
     return _.get(DENDRON_CONFIG, conf)?.desc;
   };
+
+  /**
+   * Given an config object and an optional array of lodash property path,
+   * omit the properties from the object and flatten it
+   * The result will be a flat array of path-value pairs
+   *
+   * Each pair will contain a path and a value.
+   * The value is either a primitive value, or a stringified array.
+   *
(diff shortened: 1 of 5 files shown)


packages/engine-test-utils/src/__tests__/engine-server/configUtils.spec.ts
```diff
@@ -9,6 +9,64 @@ function getConfig(
...
+      });
+    });
+    describe("WHEN v4 config is given", () => {
+      test("THEN return empty list", () => {
+        const config = ConfigUtils.genDefaultV4Config();
+        const output = ConfigUtils.findDifference({ config });
+        expect(output.length).toEqual(0);
+      });
+    });
+    describe("WHEN changed config is given", () => {
```
kevinslin, line 27: can we also add a test where the config only differes by vaults? we don't want to emit a diff in that case

packages/common-all/src/utils/index.ts
```diff
@@ -1286,6 +1286,93 @@ export class ConfigUtils {
...
+   * The value is either a primitive value, or a stringified array.
+   *
+   * If comparing the array value of a config is unnecessary,
+   * make sure to add it to the omit path.
+   */
+  static flattenConfigObject(opts: { obj: Object; omitPaths?: string[] }) {
+    const { obj, omitPaths } = opts;
+    if (omitPaths && omitPaths.length > 0) {
+      omitPaths.forEach((path) => {
+        _.unset(obj, path);
```
kevinslin, line 1305: isn't this mutating the original config? can we make a copy?

kevinslin (changes requested):

packages/common-all/src/utils/index.ts
hikchoi: right. good catch!

hikchoi:
update:

- `InitializeWorkspace` is getting a little crowded, so adding a new event `ConfigEvents.ConfigChangeDetected`, which will track the actual diffs
- `InitializeWorkspace` will have `dendronConfigChanged` and `numConfigChanged` for ease of analysis.

`ConfigChangeDetected`'s payload will be a single stringified JSON:

e.g.)
```js
properties: {
    appVersion: '0.112.0',
    arch: 'arm64',
    changed: '[{"path":"workspace.journal.dailyDomain","value":"dailyz"},{"path":"workspace.task.taskCompleteStatus","value":"[\"x\",\"done\"]"},{"path":"workspace.enableFullHierarchyNoteTitle","value":true},{"path":"preview.enableNoteTitleForLink","value":false},{"path":"preview.enableHashesForFMTags","value":true}]',
    firstWeekSinceInstall: false,
    ideFlavor: 'Visual Studio Code',
    ideVersion: '1.70.2',
    nodeVersion: 'v16.13.2',
    userAgent: 'vscode',
    vscodeSessionId: redacted
  }
```

PR #3522, opened 2022-09-14
chore: telemetry in web ext

jonathanyeung:
## chore: telemetry in web ext

Adds telemetry to the web extension. 

### Details
- Provides scaffolding for telemetry in web extension and DI based telemetry client injection, which we can gradually converge with the local extension.
- Used Segment's HTTP API instead of their browser library because the browser library requires 'window', which is unavailable in the web ext sandbox. The HTTP API is so simple that it might be worth eventually tossing out the node implementation and converging to just use the HTTP version. It's simple enough that a library isn't really required.
- Telemetry is safely separated from ide-prod telemetry in the following ways:
  - uploads to a different segment source called 'dendron_web_extension': https://app.segment.com/dendron/sources/dendron_web_extension/debugger.  This hasn't been hooked up to amplitude yet.
  -  adds a new identify prop called 'appHost' from vscode.env. This has values like 'desktop', 'codespaces', 'github.dev': See https://code.visualstudio.com/api/references/vscode-api#env
  - The web ext reports under a new type `AppNames.CODE_WEB`
- Telemetry added in web ext: an analogous `InitializeWorkspace` event, + events for the 3 commands.
- I'm attempting to link a local user to a web ext instance by the `vscode.context.globalState` mechanism. We can test in nightly to see if it syncs properly.

After we do some analysis and scope out how the telem looks, we can make decisions on whether it makes sense to consolidate to 'ide-prod' or not.

TODO: I'll update our airtable telem after an initial pass at the review.


packages/plugin-core/src/_extension.ts
```diff
@@ -166,6 +167,14 @@ export async function _activate(
       sessionId: AnalyticsUtils.getSessionId(),
       release: AnalyticsUtils.getVSCodeSentryRelease(),
     });
+
+    // Temp: store the user's anonymous ID into global state so that we can link
+    // local ext users to web ext users:
+    context.globalState.setKeysForSync([GLOBAL_STATE_KEYS.ANONYMOUS_ID]);
```
kevinslin, line 173: we have a state service to set state > [[../packages/plugin-core/src/services/stateService.ts]]
can we undeprecate it and put it in there instead?

packages/plugin-core/src/telemetry/common/VersionProvider.ts
```diff
@@ -0,0 +1,15 @@
+import * as vscode from "vscode";
+
+export class VersionProvider {
```
kevinslin, line 3: can we make use of [[../packages/plugin-core/src/versionProvider.ts]] or refactor it?

packages/plugin-core/src/telemetry/web/WebTelemetryClient.ts
```diff
@@ -0,0 +1,99 @@
...
+  constructor(
+    @inject("anonymousId") private anonymousId: string,
+    @inject("extVersion") private extVersion: string
+  ) {}
+
+  /**
+   * This key talks to the 'Dendron-Web-Extension' source in Segment. NOTE: this
+   * is different from the 'ide-prod' source.
+   */
+  private DENDRON_WEB_EXTENSION_SEGMENT_WRITE_KEY =
```
kevinslin, line 25: can we obfuscate this like we do currently with the regular segment key?

packages/plugin-core/src/_extension.ts
kevinslin: we have a state service to set state > [[../packages/plugin-core/src/services/stateService.ts]]
can we undeprecate & implement the logic from there?

packages/plugin-core/src/web/extension.ts
```diff
@@ -95,3 +98,12 @@ async function setupTreeView(context: vscode.ExtensionContext) {
...
   });
 }
+
+async function reportActivationTelemetry() {
+  const telemetryClient =
+    container.resolve<ITelemetryClient>("ITelemetryClient");
+
+  await telemetryClient.identify();
+  // TODO: Add workspace properties later.
+  await telemetryClient.track(VSCodeEvents.InitializeWorkspace);
```
kevinslin, line 108: do we know that the user has initialized a workspace? if they have dendron installed and open up a regular code project, would this not also trigger?

kevinslin (changes requested):

packages/plugin-core/src/telemetry/web/WebTelemetryClient.ts
jonathanyeung: How is this obfuscated? Aren't they just here in plaintext? https://github.com/dendronhq/dendron/blob/766ba6db92e5b17cd6e9d24b478e5249e9acf3c4/packages/common-all/src/config.ts#L24

packages/plugin-core/src/telemetry/web/getAnonymousId.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  if (resp.error) {
+    throw resp.error;
+  }
+
+  if (resp.data) {
+    return resp.data;
+  } else {
+    const newId = genUUID();
+    await storage.write(GLOBAL_STATE_KEYS.ANONYMOUS_ID, newId);
```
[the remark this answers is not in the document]
jonathanyeung: we will get some divergence if a user opens dendron on web before opening dendron locally with the released version.  In that case, the web version will report a different anonymous id compared to the local for that particular user.  I don't think it's a big deal though.

packages/plugin-core/src/_extension.ts
jonathanyeung: I'd rather not use State Service because it's a singleton. I think having a enum of key values is enough.

packages/plugin-core/src/telemetry/common/VersionProvider.ts
jonathanyeung: This class isn't necessary, I'll remove it and just inline the code 

This is the better methodology to get version though. It's a bit too much work to refactor right now since the existing Version Provider is static - I'm going to mark it as deprecated.

packages/plugin-core/src/web/extension.ts
jonathanyeung: This will only report if they're inside a Dendron workspace, but it's more an artifact of some incorrect logic in the initialization flow.  I'll make a TODO note to fix the initialization bug.

packages/plugin-core/src/utils/common/getStageFromPkgJson.ts
```diff
@@ -0,0 +1,17 @@
+import { Stage } from "@dendronhq/common-all";
+
+/**
+ * Extrapolates the stage from the 'name' property in the package.json manifest.
+ * For node environments, getStage() is sufficient, but in environments that
+ * aren't node based (where process.env is unavailable), this method suffices.
+ * NOTE: This method does not return 'test'
+ */
+export function getStageFromPkgJson(packageJSON: any): Stage {
```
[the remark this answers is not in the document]
jonathanyeung: Agree that webpack is probably the more correct thing to do, but it's also a lot of additional work + difficult testing.  We would need a separate webpack config for development, nightly, and production + modify all the build scripts accordingly.  

I also don't think examining the package manifest is the _worst_ methodology - can we defer webpack for now?  I'll convert this to a private method within web logic so it's not used elsewhere.

jonathanyeung:
I need to rework how some of the injection works - the `test-plugin-web` pass is now failing.

packages/plugin-core/src/telemetry/web/getAnonymousId.ts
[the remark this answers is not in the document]
kevinslin: can we have some logic locally?

if we open locally and there is an id that is different from the local, can we track this as an event?
```ts
track(MultipleDeviceIds, ids: [...])
```

we can then use identify to merge the identities

kevinslin (changes requested):

packages/plugin-core/src/telemetry/common/VersionProvider.ts
jonathanyeung: I'm not quite following - this methodology requires extension context - the places that invoke the current static `VersionProvider` don't have access to extension context, so it's not easy to refactor that without more changes.

packages/plugin-core/src/utils/common/getStageFromPkgJson.ts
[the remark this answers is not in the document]
jonathanyeung: So `process.env.STAGE` will be set correctly whenever webpack runs at compile time? Can you point me to where we set that environment variable?

packages/plugin-core/src/web/extension.ts
jonathanyeung: The thing we discussed a little while ago where `activate` doesn't get run again when a new folder gets opened in `vscode.dev`, which prevents us from initializing properly in that scenario.  So right now, Dendron web only really works in the `github.dev` hitting `.` flow.

kevinslin: got it. can we add the TODO in the code?

jonathanyeung: I added it in the latest commit (line 29 on this file)

packages/plugin-core/src/utils/common/getStageFromPkgJson.ts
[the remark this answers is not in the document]
kevinslin: it will be if we set it. we currently don't but webpack automatically sets NODE_ENV. https://webpack.js.org/configuration/mode/

we also inspect `NODE_ENV` in src/env.ts for when calling `getStage`

packages/plugin-core/src/telemetry/common/VersionProvider.ts
kevinslin: yeah, it will take some refactoring work. we don't have to do all of it but can we make some progress (eg. refactor one of the methods)?

jonathanyeung: I'm going to defer for now, as it's getting out of scope of this change - all 4 references to `VersionProvider.version()` are inside static methods themselves.  Any reworking of those will create circular dependencies.

packages/plugin-core/src/utils/common/getStageFromPkgJson.ts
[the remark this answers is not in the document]
jonathanyeung: as discussed offline, going to try `getStage`

jonathanyeung:
> I need to rework how some of the injection works - the `test-plugin-web` pass is now failing.

resolved now

kevinslin:
was this a race condition or a legitimate failure? https://github.com/dendronhq/dendron/actions/runs/3067223812/jobs/4953278963

PR #3523, opened 2022-09-14
fix: click into "empty area" on sidebar submenuitem

namjul:
This PR aim is to make clicks into the "empty area" of sidebar submenuitems not do nothing and rather to something.
The something could be expanding the submenuitem or opening the page. I decided for expanding since that is the default behavior of the antd Menu component. I think it makes sense to expect/intiut that the user wants to expand the submenu when he has not clicked directly on the link text.

Context: [[dendron://private/task.2022.08.21.nav-bar-should-navigate-when-menu-is-clicked]]

Demostration of the issue: https://www.loom.com/share/4998f4d983e94ee8b2ecb55a3550a6ec
Demostration of the fix: https://www.loom.com/share/266547466558423ba20b9b775cbcf405

# Pull Request Checklist

### DIFF
--- a/packages/nextjs-template/components/DendronTreeMenu.tsx
+++ b/packages/nextjs-template/components/DendronTreeMenu.tsx
@@ -177,7 +177,7 @@ function MenuView({
           title={
             <MenuItemTitle
               menu={menu}
-              noteIndex={noteIndex!}
+              noteIndex={noteIndex}
               onSubMenuSelect={onSubMenuSelect}
             />
           }
@@ -200,7 +200,7 @@ function MenuView({
       <MenuItem key={menu.key} icon={menu.icon}>
         <MenuItemTitle
           menu={menu}
-          noteIndex={noteIndex!}
+          noteIndex={noteIndex}
           onSubMenuSelect={onSubMenuSelect}
         />
       </MenuItem>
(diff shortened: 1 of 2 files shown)


packages/nextjs-template/components/DendronTreeMenu.tsx
```diff
@@ -232,8 +245,16 @@ function MenuItemTitle(props: Partial<NoteData> & { menu: DataNode }) {
         href={getNoteUrl(props.menu.key as string, {
           noteIndex: props.noteIndex!,
         })}
+        passHref
       >
-        {props.menu.title}
+        <a
+          href="dummy"
```
namjul, line 251: {https://github.com/vercel/next.js/discussions/32233#discussioncomment-1766768

kevinslin (approved):

packages/nextjs-template/components/DendronTreeMenu.tsx
```diff
@@ -232,8 +246,18 @@ function MenuItemTitle(props: Partial<NoteData> & { menu: DataNode }) {
         href={getNoteUrl(props.menu.key as string, {
           noteIndex: props.noteIndex!,
```
hikchoi, line 247: I think maybe `getNoteUrl` should allow for `props.noteIndex` to be undefined and handle it appropriately instead of us having to pass it along from the top and doing a non-null assertion each time.

hikchoi (changes requested):
a small nitpick, but otherwise looks good! I was confused with this behavior in the past personally.

packages/nextjs-template/components/DendronTreeMenu.tsx
namjul: yes, good thing 4d64edb3e

hikchoi (approved):

PR #3525, opened 2022-09-14
fix(publish): display headings pleasantly when containing non-textonly content types 

namjul:
- will fix
  - [[dendron://incubator/task.2022.08.07.code-in-heading-breaks-the-formatting]]
  - [[dendron://private/task.2022.08.29.revisit-headings-whitespace]]

# Pull Request Checklist


kevinslin:
for the ui changes, can we also add playwright tests when this is ready?

kevinslin:
looks like snapshot tests are failing

hikchoi (changes requested):
Just tested it out locally and fixes the issue!

Looks good other than the snapshot fails

kevinslin:
looks like playwright test are failing?

namjul:
> looks like playwright test are failing?

The error https://github.com/dendronhq/dendron/actions/runs/3089152419/jobs/4996433156#step:10:32 gets resolved when I revert https://github.com/dendronhq/dendron/commit/27662dc9de#diff-48c116cf8058e55ab46b8ace6daf7fad4141f3f1e1d1c089b8458ec0135547a3L52-R56. The actuall error happens in `DendronTOC` and possibly conflicts by having antd using 17.x `@types/react` and `DendonTOC` 18.x.

What was the reason we upgraded `@types/react` ?

hikchoi (changes requested):
looks like either we have to revert the `@types/react` version bump, or bump up antd to at least `4.20`, since that is when they fixed the compatibility issue with `@types/react` 18

(antd commit that fixed it I _think_: https://github.com/ant-design/ant-design/commit/5b8e4e51e95c91efe4d57aa08b57fa6401d54111)

Do we have a reason to stick with the current version of antd?

kevinslin:
we can either upgrade antd or downgrade types/react in nextjs. either option works, we might run into subtle differences in rendering if upgrading antd

hikchoi:
There are quite a few minor releases we are jumping up, and I see a lot of deprecations in their releases. Bumping antd up in this PR might be a bit of a scope creep as we don't know what will break yet. 

One vote from me for downgrading the types and revisiting it later in a separate task 🙆🏼

namjul:
@kevinslin @hikchoi I decided to adress this in a seperate PR https://github.com/dendronhq/dendron/pull/3561.

hikchoi (approved):

kevinslin (approved):

PR #3528, opened 2022-09-14
chore: ab test for video in launch tutorial welcome screen

Harshita-mindfire:
This PR aims to add an AB test for 1 min feature highlights video in the Welcome screen displayed after running `Launch Tutorial` command. 

- used `webm` format for video, with `.mp4` there was no sound.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/assets/dendron-ws/vault/welcome_video.html
+++ b/packages/plugin-core/assets/dendron-ws/vault/welcome_video.html
@@ -1,0 +1,128 @@
+<!DOCTYPE html>
+<html lang="en">
+  <style>
+    * {
+      font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+    }
+    h1 {
+      text-align: center;
+      font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+      margin-bottom: 0px;
+      margin-top: 0px;
+      padding-right: 20px;
+    }
+    .container{
+      text-align: center;
+    }
+    .subtitle {
+      margin-top: 12px;
+    }
+    #btn_container {
+      padding-top: 20px;
+      padding-bottom: 50px;
+    }
+    .initialize_btn {
+      border-radius: 2px;
+      background-color: rgb(0, 168, 36);
+      border: none;
+      color: #FFFFFF;
+      text-align: center;
+      font-size: 22px;
+      padding: 8px;
+      width: 220px;
+      transition: all 0.4s;
+      cursor: pointer;
+      margin: 25px;
+    }
+    .initialize_btn span {
+      cursor: pointer;
+      display: inline-block;
+      position: relative;
+      transition: 0.4s;
+    }
+    .initialize_btn span:after {
+      content: '\00bb';
+      position: absolute;
+      opacity: 0;
+      top: 0;
+      right: -20px;
+      transition: 0.5s;
+    }
+    .initialize_btn:hover {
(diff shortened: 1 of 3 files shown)


jonathanyeung:
As discussed offline, if we can't figure out a way to embed, I'd rather us not run this experiment rather than include the file as part of our extension binary.  I think that the binary size will adversely affect download rates and it's also not going to publish to ovsx as it stands (32 mb single file size limit).  @kevinslin - thoughts?

Harshita-mindfire:
one possible workaround is uploading the video to google drive. This works fine in preview with a valid id
```html
<video preload="true" controls autoplay>
  <source
    src="https://drive.google.com/uc?export=download&amp;id=<insert-video-drive-id-here>"
    type="video/webm"
  />
</video>
```

packages/plugin-core/assets/dendron-ws/vault/welcome_video.html
```diff
@@ -0,0 +1,128 @@
...
+  </div>
+    <br>
+    <div id="btn_container" class="container">
+      <a class="initialize_btn" id="btn_id" target="_blank" rel="nofollow"><span>Get Started</a>
+    </div>
+    
+    <div id="gif-ctn">
+      <video controls autoplay>
+        <source
+        src="https://org-dendron-public-assets.s3.amazonaws.com/images/dendron_feature_highlights.webm"
```
kevinslin, line 108: can we use the cloudfront link instead? will load faster > https://d2q204iup008xl.cloudfront.net/images/dendron_feature_highlights.webm

packages/plugin-core/src/WelcomeUtils.ts
```diff
@@ -21,14 +23,33 @@ async function initWorkspace() {
...
-      "dendron-ws",
-      "vault",
-      "welcome.html"
+    let content: string;
+    const ABUserGroup = _2022_09_VIDEO_TUTORIAL_TEST.getUserGroup(
+      SegmentClient.instance().anonymousId
     );
-    const { content } = readMD(uri.fsPath);
+    let testgroup: string;
+    const version = Number(vscode.version.substring(0, 4)); // version: x.xx.x
```
kevinslin, line 31: can we use the `semver` package to check for versions?

packages/plugin-core/assets/dendron-ws/vault/welcome_video.html
```diff
@@ -0,0 +1,128 @@
...
+    <p class="subtitle">Note, organize, and reference any amount of information.</p>
+  </div>
+    <br>
+    <div id="btn_container" class="container">
+      <a class="initialize_btn" id="btn_id" target="_blank" rel="nofollow"><span>Get Started</a>
+    </div>
+    
+    <div id="gif-ctn">
+      <video controls autoplay>
+        <source
```
kevinslin, line 107: can we make this take the full width of the screen?

kevinslin (changes requested):
video doesn't seem to load or play for me > https://www.loom.com/share/7b43a4b575e9431c82b686f3cb8a7e93

Harshita-mindfire:
> video doesn't seem to load or play for me > 

It looks like it is still downloading. Do you have any errors in console/network?

kevinslin (approved):

PR #3531, opened 2022-09-16
chore(internal): remove enableMermaid config

Harshita-mindfire:
This PR aims to remove the enableMermaid config
Docs Update PR: https://github.com/dendronhq/dendron-site/pull/651


hikchoi (approved):
Looks good!

small nit: can we remove the TODO in https://github.com/dendronhq/dendron/blob/chore/remove-enable-mermaid-config/packages/unified/src/utilsWeb.ts#L163:L163 so we don't accidentally add it back again?

jonathanyeung (approved):

PR #3533, opened 2022-09-16
updates to nextjs

kevinslin:


packages/nextjs-template/components/DendronBreadCrumb.tsx
```diff
@@ -32,10 +32,12 @@ export function DendronBreadCrumb(props: DendronCommonProps) {
   const noteParents = noteIdPareents.map((noteId) => props.notes[noteId]);
 
   return (
+    // @ts-ignore
```
jonathanyeung, line 35: by adding these, can we make the ts compiler stricter now in nextjs-template?

jonathanyeung (approved):

namjul (approved):

packages/nextjs-template/components/DendronBreadCrumb.tsx
kevinslin: we should - will take a pass at it

Issue #1207, opened 2021-08-25
Rename note command doesn't update wikilinks to the renamed note if they are inside the renamed note


ryan-p-randall:
## Please select the area your Bug applys to. (Multiple selections are Possible)

- [ ] Onboard. Initial vault setup and import/export pods 
- [ ] Create. Note creation, lookup, snippets and templates
- [x] Retrieve. Backlinks, references, graph view
- [x] Structure. Refactoring, multi-vault and schemas
- [ ] Publish. Sharing your repo with the world
- [ ] Misc (Choose this if your not sure)

## Describe the bug

The `rename note` command does not update references to a note if those notes are inside the note being renamed. (This is an odd edge case!)  

Let's say I have `note1` and `note2`, and both notes contain the wikilink `[[note1]]`. When I use the `rename note` command  to change `note1`'s file name to `note01`, the wikilink updates to `[[note01]]` inside `note2`. But it does not update inside `note1`, which is now named `note01`.  

## To Reproduce

1. Open VS Code Command Palette.
2. Use`rename note` command to rename note that references itself with a wikilink.
3. Look at the wikilink in that renamed note.
4. See that wikilink still refers to old note name. 

## Expected behavior

Wikilinks should be updated, even if they are inside the renamed note.

## Screenshots

Wikilink `[[note1]]` does not update when using `rename note` command to rename `note1` to `note01`:
<img width="491" alt="shows error" src="https://user-images.githubusercontent.com/850915/130738183-0e9a432e-a3de-4759-89ac-ed2320c41088.png">

Wikilink successfully updates in `note2`:
<img width="491" alt="shows error" src="https://user-images.githubusercontent.com/850915/130738561-ed3177b3-2961-424d-a37b-101ffc841af5.png">

## Desktop (please complete the following information)

- OS: Mac
- Editor: VS Code
- Dendron Version: 0.56.0 

## Dendron Log file

Please attach the output of `>Dev:Dendron: Open Logs` here

doscripting:
glad to see this issue has been tracked...  awesome!

aleksey-rowan:
This issue is featured in [this week's round](https://discord.com/channels/717965437182410783/739186036495876126/1017139140040593408) of [CROP](https://wiki.dendron.so/notes/tayowNP4tMvcqH7AraoQ4/).

aleksey-rowan:
The community has selected this issue as this week's CROP winner. 🎉

PR #3535, opened 2022-09-16
fix(workspace): Update backlinks after engine updates

tma66:
**Background**
Currently we do not update backlinks when a note is created/updated/deleted. This pr fills in that gap by computing the diff of links between the old and new note. This will be implemented for both the engine v3 and v2.

**Changes**
1. For a newly created note, scan through its links and create backlink if applicable
2. For a deleted note,  scan through its links and delete backlink if applicable
3. For updated note, calculate diff and add/remove backlinks
4. Created new `BacklinkUtils` to handle basic backlink methods. `LinkUtils` is intertwined with many other services so I wanted to keep it separate

### DIFF
--- a/packages/common-all/src/BacklinkUtils.ts
+++ b/packages/common-all/src/BacklinkUtils.ts
@@ -1,0 +1,59 @@
+import _ from "lodash";
+import { DLink, NotePropsMeta } from "./types";
+
+export class BacklinkUtils {
+  /**
+   * Create backlink out of link if it references another note (denoted by presence of link.to field)
+   *
+   * @param link Original link to create backlink out of
+   * @returns backlink or none if not applicable
+   */
+  static createFromDLink(
+    link: DLink
+  ): (Omit<DLink, "type"> & { type: "backlink" }) | undefined {
+    const maybeToNoteFname = link.to?.fname;
+    if (maybeToNoteFname) {
+      return {
+        from: link.from,
+        type: "backlink",
+        position: link.position,
+        value: link.value,
+      };
+    }
+    return;
+  }
+
+  /** Adds a backlink by mutating the 'note' argument in place.
+   *
+   *  @param note note that the link is pointing to. (mutated)
+   *  @param link backlink to add. */
+  static addBacklink({
+    note,
+    backlink,
+  }: {
+    note: NotePropsMeta;
+    backlink: Omit<DLink, "type"> & { type: "backlink" };
+  }): void {
+    note.links.push(backlink);
+  }
+
+  /**
+   * Remove backlink from note. If note does not contain that backlink, do nothing.
+   * Mutates note in place
+   *
+   * @param note Note to update backlinks for.
+   * @param backlink Backlink to remove
+   */
+  static removeBacklink({
+    note,
+    backlink,
+  }: {
+    note: NotePropsMeta;
+    backlink: Omit<DLink, "type"> & { type: "backlink" };
+  }): void {
+    const filteredBacklinks = note.links.filter((link) => {
+      return !_.isEqual(backlink, link);
+    });
+    note.links = filteredBacklinks;
+  }
+}
--- a/packages/common-all/src/BacklinkUtils.ts
+++ b/packages/common-all/src/BacklinkUtils.ts
@@ -28,7 +28,7 @@ export class BacklinkUtils {
    *
    *  @param note note that the link is pointing to. (mutated)
    *  @param link backlink to add. */
-  static addBacklink({
+  static addBacklinkInPlace({
     note,
     backlink,
   }: {
@@ -51,7 +51,7 @@ export class BacklinkUtils {
    * @param note Note to update backlinks for.
    * @param backlink Backlink to remove
    */
-  static removeBacklink({
+  static removeBacklinkInPlace({
     note,
     backlink,
   }: {
--- a/packages/common-all/src/BacklinkUtils.ts
+++ b/packages/common-all/src/BacklinkUtils.ts
@@ -1,4 +1,5 @@
 import _ from "lodash";
+import { DLinkUtils } from "./DLinkUtils";
 import { DLink, NotePropsMeta } from "./types";
 
 type BackLink = Omit<DLink, "type"> & { type: "backlink" };
@@ -23,6 +24,7 @@ export class BacklinkUtils {
   }
 
   /** Adds a backlink by mutating the 'note' argument in place.
+   * Check if backlink already exists before pushing
    *
    *  @param note note that the link is pointing to. (mutated)
    *  @param link backlink to add. */
@@ -33,7 +35,13 @@ export class BacklinkUtils {
     note: NotePropsMeta;
     backlink: BackLink;
   }): void {
-    note.links.push(backlink);
+    if (
+      !note.links.some((linkToCompare) =>
+        DLinkUtils.isEquivalent(backlink, linkToCompare)
+      )
+    ) {
+      note.links.push(backlink);
+    }
   }
 
   /**
--- a/packages/common-all/src/BacklinkUtils.ts
+++ b/packages/common-all/src/BacklinkUtils.ts
@@ -1,6 +1,7 @@
 import _ from "lodash";
 import { DLink, NotePropsMeta } from "./types";
 
+type BackLink = Omit<DLink, "type"> & { type: "backlink" };
 export class BacklinkUtils {
   /**
    * Create backlink out of link if it references another note (denoted by presence of link.to field)
@@ -8,9 +9,7 @@ export class BacklinkUtils {
    * @param link Original link to create backlink out of
    * @returns backlink or none if not applicable
    */
-  static createFromDLink(
-    link: DLink
-  ): (Omit<DLink, "type"> & { type: "backlink" }) | undefined {
+  static createFromDLink(link: DLink): BackLink | undefined {
     const maybeToNoteFname = link.to?.fname;
     if (maybeToNoteFname) {
       return {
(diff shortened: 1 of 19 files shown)


packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -43,7 +46,8 @@ export abstract class EngineV3Base implements ReducedDEngine {
 
   constructor(
     protected noteStore: INoteStore<string>,
-    protected logger: DLogger
+    protected logger: DLogger,
+    public vaults: DVault[]
```
jonathanyeung, line 50: I'm thinking it might be better to make vaults `protected` instead of `public`.  Not sure if using the engine is the right place to get `vaults`.  Is this possible, or does it lead to a snowball effect of changes?

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -382,4 +392,88 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
+      }
+    }
+    return [];
+  }
+
+  /**
+   * Remove backlink associated with given link that references another note (denoted by presence of link.to field)
+   * from that referenced note
+   *
+   * @param link Link potentionally referencing another note
```
jonathanyeung, line 442: nit: sp potentially

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -382,4 +392,88 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
+   */
+  protected async removeBacklink(link: DLink): Promise<NoteChangeEntry[]> {
```
jonathanyeung, line 444: aside - should we maybe undeprecate `DLink`? It's been marked as deprecate for as long as I can remember and its suggested replacement `DNoteLink` is practically the same and that interface needs some cleanup too.  @kevinslin

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -382,4 +392,88 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
+    const maybeBacklink = BacklinkUtils.createFromDLink(link);
+    if (maybeBacklink) {
+      const maybeVault = link.to?.vaultName
+        ? VaultUtils.getVaultByName({
+            vname: link.to?.vaultName,
+            vaults: this.vaults,
+          })
+        : undefined;
+      const notes = await this.noteStore.findMetaData({
+        fname: link.to!.fname!,
```
jonathanyeung, line 454: instead of `!` - can we do that same thing again as in the other review? If those aren't defined, do an early return of [] to save some backend calls.

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -382,4 +392,88 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
+        vault: maybeVault,
+      });
+      if (notes.data) {
+        return Promise.all(
+          notes.data.map(async (note) => {
+            const prevNote = _.merge(_.cloneDeep(note), {
+              body: "",
```
jonathanyeung, line 461: hm - should `NoteChangeEntry` be changed to use propsMeta instead? not sure - something to consider later.

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -382,4 +392,88 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
+    const maybeBacklink = BacklinkUtils.createFromDLink(link);
+    if (maybeBacklink) {
+      const maybeVault = link.to?.vaultName
+        ? VaultUtils.getVaultByName({
+            vname: link.to?.vaultName,
+            vaults: this.vaults,
+          })
+        : undefined;
+      const notes = await this.noteStore.findMetaData({
+        fname: link.to!.fname!,
```
jonathanyeung, line 412: same comment as below.

packages/unified/src/remark/utils.ts
```diff
@@ -365,6 +366,45 @@ const getLinkCandidates = async ({
 };
 
 export class LinkUtils {
+  static isEquivalent(linkA: DLink, linkB: DLink): boolean {
```
jonathanyeung, line 369: These methods seem more appropriate somewhere in `common-all`.

Same can be said about a lot of methods in this class... we should break down this file into methods that need to be in `unified` package and those that don't. Maybe in the next iteration of refactoring.

packages/common-all/src/BacklinkUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+   *  @param note note that the link is pointing to. (mutated)
+   *  @param link backlink to add. */
+  static addBacklink({
+    note,
+    backlink,
+  }: {
+    note: NotePropsMeta;
+    backlink: Omit<DLink, "type"> & { type: "backlink" };
+  }): void {
+    note.links.push(backlink);
```
jonathanyeung, line 37: any chance we have duplicates?

packages/plugin-core/src/test/suite-integ/CopyNoteLink.test.ts
```diff
@@ -366,7 +366,7 @@ suite("CopyNoteLink", function () {
             engine,
           });
 
-        const editor = await openNote(noteWithTarget);
+        let editor = await openNote(noteWithTarget);
```
jonathanyeung, line 369: was this change intentional? maybe I'm reading it wrong, but the second `openNote()` call seems redundant?

jonathanyeung (changes requested):
Wondering if this will maybe fix this issue? It's our most recent CROP selection: https://github.com/dendronhq/dendron/issues/1207

If not, then any changes I do on it I'll build off this review.

packages/common-all/src/BacklinkUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+export class BacklinkUtils {
+  /**
+   * Create backlink out of link if it references another note (denoted by presence of link.to field)
+   *
+   * @param link Original link to create backlink out of
+   * @returns backlink or none if not applicable
+   */
+  static createFromDLink(
+    link: DLink
+  ): (Omit<DLink, "type"> & { type: "backlink" }) | undefined {
```
namjul, line 13: Maybe extract an alias out of the return types. Something like `type BackDLink = Omit<DLink, "type"> & { type: "backlink" }`

packages/plugin-core/src/test/suite-integ/CopyNoteLink.test.ts
tma66: This test was failing on github every now and then; for some reason, the editor was still sometimes reading the previous document. I added this change in hopes that it fixes that.

packages/common-all/src/BacklinkUtils.ts
tma66: Definitely could happen. Do you think its worth the extra runtime to perform a dup check?

packages/unified/src/remark/utils.ts
tma66: Makes sense. I moved `BacklinksUtil` into common-all for this reason. Let's do it in another step of refactoring

packages/common-all/src/engine/EngineV3Base.ts
tma66: ah good reminder

packages/common-all/src/engine/EngineV3Base.ts
tma66: There are places where the note body is needed. Right now it'll be messier to change so leaving as is for now...

packages/common-all/src/engine/EngineV3Base.ts
tma66: Right now its part of `DCommonProps` so changing to `protected` isn't possible w/o another refactoring :/

packages/common-all/src/BacklinkUtils.ts
tma66: Sure, can do

tma66:
> Wondering if this will maybe fix this issue? It's our most recent CROP selection: #1207
> 
> If not, then any changes I do on it I'll build off this review.

Actually this will fix that bug after this latest version. Just tested it

packages/common-all/src/BacklinkUtils.ts
jonathanyeung: I think it's worth doing the check (shouldn't be too expensive).  Otherwise we may have funky UI in the backlinks panel and potentially in other places as well.

packages/common-all/src/engine/EngineV3Base.ts
kevinslin: no, we should continue to remove it

we use `DNoteLink` in new methods, the key difference is the location of the `alias` field (see references of dlink2DNoteLink)
    - DLink has an `alias` field, it's not supposed to be on the link field but actually in the `from/to` field
    - `DNoteLoc` is a super set of `DLoc` (has an `alias` field)
    - `DNoteLink` is used in other places

packages/common-all/src/BacklinkUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+    return;
+  }
+
+  /** Adds a backlink by mutating the 'note' argument in place.
+   * Check if backlink already exists before pushing
+   *
+   *  @param note note that the link is pointing to. (mutated)
+   *  @param link backlink to add. */
+  static addBacklink({
```
kevinslin, line 31: can we rename this to `addBacklinkInplace` to make it obvious?

kevinslin:
looks like theres a compiler issue

kevinslin (approved):

PR #3538, opened 2022-09-18
chore/sqlite to enginev3

kevinslin:
- add sqlite store
- fix bad error message
- skip test


packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -910,11 +913,47 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
   private async initNotes(
     schemas: SchemaModuleDict
   ): Promise<RespWithOptError<NotePropsByIdDict>> {
-    const ctx = "DEngine:initNotes";
-    this.logger.info({ ctx, msg: "enter" });
+    const ctx = "DEngineV3:initNotes";
     let errors: IDendronError[] = [];
     let notesFname: NotePropsByFnameDict = {};
     const start = process.hrtime();
+    const enableSQLITE =
```
jonathanyeung, line 920: `DendronEngineV3` shouldn't be aware of anything related to SQLite - it should only know about the `IDataStore` interface, otherwise we're losing the benefits we gained by creating those interfaces. We need to pull all init logic out of this class/file.

jonathanyeung (changes requested):

tma66:
> FileStore

FileStore is never used. But kept there for backwards compatibility until we clean it up

hikchoi (changes requested):
Same comment about the init process that @jonathanyeung 

if I'm understanding it correctly, sqlite backend wouldn't work currently with the web extension unless we make it work in memory and in the browser. So I think it makes sense that this is tied to `FileStorage` for the time being.

packages/engine-server/src/DendronEngineV3.ts
hikchoi: What happens if a user has the SQLite setting flipped in their configs and is using the web extension? wouldn't it cause problem at this point?

PR #3539, opened 2022-09-19
fix(sync): next js export pod error

jonathanyeung:
## fix(sync): next js export pod error

Fixing a few issues in Next JS Export and the playwright test pass:
- adding back 'handleDup' logic in publishing
- adding a missing `wsRoot` parameter in NextJS Export Pod
- Fixing a compile warning/error in `DendronNotePage.tsx`

PR #3541, opened 2022-09-19
Chore/separate out ci workflows

jonathanyeung:
## Chore: separate out CI playwright workflows

This change pulls the playwright test pass into a separate job to make it easier for us to spot failures in it. It was previously running as part of the CLI-14 pass.

NOTE TO REVIEWERS: the only relevant changes are in `ci.yml`.  The other changes are artifacts of a separate PR here https://github.com/dendronhq/dendron/pull/3539, but I wanted to include some test fixes in my branch here to make sure the playwright tests were working correctly.

There are some duplicate GH action jobs here, but I tried using [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow), but it seems a bit immature at the moment so I decided not to continue. (See https://stackoverflow.com/questions/71046096/reusable-workflows-with-local-actions)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,27 +102,6 @@ jobs:
         run: |
           Xvfb :99 -screen 0 1024x768x24 > /dev/null 2>&1 &
           echo 'DISPLAY=:99.0' >>$GITHUB_ENV
-
-      - name: Export test-workspace
-        if: runner.os == 'Linux' && matrix.suite == 'cli' && matrix.node-version == '14.x'
-        working-directory: ./test-workspace
-        run:  ../node_modules/.bin/dendron exportPod --podId dendron.nextjs --config "dest=../packages/nextjs-template/"
-
-      - name: Install Playwright dependencies
-        if: runner.os == 'Linux' && matrix.suite == 'cli' && matrix.node-version == '14.x'
-        run: npx playwright install --with-deps
-
-      - name: Run Playwright tests
-        if: runner.os == 'Linux' && matrix.suite == 'cli' && matrix.node-version == '14.x'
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ on:
       - "*"
       - "*/*"
 
+env:
```
jonathanyeung, line 17: note to reviewers: only changes in this file (ci.yml) are relevant.  Everything else can be ignored in this review.

namjul (approved):
Nice, love that.

Harshita-mindfire (approved):

kevinslin (approved):

.github/workflows/ci.yml
```diff
@@ -172,3 +148,63 @@ jobs:
 
       - name: Run Plugin-Web tests
         run: yarn ci:test:plugin-web
+
+  test-playwright:
+    strategy:
```
hikchoi, line 153: the github workflow json schema seems to require `matrix` as a property for `strategy`, and we are not defining a matrix here.

If we are not running this in a matrix I don't think we need the `strategy`

hikchoi (changes requested):

.github/workflows/ci.yml
hikchoi: ref: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategy

jonathanyeung: thanks for pointing it out, removed here + in the web tests.

hikchoi (approved):

PR #3543, opened 2022-09-19
fix(markdown): same page header links

jonathanyeung:
## fix(markdown): same page header links

Fixing a subtle regression introduced with https://github.com/dendronhq/dendron/pull/3497. Markdown links that point to a header on the same page now work again in Dendron Preview.


kevinslin (approved):

hikchoi (approved):
Good catch! Thanks for the fix.

Might want to re-run CI since a few are failing.

PR #3544, opened 2022-09-19
feat(workspace): copy as command

Harshita-mindfire:
This PR aims to add a Copy As command. It is similar to Export pod command, with a predefined config. 
Presently it supports two formats: JSON and markdown. The formatted note is copied to the clipboard.

- Task: [[Copy as Command|dendron://private/task.2022.07.31.copy-as-command]]
- Docs: https://github.com/dendronhq/dendron-site/pull/655

TODO: copy as for a selection in note body

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs

### DIFF
--- a/packages/plugin-core/src/commands/CopyAsCommand.ts
+++ b/packages/plugin-core/src/commands/CopyAsCommand.ts
@@ -1,4 +1,4 @@
-import { assertUnreachable } from "@dendronhq/common-all";
+import { DendronError } from "@dendronhq/common-all";
 import {
   CopyAsFormat,
   getAllCopyAsFormat,
@@ -68,7 +68,9 @@ export class CopyAsCommand extends BasicCommand<
         return PodCommandFactory.createPodCommandForStoredConfig({ config });
       }
       default:
-        assertUnreachable(format);
+        throw new DendronError({
+          message: `${format} is not a valid copy as format. If you are using a keybinding, make sure the argument is one of the following values: ${getAllCopyAsFormat()}`,
+        });
     }
   }
 
--- a/packages/plugin-core/src/commands/CopyAsCommand.ts
+++ b/packages/plugin-core/src/commands/CopyAsCommand.ts
@@ -6,9 +6,11 @@ import {
   PodExportScope,
   PodV2Types,
 } from "@dendronhq/pods-core";
+import _ from "lodash";
 import { PodCommandFactory } from "../components/pods/PodCommandFactory";
 import { PodUIControls } from "../components/pods/PodControls";
 import { DENDRON_COMMANDS } from "../constants";
+import { VSCodeUtils } from "../vsCodeUtils";
 import { BaseCommand, CodeCommandInstance } from "./base";
 
 type CommandOutput = void;
@@ -30,6 +32,13 @@ export class CopyAsCommand extends BaseCommand<
   constructor(_name?: string) {
     super(_name);
     this.format = getAllCopyAsFormat();
+  }
+
+  async sanityCheck(_opts?: Partial<CodeCommandInstance> | undefined) {
+    if (_.isUndefined(VSCodeUtils.getActiveTextEditor())) {
+      return "you must have a note open to execute this command";
+    }
+    return;
   }
 
   async gatherInputs() {
--- a/packages/plugin-core/src/commands/CopyAsCommand.ts
+++ b/packages/plugin-core/src/commands/CopyAsCommand.ts
@@ -1,0 +1,66 @@
+import { assertUnreachable } from "@dendronhq/common-all";
+import {
+  CopyAsFormat,
+  getAllCopyAsFormat,
+  JSONV2PodConfig,
+  PodExportScope,
+  PodV2Types,
+} from "@dendronhq/pods-core";
+import { PodCommandFactory } from "../components/pods/PodCommandFactory";
+import { PodUIControls } from "../components/pods/PodControls";
+import { DENDRON_COMMANDS } from "../constants";
+import { BaseCommand, CodeCommandInstance } from "./base";
+
+type CommandOutput = void;
+type CommandInput = CodeCommandInstance;
+type CommandOpts = CodeCommandInstance;
+
+/**
+ * Command that will find the appropriate export command to run, and then run
+ * it. This is the UI entry point for all export pod functionality.
+ */
+export class CopyAsCommand extends BaseCommand<
+  CommandOpts,
+  CommandOutput,
+  CommandInput
+> {
+  public format: CopyAsFormat[];
+  key = DENDRON_COMMANDS.COPY_AS.key;
+
+  constructor(_name?: string) {
+    super(_name);
+    this.format = getAllCopyAsFormat();
+  }
+
+  async gatherInputs() {
+    const format = await PodUIControls.promtToSelectCopyAsFormat();
+
+    if (!format) {
+      return;
+    }
+    switch (format) {
+      case CopyAsFormat.JSON: {
+        const config: JSONV2PodConfig = {
+          destination: "clipboard",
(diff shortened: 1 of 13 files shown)


namjul (approved):
Tested locally and worked great!

packages/plugin-core/src/components/pods/PodCommandFactory.ts
```diff
@@ -22,27 +22,46 @@ export class PodCommandFactory {
...
-      storedConfig.exportScope = exportScope;
+    let podType: PodV2Types;
+    let storedConfig: ExportPodConfigurationV2 | undefined;
+    if (config) {
+      podType = config.podType;
+      storedConfig = config;
+    } else {
+      storedConfig = PodV2ConfigManager.getPodConfigById({
+        podsDir: path.join(getExtension().podsDir, "custom"),
+        opts: configId!,
```
kevinslin, line 48: can we avoid type assertion here? rather, lets check if configId exist and throw an error if it does not

packages/plugin-core/src/components/pods/PodCommandFactory.ts
```diff
@@ -22,27 +22,46 @@ export class PodCommandFactory {
...
+      });
+
+      if (!storedConfig) {
+        throw new DendronError({
+          message: `No pod config with id ${configId!.podId} found.`,
```
kevinslin, line 53: lets not do type assertions here

packages/plugin-core/src/commands/CopyAsCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+      }
+      default:
+        assertUnreachable(format);
+    }
+  }
+
+  /**
+   * no-op
+   */
+  async enrichInputs(inputs: CommandInput): Promise<CommandOpts | undefined> {
```
kevinslin, line 78: if you extend from `BasicCommand`, you can remove this

packages/plugin-core/src/constants.ts
```diff
@@ -513,6 +513,11 @@ export const DENDRON_COMMANDS: { [key: string]: CommandEntry } = {
     title: `${CMD_PREFIX} Copy Codespace URL`,
     when: `${DendronContext.PLUGIN_ACTIVE} && shellExecutionSupported`,
   },
+  COPY_AS: {
```
kevinslin, line 516: can we use `cmd|ctrl+k cmd|ctrl+c` as keyboard shortcut for this?

kevinslin (changes requested):

packages/pods-core/src/index.ts
```diff
@@ -73,4 +73,13 @@ export function getAllImportPods(): PodClassEntryV4[] {
   ];
 }
 
+export enum CopyAsFormat {
+  "JSON" = "JSON",
+  "MARKDOWN" = "Markdown",
+}
+
+export function getAllCopyAsFormat(): CopyAsFormat[] {
```
hikchoi, line 81: nit: since `CopyAsFormat` is a string enum we can grab all the values with `Object.values(CopyAsFormat)`

packages/plugin-core/src/components/pods/PodControls.ts
```diff
@@ -552,4 +554,23 @@ export class PodUIControls {
     });
     return podIdQuickPick?.label;
   }
+
+  /**
+   * Prompt user to select the copy as format
+   */
+  public static async promtToSelectCopyAsFormat(): Promise<
+    CopyAsFormat | undefined
```
hikchoi, line 562: Is there ever going to be a case where the format is undefined?

packages/plugin-core/src/commands/CopyAsCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  async sanityCheck(_opts?: Partial<CodeCommandInstance> | undefined) {
+    if (_.isUndefined(VSCodeUtils.getActiveTextEditor())) {
+      return "you must have a note open to execute this command";
+    }
+    return;
+  }
+
+  async gatherInputs() {
```
hikchoi, line 45: low hanging enhancement idea for later:
allow command opts to specify the format so we can define keybindings for each `CopyAsFormat`. (could be tied to a context menu this way too)

hikchoi (changes requested):
left some comments, and a small nit.

packages/plugin-core/src/components/pods/PodControls.ts
Harshita-mindfire: When user press `Esc`

packages/plugin-core/src/constants.ts
Harshita-mindfire: By default, in windows, this is a shortcut to add a comment line
`{ "key": "ctrl+k ctrl+c",         "command": "editor.action.addCommentLine",`

jonathanyeung (approved):

packages/plugin-core/src/commands/CopyAsCommand.ts
Harshita-mindfire: https://github.com/dendronhq/dendron/pull/3544/commits/0ec47f1228fcfa90c129054a6e1232fa86030e5c

packages/plugin-core/src/constants.ts
kevinslin: yep. i figured we're not going to win the battle of not conflicting with keyboard shortcuts. rather, just go for conflicts of items that are not used a lot. the `ctrl+/` shortcut is the main one used for comments and we found that people generally don't set shortcuts themselves. for new commands going forward, lets try to find a shortcut for them in the `ctrl+k/cmd+k` namespace that doesn't conflict with a popular command

packages/plugin-core/src/commands/CopyAsCommand.ts
```diff
@@ -0,0 +1,78 @@
...
+type CommandOpts = CodeCommandInstance;
+
+/**
+ * Command that will find the appropriate export command to run, and then run
+ * it. This is the UI entry point for all export pod functionality.
+ */
+export class CopyAsCommand extends BasicCommand<
+  CommandOpts,
+  CommandOutput,
+  CopyAsFormat
```
hikchoi, line 27: Can we keep `CommandInput` like we usually do and give it a property `format`?

as-is, keybindings have to be defined like this:

```
  {
    "command": "dendron.copyAs",
    "key": "cmd+k m",
    "when": "dendron:pluginActive",
    "args": "Markdown"
  }
```

Instead I think we want this:

```
  {
    "command": "dendron.copyAs",
    "key": "cmd+k m",
    "when": "dendron:pluginActive",
    "args": {
        "format": "Markdown"
    }
  }
```

packages/plugin-core/src/commands/CopyAsCommand.ts
```diff
@@ -0,0 +1,78 @@
...
+        const config: MarkdownV2PodConfig = {
+          destination: "clipboard",
+          exportScope: PodExportScope.Note,
+          podType: PodV2Types.MarkdownExportV2,
+          podId: "copyAs.markdown", // dummy value, required property
+        };
+        return PodCommandFactory.createPodCommandForStoredConfig({ config });
+      }
+      default:
+        assertUnreachable(format);
```
hikchoi, line 71: When defining a keybinding and mispelling the format args the user can reach this, but the error message isn't informing.

Can we instead pop an error message saying "X" is not a valid copy as format and maybe suggest to check keybindings.json to make sure there's no typo? since that's the only way we can reach this path I _think_

kevinslin (approved):

Issue #2393, opened 2022-02-10
Add auto-complete when renaming with `F2`


aleksey-rowan:
## Please select if your request is either something new or an enhancement ##

- [x] Enhancement of an existing Feature.
- [ ] Request of a new feature.

## Please select the area your request applies to. (Multiple selections are Possible. You can leave blank if you're not sure.) ##

- [ ] Workspace - VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup - Dendron's Lookup Command
- [ ] Views - Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema - Dendron Schemas
- [ ] Pod - Data import from / export to Dendron
- [ ] Publish - External Site Publish
- [ ] Markdown - Markdown features, such as syntax support and features

## Is your feature request related to a problem? Please describe ##

When renaming a link under the cursor using `F2 / rename symbol`, the popup doesn't support auto-complete for file names like the lookup command.

## Describe the solution you'd like ##

The rename popup has the same auto-complete as the lookup.

## Additional context ##

Spawned off of https://github.com/dendronhq/dendron/issues/1694.

jonathanyeung:
HI @aleksey-rowan - this is a good idea, but unfortunately we don't actually own this control / UI for F2 rename on a link - it's part of VSCode.  I don't think we can alter this behavior.

hikchoi:
We are able to add autocompletion in lookup because we do our own indexing of notes in the quickpick, and quickpick exposes an API that allows changing the current value dynamically.

Unfortunately, the `RenameProvider` doesn't expose this.

This is a very brief way of how it works:

1. `RenameProvider.prepareRename` API lets extension authors decide how / what to extract from the cursor position / selection. How we implement `prepareRename` will decide the initial value that will be filled when pressing <kbd>F2</kbd>.
2. `RenameProvider.provideRenameEdits` API lets extension authos how the actual rename operations should happen. It will return a `vscode.WorkspaceEdit` which is vscode's construct of what an "edit" is. We however return a empty `WorkspaceEdit` because we use our own engine to perform rename operations.

Between 1 and 2, nothing is exposed where the extension author has control over the input box. We can intercept and try to simulate it by handing the control over to lookup before we proceed, but this probably isn't going to be a good user experience.

I did a quick search of the issues in the vscode repo, but I don't see any requests / proposal on this matter.

hikchoi:
[...]

PR #3559, opened 2022-09-21
enhance(lookup): tab autocomplete in all lookup controls

jonathanyeung:
## enhance(lookup): tab autocomplete in all lookup controls

This change adds tab autocomplete to all lookup controls, as well as in the web extension. 

Previously, if a command wanted tab autocomplete, it needed to implement an `onAutoComplete` interface, and then register itself with a singleton registrar during init. Any places that wanted to run the command would need to talk to the registrar and retrieve the right command instance out of it.

Instead, the mechanism now uses the `vscode.event` / `eventEmitter` pattern.  If you want to get notified of when the 'tab' key gets hit, you just get an `onAutocomplete` event out of the registrar (still a singleton for now), and then register /unregister your own callback in your own class. This simplifies the flow a lot and makes it much easier to add tab autocomplete functionality anywhere.

Related: https://github.com/dendronhq/dendron/issues/2393


kevinslin:
can we add tests for this?

packages/plugin-core/src/web/extension.ts
```diff
@@ -52,6 +53,21 @@ async function setupCommands(context: vscode.ExtensionContext) {
...
+    NoteLookupAutoCompleteCommand
+  );
+  const tabAutoCompletekey = DENDRON_COMMANDS.LOOKUP_NOTE_AUTO_COMPLETE.key;
+
+  if (!existingCommands.includes(tabAutoCompletekey))
+    context.subscriptions.push(
+      vscode.commands.registerCommand(
+        tabAutoCompletekey,
+        async (_args: any) => {
+          await TabAutoCompleteCommand.run();
```
kevinslin, line 66: we can drop the await here and just return this, no?

kevinslin (commented):
nice! big improvement in dx as well as functionality

jonathanyeung:
tests added

kevinslin (approved):

PR #3560, opened 2022-09-21
chore(workspace): Remove engine.notes references Part 2

tma66:
Second pass to remove `engine.notes` references in our code.

TODO:
- Remove references to `getNoteByFnameFromEngine`


packages/engine-test-utils/src/__tests__/engine-server/DendronEngineV3.spec.ts
```diff
@@ -1,20 +1,34 @@
...
 import {
+  getPreset,
   ENGINE_HOOKS,
   ENGINE_V3_PRESETS,
   ENGINE_V3_PRESETS_MULTI,
 } from "../../presets";
 
 const createEngine = createEngineV3FromEngine;
 
+test.skip("bond", async () => {
```
namjul, line 21: Note that this is skiped. In case this is not intended.

namjul (approved):
skimmed through and lgtm

packages/common-all/src/types/store.ts
```diff
@@ -16,6 +16,7 @@ export type FindNoteOpts = {
   // If vault is provided, filter results so that only notes with matching vault is returned
   vault?: DVault;
   // If true, exclude stubs from results. Otherwise, include stub notes
+  // WARNING: If false and no other parameters are set, this will return all notes in the engine
```
jonathanyeung, line 19: I think we need to be very careful about exposing this functionality of getting all notes, or else it'll start getting used and we'd be kinda back to engine v2.

packages/common-server/src/StateValidator.ts
```diff
@@ -22,8 +23,9 @@ export class StateValidator {
         const rootNote = (await engine.findNotes({ fname: "root", vault }))[0];
         try {
           if (rootNote) {
+            const engineNotes = await engine.findNotes({ excludeStub: true });
```
jonathanyeung, line 26: We can probably remove this `StateValidator` util once we're on V3? Since there's just a single state.

packages/plugin-core/src/utils/ExtensionUtils.ts
```diff
@@ -313,7 +313,8 @@ export class ExtensionUtils {
       type: workspaceType,
       config: dendronConfig,
     } = workspace;
-    let numNotes = _.size(engine.notes);
+    const notes = await engine.findNotes({ excludeStub: false });
```
jonathanyeung, line 316: How long does this take for org-workspace? We should change to `findNotesMeta` in as many places where possible.

Or in this case, just delete the telemetry for now until we have a better query model.

packages/plugin-core/src/utils/md.ts
```diff
@@ -494,9 +494,10 @@ export async function findReferencesById(id: string) {
     return;
   }
 
+  const engineNotes = await engine.findNotes({ excludeStub: true });
   const notesWithRefs = NoteUtils.getNotesWithLinkTo({
     note,
-    notes: engine.notes,
+    notes: engineNotes,
```
jonathanyeung, line 500: We should move this link finding logic to the backend (later).  Maybe need to check on performance in `nightly` to see if it's acceptable - does backlinks panel still work in org-workspace?

jonathanyeung (changes requested):
Nice work - a few concerns about performance, but we can see how things look in `nightly`.  We should leverage the `meta` methods wherever possible, which should be in post places where we're doing `findNotes()`

packages/engine-test-utils/src/__tests__/engine-server/DendronEngineV3.spec.ts
tma66: Good catch, forgot to remove it

packages/common-all/src/types/store.ts
tma66: Yea, unfortunately there are some use cases that require all the notes (or requires a filter that we don't support). But good reminder on the `findNotesMeta`. I'll fix a few instances here and do a pass converting from NoteProps to NotePropsMeta where feasible

packages/common-server/src/StateValidator.ts
tma66: It'll still be useful for validating the metadata (making sure children and parent props are correct compared to the filesystem)

packages/plugin-core/src/utils/ExtensionUtils.ts
tma66: Right now its still using `enginev2` so its reading from memory still. When we swap to v3, we should keep an eye on the time.

kevinslin (approved):

jonathanyeung (approved):

PR #3563, opened 2022-09-21
feat(lookup): Add `Create New with Template` label to note lookup

hikchoi:
# feat(lookup): Add `Create New with Template` label to note lookup

This PR:
- Adds `LabelUtils` which given a string label, add highlight text that has an expiration date.
- Add AB test to experiment efficacy of the quickpick label highlighting
- Refactors template selection logic of `ApplyTemplateCommand` to a reusable component `QuickPickTemplateSelector`
- Fix some issues with `MoveSelectionToCommand`
- Add `Create New with Template` label in Note Lookup command.
  - This ignores `selectionExtract`, but `selection2link` works
  - This ignores any onCreate behavior, so `Create Task Note` button will be ignored.

## Demonstration

image

image

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/LabelUtils.ts
+++ b/packages/common-all/src/LabelUtils.ts
@@ -1,10 +1,5 @@
 import { DateTime } from "luxon";
 
-export type DendronLabel = {
-  value: string;
-};
-export const DENDRON_LABELS = new Map<string, DendronLabel>([]);
-
 /**
  * Simple utility class for generating labels with a time constrained _highlight_
  * Use it to give a certain label that is exposed in the UI some highlighting text
@@ -21,14 +16,16 @@ export class LabelUtils {
     highlight: {
       value: string;
       location: "prefix" | "suffix";
-      expirationTime: DateTime;
+      expirationDate: Date;
     };
   }) {
     const { value, highlight } = opts;
     let prefix = "";
     let suffix = "";
     const shouldHighlight =
-      highlight.expirationTime.diffNow().as("milliseconds") > 0;
+      DateTime.fromJSDate(highlight.expirationDate)
+        .diffNow()
+        .as("milliseconds") > 0;
     if (shouldHighlight) {
       if (highlight.location === "prefix") {
         prefix = `${highlight.value} `;
--- a/packages/common-all/src/LabelUtils.ts
+++ b/packages/common-all/src/LabelUtils.ts
@@ -1,0 +1,43 @@
+import { DateTime } from "luxon";
+
+export type DendronLabel = {
+  value: string;
+};
+export const DENDRON_LABELS = new Map<string, DendronLabel>([]);
+
+/**
+ * Simple utility class for generating labels with a time constrained _highlight_
+ * Use it to give a certain label that is exposed in the UI some highlighting text
+ * for a short amount of time.
+ *
+ * TODO: Depending on the usage, make this a service that can be
+ * generally used to provide all public facing UI text
+ * TODO: Make this a central place to manage all labels and their durations
+ * TODO: Consolidate all constants defined across our codebase here.
+ */
+export class LabelUtils {
+  static createLabelWithHighlight(opts: {
+    value: string;
+    highlight: {
+      value: string;
+      location: "prefix" | "suffix";
+      expirationTime: DateTime;
+    };
+  }) {
+    const { value, highlight } = opts;
+    let prefix = "";
+    let suffix = "";
+    const shouldHighlight =
+      highlight.expirationTime.diffNow().as("milliseconds") > 0;
+    if (shouldHighlight) {
+      if (highlight.location === "prefix") {
+        prefix = `${highlight.value} `;
+      }
+      if (highlight.location === "suffix") {
+        suffix = ` ${highlight.value}`;
+      }
+    }
+    const out = `${prefix}${value}${suffix}`;
+    return out;
+  }
+}
--- a/packages/common-all/src/abTests.ts
+++ b/packages/common-all/src/abTests.ts
@@ -92,24 +92,22 @@ export const _2022_09_VIDEO_TUTORIAL_TEST = new ABTest(
 /**
  * Experiment to test highlights in quickpick labels
  */
-export enum QuickPickLabelHighlightTestGroups {
+export enum CreateNewWithTemplateQuickPickLabelHighlightTestGroups {
   "none" = "none",
   "label" = "label",
 }
 
-export const _2022_09_QUICKPICK_LABEL_HIGHLIGHT_TEST = new ABTest(
-  "2022-09-QuickPickLabelHighlightTest",
-  [
+export const _2022_09_CREATE_NEW_WITH_TEMPLATE_QUICKPICK_LABEL_HIGHLIGHT_TEST =
+  new ABTest("2022-09-CreateNewWithTemplateQuickPickLabelHighlightTest", [
     {
-      name: QuickPickLabelHighlightTestGroups.none,
+      name: CreateNewWithTemplateQuickPickLabelHighlightTestGroups.none,
       weight: 1,
     },
     {
-      name: QuickPickLabelHighlightTestGroups.label,
+      name: CreateNewWithTemplateQuickPickLabelHighlightTestGroups.label,
       weight: 1,
     },
-  ]
-);
+  ]);
 
 /**
  * Tutorial type of our ever-running / up to date main tutorial.
--- a/packages/common-all/src/abTests.ts
+++ b/packages/common-all/src/abTests.ts
@@ -90,6 +90,28 @@ export const _2022_09_VIDEO_TUTORIAL_TEST = new ABTest(
 );
 
 /**
+ * Experiment to test highlights in quickpick labels
+ */
+export enum QuickPickLabelHighlightTestGroups {
+  "none" = "none",
+  "label" = "label",
+}
+
+export const _2022_09_QUICKPICK_LABEL_HIGHLIGHT_TEST = new ABTest(
+  "2022-09-QuickPickLabelHighlightTest",
+  [
+    {
+      name: QuickPickLabelHighlightTestGroups.none,
+      weight: 1,
+    },
+    {
+      name: QuickPickLabelHighlightTestGroups.label,
+      weight: 1,
+    },
+  ]
+);
+
+/**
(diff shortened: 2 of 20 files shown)


hikchoi:
pending:

- [ ] update `dendron.topic.templates`
- [ ] update `dendron.topic.lookup.create`

hikchoi:
ask for reviewers:

If you have any better way to highlight the label, let me know. Using a codicon is the most we can do within the quickpick item. I couldn't find anything that could signify _new_ in the built in codicons. We could create a custom web font with a glyph that contains a better icon, but I think that's beyond what I can do in a timely manner 😅

jonathanyeung:
> If you have any better way to highlight the label, let me know. 

Maybe the beaker? to represent experimental?  Actually, I don't think the smiley you picked is half-bad - it gets the point across.

packages/plugin-core/src/components/lookup/LookupProviderV3Interface.ts
```diff
@@ -53,7 +53,15 @@ export type OnUpdatePickerItemsOpts = {
 };
 
 export type ILookupProviderOptsV3 = {
+  /**
+   * should provide `Create New`
+   */
   allowNewNote: boolean;
+  /**
+   * should provide `Create New with Template`
+   * `allowNewNote` should be true for this to take effect.
```
jonathanyeung, line 62: can we add in the comments what the default value is since it's an optional parameter? 

also nit: instead of 'should be true', we should say 'must be true'.

packages/plugin-core/src/components/lookup/TemplateSelector.ts
```diff
@@ -0,0 +1,89 @@
...
+  DLogger,
+  NoteProps,
+} from "@dendronhq/common-all";
+import { ExtensionProvider } from "../../ExtensionProvider";
+import { NoteLookupProviderUtils } from "./NoteLookupProviderUtils";
+import { HistoryEvent } from "@dendronhq/engine-server";
+import { Logger } from "../../logger";
+import * as vscode from "vscode";
+
+export interface TemplateSelector {
```
jonathanyeung, line 13: maybe no need for an interface until we have more than one implementation?

packages/common-all/src/abTests.ts
```diff
@@ -89,6 +89,28 @@ export const _2022_09_VIDEO_TUTORIAL_TEST = new ABTest(
...
 
+/**
+ * Experiment to test highlights in quickpick labels
+ */
+export enum QuickPickLabelHighlightTestGroups {
+  "none" = "none",
+  "label" = "label",
+}
+
+export const _2022_09_QUICKPICK_LABEL_HIGHLIGHT_TEST = new ABTest(
```
jonathanyeung, line 100: nit: can we have somewhere in the test name that this concerns the apply-template row in the quickpick? Otherwise it's harder to know what experiemnt 'quickpick-label' is referring to

packages/plugin-core/src/commands/MoveSelectionToCommand.ts
```diff
@@ -112,8 +112,12 @@ export class MoveSelectionToCommand extends BasicCommand<
 
   private createLookupProvider(opts: { activeNote: NoteProps | undefined }) {
     const { activeNote } = opts;
-    return this.extension.noteLookupProviderFactory.create(this.key, {
+    // the id here should be "lookup" as long as we are supplying this
+    // to the lookup command.
+    // TODO: give it its own id once we refactor.
```
jonathanyeung, line 117: After refactoring, I don't think we need any ID at all - anyway, irrelevant for now.

packages/plugin-core/src/components/lookup/utils.ts
```diff
@@ -437,6 +438,15 @@ export class PickerUtilsV2 {
     }
   }
 
+  static isCreateNewNoteWithTemplatePick(
```
jonathanyeung, line 441: should this be `Picked`?

jonathanyeung (changes requested):
Minor nits - should be cool!

packages/plugin-core/src/components/lookup/TemplateSelector.ts
Harshita-mindfire: If we decide to remove the interface, can we also update the filename to `QuickPickTemplateSelector`

kevinslin:
> > If you have any better way to highlight the label, let me know.
> 
> Maybe the beaker? to represent experimental? Actually, I don't think the smiley you picked is half-bad - it gets the point across.

i vote for the beaker. also lets drop the exclamation. `{beaker} [new] {feature}`

packages/common-all/src/LabelUtils.ts
```diff
@@ -0,0 +1,43 @@
...
+ * TODO: Make this a central place to manage all labels and their durations
+ * TODO: Consolidate all constants defined across our codebase here.
+ */
+export class LabelUtils {
+  static createLabelWithHighlight(opts: {
+    value: string;
+    highlight: {
+      value: string;
+      location: "prefix" | "suffix";
+      expirationTime: DateTime;
```
kevinslin, line 24: nitpick: i think time is too granular. `expirationDate` shoudl be enough with a type of `Date`

kevinslin (changes requested):

packages/plugin-core/src/components/lookup/TemplateSelector.ts
hikchoi: makes sense. went with a similar pattern with [HierarchySelector](https://github.com/dendronhq/dendron/blob/feat/create-new-with-template/packages/plugin-core/src/components/lookup/HierarchySelector.ts#L10:L10) but the interface is not needed here too

packages/plugin-core/src/components/lookup/utils.ts
hikchoi: went with `Pick` since the counterpart is named `isCreateNewNotePick` but I can fix both.

packages/plugin-core/src/components/lookup/NotePickerUtils.ts
```diff
@@ -81,6 +93,44 @@ export class NotePickerUtils {
...
+    );
+
+    let label: string;
+    if (ABUserGroup === QuickPickLabelHighlightTestGroups.label) {
+      label = LabelUtils.createLabelWithHighlight({
+        value: CREATE_NEW_WITH_TEMPLATE_LABEL,
+        highlight: {
+          value: "$(smiley) [New!] ",
+          location: "prefix",
+          expirationTime: DateTime.fromISO("2022-10-01"),
```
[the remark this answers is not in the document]
hikchoi: At first I actually thought about tying this to the metadata so the expiration depends on the user actually seeing it, but I don't think that's a good way to approach feature highlighting here.

A feature is _new_ for a certain amount of period after the release. After that, it is just a regular feature that has been out for a while.

The problem we are tackling with this is "users are not aware of new features right away when they release", and after that, the problem statement changes to "users are not discovering our features". I would say those are two different problems that should be dealt in different ways.

I do understand where you're coming from though. But even when we change this so that it expires depending on the user seeing it or not, we still want the hard coded expiration date.
By doing that we are containing this to a specific time frame (our opinion on how long a feature belongs in _new_), eliminating the need to come back and clean it up (otherwise all new and returning users will be bombarded with features highlighted as _new_), and still basing the highlight on the fact that it was viewed by the user.

> i would increase the expiration time to a month. 
Got it. I think a month is a good time frame for us to use as default if we decide to use this highlighting feature more in the future as well.

packages/plugin-core/src/components/lookup/QuickPickTemplateSelector.ts
```diff
@@ -0,0 +1,101 @@
...
+        },
+      });
+      controller.show({
+        title: "Select template to apply",
+        placeholder: "template",
+        provider,
+        initialValue,
+      });
+
+      VSCodeUtils.setContext(DendronContext.NOTE_LOOK_UP_ACTIVE, true);
```
hikchoi, line 86: @jonathanyeung 

Doing some cleanup here after a rebase to apply the changes from https://github.com/dendronhq/dendron/pull/3559/ to work here as well.

This code as-is works if `QuickPickTemplateSelector` is used in the context of another command (i.e. `ApplyTemplateCommand`), but not if this is part of `NoteLookupCommand`.

This is because as part of `onDidHide`, the lookup controller sets `DendronContext.NOTE_LOOKUP_ACTIVE` to false, and that timing happens to be after the quickpick for template selection is shown. This results in the template selector not autocompletable.

Since we are relying on the global context state, it's a bit tricky to handle this if another lookup has to be chained. Any pointers on how to tackle this?

jonathanyeung: One possible 'solution' is to remove the conditional check on tab autocomplete of having the note-lookup-active flag set.  This _theoretically_ shouldn't be a problem, b/c it uses a pub/sub event, and if there are no lookup controls around when someone is hitting tab, then no callback gets invoked.

The potential concern with that approach is I'm not 100% confident all of our lookup controller cleanup is working properly.  I don't think the side effects should be too bad if some controller _is_ still hanging around? Maybe something we can test w/ nightly. 

Maybe this is a defeatist mentality haha, but I also wouldn't think it's so bad if tab autocomplete doesn't work in this one control - it's still an improvement from where we are today.

Another reason to refactor the lookup code...

hikchoi: > This theoretically shouldn't be a problem, b/c it uses a pub/sub event, and if there are no lookup controls around when someone is hitting tab, then no callback gets invoked.

But we are contributing a default keybinding for the autocomplete command (as `tab`) as an extension, and without that context key would we be blocking tabs completely?

hikchoi: Resolved by removing the `onDidHide` behavior of the quickpick and handling it during command cleanup

kevinslin (approved):

PR #3575, opened 2022-09-22
enhance(publish): make search mode as default for search bar in published site

Harshita-mindfire:
This PR aims to make search as default mode in a published site. It also adds a config option `searchDefaultBehavior` which can take two values: either `search` or `lookup`.  Added back the `fetchRootResults` in `NoteLookupUtils`. This is still needed in nextjs-template
```yml
publishing:
  searchDefaultBehavior: "search|lookup"
```
- if set to `lookup`, default lookup, add a `?` to toggle search
- if set to `search`, default search, add a `/` to toggle lookup

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/constants/configs/publishing.ts
+++ b/packages/common-all/src/constants/configs/publishing.ts
@@ -279,7 +279,7 @@ export const PUBLISHING: DendronConfigEntryCollection<DendronPublishingConfig> =
       desc: "Define custom sidebar",
       value: undefined,
     },
-    searchDefaultBehavior: {
+    searchMode: {
       label: "Default search mode for publishing",
       desc: "Default behavior of search bar in published site",
     },
--- a/packages/common-all/src/constants/configs/publishing.ts
+++ b/packages/common-all/src/constants/configs/publishing.ts
@@ -279,4 +279,8 @@ export const PUBLISHING: DendronConfigEntryCollection<DendronPublishingConfig> =
       desc: "Define custom sidebar",
       value: undefined,
     },
+    searchDefaultBehavior: {
+      label: "Default search mode for publishing",
+      desc: "Default behavior of search bar in published site",
+    },
   };
--- a/packages/common-all/src/types/configs/publishing/publishing.ts
+++ b/packages/common-all/src/types/configs/publishing/publishing.ts
@@ -56,7 +56,7 @@ export type DendronPublishingConfig = {
   siteBanner?: string;
   giscus?: GiscusConfig;
   sidebarPath?: string | false;
-  searchDefaultBehavior?: SearchMode;
+  searchMode?: SearchMode;
 };
 
 export type CleanDendronPublishingConfig = DendronPublishingConfig &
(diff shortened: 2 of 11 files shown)


kevinslin (approved):

packages/common-all/src/constants/configs/publishing.ts
```diff
@@ -279,4 +279,8 @@ export const PUBLISHING: DendronConfigEntryCollection<DendronPublishingConfig> =
       desc: "Define custom sidebar",
       value: undefined,
     },
+    searchDefaultBehavior: {
```
hikchoi, line 282: nit: maybe `defaultSearchBehavior`? or simply `searchMode`? (since lot of our configs are there to provide configurable defaults but don't have `default` in their name)

hikchoi (changes requested):
small nit on config naming but otherwise LGTM.

packages/common-all/src/types/configs/publishing/publishing.ts
```diff
@@ -109,5 +115,6 @@ export function genDefaultPublishingConfig(): DendronPublishingConfig {
     enableRandomlyColoredTags: true,
     enableTaskNotes: true,
     enablePrettyLinks: true,
+    searchMode: SearchMode.SEARCH,
```
namjul, line 118: If we make the default `SearchMode.SEARCH`, aren't we then changing the current default which is lookup? It looks to me that we should set it to `SearchMode.LOOKUP` so that we don't change the default behavior.

packages/nextjs-template/components/DendronSearch.tsx
```diff
@@ -255,13 +265,13 @@ function DendronSearchComponent(props: DendronCommonProps & SearchProps) {
       value={searchQueryValue}
       getPopupContainer={(trigger) => trigger.parentElement}
       // @ts-ignore
-      onClick={results === SearchMode.SEARCH_MODE ? () => null : onClickLookup}
+      onClick={searchMode === SearchMode.SEARCH ? () => null : onClickLookup}
```
namjul, line 268: I am wondering what the `onClick` actually do. Quickly playing on site it does nothing or rather on the first click it opens and suddenly closes again.

packages/common-all/src/types/configs/publishing/publishing.ts
Harshita-mindfire: We actually want to change the default behavior to search instead of lookup

packages/nextjs-template/components/DendronSearch.tsx
Harshita-mindfire: onclick, when the mode is Lookup(present behavior) fetches notes and updates the autocomplete with the results somewhat similar to how lookup behaves when no note is open in local workspace. The behavior is sketchy sometimes, I can take a look at this in a followup PR

PR #3576, opened 2022-09-22
chore(workspace): Use getNoteMeta and findNotesMeta where possible

tma66:
Made a quick sweep to convert some engine calls to the more efficient *Meta calls.


jonathanyeung (approved):
Nice!

As long as the code compiles, I think we should be good right? Else any references to `note.body` would cause errors.

PR #3577, opened 2022-09-23
chore: remove circular deps in common all

jonathanyeung:
## chore: remove circular deps in common all

This change removes all circular dependencies in common-all. 

When working on this change https://github.com/dendronhq/dendron/pull/3572, all the CLI tests started failing because some changes in common-all caused import resolution to fail b/c of circular dependencies in common-all. 

Functionally, the code should be exactly the same as before - all I did was move stuff around as-is.  The one exception is that I found two definitions for `DendronDevConfig` - I wasn't 100% sure which one we actually want.  @hikchoi - I left a comment in the PR itself. 

This change also introduces a new madge check for common-all in addition to the one we have in plugin-core.  I'll soon add checks for all of our packages to limit our circ deps to the current counts.

On a side note... common-all is due for a clean up.  There are duplicated types, nearly identical v1 and v2 types, and files are a bit all over the place.

### DIFF
--- a/packages/common-all/src/types/configs/dev/DendronDevConfig.ts
+++ b/packages/common-all/src/types/configs/dev/DendronDevConfig.ts
@@ -26,6 +26,12 @@ export type DendronDevConfig = {
    * Enable new preview as default
    */
   enablePreviewV2?: boolean;
+  /** Force the use of a specific type of watcher.
+   *
+   * - plugin: Uses VSCode's builtin watcher
+   * - engine: Uses the engine watcher, watching the files directly without VSCode
+   */
+  forceWatcherType?: "plugin" | "engine";
   /**
    * Enable export pod v2
    */
--- a/hooks/pre-push.js
+++ b/hooks/pre-push.js
@@ -8,25 +8,22 @@ const path = require("path");
  * Uses madge to check for circular dependencies. Issues soft warning if
  * circular dependencies detected
  */
-function checkCircularDependencies() {
-  const CIRCULAR_DEP_THRESHOLD = 1; // Lower this each time we fix a circular dependency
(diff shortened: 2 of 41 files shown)


packages/common-all/src/types/configs/dev/DendronDevConfig.ts
```diff
@@ -26,6 +26,12 @@ export type DendronDevConfig = {
    * Enable new preview as default
    */
   enablePreviewV2?: boolean;
+  /** Force the use of a specific type of watcher.
+   *
+   * - plugin: Uses VSCode's builtin watcher
+   * - engine: Uses the engine watcher, watching the files directly without VSCode
+   */
+  forceWatcherType?: "plugin" | "engine";
```
jonathanyeung, line 34: @hikchoi - we had duplicated types for `DendronDevConfig` - the only difference between them is that the one we were exporting had this `forceWatcherType` property.  Is it fine to include it / remove it? 

FWIW - longer term I would prefer to remove it, but it looks like there are a few references to it in plugin-core and elsewhere.

hikchoi: this seems to be added to the old config type while config migration was in process (and the dev namespace itself was never migrated because there was nothing to migrate).

so `DendronDevConfig.ts` was never exported and was there as a placeholder during config migration. (plan was to clean up the old type that was actually exported and export this instead when that happens)

so yeah, for now with this change we should do this.

tangentially, I think this was added as part of Native workspaces, but after a quick skim through the code I don't think we need chokidar to watch files at all, so we can remove this in the future.

Sorry about the confusion with these remnant config types. Looking at the stats I think we can proceed with closing the loop on config migration on October finally so I'll be cleaning the rest of this up.

hikchoi (approved):

PR #3579, opened 2022-09-23
chore: more logging around engine init

kevinslin:
chore: more logging around engine init

### DIFF
--- a/packages/engine-server/src/drivers/file/storev2.ts
+++ b/packages/engine-server/src/drivers/file/storev2.ts
@@ -116,6 +116,7 @@ export class FileStorage implements DStore {
   }
 
   async init(): Promise<DEngineInitResp> {
+    const ctx = "Store:init";
     let errors: IDendronError<any>[] = [];
     try {
       const resp = await this.initSchema();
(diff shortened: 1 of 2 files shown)


packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -116,6 +116,7 @@ export class FileStorage implements DStore {
   }
 
   async init(): Promise<DEngineInitResp> {
+    const ctx = "Store:init";
```
hikchoi, line 119: super duper nit, but I found it more easier to read logs when the `ctx` strings match the exact class name it's in

```suggestion
    const ctx = "FileStorage:init";
```

hikchoi (changes requested):

packages/engine-server/src/drivers/file/storev2.ts
hikchoi: not really an issue right now because we don't have any other places that use the same ctx string, but that could change in the future. Having a uniquely identifiable ctx string would help a lot with pinpointing the cause.

jonathanyeung (approved):

PR #3581, opened 2022-09-23
chore(workspace): Migrate getNoteByFnameFromEngine to engine methods

tma66:
This PR migrates most of the `NoteUtils.getNoteByFnameFromEngine` to use the proper engine methods.

TODO: Migrate getNotesByFnameFromEngine

### DIFF
--- a/packages/plugin-core/src/commands/CreateTask.ts
+++ b/packages/plugin-core/src/commands/CreateTask.ts
@@ -39,7 +39,7 @@ export class CreateTaskCommand extends BasicCommand<
     const { createTaskSelectionType, addBehavior } =
       ConfigUtils.getTask(config);
 
-    await maybeSendMeetingNoteTelemetry("task");
+    maybeSendMeetingNoteTelemetry("task");
 
     return {
       lookup: AutoCompletableRegistrar.getNoteLookupCmd().run({
--- a/packages/plugin-core/src/commands/CreateTask.ts
+++ b/packages/plugin-core/src/commands/CreateTask.ts
@@ -39,7 +39,7 @@ export class CreateTaskCommand extends BasicCommand<
     const { createTaskSelectionType, addBehavior } =
       ConfigUtils.getTask(config);
 
-    maybeSendMeetingNoteTelemetry("task");
+    await maybeSendMeetingNoteTelemetry("task");
 
     return {
       lookup: AutoCompletableRegistrar.getNoteLookupCmd().run({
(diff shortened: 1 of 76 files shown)


packages/plugin-core/src/components/views/NoteGraphViewFactory.ts
```diff
@@ -72,12 +72,14 @@ export class NoteGraphPanelFactory {
...
       this._onEngineNoteStateChangedDisposable =
-        this._engineEvents.onEngineNoteStateChanged((noteChangeEntry) => {
+        this._engineEvents.onEngineNoteStateChanged(async (noteChangeEntry) => {
           const ctx = "NoteGraphViewFactoryEngineNoteStateChanged";
           Logger.info({ ctx });
           if (this._panel && this._panel.visible) {
-            noteChangeEntry.map((changeEntry) =>
-              this.refresh(changeEntry.note)
+            await Promise.all(
+              noteChangeEntry.map((changeEntry) =>
```
kevinslin, line 80: you need to return something from this map for the await to trigger

packages/plugin-core/src/commands/CreateTask.ts
```diff
@@ -39,7 +39,7 @@ export class CreateTaskCommand extends BasicCommand<
     const { createTaskSelectionType, addBehavior } =
       ConfigUtils.getTask(config);
 
-    maybeSendMeetingNoteTelemetry("task");
+    await maybeSendMeetingNoteTelemetry("task");
```
jonathanyeung, line 42: nit: don't await this - we don't want to slow down note creation just to send telemetry. (although the part of the `maybeSendMeetingNoteTelemetry` function that is actually awaited should be trivially fast).  But in general - just cuz something is awaitable, doesn't mean it should be awaited on.

packages/plugin-core/src/commands/GotoNote.ts
```diff
@@ -315,7 +315,7 @@ export class GotoNoteCommand extends BasicCommand<
 
             // check if we should send meeting note telemetry.
             const type = qs.startsWith("user.") ? "userTag" : "general";
-            maybeSendMeetingNoteTelemetry(type);
+            await maybeSendMeetingNoteTelemetry(type);
```
jonathanyeung, line 318: same comment as before.

jonathanyeung (approved):

kevinslin (approved):

PR #3587, opened 2022-09-27
chore: add telemetry and inform if localhost is blocked

Harshita-mindfire:
This PR aims to add a check to see if localhost is available at the time of Dendron init. Incase of blocked localhost, we send a warning toaster reporting the same with a link to see [troubleshooting docs](https://wiki.dendron.so/notes/a6c03f9b-8959-4d67-8394-4d204ab69bfe/#whitelisting-localhost).
- Task: [[show error message when localhost is blocked|task.2022.09.13.show-error-message-when-localhost-is-blocked]]

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/utils/StartupUtils.ts
+++ b/packages/plugin-core/src/utils/StartupUtils.ts
@@ -15,6 +15,7 @@ import { DConfig, readMD } from "@dendronhq/common-server";
 import {
   DEPRECATED_PATHS,
   DoctorActionsEnum,
+  execa,
   InactvieUserMsgStatusEnum,
   MetadataService,
   MigrationChangeSetStatus,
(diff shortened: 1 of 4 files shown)


packages/plugin-core/src/utils/StartupUtils.ts
```diff
@@ -544,4 +545,28 @@ export class StartupUtils {
 
     MetadataService.instance().v100ReleaseMessageShown = true;
   }
+
+  /**
+   * this method pings the localhost and checks if it is available. Incase local is blocked off,
+   * displays a toaster with a link to troubleshooting docs
+   */
+  static async showWhitelistingLocalhostDocsIfNecessary() {
+    const { failed } = await execa("ping", ["127.0.0.1"]);
```
kevinslin, line 554: can we create a test for this? also - is ping available for powershell?

jonathanyeung: A test that pings local host? Is that going to be a reliable test?

jonathanyeung (approved):

packages/plugin-core/src/utils/StartupUtils.ts
hikchoi: Looks like powershell has an applet called [Test-Connection](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/test-connection?view=powershell-7.2)

hikchoi: as long as we cover all platforms (make it work in powershell), I think we can just mock the response coming from the ping for testing. No use in testing the external dependency itself.

packages/plugin-core/src/utils/StartupUtils.ts
```diff
@@ -544,4 +545,28 @@ export class StartupUtils {
...
+        )
+        .then((resp) => {
+          if (resp === "Open troubleshooting docs") {
+            vscode.commands.executeCommand(
+              "vscode.open",
+              "https://wiki.dendron.so/notes/a6c03f9b-8959-4d67-8394-4d204ab69bfe/#whitelisting-localhost"
+            );
+          }
+        });
+      AnalyticsUtils.track(ExtensionEvents.LocalhostBlocked);
```
hikchoi, line 569: To better measure if users are actually going through troubleshooting (at least to see if they are opening the link), maybe we should divide the event? (one for notice shown, one for accepted / rejected)

hikchoi (changes requested):

packages/plugin-core/src/utils/StartupUtils.ts
Harshita-mindfire: powershell does allow me to ping localhost
image

kevinslin (approved):

PR #3593, opened 2022-09-27
chore(workspace): Migrate getNotesByFname to engine methods

tma66:
Similar to https://github.com/dendronhq/dendron/pull/3581, except this PR migrates the `getNotesByFnameFromEngine`

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1052,7 +1052,7 @@ export class NoteUtils {
     };
   }
 
-  static toNoteLoc(note: NoteProps): DNoteLoc {
+  static toNoteLoc(note: NotePropsMeta): DNoteLoc {
     const { fname, id, vault } = note;
     return {
       fname,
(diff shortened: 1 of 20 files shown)


packages/common-all/src/dnode.ts
```diff
@@ -1218,48 +1218,43 @@ export class NoteUtils {
...
    */
-  static *ancestors(opts: {
+  static async ancestors(opts: {
     fname: string;
     vault?: DVault;
     engine: DEngineClient;
     includeSelf?: boolean;
     nonStubOnly?: boolean;
-  }): Generator<NoteProps> {
+  }): Promise<NotePropsMeta | undefined> {
```
jonathanyeung, line 1227: FWIW, there's an AsyncGenerator that also works here, but if the new signature works then that's also fine. 

```typescript
  static async *ancestors(opts: {
    fname: string;
    vault?: DVault;
    engine: DEngineClient;
    includeSelf?: boolean;
    nonStubOnly?: boolean;
  }): AsyncGenerator<NoteProps> {
```

jonathanyeung (approved):

packages/common-all/src/dnode.ts
tma66: Got it. I'm not sure if generator is needed here. The only use case I see is for the `color` method which it seems you've commented out for now lol

Harshita-mindfire (approved):
lgtm, apart from a merge conflict that sneaked in.

PR #3606, opened 2022-09-29
chore: info logs set to error

kevinslin:
chore: info logs set to error

correct some bad log levels


jonathanyeung (approved):
nice, just noticed a flood of these recently as well.

hikchoi (approved):

PR #3609, opened 2022-09-29
chore: revert onWillSave method to be sync

Harshita-mindfire:
Reverted onWillSaveNote method to be sync. This adds back the deprecated `getNoteByFnameFromEngine` reference in `onWillSaveNote` method

### DIFF
--- a/packages/plugin-core/src/WorkspaceWatcher.ts
+++ b/packages/plugin-core/src/WorkspaceWatcher.ts
@@ -296,22 +296,24 @@ export class WorkspaceWatcher {
    * When saving a note, do some book keeping
    * - update the `updated` time in frontmatter
    * - update the note metadata in the engine
+   *
+   * this method needs to be sync since event.WaitUntil can be called
+   * in an asynchronous manner.
    * @param event
    * @returns
    */
-  private async onWillSaveNote(event: TextDocumentWillSaveEvent) {
+  private onWillSaveNote(event: TextDocumentWillSaveEvent) {
     const ctx = "WorkspaceWatcher:onWillSaveNote";
     const uri = event.document.uri;
     const engine = this._extension.getEngine();
     const fname = path.basename(uri.fsPath, ".md");
     const now = Time.now().toMillis();
 
-    const note = (
-      await engine.findNotes({
-        fname,
-        vault: this._extension.wsUtils.getVaultFromUri(uri),
-      })
-    )[0];
+    const note = NoteUtils.getNoteByFnameFromEngine({
+      fname,
+      vault: this._extension.wsUtils.getVaultFromUri(uri),
+      engine,
+    });
 
     // If we can't find the note, don't do anything
     if (!note) {
--- a/packages/plugin-core/src/WorkspaceWatcher.ts
+++ b/packages/plugin-core/src/WorkspaceWatcher.ts
@@ -254,7 +254,7 @@ export class WorkspaceWatcher {
    * @param event
    * @returns
    */
-  async onWillSaveTextDocument(event: TextDocumentWillSaveEvent) {
+  onWillSaveTextDocument(event: TextDocumentWillSaveEvent) {
     try {
       const ctx = "WorkspaceWatcher:onWillSaveTextDocument";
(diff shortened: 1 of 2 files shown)


packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -296,22 +296,24 @@ export class WorkspaceWatcher {
...
     const fname = path.basename(uri.fsPath, ".md");
     const now = Time.now().toMillis();
 
-    const note = (
-      await engine.findNotes({
-        fname,
-        vault: this._extension.wsUtils.getVaultFromUri(uri),
-      })
-    )[0];
+    const note = NoteUtils.getNoteByFnameFromEngine({
```
jonathanyeung, line 312: for the engine v3 migration, we'll need to get rid of this call. So I think what we can do here is to stuff all the async function calls into the promise currently at the bottom of `onWillSaveNote`: 

```typescript
      const p = new Promise(async (resolve) => {
        note.updated = now;
        await engine.writeNote(note, { metaOnly: true });
        return resolve(changes);
      });
      event.waitUntil(p);
```

Basically move all the logic currently in this function into the promise

hikchoi: One thing to note before adding everything into `waitUntil`:

```
 Protection against misbehaving listeners is implemented as such:

    there is an overall time budget that all listeners share and if that is exhausted no further listener is called
    listeners that take a long time or produce errors frequently will not be called anymore

The current thresholds are 1.5 seconds as overall time budget and a listener can misbehave 3 times before being ignored.
```

jonathanyeung: Good call on pointing that out. For a local engine, I imagine we'll be done with everything in <100 ms. But if we have a remote engine, then we'll need to rethink how this flow works.  In a remote engine, maybe we won't implement this `onWillSave` event in vscode and have it handled elsewhere.

Harshita-mindfire: The if condition inside which the waitUntil is present, is depend on `note`. The note that we are fetching asynchronously.

kevinslin: can we pass in the note object in `onWillSaveTextDocument`? move the logic of checking for a note there

hikchoi: > can we pass in the note object in onWillSaveTextDocument? move the logic of checking for a note there

now that you mention it, `onWillSaveTextDocument` should also probably be sync 😅 (nothing in that method is awaited and it is now  synchronous).

everything that happens in the `onWillSaveTextDocument` block that is async needs to be inside a `waitUntil`

Harshita-mindfire: > can we pass in the note object in onWillSaveTextDocument? move the logic of checking for a note there

one of the method in the call chain has to be async for this, throws the same error.

packages/plugin-core/src/WorkspaceWatcher.ts
```diff
@@ -296,60 +296,59 @@ export class WorkspaceWatcher {
...
         await engine.writeNote(note, { metaOnly: true });
-        return resolve(changes);
-      });
-      event.waitUntil(p);
-    }
+      }
+      return resolve(changes);
+    });
+    event.waitUntil(promise);
     return { changes };
```
jonathanyeung, line 352: Functionally there's nothing wrong with this since no one is using the return value, but let's just return void instead of {changes}.  Right now, changes will always be an empty array because it's getting populated asynchronously after this function returns.

For the same reason, the promise can also resolve void instead of `resolve(changes)`

jonathanyeung (changes requested):

packages/plugin-core/src/WorkspaceWatcher.ts
jonathanyeung: let's defer for now due to existing testing considerations.

jonathanyeung (approved):

PR #3616, opened 2022-09-30
chore(internal): track remote information on init

hikchoi:
# chore(internal): track remote information on init
This PR:
- adds a new event `GitEvents.TopLevelRepoFound`, which tracks the hash of the remote url if the workspace has a top level remote repo.
- For remotely tracked self contained vault and workspace vault, this would be the repo at wsRoot.
- If a repo is not found in wsRoot, either it is not tracked remotely, or in the case of a workspace, individual vaults may be tracked by not the workspace itself. We don't count this as a "top repo" as it is ambiguous. 

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -876,6 +876,22 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
     return contributors.filter(isNotUndefined);
   }
 
+  async getToplevelRepo() {
+    // if self contained vault,
+    //   if tracked remotely, wsRoot should be a repo
+    //   else, nothing to get. undefined
+    // if not self contained,
+    //   if tracked remotely, wsRoot should be a repo
+    //   else, top level repo is ambiguous even if the vaults are remotely tracked.
+    const git = new Git({ localUrl: this.wsRoot });
+    try {
+      const remote = await git.getRemoteUrl();
+      console.log({ remote });
+    } finally {
+      console.log("foo");
+    }
+  }
+
   async commitAndAddAll({
     engine,
   }: {
--- a/packages/engine-server/src/workspace/service.ts
+++ b/packages/engine-server/src/workspace/service.ts
@@ -885,7 +885,7 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
    *   In this case, it is ambiguous what the top level is, and we assume the top level is not tracked remotely.
    * @returns remote url or undefined
    */
-  async getToplevelRemoteUrl(): Promise<string | undefined> {
+  async getTopLevelRemoteUrl(): Promise<string | undefined> {
     const git = new Git({ localUrl: this.wsRoot });
     const remoteUrl = await git.getRemoteUrl();
     return remoteUrl;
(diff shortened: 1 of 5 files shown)


Harshita-mindfire (approved):
code lgtm, a ques: why are we tracking this info?

Harshita-mindfire:
>  ques: why are we tracking this info?

nv, read the task : [[task.2022.09.13.track-git-repos-on-initialize]]

namjul (approved):

packages/engine-server/src/workspace/service.ts
```diff
@@ -876,6 +876,21 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
+   * If self contained vault, workspace root should be the top level if remotely tracked.
+   * If not self contained, workspace root should be the top level if remotely tracked.
+   * If not self contained vault, and workspace root doesn't have a remote url,
+   *   This means nothing is remotely tracked, or some vaults in the workspace is tracked, not the workspace itself.
+   *   In this case, it is ambiguous what the top level is, and we assume the top level is not tracked remotely.
+   * @returns remote url or undefined
+   */
+  async getTopLevelRemoteUrl(): Promise<string | undefined> {
+    const git = new Git({ localUrl: this.wsRoot });
+    const remoteUrl = await git.getRemoteUrl();
```
kevinslin, line 890: will this throw an exception if there is no repo?

packages/plugin-core/src/test/suite-integ/workspaceActivator.test.ts
```diff
@@ -0,0 +1,153 @@
...
+import { ExtensionProvider } from "../../ExtensionProvider";
+import { describeMultiWS } from "../testUtilsV3";
+import { describe } from "mocha";
+import { expect } from "../testUtilsv2";
+import { trackTopLevelRepoFound } from "../../workspace/workspaceActivator";
+import sinon from "sinon";
+import _ from "lodash";
+
+suite("workspaceActivator", function () {
+  describe("trackTopLevelRepoFound", () => {
```
kevinslin, line 12: can we add a test case when there is no repo found?

packages/engine-server/src/workspace/service.ts
hikchoi: it will not. `Git` will try to get the remote, and if it fails `remote` is `undefined`, and in turn `remoteUrl` will be undefined

packages/plugin-core/src/test/suite-integ/workspaceActivator.test.ts
hikchoi: I initially omitted the no repo case because I assumed this would all be covered by unit tests in `git.spec.ts`, and because this integ test was already feeling a bit redundant.

> covered by unit tests in `git.spec.ts`

This was not the case, so I added a few unit tests that needs to be covered in order to verify `trackTopLevelRepoFound` works.

> integ test was already feeling a bit redundant

and also changed my mind on this after realizing we don't have good coverage in unit tests (at least the ones related to the `Git` class). So I also added the `no repo` case in the integ tests.

PR #3617, opened 2022-09-30
prototype query interface for engine

jonathanyeung:
## prototype query interface for engine

Some prototype code with some ideas about how to prep our engine for SQLite. Note - code isn't close to compiling yet :) 

Rough Ideas being illustrated:
- Adding a new interface to support the `query` mechanism. This will enable us to remove the query implementation details from `EngineV3`. Note @namjul - I tried using `neverthrow` in the interface definition here.
- Replacement of the current dictionary based `NoteMetadataStore` with a version that incorporates Fuse. @kevinslin - this was what you mentioned during our Wednesday chat; after looking more closely at the Fuse code, I think this is the sensible thing to do.
- Removal of fuseEngine references from `enginev3` and `enginev3base` - this is possible now cuz it's abstracted via the `IDataStore` & `IDataQueryable` interfaces.
- Starting to tease out the init logic of enginev3 into a more componetized / injectable manner.  This can then let us incorporate SQLlite more cleanly.  This logic is in `EngineV3Factory.ts` and was forked off of the `init()` methods in the current `EngineV3` class. This is the roughest part of the review - definitely doesn't compile yet.


packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -291,37 +290,49 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
+      .map(async (items) => {
+        if (items.length === 0) {
+          return { data: [] };
+        }
 
-    const notes = await Promise.all(
-      items.map((ent) => this.noteStore.get(ent.id)) // TODO: Should be using metadata instead
-    );
+        const notes = await Promise.all(
+          items.map((ent) => this.noteStore.get(ent.id)) // TODO: Should be using metadata instead
```
namjul, line 302: Reading this I was wondering what if `noteStore.get` throws or returns and `RespV3` with an error.
Read thw whole thread https://github.com/supermacro/neverthrow/issues/113 (its not long) for a discussion around that.

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -143,14 +143,15 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
       vaults: ConfigUtils.getVaults(config),
       noteStore: new NoteStore(
         fileStore,
-        new NoteMetadataStore(),
+        new FuseMetadataStore(),
```
kevinslin, line 146: its a problem right now that the `FuseMetadataStore`  is initialized twice. the SQLite metadata store needs to pull down binaries and initialize tables on the initial load. either we 
- introduce an `init` method to the metadatastore/queryEngine 
- initialize it/inject it outside of DendronEngine 
- combine metadatastore and queryEngine

jonathanyeung: Sorry for the confusion - this `create` method was just scratch work - I ended up moving the `create()` method intended for use into `EngineV3Factory.ts`, where a single FuseMetadatStore is used. 

But all the code is still just prototyping.

packages/common-all/src/store/FuseMetadataStore.ts
```diff
@@ -0,0 +1,232 @@
...
+  /**
+   * See {@link IDataStore.find}
+   */
+  async find(opts: FindNoteOpts): Promise<RespV3<NotePropsMeta[]>> {
+    if (!opts.fname) {
+      return { data: [] }; // JYTODO
+    }
+
+    const searchResults = this.notesIndex.search({
+      fname: `=${opts.fname}`,
```
jonathanyeung, line 103: @kevinslin as a heads up - I've done zero testing on these search strings; just put something in here after reading fuse documentation

kevinslin:
cleaning up stale prs

PR #3618, opened 2022-09-30
chore(internal): Deprecate V4 config and remove intermediate types

hikchoi:
WIP

# Pull Request Checklist


hikchoi:
NOTE: this PR still has some pending tasks to be done, but since the diffs are getting huge it would be great to do an initial pass of reviews at this point.

pending tasks:
- [ ] gracefully handle v4 configs without forcing or prompting for a migration

packages/engine-server/src/workspace/service.ts
```diff
@@ -1594,41 +1593,41 @@ export class WorkspaceService implements Disposable, IWorkspaceService {
...
-      // if changes were made, use updated changes in subsequent configuration
-      if (!_.isEmpty(changes)) {
-        const { data } = _.last(changes)!;
-        dendronConfig = data.dendronConfig;
-      }
-    }
-
-    return changes;
-  }
+  // async runConfigMigrationIfNecessary({
```
jonathanyeung, line 1596: let's just delete if we don't need it?

jonathanyeung (commented):
lgtm apart from some merge conflicts.  good to merge if CI pass looks clean, nice job!

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -977,7 +977,9 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
         const cachePath = path.join(vpath, CONSTANTS.DENDRON_CACHE_FILE);
         const notesCache = new NotesFileSystemCache({
           cachePath,
-          noCaching: DConfig.readConfigSync(this.wsRoot).noCaching,
+          // noCaching: DConfig.readConfigSync(this.wsRoot).noCaching,
+          // TODO: clean up
+          noCaching: false,
```
kevinslin, line 982: can we add a next task to get rid of this option? caching should always be on

packages/engine-server/src/enginev2.ts
```diff
@@ -96,12 +96,13 @@ type CachedPreview = {
...
 function createRenderedCache(
-  config: IntermediateDendronConfig,
+  config: DendronConfig,
   logger: DLogger
 ): Cache<string, CachedPreview> {
   const ctx = "createRenderedCache";
 
-  if (config.noCaching) {
+  // TODO: remove null cache related logics later.
+  if (/** config.noCaching**/ false) {
```
kevinslin, line 105: can we just remove this?

kevinslin (approved):
awesome job. just some small comments. feel free to merge after fixing the merge conflict

packages/engine-server/src/workspace/service.ts
hikchoi: I left it there to move / re-use it in the still-pending task to handle v4 configs gracefully, but yeah, this will be cleaned up.

packages/engine-server/src/DendronEngineV3.ts
hikchoi: Git it. I left it there to avoid scope creep.

hikchoi:
Just a clarification that this is still incomplete and we probably should not merge this before we handle the v4 users.

I'll re-request a review after that part is implemented.

kevinslin:
can we merge this and then handle v4 users as the next task? (as long as get it in by the next release in two weeks time, this should be good)

PR #3619, opened 2022-09-30
enhance(sync): selection export scope for pods v2

Harshita-mindfire:
This PR aims to support selection scope while exporting a note with export pod v2. This also benefits the `Copy As` command. 

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
+++ b/packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
@@ -283,10 +283,13 @@ export abstract class BaseExportPodCommand<
               const filteredPayload = opts.payload.filter(
                 (note) => note.fname !== savedNote.fname
               );
-              await this.executeExportNotes({
-                ...opts,
-                payload: filteredPayload.concat(savedNote),
-              });
+              // export only selection if present instead of entire body
+              const exportOpts =
+                await this.getPropsForSelectionScopeIfNecessary({
+                  ...opts,
+                  payload: filteredPayload.concat(savedNote),
+                });
+              await this.executeExportNotes(exportOpts);
             }
           }
         );
@@ -303,8 +306,9 @@ export abstract class BaseExportPodCommand<
 
       return true;
     } else {
+      const exportOpts = await this.getPropsForSelectionScopeIfNecessary(opts);
       // Save is not needed. Go straight to exporting
-      await this.executeExportNotes(opts);
+      await this.executeExportNotes(exportOpts);
       return false;
     }
   }
@@ -350,7 +354,7 @@ export abstract class BaseExportPodCommand<
     if (!maybeNote) {
       vscode.window.showErrorMessage("couldn't find the note somehow");
     }
-    return [maybeNote!];
+    return [maybeNote];
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/plugin-core/src/commands/pods/BaseExportPodCommand.ts
```diff
@@ -283,10 +283,13 @@ export abstract class BaseExportPodCommand<
               const filteredPayload = opts.payload.filter(
                 (note) => note.fname !== savedNote.fname
               );
-              await this.executeExportNotes({
-                ...opts,
-                payload: filteredPayload.concat(savedNote),
-              });
+              // export only selection if present instead of entire body
+              const exportOpts =
```
kevinslin, line 287: can we raise an warning if nothing is selected?

Harshita-mindfire: We don't have a separate scope for `clipboard`. We export the selected text if the export scope is `note` and selection is present. Not very sure if that would be the right call to raise a warning on each note export.

kevinslin: can we modify this so we actually create a new scope "selection" and only apply this on selection? 

this way, we can be specific about the selection case. we also don't want to accidentally export the selection when the user meant to export the note.

copy as should then always set the `selection` scope

Harshita-mindfire: got it

Harshita-mindfire: [warning toaster](https://github.com/dendronhq/dendron/pull/3619/files#diff-b300d2a5c67b0dd9dd53963dac70b830cd1ae68b29c8de1cc7daeee764cd40bcR409)

namjul (approved):
just read it through without trying locally and lgtm.

Harshita-mindfire:
docs: https://github.com/dendronhq/dendron-site/pull/655/files

hikchoi (changes requested):
Tried testing this out locally, but it looks like activation silently fails (no errors, did not activate).

packages/plugin-core/src/components/pods/PodControls.ts
```diff
@@ -530,6 +533,9 @@ export class PodUIControls {
       case PodExportScope.Workspace:
         return "Exports all notes in the Dendron workspace";
 
+      case PodExportScope.Selection:
+        return "Export the selection from currently opened note";
```
jonathanyeung, line 537: nit: maybe "selected text" instead of "selection" just to make it clearer.

jonathanyeung (approved):
re: @hikchoi 's comment - is the init failure unique to this change commit or was it introduced elsewhere?  Otherwise lgtm

hikchoi:
> re: @hikchoi 's comment - is the init failure unique to this change commit or was it introduced elsewhere? Otherwise lgtm

looks like this is because of the (already fixed) regression with `showWhitelistingLocalhostDocsIfNecessary`.

I rebased it locally and tried again and it activated correctly.

hikchoi (approved):
Just tried it out locally (had to rebase master to fix the activation issue), and looks good. `copy as` and export pod works great!

PR #3630, opened 2022-10-05
chore(internal): add neverthrow package

namjul:
- chore(internal): add `neverthrow` package
- chore(internal): use `neverthrow` in sidebar

# Pull Request Checklist


namjul:
I was trying to add https://github.com/mdbetancourt/eslint-plugin-neverthrow. It would warn when calling a Result returning function but than not checking for an error. It it looks like our current eslint setup is not typescript aware which we would need to setup first https://typescript-eslint.io/docs/linting/typed-linting/. Just mentioning and leaving this now as is.

kevinslin:
this looks good but looks like the playwright test is failing. can you investigate that in a separate pr?

kevinslin (approved):

packages/common-all/src/utils/index.ts
```diff
@@ -52,6 +52,20 @@ import {
 } from "../types/configs/dendronConfigLegacy";
 import { DVault } from "../types/DVault";
 
+export {
```
hikchoi, line 55: if we are re-exporting this from common-all to other packages, maybe we should export them as a whole to prevent name collision?

hikchoi: on second thought, that might complicate things a bit. Probably fine until we actually have problems with the identifiers collide.

hikchoi (approved):

packages/common-all/src/utils/index.ts
namjul: Exporting them directly we communicate that they are part of a [Shared Kernel](http://ddd.fed.wiki.org/view/shared-kernel) rather something extra (like under the neverthrow namespace). But I guess we are doing that already by making it part of `common-all`. From an ergonimics POV I like not having to first import a symbol and then needing to destructure it to access the parts. All having approved the current state I will merge this now as it is.

PR #3632, opened 2022-10-05
fix(workspace): go to definition for wikilink with header

Harshita-mindfire:
This PR aims to fix the issue with `Go to Definition` not working correctly when used on wikilinks with header.
- [bug](https://www.loom.com/share/00570ddc70244549a6c1748e4d91dce5)
- task: [[dendron://private/task.2022.10.02.go-to-definition-doesnt-work-inside-the-same-header-on-newly-created-notes]]

The Reference Hover provider creates a new note with same name but id with prefix `fake-` to render hover. This PR aims to filter those references in definition provider

### DIFF
--- a/packages/plugin-core/src/features/DefinitionProvider.ts
+++ b/packages/plugin-core/src/features/DefinitionProvider.ts
@@ -96,7 +96,7 @@ export default class DefinitionProvider implements vscode.DefinitionProvider {
           Logger.error({ msg: `${refAtPos.vaultName} is not defined` });
         }
       }
-      const notes = await (
+      const notes = (
         await engine.findNotesMeta({ fname: refAtPos.ref, vault })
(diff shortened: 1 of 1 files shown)


packages/plugin-core/src/features/DefinitionProvider.ts
```diff
@@ -96,7 +96,9 @@ export default class DefinitionProvider implements vscode.DefinitionProvider {
           Logger.error({ msg: `${refAtPos.vaultName} is not defined` });
         }
       }
-      const notes = await engine.findNotesMeta({ fname: refAtPos.ref, vault });
+      const notes = await (
```
kevinslin, line 99: you don't need the outer await here

Harshita-mindfire: 🤦

jonathanyeung (approved):

kevinslin (approved):

Issue #3605, opened 2022-09-29
Custom color tags no longer working


EngineerFaunce:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [x] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [x] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

The `color` attribute in a tag's frontmatter no longer works.

## To Reproduce

1. Create a tag with the `color` attribute in the frontmatter set to some value (I used `color: "#cc99ff"`
2. Reload window
3. Add tag to a note

## Expected behavior

Tags to display the color defined in their frontmatter.

## Screenshots

Tag with `color` attribute set:
![image](https://user-images.githubusercontent.com/16312436/192919452-f9dd5e8e-f6e4-49bb-99ad-ac1f8620621d.png)

Tag as it appears in a note:
![image](https://user-images.githubusercontent.com/16312436/192919494-4e3c5647-33a1-41fc-a47e-c1a35b071bb0.png)

Actual color:
![image](https://user-images.githubusercontent.com/16312436/192919647-c87570d7-37fc-46c6-89c8-f73857c28ab1.png)

## Desktop (please complete the following information)

- OS: Windows 10
- Versions (all that apply to bug):
  - Dendron Extension: v0.114.0
  - VS Code version: 1.71.2

## Additional context

Example from the dendron documentation: https://wiki.dendron.so/notes/zDqYUhNE10HXlmStfqwVD/

## Dendron Log file

[dendron.log](https://github.com/dendronhq/dendron/files/9670102/dendron.log)

kevinslin:
thanks for the issue, we're looking into this!

Harshita-mindfire:
This should be fixed after Dendron v0.116. Feel free to comment here if this is still an issue.

EngineerFaunce:
I've updated to Dendron v0.116.0, but the custom color tag still isn't working:

![image](https://user-images.githubusercontent.com/16312436/198856517-fa3030ec-c883-4e79-9385-0f59e93dd9e6.png)

[...]

PR #3637, opened 2022-10-06
fix(workspace): custom color decoration for hashtags

Harshita-mindfire:
Related issue: https://github.com/dendronhq/dendron/issues/3605
This PR aims to fix the issue with Dendron not recognizing color code set at the frontmatter.

Previously, the NoteUtils's color method looked up the ancestors of fname provided and returned configured/generated color.
With this fix, we provide note to render(one with the frontmatter) to color method and it returns the generated/configured color accordingly.

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -1185,11 +1185,7 @@ export class NoteUtils {
    * @param fname The fname of note that you want to get the color of.
    * @returns The color, and whether this color was randomly generated or explicitly defined.
    */
-  static color(opts: {
-    fname: string;
-    vault?: DVault;
-    note?: NoteProps | NotePropsMeta;
-  }): {
+  static color(opts: { fname: string; vault?: DVault; note?: NotePropsMeta }): {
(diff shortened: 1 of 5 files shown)


kevinslin:
can we add a test for this?

packages/common-all/src/dnode.ts
```diff
@@ -1185,16 +1185,24 @@ export class NoteUtils {
    * @param fname The fname of note that you want to get the color of.
    * @returns The color, and whether this color was randomly generated or explicitly defined.
    */
-  static color(opts: { fname: string; vault?: DVault }): {
+  static color(opts: {
+    fname: string;
+    vault?: DVault;
+    note?: NoteProps | NotePropsMeta;
```
jonathanyeung, line 1191: nit: `note?: NotePropsMeta` should be sufficient here since the type is a strict subset of `NoteProps`

jonathanyeung (approved):
Thanks for the fix!

namjul (approved):

PR #3639, opened 2022-10-06
chore: improve eslint setup

namjul:
The aim of this PR is the solve the following points:

- get rid of `WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.` warning
- ~~allow for [type linting](https://typescript-eslint.io/docs/linting/typed-linting/)~~
- remove blacklisted packages https://github.com/dendronhq/dendron/blob/master/.eslintignore#L6-L8
- ~~maybe add https://github.com/supermacro/neverthrow#recommended-use-eslint-plugin-neverthrow~~

The current result of  root-level `yarn lint` (currenlty not used anywhere and only available for manual checks):

```sh
...
✖ 125 problems (85 errors, 40 warnings)
  30 errors and 0 warnings potentially fixable with the `--fix` option.
```

# Pull Request Checklist


namjul:
@kevinslin `yarn lint` reports now:

```sh
✖ 166 problems (124 errors, 42 warnings)
  35 errors and 0 warnings potentially fixable with the `--fix` option.
```

I still need to fix some errors to make the ci work. additionality to that shall I take on the next step and cleanup all the rest of the erros from the `lint` command?. And further, are there plans to use the `lint` command somewhere like inside CI e.g. https://github.com/facebook/docusaurus/blob/main/.github/workflows/lint.yml

.eslintrc.js
```diff
@@ -5,19 +5,23 @@ module.exports = {
...
     jest: true,
   },
-  extends: ["plugin:react/recommended", "airbnb", "airbnb/hooks", "prettier"],
+  extends: [
+    "plugin:react/recommended",
+    "plugin:@typescript-eslint/eslint-recommended",
+    "airbnb",
+    "airbnb/hooks",
+    "prettier",
+    "plugin:neverthrow/recommended",
```
kevinslin, line 14: will this throw an error if neverthrow is not used? if so, its going to cause too many errors. would not include

package.json
```diff
@@ -138,6 +138,7 @@
     "@dendronhq/conventional-changelog-dendron": "^5.0.11",
     "chalk": "^4.1.1",
     "cross-env": "^7.0.3",
+    "eslint-plugin-neverthrow": "^1.1.4",
```
kevinslin, line 141: see comment about this causing too much churn in the codebase. if we do this, would only run this on new code reviews

kevinslin: regardless, would leave it as a separate pr and not combine with this one

tsconfig.json
```diff
@@ -1,41 +1,9 @@
...
-        "packages/engine-server/src"
-      ],
-      "@dendronhq/plugin-core": [
-        "packages/plugin-core/src"
-      ],
-      "@dendronhq/pods-core": [
-        "packages/pods-core/src"
-      ]
-    }
+    "noImplicitAny": true
```
kevinslin, line 7: the paths were necessary for vscode to be able to generate import statements. is that no longer the case?

kevinslin:
> @kevinslin `yarn lint` reports now:
> 
> ```shell
> ✖ 166 problems (124 errors, 42 warnings)
>   35 errors and 0 warnings potentially fixable with the `--fix` option.
> ```
> 
> I still need to fix some errors to make the ci work. additionality to that shall I take on the next step and cleanup all the rest of the erros from the `lint` command?. And further, are there plans to use the `lint` command somewhere like inside CI e.g. https://github.com/facebook/docusaurus/blob/main/.github/workflows/lint.yml

first tenant is to not add more friction to the dev process. would not introduce the `lint` command unless it can reliably detect **new** issues. 

do lint errors currently prevent commit/push hooks? if so, then we'll need to fix. if not, would also split that off into a separate pr

.eslintrc.js
namjul: agree, removed it [`9bf6aa8` (#3639)](https://github.com/dendronhq/dendron/pull/3639/commits/9bf6aa8febb3f8193b56bdd15a67d77f705f5ae4)

package.json
namjul: as long we cannot use typeaware eslint rules this neverthrow plugin has no effects. its now disabled because it results in Out of Memory error https://github.com/dendronhq/dendron/pull/3639/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R23
we can introduce the plugin some time in the future when typeaware linting works.

tsconfig.json
namjul: you are right we need the paths that imports are resolved correctly in the editor. its the same in vscode and neovim.

namjul: https://github.com/dendronhq/dendron/pull/3639/commits/599edd31a48c1aca19245c8532fa880e4ed84f14

kevinslin (approved):
looks good - can you add an async about potential eslint changes when people push after rebasing?

Issue #3578, opened 2022-09-23
Task notes do not render correctly inside note references (in the preview)


samuelxyz:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [x] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [x] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

Task notes do not show information (checkbox, prio, due date etc) when embedded in a note reference. I found this in the note preview, haven't tested with publish. Aggravatingly, in another note with the same reference, *one* of the tasks renders correctly. This persists through reloading the window and restarting VSCode.

## To Reproduce

1. Create a task note (call it note A)
2. [[Link]] to note A from another note B
3. ![[Reference]] to note B from another note C
4. View the note preview of C

## Expected behavior

Link to task note, inside the note reference, shows additional information (checkbox, prio, due date, etc) beyond just the name.

## Screenshots

Preview of a note reference showing the bugged links:

<img width="204" alt="image" src="https://user-images.githubusercontent.com/41557313/191991923-335e6872-eed9-4989-b0fa-cf72a7273f26.png">

A note reference in a different file renders one of the tasks but not the others:

<img width="279" alt="image" src="https://user-images.githubusercontent.com/41557313/191992677-e6eb27cf-2bf3-4612-872f-9df67720c724.png">

Referenced note seems to render correctly on its own:

<img width="288" alt="image" src="https://user-images.githubusercontent.com/41557313/191992225-b17a5f46-7829-4a6b-9e5e-3fdfbd20fbd6.png">

## Desktop (please complete the following information)

- OS: Windows 11 (Windows_NT x64 10.0.22000)
- Browser (if with published websites; include version): [e.g. chrome, safari]
- Versions (all that apply to bug):
  - Dendron Extension: 0.113.1
  - VS Code version: 1.71.2
  - Dendron CLI:

## Additional context

Frontmatter of the first task in the screenshots:

```
---
id: 0xquc81kkfnp64r365ilg75
title: Roommate Agreement
desc: ''
updated: 1662224361382
created: 1662151842058
status: ''
[...]

PR #3640, opened 2022-10-06
fix(preview): task notes inside of note references should render correctly

kevinslin:
fix(preview): task notes inside of note references should render correctly #3640 

This is related to https://github.com/dendronhq/dendron/issues/3578


jonathanyeung (approved):
Nice catch, thanks for fixing!

PR #3641, opened 2022-10-06
enhance(workspace): Add last of missing methods to engine v3 + remove notes

tma66:
**Changes**
1. Add last remaining methods to the engine v3 (renderNote, getNoteBlocks, getDecorations)
2. Modified `isCachedPreviewUpToDate` logic so that it returns early if a note ref is newer (vs getting latest timestamp of all refs and comparing)
3. Removed `notes` and `notefnames` from engine interface
4. Removed `queryNotesSync` as its redundant


packages/common-all/src/types/typesv2.ts
```diff
@@ -329,18 +329,6 @@ export type BulkWriteNotesOpts = {
 // === Engine and Store Main
 
 export type DCommonProps = {
-  /**
-   * @deprecated
-   * For access, see {@link DEngine.getNote}
-   * Dictionary where key is the note id.
-   */
-  notes: NotePropsByIdDict;
```
jonathanyeung, line 337: woo hoo!

packages/engine-server/src/enginev2.ts
```diff
@@ -589,13 +571,86 @@ export class DendronEngineV2 implements DEngine {
     }
     // TODO: Add another check to see if backlinks have changed
 
-    return (
-      cachedPreview.updated >=
-      NoteUtils.getLatestUpdateTimeOfPreviewNoteTree({
-        rootNote: note,
-        notes: this.notes,
+    const visitedIds = new Set<string>();
+    return this._isCachedPreviewUpToDate({
```
jonathanyeung, line 575: Are the changes to this class bug fixes for regressions from refactoring?

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -1022,6 +1166,45 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
     };
   }
 
+  private createRenderedCache(
```
jonathanyeung, line 1169: I'm not opposed to leaving it since you've already added it, but I think it makes sense to get rid of the preview caching logic.  I'm not convinced it's actually causing any perceivable perf gains and it makes the render logic more complex without a good reason. 

From profiling I did in the past, if we can't render a note in <50 ms then we have a bug in our rendering logic and that's where we should make a fix.

packages/common-all/src/types/typesv2.ts
```diff
@@ -552,6 +531,18 @@ export type DEngineClient = Omit<DEngine, "store">;
 
 export type DStore = DCommonProps &
   DCommonMethods & {
+    /**
+     * @deprecated
```
jonathanyeung, line 535: Do we still need to add these back?

packages/plugin-core/src/services/EngineAPIServiceInterface.ts
```diff
@@ -39,16 +37,6 @@ import {
 
 export interface IEngineAPIService {
   trustedWorkspace: boolean;
-  /**
-   * @deprecated
-   * For accessing a specific note by id, see {@link IEngineAPIService.getNote}
-   * If you need all notes, avoid modifying any note as this will cause unintended changes on the store side
-   */
-  notes: NotePropsByIdDict;
```
jonathanyeung, line 47: woohoo!

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -859,21 +855,169 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
...
+    flavor,
+    dest,
+  }: RenderNoteOpts): Promise<RenderNoteResp> {
+    const ctx = "DEngine:renderNote";
+
+    // If provided, we render the given note entirely. Otherwise find the note in workspace.
+    if (!note) {
+      note = (await this.getNote(id)).data;
+    } else {
+      // `procRehype` needs the note to be in the engine, so we have to add it in case it's a dummy note
```
jonathanyeung, line 873: This actually shouldn't be necessary anymore since we stopped using the engine in `procRehype` - can we try removing this `this.writeNote` and see if it still works? 

If it's too much for now, maybe add a `TODO:` that captures this.

packages/engine-server/src/enginev2.ts
tma66: Not a bug fix; I just wanted to keep it consistent with v3 (even though we're gonna remove this shortly) and it didn't take a long time to change

packages/engine-server/src/DendronEngineV3.ts
tma66: I'm not familiar enough to make a decision here, but I would definitely lean towards simplicity if we can. Any thoughts @kevinslin

packages/common-all/src/types/typesv2.ts
tma66: This is needed for `storev2` to work as `enginev2` calls store.notes. This will be removed once both are removed

packages/engine-server/src/DendronEngineV3.ts
kevinslin: i would leave it in for now and remove it in a subsequent pr if we don't need it

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -298,7 +298,7 @@ export abstract class EngineV3Base implements ReducedDEngine {
     });
 
     if (items.length === 0) {
-      return { data: [] };
+      return [];
```
kevinslin, line 301: why does `queryNotes` not return a `RespV3` object like all the other engine methods?

kevinslin: (not a blocker - but we should consolidate the return methods in a subsequent PR if this isn't intentional)

kevinslin (approved):
looks good - a comment about queryNote interface which doesn't need to be updated in this PR

packages/common-all/src/engine/EngineV3Base.ts
tma66: I wanted `queryNotes` to behave the same as `findNotes` which currently also returns a list of results. We can change it all to RespV3 later if you feel its necessary.

jonathanyeung (approved):

packages/engine-server/src/DendronEngineV3.ts
hikchoi: just adding to the conversation that I have triaged some issues with preview where some nested note references were not being rendered due to caching issues.

PR #3642, opened 2022-10-07
chore/query interface kevin

kevinslin:
chore: implement IQueryStore

https://github.com/dendronhq/dendron/pull/3642

This is a slightly scaled back version of `prototype-query-interface-for-engine`
We are introducing an `IQueryStore` (formerly `IQueryable`) that is responsible for proxying all free text queries. 

Combining metadata store with the query store is still on the table but I wanted to limit the scope of this change and only combine it if necessary (will be better apparent when we try to merge SQLite)
Note that `IQueryStore` is non-generic by design, it has separate commands for `queryNote` and `querySchema` and returns `NotePropsMeta` for notes. 
Until we have need for a more generic interface, the current design optimizes for ease of integration. 

Also did some refactoring: 
- EngineBaseV3 now handles `queryNotes` for `DendronEngineV3`
- constructor for `EngineBaseV3` takes an object instead of positional params for easier refactoring

### DIFF
--- a/packages/common-all/src/engine/EngineV3Base.ts
+++ b/packages/common-all/src/engine/EngineV3Base.ts
@@ -6,7 +6,7 @@ import { ERROR_SEVERITY, ERROR_STATUS } from "../constants";
 import { DLogger } from "../DLogger";
 import { DNodeUtils, NoteUtils } from "../dnode";
 import { DendronCompositeError, DendronError } from "../error";
-import { IQueryStore, INoteStore, FuseMetadataStore } from "../store";
+import { INoteStore, IQueryStore } from "../store";
 import {
   BulkGetNoteMetaResp,
   BulkGetNoteResp,
--- a/packages/common-all/src/engine/EngineV3Base.ts
+++ b/packages/common-all/src/engine/EngineV3Base.ts
@@ -43,6 +43,7 @@ import { VaultUtils } from "../vault";
  * DendronEngineV3Web
  */
 export abstract class EngineV3Base implements ReducedDEngine {
+  // TODO: remove
   protected abstract fuseEngine: FuseEngine;
   protected noteStore;
   protected noteQueryable;
@@ -296,29 +297,28 @@ export abstract class EngineV3Base implements ReducedDEngine {
   async queryNotes(opts: QueryNotesOpts): Promise<QueryNotesResp> {
     // const ctx = "Engine:queryNotes";
     const { qs, vault, onlyDirectChildren, originalQS } = opts;
-
     // Need to ignore this because the engine stringifies this property, so the types are incorrect.
     // @ts-ignore
     if (vault?.selfContained === "true" || vault?.selfContained === "false")
       vault.selfContained = vault.selfContained === "true";
 
-    const items = await this.fuseEngine.queryNote({
-      qs,
+    const response = await this.noteQueryable.query(qs, {
       onlyDirectChildren,
       originalQS,
     });
-
+    if (response.isErr()) {
+      return {
+        error: response.error,
+      };
+    }
+    const items = response.value;
     if (items.length === 0) {
       return [];
     }
 
-    const notes = await Promise.all(
-      items.map((ent) => this.noteStore.get(ent.id)) // TODO: Should be using metadata instead
-    );
+    const notes = await this.noteStore.bulkGet(items.map((ent) => ent.id));
 
     let modifiedNotes;
-    // let notes = items.map((ent) => this.notes[ent.id]);
-    // if (!_.isUndefined(vault)) {
     modifiedNotes = notes
       .filter((ent) => _.isUndefined(ent.error))
       .map((resp) => resp.data!);
--- a/packages/common-all/src/engine/EngineV3Base.ts
+++ b/packages/common-all/src/engine/EngineV3Base.ts
@@ -6,7 +6,6 @@ import { ERROR_SEVERITY, ERROR_STATUS } from "../constants";
 import { DLogger } from "../DLogger";
 import { DNodeUtils, NoteUtils } from "../dnode";
 import { DendronCompositeError, DendronError } from "../error";
-import { FuseEngine } from "../FuseEngine";
 import { INoteStore } from "../store";
 import { INoteQueryable } from "../store/IDataQuery";
 import {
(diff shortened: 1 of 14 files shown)


packages/common-all/src/types/ReducedDEngine.ts
```diff
@@ -5,6 +5,7 @@ import { DEngine } from "./typesv2";
  */
 export type ReducedDEngine = Pick<
   DEngine,
+  | "wsRoot"
```
jonathanyeung, line 8: I don't think we should expose these kinds of properties anymore - engine interfaces should be focused on note/schema CRUD only, and any access to other things should be provided elsewhere.  This will help us avoid big monolithic objects like `DExtension`.

packages/common-all/src/store/IDataQuery.ts
```diff
@@ -0,0 +1,39 @@
...
+export interface IQueryStore {
+  queryNotes(
+    qs: string,
+    opts: INoteQueryOpts
+  ): ResultAsync<NotePropsMeta[], DendronError>;
+  querySchemas(
+    qs: string,
+    opts?: INoteQueryOpts
+  ): ResultAsync<{ id: string }[], DendronError>;
+  updateNotesIndex(changes: NoteChangeEntry[]): ResultAsync<void, DendronError>;
```
jonathanyeung, line 29: IMO we shouldn't be exposing these `update`/`replace`/`remove` methods, because they're exposing too much implementation detail through the interface.  For example - why does an `IQueryStore` need to use indexes? This interface assumes it, which greatly reduces the utility of the interface.

I get that if we decide to not combine the Fuse logic with the metadata store, we won't be able to remove the current Fuse references from `EngineV3*` classes without adding some methods like these - but then at that point, we aren't _really_ removing Fuse details from those classes since all the assumptions on how the store works still exist through this interface.

As I was coding this up, I also realized that a queryStore implementation would be hard to decouple from the metadata store CRUD methods, so perhaps a more sensible thing to do here is to have IQueryStore extend `IDataStore<string, NotePropsMeta>`.  Not sure yet - dunno if it's more sensible that way vs having `IQueryStore` just expose `queryNotes`/`querySchema`. 

Unrelated - but can we make this file name align with the interface name as per our new convention? i.e. `IQueryStore.ts`

packages/common-all/src/store/IDataQuery.ts
```diff
@@ -0,0 +1,39 @@
...
+import {
+  DendronError,
+  NoteChangeEntry,
+  NotePropsByIdDict,
+  NotePropsMeta,
+  SchemaModuleDict,
+  SchemaModuleProps,
+} from "..";
+
+export type INoteQueryOpts = {
```
jonathanyeung, line 11: I don't think we should expose `onlyDirectChildren` - it adds too much bespoke logic into the store/engine side of things and muddies the interface.  Instead, I think it's better to have the calling side apply these kinds of filtering. Same goes with the word `originalQS` - it should just be `queryString`

jonathanyeung (changes requested):

packages/common-all/src/store/IDataQuery.ts
kevinslin: i think all of these methods will still apply to any querystore. i don't want to create something generic until we have at least two implementations of it (aka - once we implement the SQLite interface, we will have a better idea of where to draw the lines). 

my thinking here is that this is the intermediary step. the interface is not *finalized* until we also implement the SQLIte store behind this interface.

ack on changing the name

packages/common-all/src/store/IDataQuery.ts
kevinslin: it might make sense but see comment above - i'm trying to avoid large refactorings with this change. this is meant to be an incremental refactor to get us to the desired state of having a more fully abstracted store

packages/common-all/src/types/ReducedDEngine.ts
kevinslin: so i needed to expose this because we need `wsRoot` to compare vaults for equality (sometimes vaults can have absolute file paths and so we need the wsRoot to compare them). we can mark it as protected if we don't want others to use them

packages/common-all/src/store/IDataQuery.ts
hikchoi: What is the benefit we are gaining from incrementally introducing a non-finalized interface when we only have one implementation of it? wouldn't it be better to wait for a second implementation to come, and then with the two extrapolate a generic interface instead of introducing more things to clean up in the future?

I can understand introducing some non-finalized methods in the interface if we are already dealing with more than one implementation of it. But I don't feel like we are gaining anything from incrementally refactoring at this point of time.

What is the timeline for implementing the SQLite store? If the gap between now and then is large enough, eventually these non-generics will be used in many places as if they were final. We've seen this in many places that we allowed intermediary non-finalized refactors (notably in config, though we could argue that it was necessary in config), and the cleanup was a heavy task. Given our total bandwidth, we are always hesitant on budgeting out the time to clean things up, which makes the matter even worse.

IMO, it would be okay if this PR was followed right away with the discussion for the SQLite store, and we immediately resolve the non-generic nature of it, but not if we don't have a reasonable ETA on when we are going to introduce the SQLite store.

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -285,41 +296,39 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
       vault.selfContained = vault.selfContained === "true";
 
-    const items = await this.fuseEngine.queryNote({
-      qs,
+    const response = await this.queryStore.queryNotes(qs, {
       onlyDirectChildren,
       originalQS,
     });
-
+    if (response.isErr()) {
```
namjul, line 308: On deciding which words/symbols to use to point to the underlying type I got used to use the term "result" to mean its a `Result` type or "resp" to mean to point to the `RespV3` type. Not wanting to do unnecessary bikeshedding here but wondering if we can converge on a name for this.

packages/common-all/src/store/IDataQuery.ts
jonathanyeung: @hikchoi - agree with your concerns.  I'll have an async with a proposal out soon on how to proceed.

jonathanyeung (approved):

packages/common-all/src/engine/EngineV3Base.ts
kevinslin: got it -> i'm good with this and will start adopting. can you add it to the conventions doc if its not already there?

PR #3646, opened 2022-10-07
Chore/circ dep checks for all packages

jonathanyeung:
## chore: circ dep checks for all packages

Adds madge circular dependency checks to all relevant non-test packages to the husky pre push hook.

I also fixed a few low-hanging fruit circular dependencies.


kevinslin (commented):
looks good - does this affect the latency of doing a push? 
if so, can we run all these checks in parallel?

hikchoi (approved):
lgtm.

maybe it's time we turn this into a CI run if we are starting to be concerned with the push latency. (but I think it's worth having it in pre-push when we are actively trying to keep it down)

jonathanyeung:
It still feels fast enough - not concerned about it at the moment, but as @hikchoi mentioned we can integrate it into CI later (which would be more robust anyway).

PR #3647, opened 2022-10-10
fix(internal): correct reading of error tack

namjul:
This PR's aim is to fix an error that occurs while [building nextjs-template](https://github.com/dendronhq/dendron/actions/runs/3201841712/jobs/5230211157#step:10:27).
These changes where missed when doing https://github.com/dendronhq/dendron/pull/3630

In addition this PR does a cleanup in `nextjs-template` so that it is possible to let it run through `run lerna:typecheck`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "format": "lerna exec --parallel -- yarn format",
     "format:pkg": "prettier --write",
     "lint": "eslint . --ext ts,tsx",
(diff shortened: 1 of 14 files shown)


kevinslin (approved):

Harshita-mindfire:
@namjul updating this changelog title to `chore(internal): apply correct accessing of error track` in the weekly journal

PR #3649, opened 2022-10-10
chore(publish): streamline dendron config availablity

namjul:
The aim of this PR is to add a validation layer for dendron config

[[dendron://private/task.2022.10.04.nextjs-template-dendronyml-validation]]

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [ ] lg: screen ≥ 992px
    - [ ] safari
    - [ ] firefox
    - [ ] chrome

## Docs

### Extended


hikchoi:
I'll wrap up #3618 asap so it doesn't block this! 🙏🏼

packages/common-all/src/parse.ts
```diff
@@ -26,7 +23,7 @@ export const parse = <T extends z.ZodTypeAny>(
       new DendronError({
         message: [
           ...(msg ? [msg] : []),
-          JSON.stringify(formatErrors(parsed.error.format()), null, 2),
```
namjul, line 29: Simplifying here the formatting. Lets return to formatting at a later moment where we actually define how that should look like.

packages/common-all/src/utils/index.ts
```diff
@@ -1385,6 +1402,77 @@ export class ConfigUtils {
...
+        // hierarchy?: { [key: string]: HierarchyConfig };
+        // duplicateNoteBehavior?: DuplicateNoteBehavior;
+        writeStubs: z.boolean(),
+        seo: z.object({}).passthrough(),
+        github: z.object({ enableEditLink: z.boolean() }).passthrough(),
+        // theme?: Theme;
+        // segmentKey?: string;
+        // cognitoUserPoolId?: string;
+        // cognitoClientId?: string;
+        enablePrettyLinks: z.boolean(),
```
namjul, line 1443: For now I have only included the required properties.

packages/common-all/src/utils/index.ts
```diff
@@ -1485,3 +1573,30 @@ export function globMatch(patterns: string[] | string, fname: string): boolean {
...
+  static fromStr(str: string, overwriteDuplicate?: boolean) {
+    return this.load(str, {
+      schema: YAML.JSON_SCHEMA,
+      json: overwriteDuplicate ?? false,
+    }) as Result<AnyJson, DendronError>;
+  }
+  static toStr(data: any) {
+    return this.dump(data, { indent: 4, schema: YAML.JSON_SCHEMA });
+  }
+}
```
namjul, line 1602: We could also extract these into its own file.

packages/common-all/src/types/configs/publishing/publishing.ts
```diff
@@ -89,6 +92,52 @@ export type GoogleAnalyticsConfig = {
...
+      // cognitoUserPoolId?: string;
+      // cognitoClientId?: string;
+      enablePrettyLinks: z.boolean().default(true),
+      // siteBanner?: string;
+      // giscus?: GiscusConfig;
+      // sidebarPath?: string | false;
+      searchMode: searchModeSchema.optional().default(SearchMode.LOOKUP),
+    })
+    .passthrough()
+);
```
namjul, line 139: Now we define the default values in 2 locations. One option would be to use the schema inside `genDefaultPublishingConfig` like so: 

```ts
export function genDefaultPublishingConfig(): DendronPublishingConfig {
  const defaultConfigResult = parse(publishingSchema, {})
  if (defaultConfigResult.isErr()) {
    throw new DendronError({
      message: ErrorMessages.formatShouldNeverOccurMsg(),
    });
  }
  return defaultConfigResult.value
}
```

This creats a small overhead since now we need to parse for accesing the default values. We could do this only once the file gets imported.

@kevinslin @jonathanyeung @hikchoi

packages/common-server/src/files.ts
```diff
@@ -210,3 +214,24 @@ export function removeMDExtension(nodePath: string) {
   }
   return nodePath;
 }
+
+export function readToJson(path: string) {
```
kevinslin, line 218: nitpick - why not just keep this as `readJSON`?

packages/common-all/src/types/configs/publishing/publishing.ts
kevinslin: @hikchoi has plans of a config service. we can do the parsing once when that service is in place. right now - i would just definite it in two places and make sure `[[dendron://dendron.docs/pkg.common-all.dev.cook.add-new-config]]` is updated

packages/nextjs-template/utils/build.ts
```diff
@@ -97,12 +102,24 @@ export function getNoteMeta(id: string): Promise<NoteProps> {
 }
 
 let _CONFIG_CACHE: DendronConfig | undefined;
-export function getConfig(): Promise<DendronConfig> {
+export async function getConfig(): Promise<DendronConfig> {
```
kevinslin, line 105: instead of defining this in `getConfig`, we should add the check in https://github.com/dendronhq/dendron/blob/fix/publish-nullish-publishing-config-field-runtime-error/packages/pods-core/src/builtin/NextjsExportPod.ts#L492:L492

its during the building stage that the config is inspected and acted upon

packages/common-all/src/utils/index.ts
```diff
@@ -6,7 +6,22 @@ import minimatch from "minimatch";
 import path from "path";
 import querystring from "querystring";
 import semver from "semver";
+import YAML from "js-yaml";
```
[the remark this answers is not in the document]
namjul: yea I did not see that. I moved the new yaml utils into common-server here [`278d893` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/278d8935684e86da23b5a54d8abd3ea3c6ea3222)

I think they would actually fit well under common-all because `js-yaml` does actually not depend on only server code. I just saw also that there exist other usages of `js-yaml` inside `common-all`. So I am considering moving that back into `common-all` and maybe out of `common-server`

packages/common-all/src/types/configs/publishing/publishing.ts
namjul: Maybe we should rename the "term" schema, as it clashes with our exsiting semantic meaning with "Dendron Schemas".

namjul:
@kevinslin I played with an idea in this commit [`3c1090e` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/3c1090ee653e41dda4f6fab7e5126a2fe0f30377). Its about using es6 modules instead of classes for namespacing purposes. I saw a couple of times the pattern of using classes and maybe we are having a "leakage" here where we started using them for defining module API surfaces. Do we want to address this pattern and add an explicit intend into the dev docs.

packages/common-all/src/types/configs/publishing/publishing.ts
```diff
@@ -89,6 +92,52 @@ export type GoogleAnalyticsConfig = {
...
+      copyAssets: z.boolean().default(true),
+      //
+      // canonicalBaseUrl?: string;
+      // customHeaderPath?: string;
+      // ga?: GoogleAnalyticsConfig;
+      // logoPath?: string;
+      // siteFaviconPath?: string;
+      // siteIndex?: string;
+      siteHierarchies: z.array(z.string()).default(["root"]),
+      writeStubs: z.boolean().default(false),
```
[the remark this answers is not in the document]
namjul: I had to look that up ;) so it will allow the input to be `undefined` but the output to be non-optional. So it will not throw but return the field (after being parsed) with the default value.

I added this to [[dendron://dendron.docs/pkg.common-all.ref.schemas#optional-vs-default]]

packages/common-all/src/utils/index.ts
```diff
@@ -6,7 +6,22 @@ import minimatch from "minimatch";
...
+  err,
+  Err,
+  Result,
+  okAsync,
+  errAsync,
+  ResultAsync,
+  fromThrowable,
+  fromPromise,
+  fromSafePromise,
+} from "neverthrow";
```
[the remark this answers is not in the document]
namjul: b43c5e981

packages/common-server/src/files.ts
namjul: I don't have a good answer ;) [`8cd6916` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/8cd6916f29f967f24d0e2b80d148bfc5948f348f). I created a tweet around this lol https://twitter.com/Namgnoi/status/1581974495107043328

packages/nextjs-template/utils/build.ts
namjul: Yes this extra parsing should not be necessary since it is already done in `NExtjsExportPod`. I removed that change in nextjs-template [`27cb739` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/27cb739d6762447403630e42b3016e4059d401bb)

packages/common-all/src/sidebar.ts
```diff
@@ -1,5 +1,5 @@
 import _ from "lodash";
-import { z } from "zod";
+import { z } from "./util/zodUtil";
```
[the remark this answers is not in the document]
namjul: I removed the `zodUtil` and moved zod/schema/parse related utils into the `parse.ts` file. I think this way we achive a higher cohesion around these javascript modules.

[`3758646` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/3758646d894cd1474e8b16af1225475039016a8c)
[`b318250` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/b31825075871ad1683f9869f54ba4fdeb3339bc5)

packages/common-all/src/utils/index.ts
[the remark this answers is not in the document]
namjul: I moved the new yaml utils into `common-all` https://github.com/dendronhq/dendron/pull/3649/commits/8cc9abdbe0e46d434262d00d64752ae802184d8d

These are all the other uses of `js-yaml` which are not inside `common-server` and therefore access it because of monorepo and not because its an actual depedency.
```
packages/engine-server/src/drivers/file/schemaParser.ts|11 col 19| import YAML from "js-yaml";
packages/unified/src/remark/extendedImage.ts|8 col 19| import YAML from "js-yaml";
packages/common-server/src/filesv2.ts|22 col 36| import YAML, { JSON_SCHEMA } from "js-yaml";
packages/common-server/src/files.ts|3 col 19| import YAML from "js-yaml";
packages/plugin-core/src/web/injection-providers/getWorkspaceConfig.ts|3 col 19| import YAML from "js-yaml";
packages/plugin-core/src/web/test/helpers/WorkspaceHelpers.ts|3 col 19| import YAML from "js-yaml";
packages/unified/src/decorations/diagnostics.ts|12 col 19| import YAML from "js-yaml";
```

I find this use https://github.com/Thinkmill/manypkg#checks which is a kind of monorepo linter that would prevent these kind of things.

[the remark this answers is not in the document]
namjul: I would suggest to fix them, but in a seperate PR.

packages/common-all/src/types/configs/publishing/publishing.ts
```diff
@@ -89,6 +92,52 @@ export type GoogleAnalyticsConfig = {
...
+      siteRootDir: z.string().default("docs"),
+      seo: seoSchema,
+      github: githubSchema,
```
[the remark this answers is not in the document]
namjul: [`1380881` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/1380881afc8ac82bee9d137099f6ec6ddb403ff8)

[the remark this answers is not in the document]
namjul: While coding on that change I found some duplication around gitutils. I took the initiative to  start some refactoring around that here https://github.com/dendronhq/dendron/pull/3681

packages/common-all/src/types/configs/publishing/publishing.ts
kevinslin: can we refer to types enforced by zod as zod schemas vs dendron schemas?

packages/common-all/src/utils/index.ts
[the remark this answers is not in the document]
kevinslin: sounds good

packages/common-server/src/files.ts
kevinslin: haha, nice

kevinslin:
> @kevinslin I played with an idea in this commit [`3c1090e` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/3c1090ee653e41dda4f6fab7e5126a2fe0f30377). Its about using es6 modules instead of classes for namespacing purposes. I saw a couple of times the pattern of using classes and maybe we are having a "leakage" here where we started using them for defining module API surfaces. Do we want to address this pattern and add an explicit intend into the dev docs.

it looks like that commit is no longer available

kevinslin (approved):

namjul:
> > @kevinslin I played with an idea in this commit [`3c1090e` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/3c1090ee653e41dda4f6fab7e5126a2fe0f30377). Its about using es6 modules instead of classes for namespacing purposes. I saw a couple of times the pattern of using classes and maybe we are having a "leakage" here where we started using them for defining module API surfaces. Do we want to address this pattern and add an explicit intend into the dev docs.
> 
> it looks like that commit is no longer available

This is the commit: [`364b646` (#3649)](https://github.com/dendronhq/dendron/pull/3649/commits/364b646381d19ff019c97a00353310a0c4ce5009)

Issue #3648, opened 2022-10-10
Vault Remove command doesn't updates the folders in code-workspace file


Harshita-mindfire:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

`Vault Remove command does not remove the folder from code workspace file.

## To Reproduce

1. In a multi vault workspace, run Vault Remove command
2. see vault is removed from dendron.yml, but not from code workspace.

## Expected behavior
vault folder should be removed from both code workspace and dendron.yml

## Desktop (please complete the following information)

- Browser (if with published websites; include version): [e.g. chrome, safari]
- Versions (all that apply to bug):
  - Dendron Extension: 0.114
  - VS Code version: 1.71
  - Dendron CLI:

## Dendron Log file

Please attach the output of `> Dendron:Dev: Open Logs` here

hikchoi:
>  This can be a windows only issue

Yup. I can't reproduce this on a mac. It indeed might be a windows issue.

hikchoi:
@Harshita-mindfire since you already assigned yourself: [[Vault Remove Command Doesnt Updates the Folders in Code Workspace File|dendron://private/task.2022.10.11.vault-remove-command-doesnt-updates-the-folders-in-code-workspace-file]]

PR #3655, opened 2022-10-12
enhance(workspace): support adding existing vault to the workspace

Harshita-mindfire:
This PR aims to 
- fix an windows path issue: https://github.com/dendronhq/dendron/issues/3648
- Adds a new `Create New Vault command`
- Adds a new `Add Existing Vault command`
- Adds a deprecation toaster for `Vault Add` command
- Updates `Vault Remove, Vault Convert` command to have {Verb} {noun} format: `Convert Vault, Remove Vault`
- Docs: https://github.com/dendronhq/dendron-site/pull/675

## Summary
- Create New Vault
when config option `enableSelfConatinedVaults` is set to false: prompt user to select destination folder, create new vault there.
when enableSelfConatinedVaults is set to true, always create a new vault at dependencies/localhost.

- Add Existing vault 
when config option `enableSelfConatinedVaults` is set to false and is a local vault, add the existing vault to the workspace, when enableSelfConatinedVaults is set to false, copy the destination folder to dependencies/localhost and update this path in `dendron.yml`

**BREAKING CHANGES**: Vault Remove, Vault Convert commands are now Remove Vault and Convert Vault respectively. Vault Add is now deprecated and will be removed in future.

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/AddExistingVaultCommand.ts
+++ b/packages/plugin-core/src/commands/AddExistingVaultCommand.ts
@@ -179,13 +179,22 @@ export class AddExistingVaultCommand extends BasicCommand<
     sourceType: VaultRemoteSource
   ): Promise<CommandOpts | undefined> {
     if (sourceType === VaultType.LOCAL) {
-      const vaultDestination = await this.gatherDestinationFolder();
-      if (!vaultDestination) return;
+      const sourcePath = await this.gatherDestinationFolder();
+      if (!sourcePath) return;
+      const placeHolder = path.basename(sourcePath);
+      const sourceName =
+        (await VSCodeUtils.showInputBox({
+          prompt: "Name of new vault (optional, press enter to skip)",
+          placeHolder,
+        })) || placeHolder;
 
-      const sourceName = await VSCodeUtils.showInputBox({
-        prompt: "Name of new vault (optional, press enter to skip)",
-        placeHolder: path.basename(vaultDestination),
-      });
+      const vaultDestination = path.join(
+        this._ext.getDWorkspace().wsRoot,
+        FOLDERS.DEPENDENCIES,
+        FOLDERS.LOCAL_DEPENDENCY,
+        sourceName
+      );
+      await fs.copy(sourcePath, vaultDestination);
 
       return {
         type: sourceType,
--- a/packages/plugin-core/src/commands/AddExistingVaultCommand.ts
+++ b/packages/plugin-core/src/commands/AddExistingVaultCommand.ts
@@ -160,7 +160,7 @@ export class AddExistingVaultCommand extends BasicCommand<
 
   async gatherDestinationFolder() {
     const defaultUri = Uri.file(this._ext.getDWorkspace().wsRoot);
-    // Prompt user where to create new vault
+    // opens the workspace root by default and prompts user to select vault
     const options: OpenDialogOptions = {
       canSelectMany: false,
       openLabel: "Select vault to add",
--- a/packages/plugin-core/src/commands/AddExistingVaultCommand.ts
+++ b/packages/plugin-core/src/commands/AddExistingVaultCommand.ts
@@ -89,10 +89,10 @@ export class AddExistingVaultCommand extends BasicCommand<
 
   /** A regular, non-self contained vault. */
   async gatherVaultStandard(
-    sourceType: VaultRemoteSource,
-    vaultDestination: string
+    sourceType: VaultRemoteSource
   ): Promise<CommandOpts | undefined> {
     let sourceName: string | undefined;
+    let vaultDestination: string | undefined;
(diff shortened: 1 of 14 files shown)


packages/plugin-core/src/commands/AddExistingVaultCommand.ts
```diff
@@ -0,0 +1,504 @@
...
+        path: vaultDestination,
+        pathRemote: remote,
+        isSelfContained: true,
+      };
+    }
+  }
+
+  async gatherInputs(): Promise<CommandOpts | undefined> {
+    const sourceTypeSelected = await VSCodeUtils.showQuickPick([
+      { label: VaultType.LOCAL, picked: true },
```
kevinslin, line 210: can we add a `description` string for both items?

- local: A local vault is a Dendron vault that is present in your computer
- remote: A remote vault is a Dendron vault that is available at a git endpoint

also add a `detail` string to show an example

- local: `eg. /home/dendron/hello-vault`
- remote: `eg. git@github.com:dendronhq/dendron-site.git`

packages/plugin-core/src/commands/AddExistingVaultCommand.ts
```diff
@@ -0,0 +1,504 @@
...
+      path: vaultDestination,
+    };
+  }
+
+  async gatherDestinationFolder() {
+    const defaultUri = Uri.file(this._ext.getDWorkspace().wsRoot);
+    // Prompt user where to create new vault
+    const options: OpenDialogOptions = {
+      canSelectMany: false,
+      openLabel: "Select vault to add",
```
kevinslin, line 151: Open an existing folder of markdown files

packages/plugin-core/src/commands/AddExistingVaultCommand.ts
```diff
@@ -0,0 +1,504 @@
...
+        await wsService.createSelfContainedVault({
+          vault,
+          addToConfig: true,
+          addToCodeWorkspace: false,
+          newVault: false,
+        });
+      } else {
+        await wsService.createVault({ vault });
+      }
+      await this.addVaultToWorkspace(vault);
```
kevinslin, line 497: this part has a lot of overlap with `CreateNewVault`. can we combine some of this common logic in a `VaultUtils` class?

packages/plugin-core/src/commands/AddExistingVaultCommand.ts
```diff
@@ -0,0 +1,504 @@
...
+  async gatherInputs(): Promise<CommandOpts | undefined> {
+    const sourceTypeSelected = await VSCodeUtils.showQuickPick([
+      { label: VaultType.LOCAL, picked: true },
+      { label: VaultType.REMOTE },
+    ]);
+    if (!sourceTypeSelected) {
+      return;
+    }
+    const sourceType = sourceTypeSelected.label;
+    const vaultDestination = await this.gatherDestinationFolder();
```
kevinslin, line 217: should not run this if picking remote vault

kevinslin (changes requested):
thanks for also taking care of some of the naming issues. left some comments

packages/plugin-core/src/commands/AddExistingVaultCommand.ts
```diff
@@ -0,0 +1,504 @@
...
+          newVault: false,
+        });
+      } else {
+        await wsService.createVault({ vault });
+      }
+      await this.addVaultToWorkspace(vault);
+      vaults = [vault];
+    }
+    window.showInformationMessage("finished adding vault");
+    await commands.executeCommand("workbench.action.reloadWindow");
```
hikchoi, line 501: maybe we should switch the order of these two, since showing an info message right before reload would remove the message.

hikchoi: same comment for the other commands as well

hikchoi (changes requested):
looks good other than some nits.

jonathanyeung (approved):

hikchoi (approved):

packages/plugin-core/src/commands/AddExistingVaultCommand.ts
Harshita-mindfire: opens the workspace root by default

packages/plugin-core/src/commands/AddExistingVaultCommand.ts
Harshita-mindfire: This didn't work out as helpful while combining because `wsService` and other common part that is present has dependency on `engine-server` which is not a dependency in common-all(VaultUtils)

PR #3678, opened 2022-10-16
enhance: experimental inline note references

kevinslin:
enhance: experimental inline note references

Renders note references inline as comments that display HTML. This makes it possible to use Dendron note references without using the preview
See loom for example: https://www.loom.com/share/458300e3a66f4753aa5cb1875bf4cd60?from_recorder=1&focus_title=1

---

This PR also addresses some other issues (aka scope creep):
- enhance(preview): slight update on hover preview behavior: don't render note title (it takes up a lot of vertical space and is not necessary since your hovering over the note)
- enhance(workspace): perf improvement when shifting focus to non-file panels. window watcher will currently call `updateDecorator` and `refreshBacklinks` on **all** editors (eg. terminal, output pane, etc) 
- fix(commands): urls were broken in hover
- chore: local build failing because of insufficient timeout between publishing to local registry and trying to install from it

### DIFF
--- a/packages/plugin-core/src/ExtensionProvider.ts
+++ b/packages/plugin-core/src/ExtensionProvider.ts
@@ -26,6 +26,10 @@ export class ExtensionProvider {
     return ExtensionProvider.extension;
   }
 
+  static getState() {
+    return ExtensionProvider.extension.getState();
+  }
+
   static getDWorkspace() {
     return ExtensionProvider.getExtension().getDWorkspace();
   }
--- a/packages/plugin-core/src/ExtensionProvider.ts
+++ b/packages/plugin-core/src/ExtensionProvider.ts
@@ -26,8 +26,8 @@ export class ExtensionProvider {
     return ExtensionProvider.extension;
   }
 
-  static getState() {
-    return ExtensionProvider.extension.getState();
+  static getCommentThreadsState() {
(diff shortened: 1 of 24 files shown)


packages/plugin-core/src/ExtensionProvider.ts
```diff
@@ -26,6 +26,10 @@ export class ExtensionProvider {
     return ExtensionProvider.extension;
   }
 
+  static getState() {
+    return ExtensionProvider.extension.getState();
```
jonathanyeung, line 30: instead of exposing a method like getState, can we just make it specific to getInlineNoteRefs to enable this feature?  This is moving us back towards a global state pattern.

Even better if we can make the inlineNoteRef data non-global.

packages/plugin-core/src/features/windowDecorations.ts
```diff
@@ -114,6 +180,120 @@ export const debouncedUpdateDecorations = debounceAsyncUntilComplete({
   trailing: true,
 });
 
+export class NoteRefComment implements Comment {
```
jonathanyeung, line 183: as per convention, can you move this to its own file - or at least the new code added here? This file is getting long.

jonathanyeung (changes requested):

jonathanyeung (approved):

PR #3680, opened 2022-10-17
chore/webpack build for sqlite

jonathanyeung:
## chore/webpack build for sqlite

Sets up our webpack build process to be able to bundle sqlite3 and its native node module properly.

How it works:
- we override the `sqlite3-binding.js` of the sqlite3 package. The original binding file doesn't work with webpack out of the box - the overridden version points to a specific path in the bundle where we manually put the native node module. 
- The native node module (the file called `node_sqlite3.node`) is platform specific (win, macos, linux; x64 vs arm), which means we need to now have platform specific versions of our extension. Thus we need to use the [Platform Specific Extensions](https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions) feature of vsce.

How the build works:
1. During binary install time, we add another step that uses node-pre-gyp to download the correct version of `node_sqlite3.node`. This is a bit of a hack - node-pre-gyp is intended to be used by the native module's repo to compile and distribute the native node package, but by adding a 'binary' prop in our package.json that mimics that of sqlite3 we can ensure that we're pulling down the right binary. If we ever need to use any additional native node modules however, we'll need to find a different mechanism as this only works if you have one dependency like this.
2. During webpack, the downloaded `node_sqlite3.node` file gets copied to the appropriate location via `CopyPlugin` and gets bundled.
3. In the Github Actions workflow, we need to publish multiple binaries for each platform and architecture flavor. We do this with a job matrix following VSCode's template here: https://github.com/microsoft/vscode-platform-specific-sample


namjul:
I discussed with @jonathanyeung last week the idea to use an wasm compiled sqlite so that we don't have to deal with different plattform specific builds. Looks like the current options for this are:
- https://github.com/rhashimoto/wa-sqlite/
- https://github.com/sql-js/sql.js/
While looking for a way how this could be used as a different data source in prisma I only found https://github.com/prisma/prisma/issues/6576 and no page in the docs around custom database connectors. Looking into their code they seem to only support hardcoded datasources https://github.com/prisma/prisma/blob/7a13f555388e2f4f8e00aab695f38b83eb9f3952/packages/client/tests/functional/_utils/providers.ts

To conclude, this option does not look to be an option with prisma.

.github/workflows/hackathon-create-test-image.yml
```diff
@@ -35,11 +35,10 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v2
         with:
-          node-version: 16.x
+          node-version: 14.x
 
       - name: Yarn Setup
         run: yarn setup
-        env: npm_config_arch: arm64
```
kevinslin, line 42: heads up that @nam has made a fix for this (you'll want to rebase)

packages/plugin-core/package.json
```diff
@@ -1429,6 +1429,17 @@
...
   },
   "icon": "media/logo.png",
+  "binary": {
+    "module_name": "node_sqlite3",
+    "module_path": "./lib/binding/napi-v{napi_build_version}-{platform}-{libc}-{arch}",
+    "host": "https://github.com/TryGhost/node-sqlite3/releases/download/",
+    "remote_path": "v5.1.2",
+    "package_name": "napi-v{napi_build_version}-{platform}-{libc}-{arch}.tar.gz",
+    "napi_versions": [
+      3,
```
kevinslin, line 1439: nitpick but do you know if we need to support node api verison 3 if our min node version is set to 16?

kevinslin (approved):
nice work! 
can you add a line in [[dendron://dendron.docs/pkg.plugin-core.dev.build]] about the extra step for local build?

jonathanyeung:
@namjul - thanks for the links and the research. I think let's try out the native modules first because I think there should be a performance advantage over the web assembly versions, but we'll need to adopt something similar for the web extension at some point anyway. 

With respect to prisma - we're actually looking to remove the ORM layer so that won't be a concern for compatibility considerations going forward.

packages/plugin-core/package.json
jonathanyeung: removed

PR #3683, opened 2022-10-18
chore: workflow update for 2 week release cycle and cleanup

jonathanyeung:
## chore: workflow update for 2 week release cycle and cleanup

A few updates to our github workflows:
- Create Release Branch now runs on Wednesdays 7pm UTC every other week
- Removing AWS upload steps which were failing from an auth issue
- Updating actions to use v3 to get rid of node 12 warnings, which are getting deprecated in Github Actions.


namjul (approved):
lgtm

PR #3687, opened 2022-10-18
chore(workspace): Bug and test fixes for Engine v3

tma66:
This is the first portion of the changes to move to engine v3. While making the switch, it uncovered several test and engine bugs that this pr will fix. The next pr will move to engine v3 under a feature flag.

**Changes**
1. Add missing candidate links feature to engine v3
2. Improve init speed of v3
3. Include `contentHash` as part of `NotePropsMeta`
4. Extend NoteStore and DataStore to be `Disposable`. This will allow the engine to refresh from clean state everytime init is called

### DIFF
--- a/packages/common-all/src/store/INoteStore.ts
+++ b/packages/common-all/src/store/INoteStore.ts
@@ -1,4 +1,5 @@
 import {
+  Disposable,
   DNoteLoc,
   NoteProps,
   NotePropsMeta,
@@ -11,7 +12,7 @@ import { FindNoteOpts } from "../types/FindNoteOpts";
 /**
  * Interface responsible for interacting with NoteProps storage layer
  */
-export interface INoteStore<K> {
+export interface INoteStore<K> extends Disposable {
   /**
    * Get NoteProps by key
    * If key is not found, return error.
--- a/packages/engine-server/src/DendronEngineV3.ts
+++ b/packages/engine-server/src/DendronEngineV3.ts
@@ -139,11 +139,7 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
 
   static create({ wsRoot, logger }: { logger?: DLogger; wsRoot: string }) {
     const LOGGER = logger || createLogger();
-    const { error, data: config } =
-      DConfig.readConfigAndApplyLocalOverrideSync(wsRoot);
-    if (error) {
-      LOGGER.error(stringifyError(error));
-    }
+    const config = DConfig.readConfigSync(wsRoot);
 
     const queryStore = new FuseQueryStore();
     const fileStore = new NodeJSFileStore();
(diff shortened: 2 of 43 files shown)


jonathanyeung (approved):

packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -172,12 +172,15 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
    * Does not throw error but returns it
    */
   async init(): Promise<DEngineInitResp> {
+    const { data: config } = DConfig.readConfigAndApplyLocalOverrideSync(
+      this.wsRoot
+    );
```
hikchoi, line 177: I'm not sure if it is safe to use `readConfigAndApplyLocalOverrideSync` yet.
We've recently identified where this causes a problem (where overrides were being merged back into `dendron.yml`) and fixed it, but there still may be places where that still happens. (related wip task: [[Make Sure Config Reads and Writes Dendronyml Correctly|dendron://private/task.2022.10.02.make-sure-config-reads-and-writes-dendronyml-correctly]])

It might be safe to keep this as-is or not apply overrides.

hikchoi: (same with all config read changes in this branch)

kevinslin (approved):
looks good - nothing to add besides existing comments

packages/common-all/src/store/INoteStore.ts
```diff
@@ -11,7 +12,7 @@ import { FindNoteOpts } from "../types/FindNoteOpts";
 /**
  * Interface responsible for interacting with NoteProps storage layer
  */
-export interface INoteStore<K> {
+export interface INoteStore<K> extends Disposable {
```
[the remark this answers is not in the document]
tma66: In the `init` method of the engine v3

packages/engine-server/src/drivers/file/storev2.ts
```diff
@@ -1541,22 +1538,19 @@ export class FileStorage implements DStore {
...
-      const notes = NoteDictsUtils.findByFname(
-        link.to.fname,
-        { notesById: this.notes, notesByFname: this.noteFnames },
-        maybeVault
-      );
+      const notes = NoteDictsUtils.findByFname({
+        fname: link.to.fname,
+        noteDicts: { notesById: this.notes, notesByFname: this.noteFnames },
+        vault: maybeVault,
+        skipCloneDeep: true,
```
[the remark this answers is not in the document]
tma66: The optimization is actually for just engine v3 when we're adding backlinks as part of the init. storev2 doesn't have this problem as its already modifying the referenced note as part of _addBacklinksImpl.

In addBacklinks of engine v3, we clone each note which isn't necessary

packages/engine-server/src/DendronEngineV3.ts
tma66: Oh got it, thanks for letting me know. I forget if one of the tests depends on the local overrides, but that can be worked around

PR #3690, opened 2022-10-19
Chore/nightly build tweaks

jonathanyeung:
## chore/nightly build tweaks

Some fixes for the nightly build:
- vsce publish doesn't support using both the --target and --packagePath arguments at the same time for some reason, so we need to specify --target at package time.
- Fixed issues with linux build logic
- Removed windows arm, which isn't available right now for sqlite3.

### DIFF
--- a/bootstrap/scripts/buildNightly.sh
+++ b/bootstrap/scripts/buildNightly.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Build plugin for CI and release
+# TODO: `buildPatch.sh` is a bad name - it should just be `buildPlugin`
+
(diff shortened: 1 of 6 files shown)


hikchoi (approved):

bootstrap/scripts/buildNightly.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# Build plugin for CI and release
+# TODO: `buildPatch.sh` is a bad name - it should just be `buildPlugin`
```
kevinslin, line 3: nitpick: remove this stale comment?

kevinslin (approved):

Issue #3701, opened 2022-10-21
hash tags are rendered as Private notes


namjul:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [x] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features

## Describe the bug

using a hashtag in this case `#test`  renders the link marked as private. but as you can see in the screenshot the tag exists.
![image](https://user-images.githubusercontent.com/328836/197188567-a9ec8351-3e5a-4de3-80c6-9b0d71fa6819.png)

using [[tags.test]] works.

## Desktop (please complete the following information)

- Browser (if with published websites; include version): [e.g. chrome, safari]
- Versions (all that apply to bug):
  - Dendron Extension:
  - VS Code version:
  - Dendron CLI:

## Dendron Log file

Please attach the output of `> Dendron:Dev: Open Logs` here

namjul:
@jonathanyeung hunch is that https://github.com/dendronhq/dendron/blob/fe4c475bfc9093e1eafd2e5f4e05805a872e6ff9/packages/unified/src/utilities/getParsingDependencyDicts.ts#L126 points to the problem

needs to be changed from hashtag.value to hashtag.fname.

namjul:
This issue caused the playwright tests to fail so I disabled the specific test case temporarily here [`478ca8e` (#3689)](https://github.com/dendronhq/dendron/pull/3689/commits/478ca8ea1470e0c059765d810a4069bc45029b3a) until we fix that.

PR #3708, opened 2022-10-24
fix(scope?): correct hashtag parsing by using relevant property

namjul:
The aim of this PR is to fix https://github.com/dendronhq/dendron/issues/3701

Additionally a temporarily disabled e2e test-case is enabled again.

# Pull Request Checklist

### DIFF
--- a/packages/unified/src/utilities/getParsingDependencyDicts.ts
+++ b/packages/unified/src/utilities/getParsingDependencyDicts.ts
@@ -129,7 +129,7 @@ function getNoteDependencies(ast: Node<Data>): DNodeCompositeKey[] {
 
   visit(ast, [DendronASTTypes.USERTAG], (noteRef: UserTag, _index) => {
     renderDependencies.push({
-      fname: noteRef.value,
+      fname: noteRef.fname,
     });
   });
 
--- a/packages/unified/src/utilities/getParsingDependencyDicts.ts
+++ b/packages/unified/src/utilities/getParsingDependencyDicts.ts
@@ -123,7 +123,7 @@ function getNoteDependencies(ast: Node<Data>): DNodeCompositeKey[] {
 
   visit(ast, [DendronASTTypes.HASHTAG], (hashtag: HashTag, _index) => {
     renderDependencies.push({
-      fname: hashtag.value,
+      fname: hashtag.fname,
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/unified/src/utilities/getParsingDependencyDicts.ts
```diff
@@ -123,7 +123,7 @@ function getNoteDependencies(ast: Node<Data>): DNodeCompositeKey[] {
 
   visit(ast, [DendronASTTypes.HASHTAG], (hashtag: HashTag, _index) => {
     renderDependencies.push({
-      fname: hashtag.value,
+      fname: hashtag.fname,
     });
   });
```
namjul, line 129: @jonathanyeung shouldn't the `USERTAG` "visit" also use `fname` instead of `value`. I mean that yes we can be sure `value` must match the `fname` value but we could also refactor this to directly check for `fname` if that is anyway what is what we are looking for.

namjul:
@jonathanyeung for some reason the CI is not triggered?

packages/unified/src/utilities/getParsingDependencyDicts.ts
jonathanyeung: Could be - not 100% sure.  One issue is that I don't think we're consistent in how we're using the fields across each AST Type (if I recall, what value gets stored in which property is a bit different across Hashtag/UserTag/Wikilink/etc.).  

Can you check 'usertag' under the debugger to see if it should be `usertag.value` or `usertag.fname`?

jonathanyeung:
> @jonathanyeung for some reason the CI is not triggered?

Huh - not sure why it's not triggering, that's strange.  Let's add an ability to manually trigger CI workflows when these situations happen.

packages/unified/src/utilities/getParsingDependencyDicts.ts
namjul: The value at https://github.com/dendronhq/dendron/blob/0db6efdbe7affa63a7a708303f2470b7dab9832e/packages/unified/src/utilities/getParsingDependencyDicts.ts#L132 is `"@example.username"` meaning that we also have an issue with usertags and this should also point to `fname` instead.

namjul: image
Yes so this issue is also appearing with usertags

namjul: fix for usertag: https://github.com/dendronhq/dendron/pull/3708/commits/3594eeaed3b0508967f15f9dfbab9fa0618f6cf6

kevinslin (approved):

PR #3709, opened 2022-10-24
chore(internal): Implement config store

hikchoi:
WIP

# Pull Request Checklist


packages/common-all/src/store/IConfigStore.ts
```diff
@@ -0,0 +1,59 @@
...
+import { IDendronError } from "../error";
+import { DendronConfig, DendronConfigValue } from "../types";
+import { DeepPartial } from "../utils";
+
+export type ConfigValue = string | number | object;
+
+export type ConfigReadMode = "override" | "default";
+export type ConfigReadOpts = {
+  mode: ConfigReadMode;
+  useCache?: boolean;
```
jonathanyeung, line 11: When would a caller not want to use a cache? Can this be safely removed from the read() params?

packages/common-all/src/store/IConfigStore.ts
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Given a dot delimited path, update the config entry value with given value in the persistant dendron config
+   * returns previous value
+   */
+  update(
+    key: string,
+    value: DendronConfigValue
+  ): ResultAsync<DendronConfigValue | undefined, IDendronError>;
+  /**
+   * Given a dot delimited path, delete the config entry in the persistant dendron config
```
jonathanyeung, line 53: nit on comment: why 'dot delimited path' - shouldn't it just be any path?

packages/common-all/src/store/IConfigStore.ts
```diff
@@ -0,0 +1,59 @@
...
+   * returns previous value
+   */
+  delete(
+    key: string
+  ): ResultAsync<DendronConfigValue | undefined, IDendronError>;
```
jonathanyeung, line 58: can this just be ResultAsync<DendronConfigValue, IDendronError>?  If the key is not present, can we propagate that back via the error instead?

packages/common-all/src/store/ConfigStore.ts
```diff
@@ -0,0 +1,336 @@
...
+    return Utils.joinPath(this._wsRoot, CONSTANTS.DENDRON_CONFIG_FILE);
+  }
+
+  constructor(fileStore: IFileStore, wsRoot: URI, homeDir: URI | undefined) {
+    this._fileStore = fileStore;
+    this._wsRoot = wsRoot;
+    this._homeDir = homeDir;
+  }
+
+  create(
```
jonathanyeung, line 29: should we maybe name this `createConfig` instead? 

Maybe just a cause of our code base conventions, but reading this makes me think that it's some factory method of `ConfigStore`.

packages/common-all/src/store/ConfigStore.ts
```diff
@@ -0,0 +1,336 @@
...
+            throw resp.error;
+          }
+          return config;
+        })
+      ),
+      (err) => err as DendronError
+    );
+  }
+
+  readRaw(): ResultAsync<
```
jonathanyeung, line 57: Hmm, maybe it's just better to do `Promise<Result<T, E>>` instead of use their ResultAsync construct. 

On the calling side, it's not too bad?  You can then just 

await readRaw().andThen()...

packages/common-all/src/store/IConfigStore.ts
```diff
@@ -0,0 +1,59 @@
...
+  useCache?: boolean;
+};
+
+export type ConfigGetOpts = ConfigReadOpts;
+
+export interface IConfigStore {
+  // entire config
+
+  /**
+   * Create a persistant dendron config
```
jonathanyeung, line 20: super nit: sp on persistent.

packages/common-all/src/store/IConfigStore.ts
hikchoi: Sorry the docstring is a bit confusing

I'm talking about the path arg you pass into `_.get()`: https://lodash.com/docs/4.17.15#get

I'll add some examples to make that clearer

packages/common-all/src/store/ConfigStore.ts
hikchoi: good point. maybe all methods should have that appended.

packages/common-all/src/store/ConfigStore.ts
hikchoi: that part is also confusing for me as well 😄
maybe I'm just using it wrong?

packages/common-all/src/store/IConfigStore.ts
namjul: Yea notice now too that we could now use the error track instead of returning `undefined`. So lets say we don't want this methode to signal an error when the key does not exist and rather signal a noop then making that more explicit using a kind of noop type instead of `undefined`. Not sure about which one makes more sense.

packages/common-all/src/store/ConfigStore.ts
```diff
@@ -0,0 +1,336 @@
+import { errAsync, okAsync, ResultAsync } from "neverthrow";
+import { DendronError, IDendronError } from "../error";
+import { URI, Utils } from "vscode-uri";
+import { ConfigReadOpts, IConfigStore } from "./IConfigStore";
+import { IFileStore } from "./IFileStore";
+import { CONSTANTS, StatusCodes } from "../constants";
+import { ConfigUtils, DeepPartial } from "../utils";
+import { DendronConfig, DendronConfigValue, DVault } from "../types";
+import { fromStr, toStr } from "../yaml";
```
namjul, line 9:
```suggestion
import * as YamlUtils from "../yaml";
```
This way the context of the functions is visible.

packages/common-all/src/store/ConfigStore.ts
```diff
@@ -0,0 +1,336 @@
...
+  ): ResultAsync<DendronConfig, IDendronError> {
+    const config: DendronConfig = ConfigUtils.genLatestConfig(defaults);
+
+    const configDumpResult = toStr(config);
+    if (configDumpResult.isErr()) {
+      return errAsync(configDumpResult.error);
+    }
+
+    const configDump = configDumpResult.value;
+    return ResultAsync.fromPromise(
```
namjul, line 40: You could shorten this to:

```suggestion
    const configResult = toStr(config).andThen((configDump) => {
      return ResultAsync.fromPromise(..)
    )
```

packages/common-all/src/store/ConfigStore.ts
namjul: > But it's not like putting the word await in front of it is that tedious imo...

Maybe it just comes down to "... the same level of expressivity and control as a regular `Result<T, E>`." 

Lets see, if you use an async function like `_fileStore.write` and we want to keep using `Result` we are forced to use try catch, with `ResultAsync` this is handled for us. This is thing about making these things easier to write.
Second thing is if about designing APIs that make use of `ResultAsnyc<..>` or stay with `Promise<Result<..>>`. 
I guess like you showed @jonathanyeung the only difference is the extra code of needing to write `async` and `await` when using `Promise`. So its still an expessivitiy thing but more about if me make that thing part of function signatures.

packages/common-all/src/store/ConfigStore.ts
```diff
@@ -0,0 +1,336 @@
...
+
+          const configLoadResult = fromStr(readPayload);
+          if (configLoadResult.isErr()) {
+            throw configLoadResult.error;
+          }
+          return configLoadResult.value as DeepPartial<DendronConfig>;
+        })
+      ),
+      (err) => err as DendronError
+    );
```
namjul, line 82: You could simplify this to:
```suggestion
          returnresp.data;
      ),
      (err) => err as DendronError
    ).andThen((readPayload) => {
          return fromStr(readPayload);
        })
    )
```

packages/common-all/src/store/ConfigStore.ts
```diff
@@ -0,0 +1,336 @@
...
+    } else {
+      const readResult = this.readWithDefaults();
+      return readResult.andThen<
+        DendronConfig,
+        IDendronError<StatusCodes | undefined>
+      >((config) => {
+        return ResultAsync.fromPromise(
+          Promise.resolve(
+            this.searchOverride().then((res) => {
+              // TODO: move validateConfig logic from common-server to common-all and add it here.
```
namjul, line 103: I guess `ConfigUtils.parse(..)` should fit here.

packages/common-all/src/store/IConfigStore.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export type ConfigValue = string | number | object;
+
+export type ConfigReadMode = "override" | "default";
+export type ConfigReadOpts = {
+  mode: ConfigReadMode;
+  useCache?: boolean;
+};
+
+export type ConfigGetOpts = ConfigReadOpts;
```
kevinslin, line 14: do we need a separate type if the underlying struct is the same?

kevinslin (commented):
looks good aside from existing comments. just one small nitpick

packages/common-all/src/store/ConfigStore.ts
jonathanyeung: > keep using Result we are forced to use try catch, with ResultAsync this is handled for us.

Ah, that's a good point. @namjul do you know of a more elegant way on the API implementation side of thing to return a ResultAsync without doing something like 

```typescript
return ResultAsync.fromPromise(
          Promise.resolve(...));
```

namjul: I don't think the `Promise.resolve` should be necessary, so this should work:
```ts
return fromPromise(this._fileStore.write(this.path, configDump))
```
and if `write` returns `ResultAsync` this becomes simply:

```ts
return this._fileStore.write(this.path, configDump)
```

hikchoi: After fiddling with these a bit more, I managed to tighten up some code here and there.

but there are some places where you just can't avoid using `ResultAsync.fromPromise` or `ResultAsync.fromSafePromise` even when we are dealing with code that does return `ResultAsync`.

There are a lot of overhead with helping `neverthrow` infer the type correctly when things get deeply nested.

I think moving forward it would be best if we just stick with the syncrhonous API of `neverthrow` and mark all of our methods async. We are giving up _some_ chaining capabilities, but I personally think it's a good compromise. IMO long chains of functions tend to not be so readable anyways.

packages/common-all/src/store/IConfigStore.ts
hikchoi: not really, preemptively distinguished them thinking it may have more props. will be cleaned up

packages/common-all/src/store/IConfigStore.ts
hikchoi: for `update`, allowing previous value to be undefined,
for `delete`, changed it so that an error result is returned

packages/common-all/src/store/IConfigStore.ts
hikchoi: The caching mechanism right now for config isn't too complex. Nothing is done when there is a miss. This is pretty much how it is implemented in `DConfig`, so for the initial phase I sticked with the way we are already doing.

I can change it so that we always read from cache if possible, but that opens up a possibility for some false negatives

packages/common-all/src/store/ConfigStore.ts
namjul: >  can't avoid using ResultAsync.fromPromise or ResultAsync.fromSafePromise even when we are dealing with code that does return ResultAsync

can you show me an example how `fromPromise` would be required when dealing with `ResultAsync`

hikchoi: @namjul see some of the methods that have a long chain in this commit: https://github.com/dendronhq/dendron/pull/3709/commits/3c37239037a792a119ea0568a28d0dc7d1b2e308

particularly something like `searchOverride`. At some point in the chain, I need to `.then()` on a result, write a callback, and then construct a new `ResultAsync` to continue on with the chaining. In that process I can't avoid using `ResultAsync.fromPromise`.

On top of that, it seems like once I call `.then` on a chain of ResultAsyncs, `neverthrow` seems to struggle with inferring the correct Result type and we need to write extra fluff to help inferring.

Compare all that to the most recent commit (https://github.com/dendronhq/dendron/pull/3709/commits/0c262747d2c174d415fb1a16555019ea5c85c581) where I changed all the method signatures to return a `Promise<Result...>`.

Now I can mark the methods async and await on parts where I would otherwise need to start a callback hell. As a result, we lose some chain-ability, but the code in my opinion is much more readable and simple.

Looks like returning a `ResultAsync` from a function that has a signature `Promise<Result...>` is allowed (see [readOverride](https://github.com/dendronhq/dendron/pull/3709/commits/0c262747d2c174d415fb1a16555019ea5c85c581#diff-f42adc20adeedcc1d2472997d9018e834d940321b3c14038d60dc563c95ef13cR114)), so for simple functions where we only do simple processing of the passed object, we can still keep using `ResultAsync`s.

For methods like `searchOverride` that has a bit more complexity to the control flow, breaking the chain and simply `await`ing an intermediate result helps so much with readability and cleaner code.

Maybe this is just me being stubborn with our usual async-await paradigm and not thinking functional enough. Happy to sync with you if you have a better suggestion / a way to keep things `ResultAsync`.

kevinslin (approved):

hikchoi:
Not sure why the window test is failing. will take a look

packages/common-all/src/store/IConfigStore.ts
jonathanyeung: The way I was thinking about it is that whether you cache or not is an implementation detail of config store that a caller shouldn't need to worry about - it should just expect to get the right value regardless of whether a cache is used or not.

hikchoi: that makes sense. I'll see what I can do here.

packages/common-all/src/store/ConfigStore.ts
jonathanyeung: I wrestled with this trying to get `searchOverride` to return a ResultAsync, but couldn't get it to work elegantly within ~30 mins...argh.  I feel like it still might be possible, but it's not that intuitive, or else we need to ramp up more with the library.

hikchoi: @jonathanyeung just curious, what was the problem you encountered here?

hikchoi: @namjul Awesome! thank you for the PR.

I now understand what you did with the use of `orElse`. 
I had doubts about neverthrow not able to be used for more complex control flows but I stand corrected. We can avoid using the `ResultAsync.fromPromise` way of doing things unless we are trying to wrap an external dependency that is not type safe.

Now that that's clear, I want to go back to my opinions on the pragmatic tradeoffs we should make:

Like @kevinslin briefly mentioned during weekly planning, even though there are still two opposing opinions, the introduction of async-await was a huge change in js/ts, and our codebase widely uses it.

`neverthrow`, with the asynchronous API, wraps `Promise<Result<T, E>>` into a `ResultAsync<T, e>`, and emulates the concept `either` in place of a Promise and allows the usual expressiveness of the synchronous `Result`s without having to `await` a piece of code (as well as sticking to its functional roots)

This is all well and good but I would say 99% of our codebase is not functional, and a full-on functional approach where we are using `neverthrow` would be inevitably slow us down. Both in terms of immediate productivity (shown here ,embarrassingly, by my inability to make this work without your help), and readability / maintainablility for the next person down the line who may or may not be familiar with `neverthrow` just yet.

I do love `neverthrow`. Like @jonathanyeung mentioned, it makes me think about the errors more seriously because errors are now something that is part of a `Result` that is returned and not something you can sweep under the rug. 

But I think if we don't allow gradual adoption of `neverthrow`, it will slowly become something we avoid using.

The compromise I made with making the signatures `Promise<Result<T, E>>`, in my opinion is a good intermediary step for the gradual adoption of `neverthrow`: 

1. Allow marking a method async by making the signature `Promise<Result<T, E>>`.
2. At a best-effort basis, try to follow the functional paradigm of `neverthrow` with the implementation, but allow our usual way of doing things in the non-functional js/ts way so we can 1. maintain a productive pace and reduce friction, and 2. have a maintainable and readable code.
3. Since `ResultAsync<T, E>` is a wrapper around `Promise<Result<T, E>>` that allows the usual chaining you get from `Result<T, E>`, if later we refactor this implementation to be functional we can always change the signature back to `ResultAsync<T, E>` and drop the async keyword. There will be no difference from the caller's perspective because they are the same thing; the difference only exists for the callee.

I know @kevinslin mentioned that this kind of mental overhead (questioning "Does this need to be `Promise<Result<T, E>>` or `ResultAsync<T, E>`?") itself is what we want to avoid, but I think this is a simple enough question that we can answer ("Am I able to come up with an implementation without `await`ing something? then use `ResultAsync<T, E>". Otherwise, default to starting with `Promise<Result<T, E>>`.), and the difference doesn't matter to the caller at all.

What do you guys think?

packages/common-all/src/store/IConfigStore.ts
hikchoi: Decided to remove caching for now. caching wasn't set up rigidly in the `DConfig` implementation, and was not used in most cases anyways.

Don't feel great having a half-baked caching mechanism be part of the config store at the moment, and there are some edge cases that need to be taken care of because of the overrides. Not worth spending time trying to make override work with caches at this moment.

packages/common-all/src/store/ConfigStore.ts
namjul: First I want to say that I appreciate this kind of exploration where we all look for a kind of synthesis around this topic. There seems to be space for allocentric processing but also allowing egocentric concerns to be named which is great :)

Regarding your proposal to start with a smooth adoption whereby we stay with the synchronous API so that we may find ourself in a more settled in place to consider `ResultAsync` at a future time, I think is a wise move. We first get to experience the general idea of a railway oriented programming style and can, like you point to, without much friction/change, adopt the `ResultAsync` "extension" when we have a bit more "skin in the game" with it.

hikchoi: Glad to have discussed this further as well. I'll wrap this discussion up and post an async about it by the end of this week, and have some dev docs written as well for others to reference in the future.

jonathanyeung: lgtm - thanks all for the thoughtful discussion!

PR #3716, opened 2022-10-26
fix(publishing): links don't show as private after nextJS export

jonathanyeung:
## fix(publishing): links don't show as private after nextJS export

This changes how we do the note rendering during NextJS Export Pod - instead of calling `getParsingDependencyDicts()` on each individual note, we just use the full set of engine notes as the dependency Dictionary. This gets rid of a performance issue of calling `getParsingDependencyDicts()` too many times during export as well as an issue where notes were getting marked as 'private' in published sites due to them not being in the dependency dictionary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,55 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [0.116.1](https://github.com/dendronhq/dendron/compare/v0.112.1...v0.116.1) (2022-10-25)
+
+
+### Bug Fixes
+
+* **publish:** correct hashtag parsing ([#3708](https://github.com/dendronhq/dendron/issues/3708)) ([ab6f2b3](https://github.com/dendronhq/dendron/commit/ab6f2b345ecb536fc0978a82427d5ba03fd8d0b0))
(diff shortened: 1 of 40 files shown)


hikchoi (commented):
hmm so the throttle was causing `engine.findNote` to only invoke with the most recent provided args and missing a bunch of calls that were needed to bring in all the dependencies?

jonathanyeung:
> hmm so the throttle was causing `engine.findNote` to only invoke with the most recent provided args and missing a bunch of calls that were needed to bring in all the dependencies?

I didn't investigate to see if the missing dependency was a regression from the throttling code or from somewhere else.

In this case, since we're doing a bulk render of all the notes in a vault, it makes sense to just start with the larger set of notes as the dependency cache and reuse it for each note's render instead of computing a separate cache for each note.

namjul (commented):
@jonathanyeung how is this PR solving the "private" marked links in addition to https://github.com/dendronhq/dendron/pull/3708.
as far as I can see the code does only the optimization of removing parsing for each note and replace it with a single get call at the beginning.

jonathanyeung:
> @jonathanyeung how is this PR solving the "private" marked links in addition to #3708. as far as I can see the code does only the optimization of removing parsing for each note and replace it with a single get call at the beginning.

@namjul - it fixes it by making sure that _all_ notes are in the parsing dependency dict. I didn't have a chance to investigate exactly why it wasn't working before though - so it's possible there are still more issues with `getParsingDependencyDicts()` or the surrounding logic.

Issue #3718, opened 2022-10-26
Deleted files being treated as a note


wookiefriseur:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [x] Workspace. VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup. Dendron's Lookup Command
- [ ] Views. Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema. Dendron Schemas
- [ ] Pod. Data import from / export to Dendron
- [ ] Publish. External Site Publish
- [ ] Markdown. Markdown features, such as syntax support and features

## Describe the bug

When opening deleted files or refactored files with their old filename an error message appears.

## To Reproduce

1. Enable git in workspace
2. Create and commit/stage note
3. Delete note through dendron
4. Go to Source Control View
5. Click on deleted file
6. Dendron treats it as a note and throws an error

Alternatively you can refactor hierarchies and deleted files will remain opened, causing the same error when selected.

## Expected behavior

Dendron should treat deleted files like they are not a note. It would prevent error messages with missing files and duplicate ids popping up.

## Desktop (please complete the following information)

- OS: Win10 64x
- Versions (all that apply to bug):
  - Dendron Extension: 0.116.0
  - VS Code version: 1.73.0-insider

## Additional context

Happens a lot in diff editor.

## Dendron Log file

```log
info: {"ctx":"WindowWatcher:onDidChangeVisibleTextEditors","editorPaths":["extension-output-dendron.dendron-#1-Dendron"]}
info: {"ctx":"refreshBacklinksChangeActiveTextEditor"}
debug: {"msg":"Note opened","fname":"delete.this.note"}
[...]

PR #3719, opened 2022-10-26
fix(workspace): Correctly handle deleted notes in editor when checking for duplicate entries

wookiefriseur:
## Commit

- [x] make sure your branch names adhere to our [branch style](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e)
- [x] make sure the commit message follows Dendron's [commit style](https://docs.dendron.so/notes/QN46JTSWpEkDkr94TJ85w)
- [x] if this pull request is addressing an existing issue, make sure to link this PR to the issue that it is resolving.
  - NOTE: make sure to read [Addressing open issues](https://docs.dendron.so/notes/adc39825-77a6-46cf-9c49-2642fcb4248e)
  - #3718

## Code

- [x] code should follow [Code Conventions](https://docs.dendron.so/notes/773e0b5a-510f-4c21-acf4-2d1ab3ed741e)

## Tests

- [x] [Write Tests](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro) 
- [x] [Confirm existing tests pass](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro)
  - partially, all pass until test suite crashes (always a few minutes in). Happens even without any code changes
- [x] [Confirm manual testing](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro) 
- [x] Common cases tested
- [ ] 1-2 Edge cases tested
- [ ] If your tests changes an existing [snapshot](https://docs.dendron.so/notes/99q7A73uGmCwu2KvSHZro)
  - [ ] [Update the Snapshot](https://docs.dendron.so/notes/u7utw5w8gyacuh3ok9ehi8j)

## Docs

- [ ] if your change reflects documentation changes, also submit a PR to [dendron-site](https://github.com/dendronhq/dendron-site) and mention the doc PR link in your current PR (NOTE: submit the PR against the `dev` branch of the dendron-site repo)

- [ ] does this change introduce a new or better way of doing things that others need to be aware of? if so, an async should be created and a process added in [Development](https://wiki.dendron.so/notes/ce9b9b29-d85b-492b-b76d-5e0c5dc52b03) or [Packages](https://wiki.dendron.so/notes/ok9ifke0zsfxnnh7xog79z5)


wookiefriseur:
I would have preferred to use https://github.com/dendronhq/dendron/blob/7059d56e23b74d04896207e1967eb4032c54ecce/packages/common-all/src/dnode.ts#L1134-L1136  but it would have been outside the plugin-core package and require to either pass a workspace reference or use node fs.

Also I did not find a better way using the VSCode API to check if a file is deleted.

An alternative would be to check in multiple spots in [WorkspaceWatcher.ts](https://github.com/dendronhq/dendron/blob/7059d56e23b74d04896207e1967eb4032c54ecce/packages/plugin-core/src/WorkspaceWatcher.ts)

packages/plugin-core/src/components/doctor/utils.ts
```diff
@@ -26,6 +26,10 @@ export class DoctorUtils {
     const { vaults, wsRoot, engine } = extension.getDWorkspace();
     let vault;
     try {
+      if ((await vscode.workspace.fs.stat(document.uri)).ctime === 0) {
```
wookiefriseur, line 29: Would have preferred to instead use something like this in L24:
```suggestion
      if(!NoteUtils.isNote(document.uri)) return;
```

Also I could not find a good alternative to `workspace.fs.stat()` in the vscode API.

hikchoi: I believe `vscode.workspace.fs.stat(document.uri)` will throw a `FileNotFound` error (see: [vscode.d.ts](https://github.com/microsoft/vscode/blob/6a62d4fbdc900506420da1aa8becf3a6354a1c1e/src/vscode-dts/vscode.d.ts#L8192)) so the if block won't be necessary.

> but it would have been outside the plugin-core package and require to either pass a workspace reference or use node fs.

A few line above we already have a check that's exactly what `NoteUtils.isNote` is doing, so I don't think this is sufficient.

For deleted notes that are still open in the editor, `document.uri.fsPath` will still end with `.md`, it's just that there isn't a file there anymore. 

So your solution of trying `vscode.workspace.fs.stat(document.uri)` and just returning when it throws would do the trick.

hikchoi (changes requested):
added a small comment on the use of `workspace.fs.stat`

hikchoi:
Could you also add a test case in `DuplicateNote.test.ts`?

Let me know if you need help setting tests up.

packages/plugin-core/src/components/doctor/utils.ts
wookiefriseur: Hey, no worries. Thanks for the response.

#### Regular Editor View
> I believe `vscode.workspace.fs.stat(document.uri)` will throw a `FileNotFound` error (see: [vscode.d.ts](https://github.com/microsoft/vscode/blob/6a62d4fbdc900506420da1aa8becf3a6354a1c1e/src/vscode-dts/vscode.d.ts#L8192)) so the if block won't be necessary.

Good point. Will change it to one line. It covers all deleted files in the regular view. 


#### Source control view
The above line does not work in Source Control view, which is why I checked for creation time:

```js
// existing
{
  ctime: 1666793183976,
  mtime: 1666793183976,
  size: 120,
  type: 1,
};

// deleted, but in source control view
{
  ctime: 0,
  mtime: 1667414348797,
  size: 0,
  permissions: undefined,
  type: 1,
};
```

But it is a bit hacky and I think I will check for uri scheme instead:

```js
// example document.uri in default view
{
  scheme: "file",
  authority: "",
  path: "/z:/Projekte/dev/dendron-example/notes/some.note.md",
  query: "",
  fragment: "",
  _formatted: "file:///z%3A/Projekte/dev/dendron-example/notes/some.note.md",
  _fsPath: "z:\\Projekte\\dev\\dendron-example\\notes\\some.note.md",
}

// example document.uri in source control view
{
  scheme: "git",
  authority: "",
  path: "/z:/Projekte/dev/dendron-example/notes/some.note.md",
  query: "{\"path\":\"z:\\\\Projekte\\\\dev\\\\dendron-example\\\\notes\\\\some.note.md\",\"ref\":\"HEAD\"}",
  fragment: "",
  _formatted: "git:/z%3A/Projekte/dev/dendron-example/notes/some.note.md?%7B%22path%22%3A%22z%3A%5C%5CProjekte%5C%5Cdev%5C%5Cdendron-example%5C%5Cnotes%5C%5Csome.note.md%22%2C%22ref%22%3A%22HEAD%22%7D",
  _fsPath: "z:\\Projekte\\dev\\dendron-example\\notes\\some.note.md",
}
```

**Question:**
Do you think there will be any problems, if I tell DoctorUtils to skip files with `git` uri scheme?
`if (document.uri.scheme === "git") return;`

Pros would be:
- no more frontmatter check on changed files in source control
- no more "duplicate id" false positives when in source control

wookiefriseur: Sorry, I think I will change or cancel my PR.

I created a test in `DuplicateNote.test.ts` and noticed that the test result is the same with and without the code changes. After some reflection what I am basically doing is just avoiding 2 calls + error message when it is a deleted note, by adding 1 unnecessary `fs` call to everything.

What I would really like to do though is to skip files in source control view, as mentioned before. It works, but I do not know how to test for that.

---

<details>
<summary>Here is my current test for deleted files</summary>

```typescript
describeSingleWS(
    "GIVEN an open file that has been deleted",
    {
      postSetupHook: ENGINE_HOOKS.setupEmpty,
    },
    () => {
      test("THEN do nothing", async () => {
        const { engine, wsRoot, vaults } = ExtensionProvider.getDWorkspace();

        await NoteTestUtilsV4.createNoteWithEngine({
          fname: "deleted",
          vault: vaults[0],
          wsRoot,
          body: "note will be deleted",
          engine,
        });

        const vaultPath = VaultUtils.getRelPath(vaults[0]);
        const deletedFilePath = path.join(wsRoot, vaultPath, "deleted.md");
        const deletedFileUri = vscode.Uri.file(deletedFilePath);
        await VSCodeUtils.openFileInEditor(deletedFileUri);

        fs.unlinkSync(deletedFilePath);

        await VSCodeUtils.openFileInEditor(deletedFileUri);
        const editor = VSCodeUtils.getActiveTextEditor();
        const document = editor?.document;

        const resp = await DoctorUtils.findDuplicateNoteFromDocument(document!);
        expect(resp).toEqual(undefined);
      });
    }
  );
```

</details>

hikchoi: I think the test case needs a duplicate note to see the effect. otherwise in both cases have no duplicate and the doctor util will do nothing.

hikchoi: > Do you think there will be any problems, if I tell DoctorUtils to skip files with git uri scheme?
> if (document.uri.scheme === "git") return;
> Pros would be:
>
>    no more frontmatter check on changed files in source control
>    no more "duplicate id" false positives when in source control

Can you elaborate? is this the editor that opens when you click on a file from the source control side panel?
To be safe with this, we would have to know if regular editors never have git scheme urls. If that is the case I say that would be fine. I can't recall anything regarding the document uri being git scheme in the regular text editor but if you can find info on this matter that would be awesome.

wookiefriseur: ## Source Control View

> Can you elaborate? is this the editor that opens when you click on a file from the source control side panel?
To be safe with this, we would have to know if regular editors never have git scheme urls. If that is the case I say that would be fine. I can't recall anything regarding the document uri being git scheme in the regular text editor but if you can find info on this matter that would be awesome.

Yeah, that's the one. I could only find out so far that an extension can implement its own scheme and the git scheme is being used by the builtin vscode git extension.

- [StackOverflow: gitURI scheme](https://stackoverflow.com/questions/56383584/whats-the-git-uri-scheme-and-is-it-documented-anywhere/71556348#71556348)
- [TextDocumentContentProvider](https://code.visualstudio.com/api/extension-guides/virtual-documents#textdocumentcontentprovider)

## Deleted Notes

> I think the test case needs a duplicate note to see the effect. otherwise in both cases have no duplicate and the doctor util will do nothing.

I created the test but as I mentioned: Dendron already checks for deleted files [a few lines later](https://github.com/dendronhq/dendron/blob/78c5b39c94a2fc3bc5b5f9ce111009b737663914/packages/plugin-core/src/components/doctor/utils.ts#L41-L46).

```typescript
    const resp = file2Note(fsPath, vault);
    if (ErrorUtils.isErrorResp(resp)) {
      // not in file system, we do nothing.
      Logger.error({ ctx, error: resp.error });
      return;
    }
```

<details>
<summary>Current Test</summary>

```typescript
  describeSingleWS(
    "GIVEN an open duplicate note that has been deleted",
    {
      postSetupHook: ENGINE_HOOKS.setupEmpty,
    },
    () => {
      test("THEN do nothing", async () => {
        const { engine, wsRoot, vaults } = ExtensionProvider.getDWorkspace();
        const vaultPath = VaultUtils.getRelPath(vaults[0]);

        await NoteTestUtilsV4.createNoteWithEngine({
          fname: "deleted",
          vault: vaults[0],
          wsRoot,
          body: "note will be deleted",
          engine,
        });
        const delNotePath = path.join(wsRoot, vaultPath, "deleted.md");
        const delNoteUri = vscode.Uri.file(delNotePath);
        const dupeNotePath = path.join(wsRoot, vaultPath, "dupe.md");
        const noteContent = fs.readFileSync(delNotePath, { encoding: "utf-8" });
        fs.writeFileSync(dupeNotePath, noteContent, { encoding: "utf-8" });

        await VSCodeUtils.openFileInEditor(delNoteUri);
        fs.unlinkSync(delNotePath);

        const editor = VSCodeUtils.getActiveTextEditor();
        const document = editor?.document;

        const resp = await DoctorUtils.findDuplicateNoteFromDocument(document!);
        expect(resp).toEqual(undefined);
      });
    }
  );
```

</details>

### Working as intended

When I first wrote the issue I thought the note was supposed to be ignored, like a file without frontmatter. The error message popup made it look to me like something broke and required action:

grafik

But now in the code I see that it is working as intended. So my change would only prevent the error popup. If that is a desired change, then I can do it with the oneliner you suggested:

```typescript
    // return if file is not a markdown
    if (!fsPath.endsWith(".md")) return;
    // return if file is in source control view
    if (document.uri.scheme === "git") return;

    const extension = ExtensionProvider.getExtension();
    const { vaults, wsRoot, engine } = extension.getDWorkspace();
    let vault;
    try {
      await vscode.workspace.fs.stat(document.uri); // Skip if file does not exist
      vault = VaultUtils.getVaultByFilePath({
        vaults,
        wsRoot,
        fsPath,
      });
    } catch (error: any) {
      // document doesn't exist in workspace.
      return;
    }
```

hikchoi (approved):
Looks good! Thanks for the fix! Hope to see more contributions in the future 🙂 

This will likely go out with the release coming in 2 weeks.

hikchoi:
note to self before merge:

`fix(workspace): Correctly handle deleted notes in editor when checking for duplicate entries`

jonathanyeung (approved):
Thanks for the contribution!

PR #3725, opened 2022-10-26
fix(workspace): duplicate note entries in lookup bar

Harshita-mindfire:
This PR aims to fix the issue with duplicate note entries in lookup bar by filtering out the note ids that starts with prefix fake.
- Bug: https://discordapp.com/channels/717965437182410783/802578902429597726/1031927164645421116
- Debug: all the three entries are of same note
image

### DIFF
--- a/packages/engine-server/src/enginev2.ts
+++ b/packages/engine-server/src/enginev2.ts
@@ -469,9 +469,6 @@ export class DendronEngineV2 implements DEngine {
     // If provided, we render the given note entirely. Otherwise find the note in workspace.
     if (!note) {
       note = this.notes[id];
-    } else {
-      // `procRehype` needs the note to be in the engine, so we have to add it in case it's a dummy note
-      this.store.updateNote(note);
     }
 
     // If note was not provided and we couldn't find it, we can't render.
--- a/packages/engine-server/src/DendronEngineV3.ts
+++ b/packages/engine-server/src/DendronEngineV3.ts
@@ -837,11 +837,7 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
     // If provided, we render the given note entirely. Otherwise find the note in workspace.
     if (!note) {
       note = (await this.getNote(id)).data;
-    } else {
-      // `procRehype` needs the note to be in the engine, so we have to add it in case it's a dummy note
-      await this.writeNote(note, { metaOnly: true });
     }
-
     // If note was not provided and we couldn't find it, we can't render.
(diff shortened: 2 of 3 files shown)


kevinslin (changes requested):
so it looks like the fake notes are being added to the engine due to  `engine.renderNote`

`engine.renderNote` will add a dummy note (temporary note created for purposes of rendering) to the engine. for rendering purposes 
logic: [[../packages/engine-server/src/enginev2.ts]]{func: renderNote}

we can do a spotfix for this but this will come up everywhere when we use `engine.findNote`. can we fix the entire problem by not adding fake notes to the engine?
we should be able pass in fake notes in `engine._renderNote` as part of note dependencies instead of adding directly to the engine 

tag @jonathanyeung to spot check what I said

jonathanyeung:
Yeah - let's do a quick investigation to see how involved it'd be to get rid of the fake-note-into-engine flow altogether.  It should be possible with the changes to unified that Kevin mentioned.  If it looks like it's not too involved of a change, let's go ahead and do that.

But the implementation you've added @Harshita-mindfire is a decent fallback.

Harshita-mindfire:
skipped adding fake note to the engine, the `_renderNote` already has a dependency on fake note to render. Hover preview works.

packages/engine-server/src/enginev2.ts
```diff
@@ -469,9 +469,6 @@ export class DendronEngineV2 implements DEngine {
     // If provided, we render the given note entirely. Otherwise find the note in workspace.
     if (!note) {
       note = this.notes[id];
-    } else {
```
kevinslin, line 472: will this be a problem if a note has a note reference to itself?

jonathanyeung: It _shouldn't_ be - the note gets included as a part of the dependency set for unified processing - but @Harshita-mindfire could you do a quick test to make sure that's the case?

jonathanyeung:
@Harshita-mindfire can you also make the corresponding change in EngineV3? 

https://github.com/dendronhq/dendron/blob/65c1bab8075723132b9dcdfee395e9bdda4c0e48/packages/engine-server/src/DendronEngineV3.ts#L928

packages/engine-server/src/enginev2.ts
Harshita-mindfire: Yep, works fine.
image

jonathanyeung (approved):

PR #3726, opened 2022-10-26
enhance(workspace): Move engineClient to v3

tma66:
This change moves engineClient to v3 for the following methods
- getNote
- getNoteMeta
- bulkGetNotes
- bulkGetNotesMeta

Similar to the backend, this will be under a dev flag `enableEngineV3`

### DIFF
--- a/packages/api-server/src/routes/note.ts
+++ b/packages/api-server/src/routes/note.ts
@@ -15,6 +15,12 @@ import {
   FindNoteOpts,
   APIRequest,
   FindNotesMetaResp,
+  GetNoteResp,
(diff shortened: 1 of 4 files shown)


jonathanyeung:
Can we also do it for `queryNotes()`?

jonathanyeung (approved):

tma66:
> Can we also do it for `queryNotes()`?
> 
> Existing changes look good.

Right now its tied with `SQLiteMetadataStore` which is different from how we implement `queryNotes` in the engine v3. I'll put this under a flag once we clean up the QueryStore/SQLite logic in v3 first

PR #3730, opened 2022-10-29
enhance(workspace): Move fuseEngine into MetadataStore

tma66:
In preparation for moving to SQLite, this pr updates the interface of the `IDataStore` to support query. We then update `NoteMetadataStore` to support query as well by moving the fuseEngine into the class. Since fuseEngine and metadatastore need to be kept in sync, it would be easier to have the metadata store handle the updates instead of the engine.

**Changes**
1. Add `query` to `NoteMetadataStore` and `SchemaMetdataStore`. This will be powered by the same fuseEngine
2. Remove fuseEngine from the engine v3
3. In the engineClient, make api call for queryNotes if under a v3 flag


jonathanyeung:
FYI (in case any of us forget)- looks like there's a test failure in test-plugin-web pass

jonathanyeung (approved):

hikchoi (approved):

PR #3742, opened 2022-11-01
fix(publish): display full breadcrumb path for sidebar excluded notes

namjul:
The aim of this PR is to fix breadcrumb display for sidebar excluded notes.

Before, the breadcrumb component for notes excluded from sidebar, only showed the note itself without the full parental hierarchy. Sidebar excluded notes happen from note properties like `nav_exclude`, `nav_exclude_children` or `has_collection`.

~~The changes in this PR imply that the logic for figuring out sidebar excluded notes is not part anymore of generating the sidebarconfig for autogenerated sidebar items. We did not point to this in the [docs](https://wiki.dendron.so/notes/qsc9k9flhxrvr4py9v8o89x/#autogenerated) so nothing needs to change there. Its now part of the view rendering logic (`DendronTreeMenu`) and therefor not anymore implicitly effecting `DendronBreadCrumb` component (which is what this PR is after).~~

# References

- [[dendron://private/task.2022.10.27.breadcrumbs-number-issue]]

# Pull Request Checklist


namjul:
I concern with the approach taken in the PR is that now the filtering of notes by the mentioned note properties is happening after `NextjsExportPod` so this will increase the json file `tree.json` inside `data` folder of `nextjs-template`.

hikchoi (approved):

namjul:
I talked with @kevinslin and found that there is a better approach to take. Namely keeping the filtering inside `sidebar.ts` meaning `tree.json` won't increase in size and we rather place the note path into the page-props while building the site.

namjul:
@kevinslin @hikchoi I made new changes with the mentioned new approach.

hikchoi (approved):
👍🏼

kevinslin (approved):

PR #3747, opened 2022-11-02
fix(workspace): completionProvider and other bulk operations for engine v3

jonathanyeung:
## fix(workspace): completionProvider and other bulk operations for engine v3

There's an issue with bulk operations (or any operations passing in an array as a query parameter) where an array param exceeding 20 items will get converted to an object. This may cause engine side logic to fail. See https://github.com/expressjs/express/issues/2661

This was causing an issue with the wikilink completion provider not providing results for bigger workspaces while on v3, and potentially other issues as well that call `bulkGet`, `bulkGetMeta`, `bulkAdd`

### DIFF
--- a/packages/api-server/src/routes/note.ts
+++ b/packages/api-server/src/routes/note.ts
@@ -53,13 +53,7 @@ router.get(
   asyncHandler(async (req: Request, res: Response<BulkGetNoteResp>) => {
     const { ids, ws } = req.query as unknown as EngineBulkGetNoteRequest;
 
-    // Express will convert the ids array into an object if it exceeds the
-    // 'arrayLimit' value set in the query parser (see {@link appModule}). In
-    // that case, convert the object back to an array
-    let processedIdPayload = ids;
-    if (!Array.isArray(processedIdPayload)) {
-      processedIdPayload = Object.values(processedIdPayload);
-    }
+    const processedIdPayload = convertToArrayIfObject(ids);
 
     const engine = await getWSEngine({ ws: ws || "" });
     ExpressUtils.setResponse(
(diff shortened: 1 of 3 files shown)


hikchoi (approved):
nice catch! maybe it's just safe to set to limit to zero and assume an object? weird that that isn't the default. very confusing 😄

packages/api-server/src/routes/note.ts
```diff
@@ -52,17 +52,38 @@ router.get(
...
   asyncHandler(async (req: Request, res: Response<BulkGetNoteResp>) => {
     const { ids, ws } = req.query as unknown as EngineBulkGetNoteRequest;
+
+    // Express will convert the ids array into an object if it exceeds the
+    // 'arrayLimit' value set in the query parser (see {@link appModule}). In
+    // that case, convert the object back to an array
+    let processedIdPayload = ids;
+    if (!Array.isArray(processedIdPayload)) {
+      processedIdPayload = Object.values(processedIdPayload);
+    }
```
namjul, line 62: Maybe extract a local function so that all occurences point to the comment explanation.
```suggestion
    // Express will convert the ids array into an object if it exceeds the
    // 'arrayLimit' value set in the query parser (see {@link appModule}). In
    // that case, convert the object back to an array
    // See: https://github.com/ljharb/qs#parsing-arrays, https://github.com/expressjs/express/issues/2661
    function toArray(ids: Record<string, string> | string[]) {
      if (!Array.isArray(ids)) {
        return Object.values(ids);
      }
      return ids
    }
```

kevinslin (approved):
this looks good. would add @namjul suggestion.

jonathanyeung:
@hikchoi 

> maybe it's just safe to set to limit to zero and assume an object? weird that that isn't the default.

This would maybe cause us to catch issues more quickly, but it may be confusing while developing the code as to why arrays are getting converted to JSON.  I think leaving as a high limit may be better for now? But maybe it'll bite us later...

I'm not really sure why there's a limit - not sure if there's some perf consideration, or maybe the URL's can be better condensed when it's an object.

hikchoi:
Most likely to avoid iterating everything when something large like 5000 comes in

PR #3744, opened 2022-11-01
neverthrow exploration

namjul:

PR #3754, opened 2022-11-04
chore: sqlite as metadata store

jonathanyeung:
## chore: sqlite as metadata store

Implementing a sqlite metadata store and sqlite cache. This is currently behind a dev flag `enableSqlite`. This implementation does not use Prisma ORM.

This implementation will supplant the use of `dendron.cache.json`, and instead do persistent storage of all metadata, including links, in a sqlite db file.

### Initialization

_This is an analog to NoteParser_:
- ✓ WHEN sqlite init is performed with a single vault THEN vault is present in the Vaults table (123 ms)
- ✓ WHEN sqlite init is performed over multiple vaults THEN all vaults are present in the Vaults table (18 ms)
- ✓ WHEN sqlite init is performed on a simple note THEN note is present in the NoteProps and VaultNotes tables (12 ms)
- ✓ WHEN sqlite init is done multiple times THEN subsequent inits don't affect DB state (18 ms)
- ✓ WHEN a new note is added to file system THEN the new note is added to NoteProps and VaultNotes tables (10 ms)
- ✓ WHEN many new notes are added to file system THEN all the new notes are added to NoteProps and VaultNotes tables (12 ms)
- ✓ WHEN a note with a wikilink is added THEN an entry is added to the links table (13 ms)
- ✓ WHEN a note with a wikilink to itself THEN an entry is added to the links table (9 ms)
- ✓ WHEN a note with a note ref is added THEN an entry is added to the links table (6 ms)
- ✓ WHEN a note with a note candidate is added THEN an entry is added to the links table (13 ms)
- ✓ WHEN a note with an unresolved wikilink is added THEN an entry is still added to the links table (16 ms)
- ✓ WHEN a note with an unresolved wikilink is added AND the wikilink is later resolved THEN the entry in the links table has source and sinks properly populated (31 ms)
- ✓ WHEN a parent and child note are added THEN a child entry is added to the links table (20 ms)
- ✓ WHEN parent, child, and grandchild notes are added THEN two child entries are added to the hierarchy table (16 ms)
- ✓ WHEN a root note is present THEN all top level notes appear as root children in the Hierarchy Table (23 ms)
- ✓ WHEN a note is modified THEN the new note state is reflected in the NoteProps table (16 ms)
- ✓ WHEN a note has been modified with new links THEN the new links are reflected in the Links table (12 ms)
- ✓ WHEN a note has been modified with links removed THEN the old links are missing from the Links table (10 ms)
- ✓ WHEN many new notes are modified THEN the all the new note states are reflected in the DB (6 ms)
- ✓ WHEN a note has been deleted THEN all references to that note are deleted from the DB (8 ms)
- ✓ WHEN a child note has been deleted THEN the parent-child row is deleted from the Hierarchy table (5 ms)
- ✓ WHEN a parent note has been deleted THEN the parent-child row is deleted from the Hierarchy table (8 ms)
- ✓ WHEN a combo of adds, updates, and deletes have happened THEN the DB state is updated correctly (8 ms)
- ✓ WHEN you have notes in two vaults with the same fname THEN Links, VaultNotes, and NoteProp tables are all correctly populated (8 ms)
- ✓ WHEN you have two vaults with cross-vault links AND the source vault is initialized first THEN the cross-vault links are properly resolved in the Links Table (6 ms)
- ✓ WHEN you have two vaults with cross-vault links AND the sink vault is initialized first THEN the cross-vault links are properly resolved in the Links Table (5 ms)
- ✓ WHEN you have a cross vault wikilink with fully qualified vault syntax THEN the cross-vault link sink points to the note in the correct vault in the Links Table (8 ms)
- ✓ WHEN you have an ambiguous cross vault wikilink that can resolve to multiple vaults THEN a link exists to each potential sink (8 ms)
- ✓ WHEN you change the ID of an existing note THEN Links, VaultNotes, and NoteProp tables are all updated correctly

### SqliteMetadataStore Integration Test with NoteStore

GIVEN a NoteStore containing a SQLiteMetadataStore internally
- ✓ WHEN workspace contains notes, THEN find and findMetadata should return correct notes (125 ms)
- ✓ WHEN workspace contains notes, THEN queryMetadata should return correct notes (47 ms)
- ✓ WHEN writing a note, THEN get, getMetadata, and queryMetadata should retrieve same note (26 ms)
- ✓ WHEN writing a stub note, THEN get and getMetadata should retrieve same note (25 ms)
- ✓ WHEN writing and deleting a note, THEN get should return CONTENT_NOT_FOUND (30 ms)
- ✓ WHEN writing a note, THEN subsequent writes should update note (24 ms)
- ✓ WHEN writing a note with the same fname, THEN content hash should change (26 ms)
- ✓ WHEN writing a note with a mismatched key, THEN error should be returned (22 ms)
- ✓ WHEN writing and getting metadata, THEN only metadata should be retrieved (22 ms)
- ✓ WHEN bulk writing metadata, THEN all metadata should be retrievable (20 ms)
- ✓ WHEN deleting a root note, THEN error should return and be CANT_DELETE_ROOT (24 ms)


packages/engine-server/src/engineClient.ts
```diff
@@ -167,7 +167,8 @@ export class DendronEngineClient implements DEngineClient, EngineEventEmitter {
       };
     }
     if (!resp.data) {
-      throw new DendronError({ message: "no data" });
+      return {} as DEngineInitResp;
+      // throw new DendronError({ message: "no data" });
```
tma66, line 171: Did you mean to remove the throw error line?

Also this is more for down the line, but if the init resp will look different based on whether we use sqllite or not, we should probably change this return type to make it more clear

packages/api-server/src/modules/workspace/index.ts
```diff
@@ -29,33 +33,47 @@ export class WorkspaceController {
...
-      return { data, error };
-    }
-    await putWS({ ws: uri, engine });
-    const duration = getDurationMilliseconds(start);
-    logger.info({ ctx, msg: "finish init", duration, uri, error });
-    let error2;
-    if (error) {
-      error2 = error2PlainObject(error);
+
+    // SQLite doesn't need to do engine.init()
```
tma66, line 54: Just heads up that there are 2 other places were engine.init is called.

1) Dendron-cli command. Right now we have flag for using v2 or v3. Not sure if we wanted to update that to use sqllite as well.

2) After running `engine.deleteSchema`, we call init again. This might cause a temp problem since we'll end up tearing down the db every time but I'm guessing that won't be necessary once the sqllite schema logic is fleshed out

packages/engine-server/src/DendronEngineV3Factory.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const dbResult = await SqliteDbFactory.createInitializedDB(
+      wsRoot,
+      ConfigUtils.getVaults(config),
+      fileStore,
+      dbFilePath
+    );
+
+    if (dbResult.isErr()) {
+      throw dbResult.error;
```
tma66, line 61: Let's log this in case it happens often.
Also, if it does error out, we can just fall back to the NoteMetadataStore

packages/api-server/src/modules/workspace/index.ts
tma66, line 54: Will we still need to run init for the SchemaStore?

packages/engine-server/src/drivers/file/index.ts
```diff
@@ -1 +1,2 @@
 export * from "./schemaParser";
+export * from "./parseAllNoteFilesForSqlite";
```
tma66, line 2: Odd name for a file lol. Most files I've seen are nouns. On purpose?

packages/engine-server/src/engineClient.ts
jonathanyeung: Forward looking, we shouldn't return a huge note payload on init (we don't need it anymore since the cache is gone on the `engineClient` side.  

But it's a bit tricky to make things compatible in this transitionary period between v2->v3. The SQLite init doesn't return anything on purpose, which is why I adjusted the logic here.  I need to do some additional scrutiny to make sure things work.

packages/engine-server/src/drivers/file/index.ts
jonathanyeung: Intentional - per our new-ish convention (follows airbnb style guide), 1 export per file only, and name the file exactly the same as the export. So in the case of a class, which most files are, it'll be a noun. But for export functions, it should match the name of the function. Makes it easier to locate where things are. 

https://docs.dendron.so/notes/773e0b5a-510f-4c21-acf4-2d1ab3ed741e/#one-export-per-file

packages/engine-server/src/drivers/sqlite/SqliteError.ts
```diff
@@ -0,0 +1 @@
+export class SqliteError extends Error {}
```
namjul, line 1: fyi: played with a different way to manage error objects here https://github.com/dendronhq/dendron/pull/3744/commits/23ac7e5582134f3a63a70fb8db91eb8353bb0109.

packages/api-server/src/modules/workspace/index.ts
jonathanyeung: Added todo comments in both places for now.

packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
```diff
@@ -0,0 +1,388 @@
...
+  HierarchyTableUtils,
+  LinksTableUtils,
+  LinkType,
+  NotePropsFtsTableUtils,
+  NotePropsTableUtils,
+  VaultNotesTableRow,
+  VaultNotesTableUtils,
+  VaultsTableUtils,
+} from "./tables";
+import { SchemaNotesTableUtils } from "./tables/SchemaNotesTableUtils";
```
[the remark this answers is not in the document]
jonathanyeung: Good catch - it's an oversight.  The Schema functionality isn't really working yet and barely tested.

packages/engine-server/src/drivers/sqlite/SqliteError.ts
jonathanyeung: Yeah I want to include a better error typing, at the very least for sql based errors.  This is obviously just a stub right now that offers no functionality. Adding a TODO comment for now.

packages/engine-server/src/drivers/file/parseAllNoteFilesForSqlite.ts
```diff
@@ -0,0 +1,927 @@
...
+ * @param files - the fsPaths of all note files to be processed
+ * @param vault - the vault in which these files belong. To process multiple
+ * vaults, call this function multiple times, once for each vault.
+ * @param db
+ * @param root
+ * @param schemas
+ * @param enableLinkCandidates
+ * @returns
+ */
+export async function parseAllNoteFilesForSqlite(
```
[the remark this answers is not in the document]
jonathanyeung: 'Delta Initialization' was just a temporary term I used to explain the proposal more easily - I'll enhance the comments here a bit though.

PR #3755, opened 2022-11-04
fix: window path slash

namjul:
This PR aims to fix an issue with using `path.posix` functions. Currently it is `undefined` in `dendron-plugin-view`. 
The reason is that it is not using the latest version of path-browserify. https://github.com/dendronhq/dendron/blob/e401a7eb0256d66495d0782ca9bbbf87a16d8779/yarn.lock#L18408 points to 0.0.1 which is not https://github.com/browserify/path-browserify @ 1.0.1. [1.0.1 does add the posix alias](https://github.com/browserify/path-browserify/releases/tag/v1.0.0). `dendron-plugin-view` currently depends on webpack@4.44.2 which implicitly depends on this older version of path-browserify. 

To solve this we use a library (https://github.com/jonschlinkert/normalize-path)  that creates valid window paths in case `posix` is ´undefined´. Another way would be to upgrade webpack but that can be a task for another time.

# Pull Request Checklist


packages/common-all/src/utils/index.ts
```diff
@@ -1186,9 +1186,27 @@ export function cleanName(name: string): string {
...
+}
+
+/**
+ * Given a path on any platform, convert it to a unix style path. Avoid using this with absolute paths.
+ */
 export function normalizeUnixPath(fsPath: string): string {
-  return path.posix.normalize(fsPath.replace(/\\/g, "/"));
+  return path.posix
+    ? path.posix.normalize(fsPath.replace(/\\/g, "/")) // NOTE: `path.posix` might be not available depending on your build system. For example at the time of writing `dendron-plugin-views` does not implement a `posix` property.
+    : slash(fsPath);
```
namjul, line 1209: Now thinking about this some more we probably should not embed `slash` in here. Its not really doing a normalization of the path and therefor should be a seperate function.

namjul: addressed here https://github.com/dendronhq/dendron/pull/3755/commits/3077f00308e59685c1834683f306b10d41328ad7

hikchoi (approved):
looks good!

PR #3761, opened 2022-11-07
fix(workspace): decoration does not display correct custom color for hashtags

Harshita-mindfire:
This PR aims to address the following issue: https://github.com/dendronhq/dendron/issues/3605#issuecomment-1297023998

Note this PR does not fix the one 'intended regression' which is that colors don't observe the hierarchy anymore (i.e. if #foo is red, then #foo.bar used to be red too, but now it'll be something else).


jonathanyeung (approved):
Oops, I think I made some wrong changes here when I was doing the v3 refactoring of unified. Nice fix!

PR #3765, opened 2022-11-08
fix(views): make calendar view not break with theme toggle

namjul:
# Pull Request Checklist


packages/plugin-core/src/views/utils.ts
```diff
@@ -144,7 +136,7 @@ export class WebViewUtils {
       // and hand it out to any other functions that need to use it.
       acquireVsCodeApi: `const vscode = acquireVsCodeApi(); window.vscode = vscode;`,
       themeMap: themeMap as WebViewThemeMap,
-      initialTheme: initialTheme ?? defaultInitialTheme,
```
jonathanyeung, line 147: What's causing the issue for calendar view? 

Does removing this cause default theme to always be light?

hikchoi: I think this is passed into a method that generates HTML, and `initialTheme` is used to override the default, which is just following the vscode theme color palette. I guess we were doing unnecessary overrides which caused some problems?

hikchoi (approved):
lgtm

packages/plugin-core/src/views/utils.ts
jonathanyeung: But this code isn't specific to calendar view - will this cause note preview to always be light?

namjul: I did an explaination what is happening in the commit message https://github.com/dendronhq/dendron/pull/3765/commits/2269cdc209030fd8c482e1b2364eb96a41b72262 but it does not really explain since it just says that codepaths are not taken. 

So the salient thing is that the codepath at https://github.com/dendronhq/dendron/blob/97499704d429fdbf084bb86ff4b25d3f4cef2eec/packages/common-server/src/etc.ts#L182 is never entered when providing a default value for `initialTheme`. Which means `prependBuiltin` is never `true` and therefore the [builtin styles](https://github.com/dendronhq/dendron/blob/97499704d429fdbf084bb86ff4b25d3f4cef2eec/packages/common-server/src/etc.ts#L73-L92) never have the higher specificity that is required for the calendar-view to work. 

Specifically 
```css
        body, .ant-layout {
          background-color: var(--vscode-editor-background);
        }
```
never applies when the `overrideTheme` value is set, which is the case with a `defaultInitialTheme`

The `defaultInitialTheme` was introduced in https://github.com/dendronhq/dendron/commit/fcf29f0f55 with the aim:
> This is to fix a bug with CSS of views other than preview that got introduced with custom preview theme support.

I checked the views behavior with the change of this PR, but could not find any abnormalities.

namjul: > Does removing this cause default theme to always be light?

The defaultTheme is derived from the body class which is dependent on the current theme style (light or dark).
https://github.com/dendronhq/dendron/blob/97499704d429fdbf084bb86ff4b25d3f4cef2eec/packages/common-server/src/etc.ts#L140

namjul: > I think this is passed into a method that generates HTML, and `initialTheme` is used to override the default, which is just following the vscode theme color palette. I guess we were doing unnecessary overrides which caused some problems?

Yes, tdoing unnecessary overrides is what looks to me as causing the issue.

namjul, line 147: Maybe we should rename that parameter to `themeOverride` to make it clear that this is only intended to fixate the theme to either light, dark or custom.

kevinslin (approved):

jonathanyeung (approved):

PR #3769, opened 2022-11-09
fix(workspace): correctly populate backlink when creating new note using go to note

hikchoi:
# fix(workspace): correctly populate backlink when creating new note using go to note

This PR:
- fixes an issue with go to note, where newly created notes using go to note (originating from a wikilink) was not getting backlinks populated.
- this happens in both engine v2 and v3 (not an engine v3 regression)
- Ideally, the client probably shouldn't be concerned with this detail. I was considering adding this logic to the engine (like the rest is done in https://github.com/dendronhq/dendron/pull/3535), but the case where we need to populate backlinks of a newly created note is very limited (pretty much only this case I can think of so far), and for a generic real-time backlink updating in the engine we would have to a O(nk) pass of all existing notes and all their links, just to find one backlink to update. So I left the logic here in the command for the time being. Maybe we can consider moving this concern to the engine when we are more confident about the perf (sqlite?).

# Pull Request Checklist
### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/plugin-core/src/commands/GoToNoteInterface.ts
+++ b/packages/plugin-core/src/commands/GoToNoteInterface.ts
@@ -19,6 +19,7 @@ export type GoToNoteCommandOpts = {
   column?: ViewColumn;
   /** added for contextual UI analytics. */
   source?: string;
+  originNote?: NoteProps;
 };
 export { GoToNoteCommandOpts as GotoNoteCommandOpts };
 
--- a/packages/plugin-core/src/commands/GoToNoteInterface.ts
+++ b/packages/plugin-core/src/commands/GoToNoteInterface.ts
@@ -19,6 +19,11 @@ export type GoToNoteCommandOpts = {
   column?: ViewColumn;
   /** added for contextual UI analytics. */
   source?: string;
+  /**
+   * the note which go to originates from.
+   * this is populated in the process of running the command
+   * and should not be passed in outside of tests.
+   */
   originNote?: NoteProps;
 };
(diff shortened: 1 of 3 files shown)


jonathanyeung:
Nice, I think this fix makes sense. 

> Ideally, the client probably shouldn't be concerned with this detail. I was considering adding this logic to the engine

With the in-progress Sqlite implementation, backlinks passed to the engine from `writeNote(noteProps)` via `noteProps.links` will actually get ignored. Instead, backlinks get calculated by doing a lookup in a Links table, searching for all notes that point to the note ID of interest. (should be O(1) with indexing). So we wouldn't need to actually spend any more computation on doing backlink updates at write time.

In that design though, I think we'll actually re-encounter this problem and this fix will no longer work since the backlinks update you do here gets ignored by the engine.  Need to do some more thinking on that.

packages/plugin-core/src/commands/GoToNoteInterface.ts
```diff
@@ -19,6 +19,7 @@ export type GoToNoteCommandOpts = {
   column?: ViewColumn;
   /** added for contextual UI analytics. */
   source?: string;
+  originNote?: NoteProps;
```
jonathanyeung, line 22: maybe add a quick comment on this prop on how it's used?

PR #3775, opened 2022-11-10
chore: fix `dendron-plugin-views` build

namjul:
This aim of this PR is to fix the build task within `dendron-plugin-views` package which currently results into the following error:

```
/home/nam/code/dendronhq/dendron/node_modules/zod-validation-error/node_modules/@swc/helpers/src/index.mjs
Can't reexport the named export '__decorate' from non EcmaScript module (only default export is available)
```

The commit 419dec79d880228e5677b005f89b52db724b4dc1 (PR: https://github.com/dendronhq/dendron/pull/3762) is responsible for this by adding a dependency
which conflicts with the webpack build in `dendron-plugin-views`. That
dependency uses withing its cjs build files `@swc/helpers` which
conflicts with the import type from webpack.
`@swc/helpers` provides a `module` entry https://github.com/swc-project/swc/blob/main/packages/swc-helpers/package.json#L6-L7 which has named reexports which gets flagged as an [error](https://github.com/webpack/webpack/blob/v4.44.2/lib/dependencies/HarmonyExportImportedSpecifierDependency.js#L372). 

Update: Found a solution which works by pinning `zod-validation-error` to an other version: https://github.com/dendronhq/dendron/pull/3775/commits/5e406cc229c9bdf559686a8febb8a1c0b63a05bc

---

Update: I tried another solution to tell webpack to prefer main entry field instead of `module´. But since that would disable webpacks tree-shaking capabilities since this requries static module structure I reverted that change.

---

The solution is to use explicitly set [`Rule.type`](https://webpack.js.org/configuration/module/#ruletype) property which was introduced in [webpack 4.0](https://github.com/webpack/webpack/releases/tag/v4.0.0):

> javascript/auto: (The default one in webpack 3) Javascript module with all module systems enabled: CommonJS, AMD, ESM
> javascript/esm handles ESM more strictly compared to javascript/auto:
>  Imported names need to exist on imported module
>  Dynamic modules (non-esm, i. e. CommonJs) can only imported via default import, everything else (including namespace import) emit errors
> In .mjs modules are javascript/esm by default

In my understanding setting the value `javascript/auto` tells webpack that it allows for `mjs` files to be not ESM strict.

This is a kind of hack and curcumvents the issue. In a proper setup `mjs` should only be handled as esm modules and nothing else.
So instead of changing how webpack sees `mjs` file we might want to
revert the change that introduced this issue.

# Pull Request Checklist

### DIFF
--- a/packages/common-all/package.json
+++ b/packages/common-all/package.json
@@ -53,7 +53,8 @@
     "title": "^3.4.4",
     "vscode-uri": "^3.0.3",
     "zod": "3.19.1",
-    "zod-validation-error": "^0.2.1"
+    "zod-validation-error": "^0.1.1",
+    "@swc/helpers": "^0.3.2"
   },
   "devDependencies": {
     "@types/fast-levenshtein": "^0.0.2",
--- a/packages/dendron-plugin-views/config/webpack.config.js
+++ b/packages/dendron-plugin-views/config/webpack.config.js
@@ -429,7 +429,6 @@ module.exports = function (webpackEnv) {
               test: /\.(js|mjs)$/,
               exclude: /@babel(?:\/|\\{1,2})runtime/,
               loader: require.resolve("babel-loader"),
-              type: "javascript/auto",
               options: {
                 babelrc: false,
                 configFile: false,
--- a/packages/dendron-plugin-views/config/webpack.config.js
+++ b/packages/dendron-plugin-views/config/webpack.config.js
@@ -298,8 +298,6 @@ module.exports = function (webpackEnv) {
       ],
     },
     resolve: {
-      // Prefer `main` over `module`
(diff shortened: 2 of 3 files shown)


kevinslin:
1. is this breaking then nightly build then if plugin-views is failing?
2. can we pin the dependency of `swc-helpers` to be compatible?

kevinslin (approved):

namjul:
> 1. is this breaking then nightly build then if plugin-views is failing?

Yes, I believe that should fail.

> 2. can we pin the dependency of `swc-helpers` to be compatible?

Found a version which works https://github.com/dendronhq/dendron/pull/3775/commits/5e406cc229c9bdf559686a8febb8a1c0b63a05bc

PR #3777, opened 2022-11-11
chore(internal): consolidate preview rendering logic

Harshita-mindfire:
This PR aims to 
- consolidate preview rendering logic of web and local extension. More info: [[Change Preview Rendering Logic on Local Ext|task.2022.11.08.change-preview-rendering-logic-on-local-ext]]
- adds an abstract method renderNote in EngineV3Base and implemented in `DendronEngineV3Web`. Removed `INoteRenderer` and `PluginNoteRenderer` files created for the same.


packages/plugin-core/src/views/common/preview/PreviewPanel.ts
```diff
@@ -52,8 +66,9 @@ export class PreviewPanel implements PreviewProxy, vscode.Disposable {
     @inject("wsRoot") private wsRoot: URI,
     private wsUtils: WSUtilsWeb,
     private webViewUtils: WebViewUtils,
-    @inject("IPreviewPanelConfig") private config: IPreviewPanelConfig,
-    @inject("INoteRenderer") private noteRenderer: INoteRenderer
+    @inject("vaults") private vaults: DVault[],
+    @inject("DendronConfig") private dendronConfig: DendronConfig,
+    @inject("ReducedDEngine") private engine: ReducedDEngine
   ) {
     this._linkHandler = linkHandler;
```
jonathanyeung, line 73: nit: ts-trick I learned recently - you can use the syntax below to avoid needing to set these properties in the constructor body. 

```typescript
    @inject("IPreviewLinkHandler") private _linkHandler: IPreviewLinkHandler,
    @inject("ITextDocumentService") private _textDocumentService: ITextDocumentService,
```

This will then cause some es-lint rules to complain about empty constructor, but I've disabled them in a pending PR.

packages/plugin-core/src/test/suite-integ/TogglePreviewLock.test.ts
```diff
@@ -5,81 +5,87 @@ import { TogglePreviewLockCommand } from "../../commands/TogglePreviewLock";
...
+            engine,
+            wsRoot,
+            vault: vaults[0],
+            fname: "preview-test",
+          });
+          await ExtensionProvider.getWSUtils().openNote(note1);
+          await previewPanel.show(note1);
         });
-        test("THEN preview should toggle to unlocked", async () => {
+        test.only("THEN preview should be locked and pristine", async () => {
```
jonathanyeung, line 54: Lingering 'test.only' statement.

packages/plugin-core/src/test/suite-integ/TogglePreviewLock.test.ts
```diff
@@ -5,81 +5,87 @@ import { TogglePreviewLockCommand } from "../../commands/TogglePreviewLock";
...
+          const wsUtils =
+            previewPanel._DO_NOT_USE_EXPOSED_FOR_TESTING_wsUtilsWeb();
```
jonathanyeung, line 56: Can you add a comment here quickly detailing the issue between having 2 different wsRoot's being used by `ExtensionProvider.getDWorkspace()` versus `container.resolve()`

packages/plugin-core/src/injection-providers/setupLocalExtContainer.ts
```diff
@@ -21,4 +43,27 @@ export async function setupLocalExtContainer(opts: {
...
+  container.register<DendronConfig>("DendronConfig", {
+    useValue: config,
+  });
+  container.register<IPreviewLinkHandler>("IPreviewLinkHandler", {
+    useClass: PreviewLinkHandler,
+  });
+  container.register<ITextDocumentService>("ITextDocumentService", {
+    useClass: TextDocumentService,
+  });
+
```
jonathanyeung, line 55: per other comment, let's define the singleton nature here: 

```typescript
  container.register<ITextDocumentService>(
    "ITextDocumentService",
    {
      useClass: TextDocumentService,
    },
    { lifecycle: Lifecycle.Singleton }
  );
```

packages/plugin-core/src/services/TextDocumentServiceFactory.ts
```diff
@@ -1,21 +1,18 @@
...
+import { container } from "tsyringe";
 
 export class TextDocumentServiceFactory {
   private static _textDocumentService: ITextDocumentService | undefined;
 
   /**
    * Instantiate TextDocumentService to be used in _extension.ts/workspace.ts
    */
-  public static create(extension: IDendronExtension): ITextDocumentService {
+  public static create(): ITextDocumentService {
```
jonathanyeung, line 11: Now that `TextDocumentService` is injectable, we don't need or want to have separate Factory classes or methods anymore - tsyringe and its containers should act as the factory. It can also manage the singleton nature of the object. 

Let's go ahead and do the following: 
1. Delete this `TextDocumentServiceFactory` class
2. Use `container.resolve()` where this Factory is being called right now (just `workspaceActivator.ts`)
3. Define the singleton behavior inside the setupContainer() function (I left another comment there on how to do that)

jonathanyeung (changes requested):

packages/plugin-core/src/components/views/PreviewViewFactory.ts
```diff
@@ -1,8 +1,7 @@
-import { IDendronExtension } from "../../dendronExtensionInterface";
-import { PreviewLinkHandler } from "./PreviewLinkHandler";
+import "reflect-metadata";
```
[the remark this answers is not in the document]
Harshita-mindfire: Removing this throws error while running testcase `Error: tsyringe requires a reflect polyfill. Please add 'import "reflect-metadata"' to the top of your entry point.`

Harshita-mindfire:
Fixing test cases

jonathanyeung (approved):

PR #3776, opened 2022-11-11
(WIP) chore/deprecate dconfig

hikchoi:
using this to run the matrix to incrementally test out changes.

PR #3781, opened 2022-11-14
fix(workspace): Correctly read and apply local override config in engine v3

hikchoi:
# fix(workspace): Correctly read and apply local override config in engine v3

This PR:
- correctly reads and applies local override in engine v3
  - I've commented in https://github.com/dendronhq/dendron/pull/3687 that using `DConfig.readConfigAndApplyLocalOverrideSync` would not be safe yet but that turns out to be an unfounded worry. Not doing so caused a regression with engine v3 where local vaults are not properly loaded. Apologies @tma66 
  - adding them back fixes the issue.
- I've also found an issue with note path resolution where if the override specifies a vault with absolute path (which works in engine v2 and there are users doing this)
  - I've added a crude resolution in `NoteStore` for this issue.

## Note
- Ultimately these should all be resolved with the migration from `DConfig` to `ConfigService`, but with our timeline for rolling out engine v3, it seems very unlikely that `ConfigService` will be ready and thoroughly tested in time (https://github.com/dendronhq/dendron/pull/3776). I deferred fixing this until after that happens, but in favor of iteration I'm going with a temporary solution rather than try to rush the switch to `ConfigService`.
- I believe this should resolve the last bit of discrepancy engine v2 and v3 has that you mentioned @kevinslin. I've tested cross vault operations between overridden vaults / normal vaults and it works as expected.

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [ ] Common cases tested
- [ ] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/common-all/src/store/NoteStore.ts
+++ b/packages/common-all/src/store/NoteStore.ts
@@ -49,18 +49,33 @@ export class NoteStore implements INoteStore<string> {
     if (metadata.error) {
       return { error: metadata.error };
     }
-
     // If note is a stub, return stub note
     if (metadata.data.stub) {
       return {
         data: { ...metadata.data, body: "" },
       };
     }
+
+    // vault.fsPath that comes from local overrides can be absolute paths (e.g. if scope is global).
+    // need to slice the absolute portion off to correctly resolve.
+    // if a relative path comes in, this will do nothing and work as intended
+    const processedVault = metadata.data.vault.fsPath.startsWith(
+      this._wsRoot.fsPath
+    )
+      ? {
+          ...metadata.data.vault,
+          fsPath: metadata.data.vault.fsPath.slice(this._wsRoot.fsPath.length),
+        }
+      : metadata.data.vault;
+
     const uri = Utils.joinPath(
       this._wsRoot,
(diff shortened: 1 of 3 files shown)


packages/common-all/src/store/NoteStore.ts
```diff
@@ -49,18 +49,33 @@ export class NoteStore implements INoteStore<string> {
...
     if (metadata.data.stub) {
       return {
         data: { ...metadata.data, body: "" },
       };
     }
+
+    // vault.fsPath that comes from local overrides can be absolute paths (e.g. if scope is global).
+    // need to slice the absolute portion off to correctly resolve.
+    // if a relative path comes in, this will do nothing and work as intended
+    const processedVault = metadata.data.vault.fsPath.startsWith(
```
kevinslin, line 62: can we add a testcase for this?

hikchoi: added test for absolute path vault case

jonathanyeung (approved):

Issue #3780, opened 2022-11-14
Rename note under user hierarchy breaks expansion links 


nickolay-kondratyev:
## Please select the area your bug applies to. (Multiple selections are Possible. You can leave blank if you're not sure.)

- [ ] Workspace.  VSCode workspace, vaults, Intellisense/autocomplete, Dendron settings
- [ ] Lookup.  Dendron's Lookup Command
- [ ] Views.  Dendron Preview, Tree View, Side Panels in the UI
- [ ] Schema.  Dendron Schemas
- [ ] Pod.  Data import from / export to Dendron
- [ ] Publish.  External Site Publish
- [ ] Markdown.  Markdown features, such as syntax support and features
- [x] Refactor

## Describe the bug

Renaming notes that are under `user.` hierarchy leads to breaking expansion links.

## To Reproduce

Create note:

- `user.Tester.th.HelloWorld` 
- `note1` 

In note1 add

```
![[user.Tester.th.HelloWorld]]
```

Use `Rename note` on `user.Tester.th.HelloWorld` to rename it to `user.Tester.th.HelloWorld2`.

Expected behavior note1 contents are expected to become:
```
![[user.Tester.th.HelloWorld2]]
```

Actual behavior note1 contents become:
```
@Tester.th.HelloWorld2
```

Expansion is broken.

## Desktop (please complete the following information)

- Browser (if with published websites; include version): [e.g. chrome, safari]
- Versions (all that apply to bug):
  - Dendron Extension: 117
  - VS Code version:
  - Dendron CLI:

Harshita-mindfire:
Thanks @nickolay-kondratyev for reporting the bug. I can repro this issue. We will ping here when this gets fixed 🙏🏻

Harshita-mindfire:
This should be fixed after Dendron version 0.118.0. Feel free to comment here if the issue still persists.

PR #3782, opened 2022-11-15
fix(workspace): rename note breaks noteref for user hierarchy

Harshita-mindfire:
Related issue: https://github.com/dendronhq/dendron/issues/3780

### Basics

### Extended
- General
### Basics

- [x] Common cases tested
- [ ] 1-2 Edge cases tested

### DIFF
--- a/packages/engine-server/src/DendronEngineV3.ts
+++ b/packages/engine-server/src/DendronEngineV3.ts
@@ -1318,7 +1318,7 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
           alias = undefined;
         }
         // for user tag links, we'll have to regenerate the alias
-        if (newLoc.fname.startsWith(USERS_HIERARCHY)) {
+        if (link.type !== "ref" && newLoc.fname.startsWith(USERS_HIERARCHY)) {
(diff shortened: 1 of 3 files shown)


packages/engine-server/src/DendronEngineV3.ts
```diff
@@ -1318,7 +1318,7 @@ export class DendronEngineV3 extends EngineV3Base implements DEngine {
           alias = undefined;
         }
         // for user tag links, we'll have to regenerate the alias
-        if (newLoc.fname.startsWith(USERS_HIERARCHY)) {
+        if (link.type !== "ref" && newLoc.fname.startsWith(USERS_HIERARCHY)) {
```
jonathanyeung, line 1321: Can you add a comment that syntax like !@john doesn't work as a note ref, which is why we need to due `link.type !== ref`

jonathanyeung (approved):

packages/engine-server/src/DendronEngineV3.ts
Harshita-mindfire: Also, instead of `!@john`, presently it was replaced with `@john`

PR #3795, opened 2022-11-23
chore(workspace): Cap number of query results from engine

tma66:
**Background**
Lookup on v3 was taking a long time (upwards of up to ~10sec). The reason is that the engine v3 queries in the following steps
1. Query results from fuse Engine
2. For each result, get note body from file system

For our org-workspace, step 1 can lead to 10k+ notes which is causing the latency issues. It's not an issue in v2 b/c the note body is all stored in memory.

**Changes**
- Cap number of results from step 1. In lookup, we cap the query results to 50 so by capping on the engine side first, it won't effect the lookup results.
- Cap on engine is arbitrarily set to 100 for now

**Todo**
- This might affect other use cases of `queryNotes` such as Doctor that might expect more than 100 results. However, I believe it makes sense to have a sensible cap when possible as we don't want to send too many notes over the api if not needed. One solution is passing a parameter if we don't want a cap
- The latency is caused by trying to read thousand of files from the file system to get the note body. For cases like lookup, the body might not be necessary. If that's the case, we can improve latency even further

### DIFF
--- a/packages/common-all/src/engine/EngineV3Base.ts
+++ b/packages/common-all/src/engine/EngineV3Base.ts
@@ -306,6 +306,8 @@ export abstract class EngineV3Base implements ReducedDEngine {
   async queryNotes(opts: QueryNotesOpts): Promise<QueryNotesResp> {
     // const ctx = "Engine:queryNotes";
     const { vault } = opts;
+    const MAX_LIMIT = 100;
+
     // Need to ignore this because the engine stringifies this property, so the types are incorrect.
     // @ts-ignore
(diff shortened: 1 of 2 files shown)


kevinslin:
yeah, lookup does not require the body. can we switch to looking up the metadata instead?

hikchoi (approved):
good catch!

tma66:
> yeah, lookup does not require the body. can we switch to looking up the metadata instead?

That should definitely be next step. Fortunately capping engine results first will reduce latency that's similar to engine v2 times

PR #3808, opened 2022-11-25
chore: mark all real notes with stub=undefined

jonathanyeung:
## chore: mark all real notes with stub=undefined

When extracting NoteProps from a string/physical file via `string2note`, omit the stub property from NoteProps even if it's set to true - a real file should never be marked as a stub.  Only non-existing nodes that are used to form a complete hierarchy should have stub=true. This is necessary for the sqlite db logic to work properly.

### DIFF
--- a/packages/common-all/src/dnode.ts
+++ b/packages/common-all/src/dnode.ts
@@ -887,7 +887,6 @@ export class NoteUtils {
     // Separate custom and builtin props
     const builtinProps = _.pick(propsWithTrait, [
       ...Object.values(DNodeExplicitPropsEnum),
-      "stub",
     ]);
 
     const { custom: customProps } = cleanProps;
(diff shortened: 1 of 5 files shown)


packages/common-all/src/dnode.ts
```diff
@@ -887,20 +887,16 @@ export class NoteUtils {
     // Separate custom and builtin props
     const builtinProps = _.pick(propsWithTrait, [
       ...Object.values(DNodeExplicitPropsEnum),
-      "stub",
     ]);
 
     const { custom: customProps } = cleanProps;
     const meta = { ...builtinProps, ...customProps };
```
namjul, line 893: Do I understand this correct that a `stub` field can still be part of this object because it could still "live" inside `customProps` but not `buildinProps` since that has now been "unpicked".

namjul (approved):

packages/common-all/src/dnode.ts
jonathanyeung: correct

PR #3814, opened 2022-11-29
enhance(workspace): Support queryNotesMeta and migrate NoteQuickInput to NoteQuickInputV2

tma66:
**Background**
Lookup currently calls `queryNotes` which returns the entire NoteProps. With engineV3, this can cause some slow down b/c we're reading from disk to get the note body. Since lookup doesn't require the body, we will support `queryNotesMeta` to return `NotePropsMeta` instead

**Changes**
1. New api route for `queryNotesMeta`
2. Migrate users to use `queryNotesMeta` instead of `queryNotes`
3. Update `NoteQuickInputV2` to include `NotePropsMeta`. This is b/c many commands need to work with `NotePropsMeta` after selection is made
4. Migrate from `NoteQuickInput` to `NoteQuickInputV2`. Since lookup doesn't need the body anymore, the data type will change as well

**Testing**
Tested on `org-workspace`. Keeping the limit to 100 notes, the response should be under 100ms

### DIFF
--- a/packages/common-all/src/engine/EngineV3Base.ts
+++ b/packages/common-all/src/engine/EngineV3Base.ts
@@ -21,6 +21,7 @@ import {
   NoteChangeEntry,
   NoteProps,
   NotePropsMeta,
+  QueryNotesMetaResp,
   QueryNotesOpts,
   QueryNotesResp,
   ReducedDEngine,
@@ -45,6 +46,8 @@ export abstract class EngineV3Base implements ReducedDEngine {
   protected logger;
   public vaults;
   public wsRoot;
+  // TODO: Make configurable
+  API_MAX_LIMIT = 100;
 
   constructor(opts: {
     noteStore: INoteStore<string>;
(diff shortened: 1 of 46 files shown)


jonathanyeung (approved):

jonathanyeung:
@Harshita-mindfire just as a heads up about this change.

packages/common-all/src/engine/EngineV3Base.ts
```diff
@@ -301,43 +304,39 @@ export abstract class EngineV3Base implements ReducedDEngine {
...
-        return VaultUtils.isEqual(vault, ent.vault, this.wsRoot);
-      });
+  /**
+   * See {@link DEngine.queryNotesMeta}
+   */
+  async queryNotesMeta(opts: QueryNotesOpts): Promise<QueryNotesMetaResp> {
+    const response = await this.noteStore.queryMetadata(opts);
+    if (response.isErr()) {
+      // TODO: need to return an error
+      return [];
```
hikchoi, line 331: can we throw here? then we can know when it goes wrong at least.
of course we can also change this to return `Promise<RespV3<QueryNotesMetaResp>>`

hikchoi (changes requested):
small comment but otherwise looks good

PR #3816, opened 2022-11-29
chore: sqlite query utils for all db access methods

jonathanyeung:
## chore: sqlite query utils for all db access methods

This change adds a `SqlQueryUtils` static utility class. This exposes analogous functions to `Database.all()`, `Database.run()`, and `Database.get()` methods of the Sqlite3 library but package the response into a ResultAsync instead of the passed-in callback of the Sqlite3 lib. This also includes enhancements to the `SqliteError` implementation to make it no longer a no-op. Finally, the response of NotePropsTableUtils.getById() was changed to return null instead of an error if the ID is not found.


jonathanyeung:
@hikchoi @tma66 - heads up on this change - it may cause some conflicts with ongoing work on sqlite stuff.

packages/engine-server/src/drivers/sqlite/SqliteQueryUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+              err.message
+            } in ${getDurationMilliseconds(
+              start
+            )} ms. Query String: ${sql.slice(0, 1000)}`
+          );
+
+          reject(err.message);
+        } else {
+          // TODO: Switch to debug.
+          logger?.info(
```
hikchoi, line 34: keeping it as `.info` for now?

hikchoi (approved):
~~looks good! think I'll just branch my task off this pr~~

edit: see new comment

packages/engine-server/src/drivers/sqlite/SqliteQueryUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+      });
+    });
+
+    return ResultAsync.fromPromise(prom, (e) => {
+      return this.getSqliteError(e as string, sql);
+    });
+  }
+
+  /**
+   * Runs Database.get() method of sqlite3 and packages the result in an
```
namjul, line 51:
```suggestion
   * Runs Database.get() method of sqlite3 and packages the result (first row) in an
```

packages/engine-server/src/drivers/sqlite/SqliteQueryUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+      });
+    });
+
+    return ResultAsync.fromPromise(prom, (e) => {
+      return this.getSqliteError(e as string, sql);
+    });
+  }
+
+  /**
+   * Runs Database.all() method of sqlite3 and packages the result in an
```
namjul, line 96:
```suggestion
   * Runs Database.all() method of sqlite3 and packages the result (array of rows) in an
```

namjul (approved):

packages/engine-server/src/drivers/sqlite/SqliteQueryUtils.ts
```diff
@@ -0,0 +1,154 @@
...
+export class SqliteQueryUtils {
+  public static run(
+    db: Database,
+    sql: string,
+    logger?: DLogger
+  ): ResultAsync<null, SqliteError> {
+    const start = process.hrtime();
+
+    const prom = new Promise<null>((resolve, reject) => {
+      db.run(sql, (err) => {
```
hikchoi, line 21: looks like this still silences errors.

I tried this:
```
  db.run(sql, (res: RunResult, err: Error) {
    ...
  }
```

running it on a same workspace that doesn't resolve to an `err` gives me a `res` that contains the actual error.

image

hikchoi (changes requested):
comment on silenced sqlite error

packages/engine-server/src/drivers/sqlite/SqliteQueryUtils.ts
jonathanyeung:
```typescript
  db.run(sql, (res: RunResult, err: Error) {
    ...
  }
```

db.run's callback only has one argument, which is err.  So if you put it as

```typescript
  db.run(sql, (err: Error) {
    ...
  }
```
then it should come back as `err`

PR #3817, opened 2022-11-29
feat(internal): web performance harness

namjul:
The aim of this PR is to add a web performance harness that measures the metrics [web vital](https://web.dev/vitals/) (a group of metrics defined by the web-vital project), [bundle size](https://github.com/ai/size-limit) and [time](https://github.com/mbalabash/estimo). 

Reports in airtable: https://airtable.com/appCWyZZPrLlVUblF/tblLRZhYUKGmScOai/viwPWgniLl5LM6gRC?blocks=bipFIuRw2Al2FLkw8 (open "Extensions" panel on the right to view through the [json editor](https://support.airtable.com/docs/json-editor-extension))

In the current state this runs as part of the test-playwright ci job therefore on every PR.

As a next step we could add a comparison,  which prints a warning (PR comment) when a threshold or budget gets exceeded.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,9 +158,6 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    container:
-      image: mcr.microsoft.com/playwright:v1.28.0-focal
-
     steps:
       - name: Checkout source
         uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,6 +198,11 @@ jobs:
         run: npx playwright install --with-deps
 
       - name: Run Playwright tests
+        env:
+          AIRTABLE_API_KEY: ${{ secrets.AIRTABLE_API_KEY }}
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -152,6 +152,8 @@ jobs:
         run: yarn ci:test:plugin-web
 
   test-playwright:
+    environment: plugin-development
```
namjul, line 155: @jonathanyeung the test-playwright job https://github.com/dendronhq/dendron/actions/runs/3638127632 failes with `Branch "chore/web-performance-harness" is not allowed to deploy to plugin-development due to environment protection rules.` Could you check what the cause of this is.

jonathanyeung: It's because of this line in the workflow: 

```yaml
  test-playwright:
    environment: plugin-development
```

Looks like we need the secrets to do the airtable upload?  Environments need to have their branches whitelisted so that they can run (to protect secret access) - our secrets in plugin-development aren't very sensitive, so I can go ahead and enable access to the env for all branches

kevinslin (approved):

PR #3830, opened 2022-12-02
chore(internal): note schema parity when sqlite backend is enabled

hikchoi:
# chore(internal): note schema parity when sqlite backend is enabled

This PR:
- properly parse and apply note schemas during engine init when `useSqlite` is enabled.
- supersedes #3822

Notes: 
- now `SchemaNotes` table is properly populated after initialization.
- currently schema initialization happens after all delta initialization steps are done. this may change once we move on to a more proper `Schemas` table in the future.
- there are a few points we can optimize (e.g. remove expensive `.get` calls used to create `NoteDicts`), but currently the init step is implemented to use the existing `SchemaUtils.matchDomain` as much as possible with minimal change to the existing way we do it. this could be done by adding sqlite specific utilities to create `NoteDicts` in one query for example. this is also left for future iteration.

# Pull Request Checklist

### Basics

### Extended
- General
- Operations
- Architecture

### Basics

### Extended

### Basics

- [x] Common cases tested
- [x] 1-2 Edge cases tested

### Extended
- CSS
    - [~] lg: screen ≥ 992px
    - [~] safari
    - [~] firefox
    - [~] chrome

## Docs

### Extended

### DIFF
--- a/packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
+++ b/packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
@@ -1,25 +1,34 @@
 import {
   DendronError,
+  DLink,
+  DLogger,
   DVault,
   ERROR_SEVERITY,
   ERROR_STATUS,
   FindNoteOpts,
   IDataStore,
   IDendronError,
+  IFileStore,
+  isNotUndefined,
+  NoteDictsUtils,
+  NoteProps,
   NotePropsMeta,
   QueryNotesOpts,
   RespV3,
+  SchemaUtils,
   StatusCodes,
 } from "@dendronhq/common-all";
 import _ from "lodash";
 import { err, errAsync, ok, okAsync, Result, ResultAsync } from "neverthrow";
 import { Database } from "sqlite3";
+import { SqliteDbFactory } from "./SqliteDbFactory";
 import {
   HierarchyTableRow,
   HierarchyTableUtils,
   LinksTableUtils,
   LinkType,
   NotePropsFtsTableUtils,
+  NotePropsTableRow,
   NotePropsTableUtils,
   VaultNotesTableRow,
   VaultNotesTableUtils,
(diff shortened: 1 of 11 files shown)


packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
```diff
@@ -133,4 +133,33 @@ export class HierarchyTableUtils {
...
+   *
+   * return all (parent, child) pairs
+   *
+   * more on recursive CTEs: https://www.sqlite.org/lang_with.html
+   */
+  public static getAllInDomain(
+    db: Database,
+    domainId: string
+  ): ResultAsync<HierarchyTableRow[], SqliteError> {
+    const sql = [
```
hikchoi, line 150: anyone have a better solution to annotate multiline sql queries?

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
```diff
@@ -5,7 +5,7 @@ import { SqliteError } from "../SqliteError";
 import { SqliteQueryUtils } from "../SqliteQueryUtils";
 
 export class HierarchyTableRow {
-  constructor(public parentId: string, public childId: string) {}
```
jonathanyeung, line 8: I had actually renamed from parent/child -> parentId/childId in my first PR in response to a code review comment.  Was it intentional to drop the 'id' part or just a merge artifact?

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
```diff
@@ -133,4 +133,33 @@ export class HierarchyTableUtils {
...
+      `  SELECT h.parent, h.child`, // 2. select a root note to anchor to. this is the domain note
+      `  FROM "Hierarchy" h`,
+      `  WHERE h.parent = "${domainId}"`, // 3. at this point cte_hierarchy has one row queued
+      `  UNION ALL`, // UNION ALL assuming we don't have dupes in the hierarchy table
+      `  SELECT h.parent, h.child`,
+      `  FROM "Hierarchy" h`, // 4. we are going to join the hierarchy table
+      `  JOIN "cte_hierarchy" c ON c.child = h.parent`, // 5. to the cte until we run out of children
+      `)`,
+      `SELECT * FROM "cte_hierarchy"`, // 6. select everything from the created view.
+    ].join("\n");
```
jonathanyeung, line 161: let's drop the "\n" and just join with a " " separator.

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
jonathanyeung: imo this is a pretty good solution. I've started to adopt it in some local changes.

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
```diff
@@ -133,4 +133,33 @@ export class HierarchyTableUtils {
...
+  ): ResultAsync<HierarchyTableRow[], SqliteError> {
+    const sql = [
+      `WITH RECURSIVE "cte_hierarchy" (parent, child) AS (`, // 1. create a view (cte)
+      `  SELECT h.parent, h.child`, // 2. select a root note to anchor to. this is the domain note
+      `  FROM "Hierarchy" h`,
+      `  WHERE h.parent = "${domainId}"`, // 3. at this point cte_hierarchy has one row queued
+      `  UNION ALL`, // UNION ALL assuming we don't have dupes in the hierarchy table
+      `  SELECT h.parent, h.child`,
+      `  FROM "Hierarchy" h`, // 4. we are going to join the hierarchy table
+      `  JOIN "cte_hierarchy" c ON c.child = h.parent`, // 5. to the cte until we run out of children
```
jonathanyeung, line 158: Can we add some infinite loop protection here? In case we have a loop between parent-child, child-parent somehow.

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
```diff
@@ -133,4 +133,33 @@ export class HierarchyTableUtils {
...
+  /**
+   * Given a domain,
+   * recursively grab all descendants of the domain
+   * until we run out
+   *
+   * return all (parent, child) pairs
+   *
+   * more on recursive CTEs: https://www.sqlite.org/lang_with.html
+   */
+  public static getAllInDomain(
```
jonathanyeung, line 146: How about `getAllDescendants()` as a name? Since we're just dealing with a concept of ID's here, I don't think we need to user the word 'domain'.

packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
```diff
@@ -30,6 +39,94 @@ import { SchemaNotesTableUtils } from "./tables/SchemaNotesTableUtils";
 export class SqliteMetadataStore implements IDataStore<string, NotePropsMeta> {
   constructor(private _db: Database, private _vaults: DVault[]) {}
 
+  /**
+   * Goes through all domains and recursively apply schemas.
+   */
+  async initSchema(fileStore: IFileStore, wsRoot: string, logger: DLogger) {
+    // this whole thing to parseAllNoteFilesForSqlite / SqliteDbFactory
```
jonathanyeung, line 46: i think the beginning "// TODO: move" part of the comment got truncated.

packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
```diff
@@ -30,6 +39,94 @@ import { SchemaNotesTableUtils } from "./tables/SchemaNotesTableUtils";
...
+                const allInDomainRows = getAllInDomainResult.value;
+                return [domainId].concat(
+                  allInDomainRows.map((row) => {
+                    return row.child;
+                  })
+                );
+              })
+            )
+          );
+          // this is too expensive
```
jonathanyeung, line 91: How long does this schema setup take for org-private? (maybe we do this measurement/tuning as a follow up)  If it helps, I left a test-case snippet that can be locally run to get an idea of how long things take here: [[user.jonathan.dev.sqlite-temp-code]]

Might need to add some additional logging though (I also am working on a pr to improve some of our logging in parseAllNoteFiles()

jonathanyeung (changes requested):

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
hikchoi: I changed it because all the queries seem to insert them as `parent` and `child`. I just assumed the tables were defined correctly.

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
hikchoi: Ideally we should do that check before inserting, but yeah. I'll add a check here for now

packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
hikchoi: right. I'll fix that.

packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
hikchoi: have to test that out. I'll let you know.

packages/engine-server/src/drivers/sqlite/tables/HierarchyTableUtils.ts
hikchoi: I'll change it back to `parentId` and `childId` and instead change the queries so the column names follow.

hikchoi:
update:

- added a limit to the max recursion by limiting the maximum hierarchy depth to 32 (arbitrary). We can extend this to worst case by setting it to the number of rows in the `Hierarchy` table.
- fixed an issue where automatically generated schemas were creating duplicate note -> schema relations in `SchemaNotes` table. I'm doing this by truncating the `SchemaNotes` table each session.

packages/engine-server/src/drivers/sqlite/SqliteMetadataStore.ts
hikchoi: Seems like `SqliteMetadataStore.initSchema` take about 2500ms consistenly on my machine with `org-workspace`.

For comparison on my machine it take about 8 seconds to for delta init + schema init to finish. so about 30% of the time is spent on `initSchema`.

jonathanyeung (approved):
nice!

PR #3832, opened 2022-12-02
chore: add job dependency in nightly workflow

Harshita-mindfire:


jonathanyeung (approved):
offline comment about leaving a comment to explain why we need job ordering.

PR #3855, opened 2022-12-08
chore(internal): add platform specific matrix to publish extension workflow

Harshita-mindfire:
This PR aims to update both `publishing-minor` and `publishing-patch` workflow to include platform-specific matrix for downloading sqlite binaries

### DIFF
--- a/.github/workflows/publish-extension-dendron-minor.yml
+++ b/.github/workflows/publish-extension-dendron-minor.yml
@@ -95,7 +95,7 @@ jobs:
           if [ ${{ matrix.platform }} == general ]; then
             yarn build:minor:remote:ci:noparam
           else 
-            yarn build:minor:remote:ci ${{ env.PUBLISHING_TARGET }}
+            yarn build:minor:local:ci ${{ env.PUBLISHING_TARGET }}
           fi
 
       - name: Upload VSIX Artifact
--- a/.github/workflows/publish-extension-dendron-minor.yml
+++ b/.github/workflows/publish-extension-dendron-minor.yml
@@ -11,6 +11,22 @@ jobs:
     environment: plugin-production
     strategy:
       fail-fast: true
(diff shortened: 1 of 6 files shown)


hikchoi (approved):

.github/workflows/publish-extension-dendron-minor.yml
```diff
@@ -64,7 +92,11 @@ jobs:
         run: |
           echo "VSIX_FILE_NAME=dendron-${{ env.DENDRON_RELEASE_VERSION }}.vsix" >> $GITHUB_ENV
           echo "VSIX_RELATIVE_PATH=./packages/plugin-core/dendron-${{ env.DENDRON_RELEASE_VERSION }}.vsix" >> $GITHUB_ENV
-          yarn build:minor:remote:ci
+          if [ ${{ matrix.platform }} == general ]; then
+            yarn build:minor:remote:ci:noparam
+          else 
+            yarn build:minor:remote:ci ${{ env.PUBLISHING_TARGET }}
```
jonathanyeung, line 98: One difference with the main flow vs nightly flow is that the main flow will also publish the packages to NPM - we can only do this once (else NPM requires a revision bump). 

I think the correct thing to do here is for the 'general' run, do the publishing to NPM as well, whereas for all others, don't do it.  I think if we adjust it to the following, then it _should_ work.  @kevinslin to chime in otherwise. 

```bash
          if [ ${{ matrix.platform }} == general ]; then
            yarn build:minor:remote:ci:noparam
          else 
            yarn build:minor:local:ci ${{ env.PUBLISHING_TARGET }}
```
The difference is making the 'else' statement version use 'local' instead of 'remote'.  We would also need to make the corresponding change in the patch flow as well.

jonathanyeung (changes requested):

